PR #180, opened 2015-01-23
Updated feature section color and margin

beaufortfrancois:
I don't know if it's only me but I tend to click on the section names because of their blue (link) colors. Hopefully, this patch addresses this issue.

Left=Before
Right=After
screenshot 2015-01-23 at 2 57 37 pm


ebidel:
:+1:

PR #181, opened 2015-01-23
Show Prefixed only when actually Prefixed

beaufortfrancois:
What do you think about not showing "Prefixed No" anymore since this is standard behavior?
This will still show "Prefixed Yes" if this is the case though.


ebidel:
Seems good! :100:

Issue #97, opened 2014-02-24
Automate UMA property list


ebidel:
The list is manually maintained in uma.py. It should pull from live UserCounter.h/cpp

Possibly from http://src.chromium.org/viewvc/chrome/trunk/src/tools/metrics/histograms/histograms.xml, which is CORs enabled

beaufortfrancois:
@ebidel May I tackle this one?

ebidel:
Totally.

beaufortfrancois:
http://src.chromium.org/viewvc/chrome/trunk/src/tools/metrics/histograms/histograms.xml hasn't been updated for 5 months since chromium switched entirely to Git.

Therefore I think I'll use https://chromium.googlesource.com/chromium/src/+/master/tools/metrics/histograms/histograms.xml?format=TEXT even though it only serves base64 encoded text for now. See https://code.google.com/p/gitiles/issues/detail?id=7 for some background

beaufortfrancois:
WIP at https://github.com/beaufortfrancois/chromium-dashboard/compare/GoogleChrome:master...beaufortfrancois:automateUMA?expand=1

ebidel:
@beaufortfrancois can we close this?

beaufortfrancois:
Yup! Doing it now!

PR #183, opened 2015-01-26
Update CSS UMA

ebidel:
R: @beaufortfrancois

### DIFF
--- a/uma.py
+++ b/uma.py
@@ -457,6 +457,11 @@ CSS_PROPERTY_BUCKETS = {
   453: 'grid',
   454: 'all',
   455: 'justify-items',
+  456: 'scroll-blocks-on',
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
Would love to LGTM!
One nit, where did you get these new UMAs?

ebidel:
The two links to the source are in uma.py. https://github.com/GoogleChrome/chromium-dashboard/blob/master/uma.py#L3

beaufortfrancois:
LGTM ;)

Do you keep track of it manually or do you have a cron job that alerts you when this file is updated?

ebidel:
Manually for now. See https://github.com/GoogleChrome/chromium-dashboard/issues/97

Issue #123, opened 2014-05-05
Server should do usage stat rounding


ebidel:

beaufortfrancois:
Yup.

PR #185, opened 2015-01-29
Truncate day percentage numbers for non whitelisted users

beaufortfrancois:
Based on our previous discussion, here's something that could suit our needs to do usage stat rounding server side.

Following same patterns for edit rights, if user is whitelisted, he'll have access to raw data for JSON endpoints while it will still show 4 decimals for UI. 
If user is not, all JSON endpoints will always truncate day percentage stats.

What do you think @ebidel @dstockwell?

BUG=#123

### DIFF
--- a/metrics.py
+++ b/metrics.py
@@ -58,6 +58,7 @@ class TimelineHandler(common.JSONHandler):
 
       memcache.set(KEY, data, time=CACHE_AGE)
 
+    data = self._clean_data(data)
     super(TimelineHandler, self).get(data)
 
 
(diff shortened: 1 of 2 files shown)


metrics.py
```diff
@@ -58,6 +58,7 @@ def get(self):
 
       memcache.set(KEY, data, time=CACHE_AGE)
 
+    data = self._clean_data(data)
```
ebidel: Should the JSON handler just call `self._clean_data(data)` instead of dropping the calls in here and below?

ebidel:
Nice! I think this will work well. One question.

metrics.py
beaufortfrancois: That's what I wanted to do initially but then I realized JSONHandler was also used in server.py to serve features.json. See https://github.com/GoogleChrome/chromium-dashboard/blob/3f8b0dc514cc86a532a8789da2539b4d117c373c/server.py#L112

ebidel: Right on :)

Issue #195, opened 2015-04-10
Add tag review process to intent to implement template


ebidel:
"We would like new APIs to go through the [tag review process](https://github.com/w3ctag/spec-reviews/issues), or provide a reason for skipping it. We hope this will reduce the chance that we will ship a poorly thought out API."

We should add a field to the new feature form for the link to the github issue (above) then update the template to include this field when users copy paste.

@beaufortfrancois interested?

PR #197, opened 2015-04-10
Update intent to implement to include a tag review

progers:
Tag, you're it! We're asking intenters to include a tag review now, or a reason why they are skipping that process.


ebidel:
Nice! I suppose we don't actually need to [demand a link](https://github.com/GoogleChrome/chromium-dashboard/issues/195) in the new feature form. WDYT?

progers:
Yeah I don't think we need to include a form entry for this, as it's mostly for the api owner process.

ebidel:
LGTM

PR #198, opened 2015-04-11
Status filter for the RSS feed

### LINKED ISSUE #101: Create a subdashboard specifically for deprecated and removed features
Goal is to have an easy, focused place for developers to get information about what features are going to be removed from Chrome.

Context: https://groups.google.com/a/chromium.org/d/msg/blink-dev/nNsZDn_utUc/clT3aVUGZaAJ

jeffposnick:
@ebidel @beaufortfrancois

This exposes a new RSS feed, https://www.chromestatus.com/features?status=Deprecated,Removed (or any other comma-separated list of `impl_status_chrome ` values). It's a little hacky since I don't believe App Engine's datastore allows for `OR`s in its queries. I'm also only doing this for the `impl_status_chrome` field—it could be made more generic, but I'm not sure if there's value in that.

This together with the existing https://www.chromestatus.com/features#deprecated|removed may be sufficient to close #101

### DIFF
--- a/server.py
+++ b/server.py
@@ -112,8 +112,7 @@ class MainHandler(common.ContentHandler, common.JSONHandler):
       elif path.endswith('.xml'): # Atom feed request.
         status = self.request.get('status', None)
         if status:
-          statuses = status.split(',')
-          feature_list = models.Feature.get_all_with_statuses(statuses)
+          feature_list = models.Feature.get_all_with_statuses(status.split(','))
         else:
           filterby = None
 
--- a/server.py
+++ b/server.py
@@ -110,29 +110,34 @@ class MainHandler(common.ContentHandler, common.JSONHandler):
         feature_list = self.__get_feature_list()
         return common.JSONHandler.get(self, feature_list, formatted=True)
       elif path.endswith('.xml'): # Atom feed request.
-        filterby = None
+        status = self.request.get('status', None)
+        if status:
+          statuses = status.split(',')
+          feature_list = models.Feature.get_all_with_statuses(statuses)
+        else:
+          filterby = None
 
-        category = self.request.get('category', None)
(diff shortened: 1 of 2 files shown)


server.py
```diff
@@ -110,29 +110,34 @@ def get(self, path, feature_id=None):
...
-              break
-
-        feature_list = models.Feature.get_all( # Memcached
-            limit=max_items,
-            filterby=filterby,
-            order='-updated')
+        status = self.request.get('status', None)
+        if status:
+          statuses = status.split(',')
+          feature_list = models.Feature.get_all_with_statuses(statuses)
```
ebidel: feature_list = models.Feature.get_all_with_statuses(status.split(','))

server.py
```diff
@@ -110,29 +110,34 @@ def get(self, path, feature_id=None):
...
-            normalized = normalized_name(v)
-            if category == normalized:
-              filterby = ('category =', k)
-              break
-
-        feature_list = models.Feature.get_all( # Memcached
-            limit=max_items,
-            filterby=filterby,
-            order='-updated')
+        status = self.request.get('status', None)
```
ebidel: remove

ebidel:
Nice! I was going to take a look at this today too. Glad I didnt :+1: 

LGTM. just the nit.

server.py
jeffposnick: Remove what? `status = self.request.get('status', None)` still seems necessary.

ebidel: Hmm, not sure. I think I was on a slow connection and this wasn't intended for here. Ignore.

ebidel: I think it was for `status = status.split(',')` line which you've inlined now.

ebidel:
LGTM

PR #209, opened 2015-05-27
Improved the labels for removed features.

phistuck:
The labels for removed features were the same as deprecated features, which is problematic and very wrong (a deprecated feature is still available, while a removed feature is simply not, which can throw exceptions).

Example of the current state -
https://www.chromestatus.com/feature/5242458724106240
Its status is "Removed", but it is displayed as "Deprecated".

### DIFF
--- a/templates/feature.html
+++ b/templates/feature.html
@@ -21,7 +21,8 @@
   <section id="name">
     <h2>
       {{ feature.name }}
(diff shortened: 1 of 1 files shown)


ebidel:
:+1: sorry for the delay

ebidel:
Deployed

Issue #214, opened 2015-07-07
Deleting a feature raises 404


beaufortfrancois:
Will investigate more...

beaufortfrancois:
@ebidel Let me know if that's ok with you.

beaufortfrancois:
This is addressed in #215. @ebidel Can we push it to live server?

ebidel:
Meant to do that yesterday! Should be deployed now.

beaufortfrancois:
Perfect. Thanks!

PR #215, opened 2015-07-07
Fixed Delete feature behaviour

beaufortfrancois:
Hello @ebidel 

This patch addresses the grunt error below:
`Fatal error: This version of vulcanize is not compatible with Polymer < 0.8. Please use vulcanize 0.7.x. File: static/elements/delete-link.html`

Moreover the `onTap` event wasn't properly prevented so I switched to `onClick` which works great now on mobile as well.

BUG=https://github.com/GoogleChrome/chromium-dashboard/issues/214

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "devDependencies": {
     "grunt": "*",
     "grunt-appengine": "^0.1.5",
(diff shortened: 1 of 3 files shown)


ebidel:
:+1:

PR #221, opened 2015-07-10
Move off deprecated Files API. Fixes #219

### LINKED ISSUE #219: Move off of the Files API
The current code is affected by the [Files API service turndown](https://cloud.google.com/appengine/docs/deprecations/files_api) which will go into effect on July 28.

It looks like the code in question is https://github.com/GoogleChrome/chromium-dashboard/blob/9edb14efdc4202ca2b01ab3fedd602f438269a76/admin.py#L171

CC: @ebidel @beaufortfrancois

ebidel:
R: @jeffposnick

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -29,7 +29,7 @@ import webapp2
 import xml.dom.minidom
 
 # Appengine imports.
-from google.appengine.api import files
+import cloudstorage
 from google.appengine.api import memcache
 from google.appengine.api import urlfetch
 from google.appengine.api import users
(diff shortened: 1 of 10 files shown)


admin.py
```diff
@@ -168,9 +168,9 @@ def get(self):
 
     if settings.PROD:
       try:
-        with files.open(BIGSTORE_BUCKET + filename, 'r') as unused_f:
+        with cloudstorage.open(BIGSTORE_BUCKET + filename, 'r') as unused_f:
           pass
-      except files.file.ExistenceError, e:
+      except Exception, e:
```
jeffposnick: How about catching `cloudstorage.errors.Error` instead of the generic `Exception`? It looks like that's the class that all the other errors in the library extend.

jeffposnick:
Reviewed, with the one comment.

admin.py
ebidel: DOne

ebidel:
This is deployed

PR #226, opened 2015-08-01
Added an edit link to the standalone feature page

phistuck:
People keep linking to the standalone page and after I sign in, I must either construct the edit link by myself, or go to the list of features and re-search for the feature, which is annoying. The solution - add an edit link in the standalone page.


phistuck:
I signed it!

ebidel:
LGTM. Thanks!

PR #228, opened 2015-08-04
Optimize algorithm for selecting milestone on scrolling the feature list

ebidel:
@jeffposnick @beaufortfrancois 

This reduces the workload on scroll events by only looking at the features below the container's scrollTop. The milestone is selected in the left nav iff a new label is scroll passed. The perf gain isn't much now, but it will be when we get more and more features

### DIFF
--- a/static/elements/chromedash-featurelist.html
+++ b/static/elements/chromedash-featurelist.html
@@ -38,7 +38,9 @@
         return false;
       }
     }
+
     var featureLiList_ = [];
+
     Polymer({
       is: 'chromedash-featurelist',
 
@@ -62,6 +64,16 @@
         search: { value: null },
         whitelisted: {
           type: Boolean,
+          value: false
+        },
+        _topmostFeatureInScrollList: {
+          type: Object,
+          readOnly: true,
+          value: function() { return {}; }
+        },
+        _milestoneWasClicked: {
+          type: Boolean,
+          readOnly: true,
           value: false
         }
       },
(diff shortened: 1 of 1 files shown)


static/elements/chromedash-featurelist.html
```diff
@@ -152,18 +164,41 @@
...
+
+        // Grab the last feature (it's at the top of the scroll container).
+        var topmostFeature = items.length ? items[items.length - 1] :
+                                            featureLiList_[0];
+
+        var idx = featureLiList_.indexOf(topmostFeature);
+         // Use next feature in list so milestone labels line up better.
+        var feature = this.features[idx + 1];
+
+        var label = feature.shipped_milestone || feature.impl_status_chrome;
```
samccone: I am wondering if this will ever bite us due to the loose truthey check.. I think we should be ok just something to think on

static/elements/chromedash-featurelist.html
```diff
@@ -152,18 +164,41 @@
...
-          if (dist < 0) {
-            dist = -dist * 1.5;
-          }
-          if (closest == null || dist < closest.dist) {
-            closest = this.filtered[i];
-            closest.dist = dist;
-          }
+
+        // Generate list of features below current scroll position.
+        var items = featureLiList_.filter(function(el) {
```
samccone: is the polyfilled? Or are we ok with this not working in old ie :)

static/elements/chromedash-featurelist.html
```diff
@@ -152,18 +164,41 @@
...
+          if (el.offsetTop < containerScrollTop) { return el; }
```
samccone: perf wise, should we throw this work into a RAF, to prevent the reflow lockup of doom?

static/elements/chromedash-featurelist.html
ebidel: IE 9 has filter. The wc.js polyfilld "work" in IE 10+, so we're covered.

samccone: :+1: sgtm :)

static/elements/chromedash-featurelist.html
ebidel: scroll events are async in chrome. If it becomes an actual bottleneck, then we can consider it.

samccone: :+1: sgtm 

TIL 

> scroll events are already async in chrome.

jeffposnick:
:+1:

PR #229, opened 2015-08-19
Perf improvements

ebidel:

- Load web fonts up front (heavy edge caching of fonts api should help)
- new endpoint for omaha proxy data. There was no sense to make the live query to the all.json endpoint if the server was already caching it
- Loading the /features list faster by Promisifying it. It was previously waiting for WCR to make the request. That slowed down intial page load.


templates/features.html
```diff
@@ -74,12 +75,66 @@
...
+
+// Fire of features.json XHR right away so data can populate faster.
+var featuresPromise = new Promise(function(resolve, reject) {
+  var url = location.hostname == 'localhost' ? '/static/js/mockdata.json' :
+                                               '/features.json';
+  var xhr = new XMLHttpRequest(); // Fill features list via AJAX.
+  xhr.open('GET', url);
+  xhr.responseType = 'json';
+  xhr.onloadend = function(e) {
+    resolve(e.target.response);
```
jeffposnick: A pattern I [like to follow](https://github.com/GoogleChrome/ioweb2015/blob/master/app/scripts/helper/request.js#L115) when Promise-ifying XHR is to check `e.target.status` for anything >= 400, and `reject()` if you get that back, since you'd presumably like to treat HTTP error responses the same way you treat XHR networking failures.

templates/features.html
```diff
@@ -94,82 +149,45 @@
...
+  featureList.metadata = chromeMetadata; // Feature list needs access to the selected milestone.
+  featureList.search = search;
 
+  featuresPromise.then(function(features) {
+    featureList.features = features;
     updateCount(featureList.filter(search.value));
     search.disabled = false;
-  };
-  xhr.send();
+  });
```
jeffposnick: There's no `catch()` here. There's no error handling in the original code, so this isn't a regression, but this might be an opportunity to add in some logic that swapped out the `loading...` with an error message.

The way the `loading...` is implemented with `::after` leaves me unsure as to how to do it cleanly. Maybe... create a `showError()` method on `<chromedash-featurelist>` and use that to swap around the internal styles?

templates/features.html
ebidel: Done

templates/features.html
ebidel: Done

ebidel:
PTAL

jeffposnick:
Feel free to ignore or defer as a future enhancement: the main page makes separate XHRs for `/features.json` and `/omaha_data`. Since the `/features.json` handler is already calling the `__get_omaha_data()` method to assign first-of-milestone annotations, how about returning that Omaha data as an additional field in the `/features.json` response? That would eliminate the need to call `/omaha_data`, right?

It would mean that third-party apps making https://www.chromestatus.com/features.json requests would get a response that's a bit more bloated, but it's an extra ~2kB on a response that's already ~90kB, so nothing egregious.

ebidel:
Possibly for the future :)

I thought about doing that but wanted to keep the metadata element separate from the features. It's responsible for just omaha data and feature are responsible for just features.

PR #231, opened 2015-08-20
Perf update round 2

ebidel:

- uses a special version of iron-icon that improves perf (not released yet)
- Improvements perf by not creating so much DOM upfront. Custom element instance count was really high with all the iron-icon (5k+). This takes us down to ~400 elements :)

In a future PR, I'll move to iron-list, everything should be even faster for page load!


samccone:
lgtm :+1: nice work @ebidel

addyosmani:
This is looking sharp :shaved_ice: LGTM2! Predictable DOM performance last I heard was sub-1000 DOM elements per page. Glad to see CD hitting way way under the threshold now. Nice work :)

PR #234, opened 2015-08-23
Iron list

ebidel:
More perf!

This further decreases load time by using iron-list to manage the feature list. In the long run, this will be a big win as the number of features goes up. On load, the features now unveil at <383ms after first paint. Only 53 custom element instances on load!

There are also some other improvements, like scroll perf (now that iron-list is managing the list) and debouncing search input.

These changes are already pushed up on https://www.chromestatus.com/features if you want to see the difference compared to the original 1.0 port: https://polymer10-perf-dot-cr-status.appspot.com/features


jeffposnick:
I'm still looking through the code, but I'm seeing styling issues with `<iron-list>`:

Double-scrollbar when the viewport is less than 910px high:
image

The `<features-list>` doesn't extend all the way to the footer if you start out with a large-ish viewport, reduce the viewport height, and then increase the height again. This is a little flaky to reproduce:
image

static/elements/chromedash-feature.html
```diff
@@ -274,20 +270,34 @@
...
+        if (target.classList.contains('tooltip') || 'tooltip' in target.dataset ||
+            target.tagName == 'A' || target.tagName == 'MULTI-LINKS') {
+          return;
+        }
       }
-      this.open = !this.open;
+
+      this.set('feature.open', !this.feature.open);
+
+      this.fire('feature-toggled', {feature: this.feature, open: this.feature.open});
```
jeffposnick: I'm having trouble understanding the distinction between this line and the `this.fire('feature-toggled', {feature: this.feature, open: this.open});` that's fired within the `_openChanged` observer. Just double-checking to make sure they serve distinct purposes.

jeffposnick:
(I noticed https://github.com/PolymerElements/iron-list/issues/73 while reading through the code, and I'm assuming that your attempts to work around that are leading to at least the second, if not also the first, sizing issue.)

static/elements/chromedash-feature.html
ebidel: Added a comment. Basically, `_toggle()` is when a user clicks open a feature card. `_openChanged` only gets invoked when the computed prop `open` changes. That will happen on page load when a feature is deep linked. Firing this event here kicks off the same logic in the main page which does the `history.pushState` etc.

ebidel:
Thanks @jeffposnick. Yea the resizing issues are from iron-list. I wanted to put in a `resize` handler for the page, but was hesitant to do so. I didn't want to forget about it when iron-list is fixed.  I'm not sure many people resize their window once the page loads, but I can add it if you want.

The list doesn't update itself when you resize the page b/c I'm setting the height explicitly in `attached`: https://github.com/GoogleChrome/chromium-dashboard/pull/234/files#diff-48bf61fdf81b40310a82e557f4ac9bb0R86.

jeffposnick:
I'm going to play around with some alternatives to https://github.com/GoogleChrome/chromium-dashboard/pull/234/files#diff-48bf61fdf81b40310a82e557f4ac9bb0R86

We should be able to get rid of the double scroll bars at least.

jeffposnick:
Two possibilities to get rid of the double-scroll bars:

Explicitly setting `box-sizing: content-box;` in the `chromedash-featurelist` style and then using

`this.$.ironlist.style.height = window.getComputedStyle(Polymer.dom(this).node).getPropertyValue('height');`

Or, subtract the top + bottom padding (5px each): `this.$.ironlist.style.height = (Polymer.dom(this).node.clientHeight - 10) + 'px';`

The second obviously is susceptible to breakage if the padding styles change.

Looks like there are also some suggestions in https://github.com/PolymerElements/iron-list/issues/73 for working around the issue entirely, which would be much nicer.

ebidel:
All these report the same dimension for me:
```
var node = Polymer.dom(this).node;
console.log(
  getComputedStyle(node).height,
  getComputedStyle(node).getPropertyValue('height'),
  node.clientHeight);
````

jeffposnick:
You'll see that unless you also set `box-sizing: content-box`, since there's a global `box-sizing: border-box` applied by default.

Here's a diff of changes against your branch that seem to resolve the issue: https://github.com/GoogleChrome/chromium-dashboard/compare/ironlist...jeffposnick:ironlist-patch

Or, you know, just subtract 10 from your calculation, since this is (hopefully) a temporary hack anyway.

ebidel:
Doh. CSS specificity issue :) Fixed

jeffposnick:
:+1:

PR #239, opened 2015-10-02
/samples

ebidel:
R: @jeffposnick 

Preview at https://polymer10-samples-dot-cr-status.appspot.com/samples

screen shot 2015-10-01 at 7 09 45 pm


addyosmani:
With Sam's comments addressed this LGTM. This is looking really nice. Great work, @ebidel!

templates/samples.html
```diff
@@ -0,0 +1,229 @@
...
+  this.filtered = features;
+};
+
+t.removeFilters = function() {
+  t.filter();
+};
+
+// Fire of samples.json XHR right away so data can populate faster.
+var samplesPromise = new Promise(function(resolve, reject) {
+  var url = location.hostname == 'localhost' ? 'https://polymer10-samples-dot-cr-status.appspot.com/samples.json' :
```
jeffposnick: `https://polymer10-samples-dot-cr-status.appspot.com/samples.json` → `https://www.chromestatus.com/samples.json`

templates/samples.html
```diff
@@ -0,0 +1,229 @@
...
+    } else {
+      reject(new Error('Samples XHR failed with status ' + e.target.statusText));
+    }
+  };
+  xhr.onerror = reject;
+  xhr.send();
+});
+
+samplesPromise.then(function(features) {
+  var re = /github.com\/GoogleChrome\/samples\/tree\/gh-pages\/(.*)/;
```
jeffposnick: There might be entries that have `googlechrome` instead of `GoogleChrome`, so use the `i` option with that RegExp.

templates/samples.html
```diff
@@ -0,0 +1,229 @@
...
+      resolve(e.target.response);
+    } else {
+      reject(new Error('Samples XHR failed with status ' + e.target.statusText));
+    }
+  };
+  xhr.onerror = reject;
+  xhr.send();
+});
+
+samplesPromise.then(function(features) {
```
jeffposnick: The tl;dr is that it's fine to do this rewriting if you'd like, but there's a historical reason why some samples link to the repo.

Some background: early on, I tended to link to the GitHub repo itself rather than the deployed version on GitHub Pages, since I wanted to direct users to the source code, and there's a link to the GitHub Pages URL in each directory's `README.md`.

Later on, we added "View Source" and "View on GitHub" links to the samples template, so it should be obvious even for inexperienced developers how to get to the underlying source code. At that point, I started linking to the GitHub Pages URL.

If the goal is to always point to the GitHub Pages URL, I should address that by cleaning up the older links in the App Engine datastore. I can do that one at a time as the older samples get migrated to the new templates.

templates/samples.html
```diff
@@ -0,0 +1,229 @@
...
+  var re = /github.com\/GoogleChrome\/samples\/tree\/gh-pages\/(.*)/;
+  features.forEach(function(f, i) {
+    f.sample_links.forEach(function(link, i) {
```
jeffposnick: `.replace()` returns the original string if there's no match. Untested, but I think this should be equivalent and is a little more straightforward:

```js
f.sample_links = f.sample_links.map(function(link) {
  return link.replace(re, 'googlechrome.github.io/samples/$1');
});
```

templates/samples.html
```diff
@@ -0,0 +1,229 @@
...
+  t.filtered = features;
+
+  $('#sample-panel').classList.remove('loading');
+
+}, function(e) {
+  console.error(e);
+});
+
+// Clear input when user clicks the 'x' button.
+search.addEventListener('search', function(e) {
```
jeffposnick: I'm having a hard time figuring our what fires the `search` event. If it's coming from a source you have control over, can you consider renaming it from `search` to something else, like `clear-search`? That seems more in keeping with what the event is actually used for.

templates/samples.html
```diff
@@ -0,0 +1,229 @@
+{% extends "base.html" %}
+{% load verbatim %}
+
+{% block html_imports %}
+  <link rel="import" href="/static/elements/samples-imports{% if VULCANIZE %}.vulcanize{% endif %}.html">
+{% endblock %}
+
+{% block rss %}
+<link rel="alternate" type="application/rss+xml" href="http://www.chromestatus.com/features.xml" title="All features" />
```
jeffposnick: Feel free to disregard if this requires any work, but can we get an RSS feed of features that have samples "for free" by using parameters to `http://www.chromestatus.com/features.xml`? If so, can you use the appropriate URL/title for that RSS feed here?

jeffposnick:
Really awesome, @ebidel!

Beyond the inline comments, the default ordering might need to be tweaked instead of what we initially talked about. After visiting the live page, I realize that what folks might do is click on the first demo link they see to try things out, and that's going to end up being a demo associated with a feature that hasn't been launched even in Canary yet, so it will almost certainly not work. So I guess ideally features that haven't made it to Chrome yet would be sorted to the bottom by default?

jpmedley:
Eric,

I like the new design. Great work! It's about time for that page to get a
makeover.

I have a couple of questions. Please redirect me if this isn't the right
forum.

Where do I click to see features for a particular release?

In addition to "View Demo" could we have a button that says "View Docs"?

Thanks

Joe Medley | Technical Writer, DevPlat | jmedley@google.com | 816-678-7195

templates/samples.html
ebidel: Gooooo catch!

templates/samples.html
ebidel: Done

templates/samples.html
ebidel: Up to you. I think we should be consistent, but I don't have a preference. Most of the time when I'm looking for demos, I want to run the demo before looking at the code. But that's just me. If we clean up the samples themselves, happy to remove this rewriting.

templates/samples.html
ebidel: Much nicee!

templates/samples.html
ebidel: Would you believe `input[type="search"` fires this!??!?! It's the event when someone clicks the "x"/

templates/samples.html
ebidel: Added

ebidel:
@jeffposnick PTAL. We could restrict the list to only show features that have shipped in a version of Chrome or do the reordering like you suggest.

@jpmedley 

> Where do I click to see features for a particular release?

Didn't build that in yet.

> In addition to "View Demo" could we have a button that says "View Docs"?

I'm hesitant to add to much to the list view. Otherwise, we're just reinventing the feature list:  chromestatus.com/features.

jeffposnick:
Thanks for addressing the feedback!

If it's straightforward, I'd go with sorting the entries so that the ones associated with any release version come first, followed at the end by the ones that are still pre-release.

ebidel:
PTAL

ebidel:
Staging is up to date: https://polymer10-samples-dot-cr-status.appspot.com/samples

jeffposnick:
Something's off with the sorting:

image

Any idea why those M47/M46 features are listed after M36? They aren't features hidden by a flag.

ebidel:
It's b/c Array.includes and spread are enabled by default but dont have a shipping milestone yet.

ebidel:
I take that back, but I'm seeing it correct when I hit the page:

screen shot 2015-10-05 at 11 45 49 am

jeffposnick:
What I'm seeing is apparently two sorted lists, one following the other, with the same entries repeated in each. Here's the shorter "Offline / Storage" example, which makes it more obvious what's going on:

image

templates/samples.html
jeffposnick: Gah.

ebidel:
PTAL. I was missing a db query.filter() that got accidentally removed. Also cleaned up some more of the UI. You can also do version filtering now (e.g. `=47`) and searches will deep link the URL.

screen shot 2015-10-05 at 3 32 58 pm

jeffposnick:
:+1:

ebidel:
Shipped https://www.chromestatus.com/samples

PR #262, opened 2016-02-03
Add support for Experimental Framework

beaufortfrancois:
This PR add support for the Experimental Framework coming soon.
See https://github.com/jpchase/ExperimentalFramework/blob/master/explainer.md for some background.

screenshot 2016-02-03 at 10 41 01 am

PR #264, opened 2016-02-04
Moved "Experimental Framework" before "Enabled by default".

beaufortfrancois:
As discussed in https://github.com/GoogleChrome/chromium-dashboard/pull/262#issuecomment-179558422, this PR moves "Experimental Framework" before "Enabled by default" in Feature form.


ebidel:
LGTM. Sorry for the delay.

Issue #269, opened 2016-03-03
Samples field does not support commas inside URLs


cbiesinger:
https://www.chromestatus.com/feature/5670760496496640

In the samples field of that feature, I have a jsbin link. Those links come with commas:
http://jsbin.com/zefuxiviso/2/edit?html,output

The dashboard splits it at the comma :(

cbiesinger:
Note that for that specific feature I've edited it the link to not have a comma, but nonetheless this is someting that would be good to have

ebidel:
Ah, good one!

@beaufortfrancois can you take this?

beaufortfrancois:
I will

beaufortfrancois:
I just sent a PR for this.

beaufortfrancois:
@cbiesinger I've changed https://www.chromestatus.com/feature/5670760496496640 to include http://jsbin.com/zefuxiviso/2/edit?html,output

cbiesinger:
Thank you!

PR #270, opened 2016-03-09
Use one URL per line for docs and samples

beaufortfrancois:
In order to make sure any doc and sample URL (even one with a comma) can be added to a feature, I've simply decided to replace comma-separated URLs with one per line. 

Moreover, this makes it even easier to read.

FIX=#269
R=@ebidel

### DIFF
--- a/models.py
+++ b/models.py
@@ -246,8 +246,8 @@ class Feature(DictModel):
     d = self.to_dict()
     #d['id'] = self.key().id
     d['owner'] = ', '.join(self.owner)
-    d['doc_links'] = ', '.join(self.doc_links)
-    d['sample_links'] = ', '.join(self.sample_links)
+    d['doc_links'] = '\r\n'.join(self.doc_links)
+    d['sample_links'] = '\r\n'.join(self.sample_links)
     d['search_tags'] = ', '.join(self.search_tags)
     return d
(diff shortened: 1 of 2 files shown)


models.py
```diff
@@ -526,12 +526,12 @@ class FeatureForm(forms.Form):
                              help_text="Prefer editor's draft.")
 
   doc_links = forms.CharField(label='Doc links', required=False, max_length=500,
-      widget=forms.Textarea(attrs={'cols': 50, 'placeholder': 'Links to documentation (comma separated)', 'maxlength': 500}),
-      help_text='Comma separated URLs')
+      widget=forms.Textarea(attrs={'cols': 50, 'placeholder': 'Links to documentation (one per line)', 'maxlength': 500}),
```
ebidel: does the input method need to change? Only the way we render them needs to change.

ebidel:
Thanks @beaufortfrancois. How does this look for features with a bunch of samples? The cards have a fixed height. Do we run into any overflow problems?

beaufortfrancois:
This only changes "Edit mode" not the rendering of the card. See below:

screenshot 2016-03-10 at 11 26 37 am

ebidel:
Got it.

PR #293, opened 2016-07-08
Remove compass. Use grunt-sass instead.

### LINKED ISSUE #292: Remove dep on compass
Use grunt-sass instead. There's no point to keep a Ruby dep.

ebidel:

Fixes #292


RByers:
Ok, cool - I don't know this code yet at all, but thanks for the heads up ;-)

jeffposnick:
Gave the new branch it a go with a local clone and the final CSS rendering looks good. 👍

PR #312, opened 2016-07-11
Improve the status description in single feature view

### LINKED ISSUE #244: Implementation Status Shows Wrong Label on Standalone page view
The following ticket shows the feature as being behind a flag in Chrome 47 for both desktop and Android when I view it on the dashboard:

https://www.chromestatus.com/features/5857463882481664

If I view the same ticket in standalone view, it makes it seem as if the feature is available in Chrome 47:

https://www.chromestatus.com/features/5857463882481664

I've confirmed with the feature owner that said feature is behind a flag on both platforms. The request is to change the label on the standalone page to accurately reflect the status of feature. In this case the Implementation status would say "behind a flag" for both desktop and Android.

NavidZ:
@RByers @mdittmer 
closes #244


templates/feature.html
```diff
@@ -82,44 +82,30 @@
...
+        {% endif %}
+        {% if feature.bug_url %}
+        (<a href="{{ feature.bug_url }}">launch bug</a>)
+        {% endif %}
+        in: </br>
+        {% if feature.shipped_milestone %}
+          - <b>Chrome for desktop release {{ feature.meta.milestone_str }}</b></br>
+        {% endif %}
+        {% if feature.shipped_android_milestone %}
+          - <b>Chrome for Android release {{ feature.shipped_android_milestone }}</b></p>
```
RByers: What do these closing 'p' tags match?

templates/feature.html
```diff
@@ -82,44 +82,30 @@
           {% else %}
             Enabled by default
           {% endif %}
-        {% endif %} in <b>desktop Chrome {{ feature.meta.milestone_str }}</b>
+        {% endif %}
+        {% if feature.bug_url %}
+        (<a href="{{ feature.bug_url }}">launch bug</a>)
+        {% endif %}
+        in: </br>
```
RByers: Do you mean just `<br>`?  br is an [empty element](https://developer.mozilla.org/en/docs/Web/HTML/Element/br).

templates/feature.html
```diff
@@ -82,44 +82,30 @@
...
+        {% if feature.shipped_milestone %}
+          - <b>Chrome for desktop release {{ feature.meta.milestone_str }}</b></br>
```
RByers: Rather than use "-" and `<br>`, this should probably be an HTML list.  i.e.
```html
<ul>
 <li>Item 1</li>
 <li>Item 2</li>
</ul>
```

RByers:
Looks reasonable overall, just a few html style nits.

If it's easy, can you also provide an example screenshot of what this looks like now?

templates/feature.html
NavidZ: I meant to write <br/>. Which one is better here? <br> or <br/>>

templates/feature.html
NavidZ: I tried to use that. Then I'm going to need to add a few css lines as well as the current css with margin=0 does't fit the content. Is it okay to add some new classes to the css file as well?

templates/feature.html
NavidZ: Missed those while copy pasting. Will remove.

NavidZ:
screenshot from 2016-07-11 16 18 08

This is the ul-li without any padding. I just realized there is also another use of this down the page. So I guess maybe this is okay that they are a bit off from the content.

templates/feature.html
RByers: Yes for sure, add classes as needed (although it may be simpler to just use a selector that relies on the dom structure).

templates/feature.html
RByers: I think in HTML empty elements CANNOT have end tags so `<br/>` is invalid.  It's the required XHTML of course though.  Anyway just write `<br>`.

templates/feature.html
RByers: But yeah this is consistent with the other UL lists here (also under the documentation section).  But it would be nice (and should be trivial) to add a CSS rule like this to align it better:
```css
#feature ul {
  list-style-position: inside;
}
```

RByers:
Thanks for the screenshot.  Can you also change the bolding?  It seems like the status ("Removed") should be bold, while each of the list items should not be.

ebidel:
This would be a good opportunity to get rid of the bullet styles (add `list-style: none`) on this page.

RByers:
Sure, eliminating the bullets entirely works for me too :-)

NavidZ:
Here is the new look adding both of those properties to the list items. @ebidel let me know if this is added under the correct tag. Is #feature which is filling the whole "content" in this page good enough or do you want me to add it somewhere to cover the whole page?

screenshot from 2016-07-12 11 02 45

ebidel:
LGTM

PR #315, opened 2016-07-12
Fixes #300

### LINKED ISSUE #300: Text insertion is shown instead of cursor when hovering over sample type buttons
Go to https://www.chromestatus.com/samples and hover over any of the left navigation buttons ('DOM', 'File APIs', etc.). A text insertion cursor is shown instead of the normal cursor that shows over buttons.

ebidel:
R: @RByers @NavidZ @mdittmer

### DIFF
--- a/templates/samples.html
+++ b/templates/samples.html
@@ -21,6 +21,11 @@ paper-input {
     padding: 12px;
   };
 }
+paper-item {
+  --paper-item: {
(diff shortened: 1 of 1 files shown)


RByers:
Hah, ok - LGTM.  I was leaving this trivial one for someone as their simplest starter bug ;-)

PR #317, opened 2016-07-12
Fixes #316

### LINKED ISSUE #316: Use Polymer lazyRegister
`window.Polymer = window.Polymer || {lazyRegister: true};` delays upfront registration work (style shimming, etc.) until instance creation. May help first paint even more...although we're using native shadow dom on the site so we already don't incur the shady dom perf hit.

ebidel:
R: @RByers @NavidZ @jeffposnick @mdittmer


RByers:
LGTM

Issue #318, opened 2016-07-12
Cannot debug parts of the site that rely on UMA


mdittmer:
There currently is no way to trick a local instance into rendering UMA-related pages. This is a problem for debugging these parts of the site. For example, #220 cannot be debugged until this is fixed.

ebidel:
I've put some time into trying to export the prod db and import it to the local dev db. It looks like there are issues with the current SDK. Filed https://code.google.com/p/googleappengine/issues/detail?id=13119

ebidel:
FYI, I've set the site up to pull from the live prod endpoints when you're on localhost. This can be overridden in settings.py by changing `PROD = False`

PR #322, opened 2016-07-12
Fix metrics view to align columns across all browsers

### LINKED ISSUE #301: CSS usage stat columns do not align with headers on safari or firefox
https://www.chromestatus.com/metrics/css/popularity

The columns don't line up with their headers. They do on Chrome.

image

RByers:
Apply flex via simple positional rules that apply to all rows of the table
instead of trying to name the elements in both the header and body that
should get flex: 1 vs. flex: 2.  The precentage heading was matching
both rules and the specificity differed depending on whether we had
native ShadowDOM.  Fixes #301


RByers:
@NavidZ PTAL

Note that I tested this by manually copying some html from the live site to the site served by my dev server as a workaround for #318.

static/sass/elements/chromedash-metrics.scss
```diff
@@ -36,13 +36,20 @@
       }
     }
 
+    > :first-child {
+      @include flex(1);
+    }
+
+    > :nth-child(2) {
+      @include flex(2);
```
NavidZ: There are two more children of stack-rank-list (one was for loading text and the other one I couldn't understand) which were hidden and didn't have sub tags either when everything is loaded. I wonder if this will have any bad effect on those if they can change during the loading or something like that. But I guess that wouldn't matter as much. I assume they will never have children so it should be good.

NavidZ:
lgtm

RByers:
@ebidel PTAL

ebidel:
LGTM

Issue #297, opened 2016-07-11
Offer better bug filing affordances from chromestatus entries


drufball:
Before a feature is launched, provide a link that goes to a new bug prepopulated with blocking:<feature bug>

After a feature is launched, provide a link that goes to a new bug prepopulated with the right component.

RByers:
I guess we'll have to augment the form / data model to include the monorail component corresponding to the feature.  This is probably a good idea anyway.  Is it worth trying to guess the component for all the existing entries, or should this just be optional (defaulting, perhaps, to Cr-Blink)?

Filed bugs should also automatically cc the feature owners.

clelland:
We can do some of the obvious ones, but it would probably be better to let the individual owners set up their own -- I'd send out a PSA to blink-dev asking folks to do it, and follow up in a couple of weeks to catch any features without components.

> Filed bugs should also automatically cc the feature owners.

Shouldn't monorail handle cc'ing feature owners? I'd be worried about naming individuals in chromestatus.com, as ownership changes all the time. Each component *should* have someone triaging it, though.

RByers:
> We can do some of the obvious ones, but it would probably be better to let the individual owners set up their own -- I'd send out a PSA to blink-dev asking folks to do it, and follow up in a couple of weeks to catch any features without components.

Wait a second - maybe asking people to supply the component is silly - we should already have it from the components assigned to the launch bug!  Perhaps we should just try to read those and copy them rather than store this in the database?  Sorry I didn't think of this sooner.  I don't know what the monorail API is like though to know if this is easy to do, but it's almost certainly better then asking devs to update all their feature entries (seems unlikely that many would do so to be honest).

> Shouldn't monorail handle cc'ing feature owners? I'd be worried about naming individuals in chromestatus.com, as ownership changes all the time. Each component should have someone triaging it, though.

Yeah but the owners are often finer grained than the monorail components.  Eg. all of input-dev uses `Blink>Input` but all features have a couple eng owners who bugs should be routed to.  They'll get to the right engineers 0-7 days sooner (given input-dev weekly triage) if we cc the feature owners.

clelland:
> Wait a second - maybe asking people to supply the component is silly - we should already have it from the components assigned to the launch bug!

Interesting... that would be good for newer features, I think. Do existing platform features have reliable data in the launch bugs (or launch bug links in the first place?)

I'll take a look at their API to see how easy that will be to read out.

[...]

PR #328, opened 2016-07-13
Ad bug creation button in list view

clelland:
This change exposes a "File new bug" button for each feature in the feature list, which links to the "new issue" page on crbug.com. If the feature has not yet launched, and has a launch bug number, then the new bug will be created as a blocker by default. Feature owners will be cc'd if they are defined on the feature.

Partially addresses #297 (Still need a better way to find the correct components to auto-assign the created bug to)


ebidel:
Can you post a screenshot?

static/elements/chromedash-feature.html
```diff
@@ -381,18 +386,43 @@
...
+    _computeCrbug: function(link) {
+      var crbugNumber = this._computeCrbugNumber(link);
+      if (crbugNumber) {
+        return '#' + crbugNumber;
       }
       return link;
+    },
+    _computeNewBugUrl: function(feature) {
+      var url = 'https://bugs.chromium.org/p/chromium/issues/entry';
+      var params = ["components=Blink"];
```
ebidel: single quotes

ebidel:
LGTM. Just the nit

clelland:
screen shot 2016-07-13 at 1 59 17 pm

static/elements/chromedash-feature.html
clelland: Done.

RByers:
This is great, thanks!

PR #330, opened 2016-07-13
Show Opera in implementation status sections

### LINKED ISSUE #267: Removed in Chrome is shown as shipped in Opera
Here's how the entry for SVGElement.offset* removal is shown:
Rendering of SVGElement.offset* entry
And these are the settings:
Settings of SVGElement.offset* entry

I was contacted by a web developer asking if this means that Opera will retain support even when it's removed from Chrome, which isn't what's going to happen.

jpchase:
This change moves the Opera release details into the implementation section, along side the Chrome release details. The Opera details are thus removed from the section for consensus & standardization section. This applies to both the features list and the standalone feature view. 

In addition, the "Implementation status" section is renamed to "Chromium status" or "Status in Chromium", as appropriate. This also applies to the new/edit feature page.

For now, the Opera icons are not used on the features list. If we like how this looks, I can do more work to improve the look, including the icon.  Here's a screenshot with the change:

image

@RByers please take a look to see if this is what you had in mind.

Fixes #267


jpchase:
Here's a screenshot of the standalone feature page with the change:

image

RByers:
Screenshots look great, thanks!  A couple minor nits on them:

- In the chromedash-feature element, the "enabled by default" is still a little confusing.  One idea would be to put the status in the heading, eg. "Chromium status: enabled by default".  Then the first line could read "Chrome desktop"
- Could be a separate issue, but can we be consistent with use of icons in the list?  Eg. the 2nd line could have "Chrome Android" text, and we could add Chrome & Opera icons to the relevant lines.
- maybe indent the list below "removed In"?  Or make it a bulletted list with list-style-position:inside ?

jpchase:
Agreed that the "enabled by default" is confusing. I wasn't sure how much we wanted to change things, but I like your suggestion.

I work on that and the other nits.

RByers:
Code LGTM.  /cc @mathiasbynens @foolip in case they have feedback on how we're representing Opera.  Big picture I see Opera as having equal footing in the chromium project as browsers like 'Chrome' and 'Android WebView'.  I.e. we have a shared feature consensus / standards process in the chromium project, but then the individual browser products may make different decisions about if/when to ship a feature.

Oh I guess this viewpoint also means the URL should be fixed to be "chromiumstatus.com" <grin>.

mathiasbynens:
> Big picture I see Opera as having equal footing in the chromium project as browsers like 'Chrome' and 'Android WebView'. I.e. we have a shared feature consensus / standards process in the chromium project, but then the individual browser products may make different decisions about if/when to ship a feature.

+1

Based on the screenshots this change looks good to me!

jpchase:
Changes :
- I've made the implementation status a separate row (bolded), with a separate row for the Chrome desktop milestone. I didn't put the status in the header as previously suggested, because I thought it looked a little busy
- I've added icons for the various products in the status section

Here's a screenshot with the changes:

image

jpchase:
Also, I didn't make the suggested changes on the /feature/ page to indent or bullet the list. Looking at the rest of the page, I think it would be better to revisit the overall style/format. For example, the consensus section uses bold labels for each vendor (no bullets). Another question is should this page match the /features/ more?

RByers:
LGTM, thanks!
@ebidel ?

RByers:
Oh one little nit: Android WebView might as well get an Android icon too.

jpchase:
Added the icon for Android Webview. Latest screenshot:

image

RByers:
Thanks, still LGTM

ebidel:
Looks great!

mathiasbynens:
Any idea when this will be deployed?

ebidel:
I'll try to get it out tomorrow. Wanted to batch up a few perf PRs and get those out together.

Issue #307, opened 2016-07-11
Explore removing milestone as the primary way features are listed


ebidel:
Need to brainstorm this more, but it's unclear if sorting features chronologically by milestone is the best way to present the list. chromestatus has evolved over time and ultimately, people just want to drill down into features, see the list of features for a given milestone, etc. 

Sorting them by milestone might not make sense anymore.

jpmedley:
I hope you mean as the default.

Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195
*If an API's not documented it doesn't exist.*

ebidel:
Yes

clelland:
I'll take a look at what I can do in this space; brainstorming ideas are welcome!

ebidel:
Rethinking this a bit for discussion....:

One things about the chronological list is that it's really good for quickly groking what features are in active dev / proposed vs. actually landing. I think anything new may want to preserver that separation. Speculative features are different from launching/ed features. Web devs mostly care about the latter.

Chronological is also useful for seeing "what's new". For example, if the list was alphabetical, you may never see newer features. That seems less useful to me.

jpchase:
I'll agree it seems more useful to group features into categories like speculative, launched/launching, stable, and "what's new". As the list of features continues to grow, it seems less useful to order/scroll the entire list of features by milestone.  Once you're past the upcoming/current/recent milestones, I think it's far more usable to be able to jump directly to an older milestone.

We could keep the list of milestones in the left nav, and add/rework the top of the nav to represent the desired categories ("what's new" and so on).

I would suggest changing the left nav so that clicking on a milestone (or category) would filter the list (a shortcut for typing a milestone filter in the short box). This replaces the behaviour of scrolling to the milestone in a chronological list. That way, users can still answer the question "what's in milestone X", but it allows the list to be sorted/grouped in different ways.

ebidel:
> I would suggest changing the left nav so that clicking on a milestone (or category) would filter the list

Might make sense to make the list a dropdown instead? That way it becomes a filtering control and we can move it near the search box

clelland:
Looking at the data model, I don't think that we have any way to answer the question "what's in milestone X?". A feature could be added behind a flag in version 50, shipped as an origin trial in 53, and stable in 55, and we would only store the latest status.

[...]

PR #334, opened 2016-07-14
Expose deep links from search ui

### LINKED ISSUE #206: Hard to deeplink into a Chrome version, also it is not discoverable via search
It is very hard to quickly get to what was in a version of Chrome.  You can create a filter for a version https://www.chromestatus.com/features#=43, however it is not discoverable.

Developers would like a single URL that they could visit that tells them what was in a version of Chrome and this URL would also be used for search engine discovery (i.e, when a user types 'what is in Chrome 42').

We could ideally also have an XML feed per version of Chrome stable, beta, dev.

NavidZ:
fixes #206

### DIFF
--- a/static/elements/chromedash-metadata.html
+++ b/static/elements/chromedash-metadata.html
@@ -55,7 +55,7 @@
       if (detail) {
         // Came from an internal click.
         this.selected = Polymer.dom(e).localTarget.dataset.version;
-        this.fire('queryChanged', {version: this.selected});
+        this.fire('query-changed', {version: this.selected});
       } else {
         // Called directly (from outside). e is a feature.
         this.selected = e.meta.milestone_str;
--- a/static/elements/chromedash-metadata.html
+++ b/static/elements/chromedash-metadata.html
@@ -55,7 +55,7 @@
       if (detail) {
         // Came from an internal click.
         this.selected = Polymer.dom(e).localTarget.dataset.version;
-        this.fire('milestoneselect', {version: this.selected});
+        this.fire('queryChanged', {version: this.selected});
       } else {
         // Called directly (from outside). e is a feature.
         this.selected = e.meta.milestone_str;
--- a/templates/features.html
+++ b/templates/features.html
@@ -94,7 +94,7 @@ if (location.hash) {
   search.value = location.hash.substr(1);
 }
 
-chromeMetadata.addEventListener('queryChanged', function(e) {
+chromeMetadata.addEventListener('query-changed', function(e) {
   search.value = '=' + e.detail.version;
   var event = new Event('input');
   search.dispatchEvent(event);
--- a/templates/features.html
+++ b/templates/features.html
@@ -96,8 +96,7 @@ if (location.hash) {
 
 chromeMetadata.addEventListener('query-changed', function(e) {
   search.value = '=' + e.detail.version;
-  var event = new Event('input');
-  search.dispatchEvent(event);
+  featureList.filter(search.value);
 });
 
 featureList.addEventListener('filtered', function(e) {
--- a/templates/features.html
+++ b/templates/features.html
@@ -94,6 +94,12 @@ if (location.hash) {
   search.value = location.hash.substr(1);
 }
 
+chromeMetadata.addEventListener('queryChanged', function(e) {
+  search.value = '=' + e.detail.version;
+  var event = new Event('input');
+  search.dispatchEvent(event);
+});
+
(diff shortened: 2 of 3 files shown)


NavidZ:
@RByers @ebidel 
I was wondering if this way of routing the metadata changed is okay or not.
I thought if eventually we change that metadata (left ui bar) to be similar to status.microsoftedge.com would this way of passing parameters out to the search bar is okay?

templates/features.html
```diff
@@ -94,6 +94,12 @@
   search.value = location.hash.substr(1);
 }
 
+chromeMetadata.addEventListener('queryChanged', function(e) {
+  search.value = '=' + e.detail.version;
+  var event = new Event('input');
+  search.dispatchEvent(event);
```
RByers: dispatching a synthetic input event seems like a bit of a hack.  I'd personally just have the two listeners call into a common "update search" function or something.

templates/features.html
```diff
@@ -116,16 +122,9 @@
...
-  if (!e.target.value && e.keyCode != 27) { // ESC
-    featureList.filter();
-    chromeMetadata.selected = null;
-  } else {
-    //if (e.keyCode == 13) {  // Enter
-      featureList.filter(e.target.value);
-    //}
-  }
+search.addEventListener('input', function(e) {
+  featureList.filter(e.target.value);
```
RByers: does this change the UX of typing into the search box at all?  'input' definitely seems better than 'keyup' (eg. I think it'll also handle paste).   What's the implication of setting `chromeMetadata.selected=null` on every key press (where it looks like we used to only do that when the input field was empty)?

NavidZ: I believe it was deselecting the left hand side options. Why only when input is empty I'm not sure. But there are some problems with the previous logic anyway. Sometime when you start typing it just sets the left hand side to "No Active Development". I assume because it goes to else and filter function somewhat sets that.

NavidZ: Regarding the UX change, I didn't notice any changes.

static/elements/chromedash-metadata.html
```diff
@@ -55,7 +55,7 @@
       if (detail) {
         // Came from an internal click.
         this.selected = Polymer.dom(e).localTarget.dataset.version;
-        this.fire('milestoneselect', {version: this.selected});
+        this.fire('queryChanged', {version: this.selected});
```
ebidel: prefer - case for events to be more consistent with native events.

`query-changed`

ebidel: You'll also want to remove https://github.com/GoogleChrome/chromium-dashboard/blob/master/static/elements/chromedash-featurelist.html#L242.

templates/features.html
ebidel: `chromeMetadata.selected = null;` is because once you start filtering, having a milestone selected makes it look like you're filtering results for that milestone. Which would be a lie :)

templates/features.html
NavidZ: But this is inside a template. So I assume it shouldn't manipulate anything directly in the context it was put it to make things clear. So just assuming there is a global function to call might not be ideal. Or maybe you have something else in mind?

NavidZ:
I removed the observer as well as changing the event name.
@ebidel right now featurelist seems to have a reference to search field as well as the metadata. Maybe if we finally want to have all the criteria be represented as a search query metadata should only change search box and search box changes featurelist removing the connection between featurelist and metadata altogether. Do you see any problem with this approach?

ebidel:
We still need the link between metadata and the feature list for the scroll position. When you scroll the list of features, the milestone gets a little marker next to it. If we remove that behavior in https://github.com/GoogleChrome/chromium-dashboard/issues/307, then yes, we can remove the metadata reference altogether.

templates/features.html
```diff
@@ -94,6 +94,12 @@
   search.value = location.hash.substr(1);
 }
 
+chromeMetadata.addEventListener('query-changed', function(e) {
+  search.value = '=' + e.detail.version;
+  var event = new Event('input');
```
ebidel: this does feel a bit hacky to me. The `input` listener doesn't up the change to `search.value`?

If no, just call `featureList.filter(search.value);` instead of firing the event. I don't see much value in arbitrating through the search box.

NavidZ: I was expecting that to dispatch. But I guess it doesn't when the value is programmatically changed. I will change it to filter it directly.

ebidel:
LGTM

PR #340, opened 2016-07-14
Eliminate hiding of citation link on feature edit form

mdittmer:
We now have cases where we wish to indicate no public signals, but provide a link. Fixes #340


ebidel:
LGTM

Issue #309, opened 2016-07-11
Track key perf metrics from the wild


RByers:
Want at least:
- Time to first render (using navigationTiming APIs presumably)
- Time to feature list being displayed
- Scroll start time

Reported back through Google Analytics.

PR #343, opened 2016-07-15
Update google analytics support

RByers:
Migrate from legacy ga.js to analytics.js (for #309)
As per [GA docs](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs).


ebidel:
👍

PR #344, opened 2016-07-15
Add analytics for features shown

RByers:
Diff on top of #343.  How do I submit a PR on top of another PR?

### DIFF
--- a/templates/features.html
+++ b/templates/features.html
@@ -88,6 +88,7 @@ var featureList = $('chromedash-featurelist');
 var chromeMetadata = $('chromedash-metadata');
 var overlay = $('chromedash-legend');
 var search = $('.search input');
+var unveilStart;
 
 // Set search box to URL deep link.
 if (location.hash) {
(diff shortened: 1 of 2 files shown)


RByers:
Note that this is following the [GA user timing docs](https://developers.google.com/analytics/devguides/collection/analyticsjs/user-timings).

templates/features.html
```diff
@@ -98,6 +99,12 @@
   $('#num-features').textContent = e.detail.count;
 });
 
+featureList.addEventListener('unveiled', function() {
+  console.timeStamp && console.timeStamp('features unveiled');
+  console.timeEnd && console.timeEnd('features unveil');
+  unveilStart && ga('send', 'timing', 'features', 'unveil', performance.now() - unveilStart);
```
ebidel: we may want to protect against ios6. IIRC, `window.performance` is defined but `'window.performance.now` isn't :( Believe they shut it off.

RByers: Ugh, good point - thanks.

RByers: Done

ebidel:
LGTM

Issue #352, opened 2016-07-20
Perf ideas


ebidel:
- [x]  lazy load new dropdown UI elements.
- [ ]  css containment on certain components
- [x]  lazy load wc.js
- [x]  use async imports if possible.
- [x]  autoprefixer for css. Slim down files!
- [x]  minify html/css/js output from vulcanize.
- [x]  use fewer `paper-*` material elements. They add weight to the page.

PR #355, opened 2016-07-20
Slim css

ebidel:
R: @jpchase @jeffposnick @beaufortfrancois @RByers 

Removes built css from git and relies on autoprefixer for vendor prefixes. Part of #352


jpchase:
LGTM.

Issue #329, opened 2016-07-13
Add a SW for offline/LoFi support?


RByers:
The data seems easily cachable, but there's some server-side templating right?  Does it make sense to try to have some sort of offline (or more important lofi) experience?  @ebidel @drufball WDYT?

If we do this, then we might as well also add a manifest to make it a PWA (not many would add it to their homescreen, but I would!).

RByers:
If we wanted a great offline experience, perhaps we could rewrite the server-side templating to be client side?

ebidel:
+1 on making a PWA. Always meant to, just never had the time.

I don't think anything will change regarding the SSR of templates. The app basically falls under the app shell approach atm and it's not a SPA. That's prime for service worker caching.

ebidel:
Fixed

PR #357, opened 2016-07-21
Add a service worker for offline support

jpchase:
For #329, this change adds a service worker to enable offline support:
- Uses sw-precache to generate the worker script, and pull down the app shell on install
- Update the base template to register the service worker on all pages
- Runtime caching to allow the /features/ pages to work offline

This is a first cut. There is more work required to support the other pages, and some files aren't being handled as expected.

For now all of the caching configuration is specified in the gulp task to generate the service worker. The Web Starter Kit uses the approach of importing a script file specify the runtime caching. Any thoughts on which approach is preferred (other than the latter better allows for custom handlers)?

Please take a look.


app.yaml
```diff
@@ -31,6 +31,14 @@ handlers:
   static_files: static/robots.txt
   upload: static/robots\.txt
 
+- url: /service-worker\.js
+  static_files: static/dist/service-worker.js
+  upload: static/dist/service-worker\.js
+  expiration: "0m"
+  secure: always
+  http_headers:
+    Vary: Accept-Encoding
```
ebidel: Not familiar with needing this. Can you explain?

gulpfile.babel.js
```diff
@@ -93,10 +98,80 @@ gulp.task('default', ['clean'], cb =>
...
+      // Scripts
+      `${staticDir}/bower_components/webcomponentsjs/webcomponents-lite.min.js`,
+      `${staticDir}/js/**/*.js`,
+      // Styles
+      `${staticDir}/css/**/*.css`,
+    ],
+    runtimeCaching: [
+      // Server-side generated content
+      {
+        urlPattern: /\/features\/\w+$/,
```
ebidel: The deep links like https://www.chromestatus.com/features/5663174342737920 are actually all client-side. I'm not sure we need to cache those. Just the /features page itself.

gulpfile.babel.js
```diff
@@ -93,10 +98,80 @@ gulp.task('default', ['clean'], cb =>
...
+      },
+      // For dynamic data (json), try the network first to get the most recent
+      // values.
+      {
+        urlPattern: /\/features.json$/,
+        handler: 'networkFirst',
+      },
+      {
+        urlPattern: /\/omaha_data$/,
+        handler: 'networkFirst',
```
ebidel: remove trailing , on these

gulpfile.babel.js
```diff
@@ -93,10 +98,80 @@ gulp.task('default', ['clean'], cb =>
...
+      },
+      // Polymer elements that may or may not be vulcanized
+      {
+        urlPattern: /\/static\/(bower_components|elements)\//,
+        handler: 'cacheFirst',
```
ebidel: remove trailing ,

ebidel: I think we should use networkFirst. https://github.com/PolymerElements/platinum-sw#cacheonly--cachefirst-defaultcachestrategy-considered-harmful

templates/base.html
```diff
@@ -83,6 +83,8 @@
   {% include "footer.html" %}
 </div>
 
+<script src="/static/js/service-worker-registration.js"></script>
```
ebidel: add `async`

static/js/service-worker-registration.js
```diff
@@ -0,0 +1,56 @@
...
+      // 127.0.0.1/8 is considered localhost for IPv4.
+      window.location.hostname.match(
+        /^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/
+      )
+    );
+
+  if ('serviceWorker' in navigator &&
+      (window.location.protocol === 'https:' || isLocalhost)) {
+    navigator.serviceWorker.register('/service-worker.js')
+    .then(function(registration) {
```
ebidel: indentation

ebidel: Do we need all this other stuff? Or are you stubbing it for later?

ebidel:
I'm fine with the globbing for now (instead of importing the generated cache list from the list of resources).

app.yaml
jpchase: I don't think it's needed. I copied from an example, and meant to go back and clean it up. I'll remove it.

gulpfile.babel.js
jpchase: Agreed. I still have to tweak this entry to get just the one page cached. There were other files getting cached incorrectly when just using `/features` as the regex.

templates/base.html
jpchase: Will do.

jeffposnick:
Thanks for putting this together! I'll take a look, too.

gulpfile.babel.js
jeffposnick: The safest way to deal with anything local and static is to make sure that it's being picked up by `sw-precache`'s `staticFileGlobs`, and then you don't need to specify it at all in the `runtimeCaching` section.

static/js/service-worker-registration.js
```diff
@@ -0,0 +1,56 @@
+/* eslint-env browser */
+(function() {
+  'use strict';
+
+  // Check to make sure service workers are supported in the current browser,
```
jeffposnick: `chromestatus.com` forcibly redirects to HTTPS, so there shouldn't be a need to implement this check.

static/js/service-worker-registration.js
jeffposnick: The only reason to keep it around is if we do want to display a "Something's updated! Please refresh." toast.

gulpfile.babel.js
```diff
@@ -93,10 +98,80 @@ gulp.task('default', ['clean'], cb =>
     'lint',
     'vulcanize',
     'scripts',
+    'generate-service-worker',
     cb
   )
 );
 
+// Copy over the scripts that are used in importScripts as part of the generate-service-worker task.
+gulp.task('copy-sw-scripts', () => {
```
jeffposnick: You don't need to do this: using `runtimeCaching` will take care of inlining the `sw-toolbox` code.

jeffposnick:
In an optimal setup, we'd be able to using the [`dynamicUrlToDependencies`](https://github.com/GoogleChrome/sw-precache#dynamicurltodependencies-objectstringarraystring) option to specify all the local files that uniquely define the contents of the top-level server-rendered URLs, like `/features`, and then be able to safely serve those cache-first. That would offer the most optimal experience, since as opposed to using a runtime caching strategy like `fastest`, clients would never have to request an update the HTML unless something actually changes.

Unfortunately, I just took a look at the different dependencies that uniquely define `/features`, and in addition to a bunch of partial/base templates (which `dynamicUrlToDependencies` doe know how to deal with), the contents are also uniquely determined by the [state of the App Engine datastore](https://github.com/GoogleChrome/chromium-dashboard/blob/master/server.py#L156). Since we can't tell at build time when the App Engine datastore has changed, `sw-precache` wouldn't be able to assign a unique hash fingerprint to `/features`, breaking the update model that `sw-precache` uses. (Hopefully that makes sense! Happy to explain in more detail.)

The summary being, I don't think we'll be able to get to the ideal, cache-first-with-hash-fingerprints service worker implementation for the dynamic HTML that the Chromium Dashboard uses. There are still some advantages of adding in a service worker with a runtime caching strategy for the HTML, like providing offline functionality. But the performance and data-saving wins won't be as great as if we could go cache-first for the navigation HTML.

static/js/service-worker-registration.js
jpchase: Good point, I'll remove it.

gulpfile.babel.js
jpchase: I left the task in for now, pending the final location for the runtime caching config (in the Gulp file vs an imported script).  I'll clean it up once the runtime caching is finalized.

ebidel:
@jeffposnick, the dashboard doesn't release that often and our assets are fairly small. WDYT about having a setup where we nuke the cache on every release? Users (re)download more than they need, but it would be an offline-first approach.

gulpfile.babel.js
jpchase: Yes, it would be my preference to use sw-precache for all the included Polymer elements. However, I'm concerned about the size of the content - how much is reasonable? Adding the vulcanized files to sw-precache, adds over 1MB to be fetched on install (total precache size 117kB -> 1.5MB, as in the see the output below).

We could:
- Precache only feature-imports.vulcanized.*, as this is part of the main view, and will have to be downloaded/cached anyway to work offline
- Use runtime caching, strategy = fastest, for the other static elements.

Partial output from sw-precache for vulcanized content:
[11:57:36] Caching static resource "static/elements/admin-imports.vulcanize.html" (39.92 kB)
[11:57:36] Caching static resource "static/elements/admin-imports.vulcanize.js" (255.38 kB)
[11:57:36] Caching static resource "static/elements/features-imports.vulcanize.html" (74.05 kB)
[11:57:36] Caching static resource "static/elements/features-imports.vulcanize.js" (303.58 kB)
[11:57:36] Caching static resource "static/elements/metrics-imports.vulcanize.html" (124.71 kB)
[11:57:36] Caching static resource "static/elements/metrics-imports.vulcanize.js" (264.49 kB)
[11:57:36] Caching static resource "static/elements/samples-imports.vulcanize.html" (47.89 kB)
[11:57:36] Caching static resource "static/elements/samples-imports.vulcanize.js" (270.76 kB)
[11:57:36] Total precache size is about 1.5 MB for 36 resources.

ebidel: The latest output is a about 942KB after my slimming PR.
```
-rw-r--r--  1 ericbidelman  eng    10K Jul 22 09:57 static/elements/admin-imports.vulcanize.html
-rw-r--r--  1 ericbidelman  eng   161K Jul 22 09:57 static/elements/admin-imports.vulcanize.js
-rw-r--r--  1 ericbidelman  eng   3.9K Jul 22 09:57 static/elements/chromedash-legend.vulcanize.html
-rw-r--r--  1 ericbidelman  eng   664B Jul 22 09:58 static/elements/chromedash-legend.vulcanize.js
-rw-r--r--  1 ericbidelman  eng    47K Jul 22 09:57 static/elements/features-imports.vulcanize.html
-rw-r--r--  1 ericbidelman  eng   208K Jul 22 09:58 static/elements/features-imports.vulcanize.js
-rw-r--r--  1 ericbidelman  eng    22K Jul 22 09:57 static/elements/metrics-imports.vulcanize.html
-rw-r--r--  1 ericbidelman  eng   173K Jul 22 09:58 static/elements/metrics-imports.vulcanize.js
-rw-r--r--  1 ericbidelman  eng    13K Jul 22 09:57 static/elements/paper-menu-button.vulcanize.html
-rw-r--r--  1 ericbidelman  eng    86K Jul 22 09:58 static/elements/paper-menu-button.vulcanize.js
-rw-r--r--  1 ericbidelman  eng    32K Jul 22 09:57 static/elements/samples-imports.vulcanize.html
-rw-r--r--  1 ericbidelman  eng   186K Jul 22 09:58 static/elements/samples-imports.vulcanize.js
```

IMO, 1M to cache an entire site (esp these days) is not a lot. Many sites are much greater than that. I have more work to do to slim down the vulcanized bundles, but we can also remove the `admin` imports. You have to be online to use those pages anyway. That'll save about 300KB right off the bat.

jeffposnick:
All of the static assets can be versioned properly by `sw-precache` and we don't have to do anything special to get that part working efficiently.

The HTML contents of at least `/features` (if not other HTML pages as well) depend on something outside of our release cycle—the current state of the App Engine datastore. The App Engine datastore could be updated at any time, outside of any formal release process.

Changing that would be a larger re-architecture, so just using a (less than optimal, but by no means awful) `runtimeCaching` strategy like `fastest` or `networkFirst` for URLs corresponding to server-generated HTML is probably the best bet for the initial implementation.

gulpfile.babel.js
jeffposnick: The problem with using a runtime strategy like `fastest` is that each and every time you request a resource, you're going to fire off an additional network request, and that will end up using bandwidth. (That network request may or may not be fulfilled by the HTTP cache, depending on the standard caching headers.)

Using the precaching and versioning strategy employed by `sw-precache`, you pay the upfront cost once, and then don't fire off any additional network requests unless the service worker detects that there's actually a change. So it ends up being more bandwidth-conscious over time.

Removing unlikely-to-be-needed bundles, like the admin bundle, definitely make sense. And as with any caching strategy, the more fine-grained the resources, the less likely that a large bundle will need to be invalidated due to a small change.

jpchase:
I've addressed various comments in the latest commits. Beyond the minor cleanups, the summary:
- All vulcanized files are pre-cached, with the exception of the admin pages
- Runtime caching is used for the server-side pages, strategy = fastest
- Runtime caching is used for data/json files, strategy = network first

The features, samples and metrics are now working offline. I couldn't fully test the metrics timeline pages due to lack of data.

ebidel:
Hey @jpchase can you rebase against master?

jpchase:
I've merged with the latest from master.

jpchase:
@ebidel Does this PR look good to merge? I don't see any outstanding comments (and Github says there are no conflicts).

ebidel:
Sorry, didn't realize this was back in my court. I want to test locally before merging, but otherwise early LGTM. Will merge if everything looks good.

ebidel:
LGTM. Works great!

PR #362, opened 2016-07-22
Remove use of extraneous elements. 

ebidel:
R: @jeffposnick @jpchase @beaufortfrancois @RByers 

A bit of code cleanup as well.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -62,17 +62,17 @@ gulp.task('styles', () => {
     .pipe(gulp.dest('static/css'));
 });
 
-// Minify the vulcanized script files
-gulp.task('scripts', () => {
-  // The script files are overwritten (as they are generated anyways)
-  //  - The 'base' option is needed to retain the relative path of each file,
-  //    allow overwriting
-  return gulp.src(
-    'static/elements/*.vulcanize.js', {base: 'static'}
-  )
-    .pipe(uglifyJS())
-    .pipe(gulp.dest('static'));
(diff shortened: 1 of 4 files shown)


gulpfile.babel.js
```diff
@@ -62,17 +62,17 @@ gulp.task('styles', () => {
...
-  // The script files are overwritten (as they are generated anyways)
-  //  - The 'base' option is needed to retain the relative path of each file,
-  //    allow overwriting
-  return gulp.src(
-    'static/elements/*.vulcanize.js', {base: 'static'}
-  )
-    .pipe(uglifyJS())
-    .pipe(gulp.dest('static'));
-});
+// // Minify the vulcanized script files
```
ebidel: Don't need this task for now since the vulcanize task takes care of minifying its own js.

jeffposnick:
You're using `fetch()` a few places here. I don't think it's the first time it's used in the codebase, but I feel like I need to point out it's [not universally supported yet](http://caniuse.com/#search=fetch). If the thinking is "It's really unlikely that anyone using the admin interface will be using Safari." then I don't have a major objection.

Alternatively, you could always pull in one of the [`fetch()` polyfills](https://github.com/github/fetch).

jeffposnick:
Other than the `fetch()` comment, 👍

ebidel:
Exactly, my thinking was that anyone hitting the /admin pages is a chromium/blink user and therefore, `fetch()` is good to go!

PR #363, opened 2016-07-25
Remove paper-button and paper-card usage

ebidel:
R: @jeffposnick @jpchase @beaufortfrancois @RByers 

Part of #352

### DIFF
--- a/templates/features.html
+++ b/templates/features.html
@@ -97,15 +97,6 @@ if (location.hash) {
   search.value = location.hash.substr(1);
 }
 
-// $('.actionlinks .blue-button').addEventListener('click', function(e) {
-//   e.preventDefault();
-
-//   // Wait for ripple to finish.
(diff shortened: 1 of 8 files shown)


templates/features.html
```diff
@@ -103,6 +97,15 @@
   search.value = location.hash.substr(1);
 }
 
+// $('.actionlinks .blue-button').addEventListener('click', function(e) {
```
beaufortfrancois: Why keep the comment there?

beaufortfrancois:
Should we remove `paper-button` as well in https://github.com/GoogleChrome/chromium-dashboard/blob/nobutton/static/sass/features/features.scss#L36?

LGTM with nit.

templates/features.html
ebidel: Done

ebidel:
Done x2

PR #368, opened 2016-07-26
Use useNativeCSSProperties

ebidel:
R: @jeffposnick @jpchase @RByers 

This reduces script time by ~150ms. Woot!

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -57,7 +57,11 @@
 {% block preload %}{% endblock %}
 
 <script>
-window.Polymer = window.Polymer || {dom: 'shadow', lazyRegister: true};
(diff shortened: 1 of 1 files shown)


jeffposnick:
You obviously know much better than I do, but will everything Polymer-y play nicely once this is flipped? https://github.com/Polymer/polymer/issues/3474 and https://github.com/Polymer/polymer/issues/3339 are both still open.

If that's not an issue for you, then 👍

ebidel:
This has already been flipped :) https://github.com/Polymer/polymer/issues/3339 is this and we're unaffected by https://github.com/Polymer/polymer/issues/3474/.

PR #381, opened 2016-08-06
Metrics

ebidel:
R: @jeffposnick @RByers @beaufortfrancois 

- Fixes issue where features.json was not actually being preloaded by template
- Adds `preconnect` and `dsn-prefetch` headers for certain resources
- Removes let/arrow funcs due to lack of Safari support :( Now works in  Safari again.
- Uses new polymer 1.6.1 `lazyRegister: "max"` for more page load boost
- Adds a POC metrics reporting lib that reports to GA using the User Timing API.
- Adds Babel support to build process.
- Inlines css on /features
- Uses `async` imports on /features to boost first render.


jeffposnick:
If you're adding Babel support, and you're using the `es2015` pugin, then shouldn't the arrow functions/`let` get transpiled to ES5, and therefore be okay from Safari's perspective? Or is it tough to get that to work in the inline `<script>` tags?

ebidel:
That's right. I didn't move out the inline scripts in the template pages so those will still be untranspiled for Safari. Eventually, I want to combine all the scripts per page (since we're not a SPA), but that's left for a later exercise.

jeffposnick:
👍

PR #382, opened 2016-08-08
Fetch data from UMA server

holte:
This change will cause feature usage data to get fetched from an export server which can query the data, rather than reading from the JSON logs that we intend to stop producing.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -42,95 +42,62 @@ import common
 import models
 import settings
 
-
-# uma.googleplex.com/data/histograms/ids-chrome-histograms.txt
-BIGSTORE_BUCKET = '/uma-dashboards/'
-BIGSTORE_RESTFUL_URI = 'https://uma-dashboards.storage.googleapis.com/'
+UMA_QUERY_SERVER = 'https://uma-export.appspot.com/chromestatus/'
 
 HISTOGRAMS_URL = 'https://chromium.googlesource.com/chromium/src/+/master/' \
     'tools/metrics/histograms/histograms.xml?format=TEXT'
 
-CSSPROPERITES_BS_HISTOGRAM_ID = str(0xbfd59b316a6c31f1)
-ANIMATIONPROPS_BS_HISTOGRAM_ID = str(0xbee14b73f4fdde73)
-FEATURE_OBSERVER_BS_HISTOGRAM_ID = str(0x2e44945129413683)
-
-PAGE_VISITS_BUCKET_ID = 52
-
-# For fetching files from the production BigStore during development.
-OAUTH2_CREDENTIALS_FILENAME = os.path.join(
-    settings.ROOT_DIR, 'scripts', 'oauth2.data')
+# Path to a local cookie to supply when running dev_appserver
+# Go to https://uma-export.appspot.com/login and login then copy the
+# value of SACSID from browser and write this to a cookie file.
+COOKIE_FILENAME = os.path.join(settings.ROOT_DIR, 'cookie')
 
 
-class YesterdayHandler(blobstore_handlers.BlobstoreDownloadHandler):
-  """Loads yesterday's UMA data from BigStore."""
+def _FetchWithCookie(url):
+  if settings.PROD:
+    return urlfetch.fetch(url)
+  try:
+    with open(COOKIE_FILENAME, 'r') as f:
+      cookie = f.readline()
+  except IOError, e:
+    logging.error(e)
+  return urlfetch.fetch(
+      url, headers={'cookie': 'SACSID=' + cookie}, deadline=60)
 
-  MODEL_CLASS = {
-    CSSPROPERITES_BS_HISTOGRAM_ID: models.StableInstance,
-    ANIMATIONPROPS_BS_HISTOGRAM_ID: models.AnimatedProperty,
-    FEATURE_OBSERVER_BS_HISTOGRAM_ID: models.FeatureObserver,
-  }
 
-  def _SaveData(self, data, yesterday, histogram_id):
-    try:
-      model_class = self.MODEL_CLASS[histogram_id]
-    except Exception, e:
-      logging.error('Invalid CSS property bucket id used: %s' % histogram_id)
-      return
+class UmaQuery(object):
+  """Reads and stores stats from UMA."""
 
-    # Response format is "bucket-bucket+1=hits".
-    # Example: 10-11=2175995,11-12=56635467,12-13=2432539420
-    #values_list = data['kTempHistograms'][CSSPROPERITES_BS_HISTOGRAM_ID]['b'].split(',')
-    values_list = data['kTempHistograms'][histogram_id]['b'].split(',')
+  def __init__(self, query_name, model_class, property_map_class):
+    self.query_name = query_name
+    self.model_class = model_class
+    self.property_map_class = property_map_class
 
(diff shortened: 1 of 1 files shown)


ebidel:
This is great. I'll give it a look!

How will the app have access to the endpoint? Right now I get 403 if I hit the constructed URL directly.

holte:
Right now the request is only allowed if it comes from:

a) An admin of the uma-export cloud project
b) google.com:chromestatus app (from X-Appengine-Inbound-Appid)

admin.py
```diff
@@ -42,183 +42,99 @@
...
-          day_percentage=(num_hits * 1.0 / total_pages)
+          day_percentage=rate
           #rolling_percentage=
           )
       entity.put()
 
+  def FetchAndSaveDate(self, date):
+    data, response_code = self._FetchData(date)
+    if response_code == 200:
+      self._SaveData(data, yesterday, histogram_id)
```
ebidel: yesterday -> date

ebidel: `histogram_id` looks left over?

admin.py
```diff
@@ -42,183 +42,99 @@
...
+          date=date,
           #hits=num_hits,
           #total_pages=total_pages,
-          day_percentage=(num_hits * 1.0 / total_pages)
+          day_percentage=rate
           #rolling_percentage=
           )
       entity.put()
 
+  def FetchAndSaveDate(self, date):
```
ebidel: `FetchAndSaveDate` -> `FetchAndSaveData`

ebidel:
Any ideas on how I should go about testing this? I believe GAE only uses the cron for the default app. It would be nice to test this flow before switching it to prod.

holte:
I think you could make a request directly to cron handler, and trigger it
that way, though that would still be modifying the same datastore as your
product.  Better would be if you have a separate cloud project for staging,
I could also whitelist that one.  I'll look and see if there is a way I can
get the right authorization provided from a dev_appserver

holte:
Ok, I added some logic that will read a cookie from file and provide that on requests when running from local dev_appserver.py.  You just need to copy the cookie value from devtools after visiting https://uma-export.appspot.com/login and save that to the cookie file.  That should let you test it with dev_appserver.py

I haven't been able to get dev_appserver.py to work for me due to some package conflicts with appengine though, so I haven't tested it myself.

ebidel:
Hey @holte, https://uma-export.appspot.com/login is 404ing for me.

holte:
If you get a 404, then you've probably logged in.  You should still be able
to grab your cookies from DevTools > Application > cookies

ebidel:
Got it!

admin.py
```diff
@@ -42,183 +42,115 @@
...
-    for val in values_list:
-      bucket_range, hits_string = val.split('=') # e.g. "10-11=2175995"
+  def _FetchData(self, date):
+    params = '?end_date=%s&day_count=1' % date.strftime('%Y%m%d')
+    result = _FetchWithCookie(UMA_QUERY_SERVER + self.query_name + params)
 
-      parts = bucket_range.split('-')
+    if (result.status_code != 200):
+      logging.error('Unable to retrieve histograms data.')
+      return [None, 404]
```
ebidel: `(None, 404)` to be consistent with below

admin.py
```diff
@@ -42,183 +42,115 @@
...
 
-      beginning_range = int(parts[0])
-      end_range = int(parts[1])
+    json_content = result.content.split('\n', 1)[1]
```
ebidel: Let's do:

```
    json_content = result.content.split('\n', 1)[1]
    j = json.loads(json_content)
    if not j.has_key('r'):
      logging.error(
          '%s results do not have an "r" key in the response' % self.query_name)
      return (None, 404)

    return (j['r'], result.status_code)
```

holte:
Fixed the issue with featureobserver.cssproperties handler.

admin.py
```diff
@@ -42,183 +42,115 @@
...
 
+  def FetchAndSaveData(self, date):
+    data, response_code = self._FetchData(date)
+    if response_code == 200:
+      self._SaveData(data, date)
+    return response_code
+
+
+UMA_QUERIES = [
+  UmaQuery(query='featureobserver',
```
holte: Done

admin.py
holte: Done

admin.py
holte: Done

admin.py
```diff
@@ -42,183 +42,115 @@
...
+    logging.error(e)
+  return urlfetch.fetch(url, headers={'cookie': 'SACSID=' + cookie})
 
-    # Response format is "bucket-bucket+1=hits".
-    # Example: 10-11=2175995,11-12=56635467,12-13=2432539420
-    #values_list = data['kTempHistograms'][CSSPROPERITES_BS_HISTOGRAM_ID]['b'].split(',')
-    values_list = data['kTempHistograms'][histogram_id]['b'].split(',')
 
-    #sum_total = int(data['kTempHistograms'][CSSPROPERITES_BS_HISTOGRAM_ID]['s']) # TODO: use this.
+class UmaQuery(Object):
```
holte: Done

admin.py
```diff
@@ -42,183 +42,115 @@
...
-      return
+def _FetchWithCookie(self, url):
+  if settings.PROD:
+    return urlfetch.fetch(url)
+  try:
+    with open(COOKIE_FILENAME, 'r') as f:
+      cookie = f.readline()
+  except IOError, e:
+    logging.error(e)
+  return urlfetch.fetch(url, headers={'cookie': 'SACSID=' + cookie})
```
holte: Done

admin.py
```diff
@@ -42,183 +42,115 @@
...
-    FEATURE_OBSERVER_BS_HISTOGRAM_ID: models.FeatureObserver,
-  }
 
-  def _SaveData(self, data, yesterday, histogram_id):
-    try:
-      model_class = self.MODEL_CLASS[histogram_id]
-    except Exception, e:
-      logging.error('Invalid CSS property bucket id used: %s' % histogram_id)
-      return
+def _FetchWithCookie(self, url):
```
holte: Done

admin.py
```diff
@@ -42,183 +42,121 @@
...
-      del properties_dict[PAGE_VISITS_BUCKET_ID]
-    else:
-      total_pages = sum(properties_dict.values())
+  def _SaveData(self, data, date):
+    property_map = self.property_map_class.get_all()
 
-    property_map = models.CssPropertyHistogram.get_all()
-    if histogram_id == FEATURE_OBSERVER_BS_HISTOGRAM_ID:
-      property_map = models.FeatureObserverHistogram.get_all()
+    for bucket_str, rate in data:
```
ebidel: `for bucket_str, rate in data.iteritems():`

ebidel:
Also confirmed `featureobserver.cssproperties` is pulling properly.

admin.py
holte: Done

ebidel:
@holte 

Seeing 403s from the server:
screen shot 2016-09-20 at 11 42 30 am

holte:
It looks like the app_id on those request are 's~cr-status' where the
export server was expecting 'google.com:chromestatus'.  Are both of those
app_id's for chromestatus or should I just make the export server only
accept the  's~cr-status'  one?

PR #385, opened 2016-08-10
Add toast for SW

ebidel:
R: @jeffposnick 

- Adds a toast for sw precaching and update cycle
- Includes the amount of data that was precached. I think this is pretty cool and we're reporting this number to Analytics over time to watch if assets grow/shrink.
- All the deferred/promise crazy is to delay with async imports loading + delaying the service worker registration on /features (so the toast doesnt come up until after the page is loaded). I was observing a janky toast (similar to IOWA).

screen shot 2016-08-10 at 4 08 58 pm


jeffposnick:
It's nice showing the precache size in the Toast, though at that point the data's already been transferred, so it's not like the user has a choice. 😁 

I'm wondering how useful it's going to be to report back the `sw-precache` size via GA for each visitor after the service worker installs. It's not like the size will vary from user agent to user agent, so I'm assuming you'd just be spamming GA with the same value over and over. Unless there's some cool way of visualizing the change in the value over time for a given unique user...

templates/base.html
```diff
@@ -62,6 +62,19 @@
 var $$ = function(selector) {
   return document.querySelectorAll(selector);
 };
+
+// Note: this version requires Promises, which are in all the
+// latest browsers except IE 11. Edge will ship them.
+// Note: native Promise.defer() is deprecated. Have to create our own.
+function deferred() {
```
jeffposnick: `defer()` is an anti-pattern:
- http://www.datchley.name/promise-patterns-anti-patterns/#3deferredantipattern
- https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern

I know `defer()`/`deferred()` snuck into the IOWA code base, but it probably shouldn't have. I think you can safely ditch it following the examples in those links.

ebidel:
I'll remove the deffered.

For the reporting to GA, I was thinking it may be interesting to see the cache size move around over time. We could visualize it using the GA tools.

templates/base.html
ebidel: Done

ebidel:
PTAL

jeffposnick:
👍

PR #387, opened 2016-08-11
Fixes #386

### LINKED ISSUE #386: Add manifest.json file

ebidel:
R: @beaufortfrancois @jeffposnick @RByers


templates/base.html
```diff
@@ -22,6 +22,10 @@
 <title>{% block page_title %}{% endblock %}{{ APP_TITLE }}</title>
 <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
 
+<link rel="manifest" href="/static/manifest.json">
+
+<meta name="theme-color" content="#standalone">
```
beaufortfrancois: Shouldn't it be `content="#366597"`?

static/manifest.json
```diff
@@ -0,0 +1,34 @@
+
+{
+  "name": "Chrome Platform Status",
+  "short_name": "Chrome status",
+  "start_url": "/features?utm_source=web_app_manifest",
+  "display": "standalone",
+  "theme_color": "#366597",
+  "background_color": "#366597",
+  "icons": [{
+    "src": "/static/img/chromium-128.png",
```
beaufortfrancois: Why chromium and not chrome there?

templates/base.html
beaufortfrancois: How about adding `<link rel="icon" sizes="192x192" href="/static/img/chrome_logo_192.png">` as well as mentioned in https://developers.google.com/web/updates/2014/11/Support-for-theme-color-in-Chrome-39-for-Android?

beaufortfrancois:
LGTM but nit

templates/base.html
ebidel: Lolz, copy and paste error. Done

static/manifest.json
ebidel: I need to switch these over from the chrome logo, just didn't have time to make the assets. Using the chrome logo makes it look like the chrome browser when it's on the homescreen :( We need something diff.

ebidel:
PTAL. There's a baller new logo!

ebidel:
+ @jeffposnick for the timezone shift.

jeffposnick:
Nice logo skillz. 👍

tophf:
clip120-fs8

Favicon in tab strip now looks like a unintelligible blob whereas the old one on the right was clear and legible. White background with a shadow as an *integral part* of an icon/logo, seriously? Please tell me it's just an unintended copypaste of the icon from a promo/presentation. If a shadow and circular background are suddenly needed somewhere on site just do it in-place via css.

ebidel:
It's best we don't reuse existing product icons. The new logo works better for add to homescreen users.

tophf:
I see the difference. What I'm talking about is favicon: if you increase the three logos and remove the white circle it'll be quite legible at 16x16px

jpmedley:
I was considering raising this issue myself.

Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195

tophf:
BTW regarding:

> new logo works better for add to homescreen users

My homescreen theme already has placeholders for icons, and there are lots of different themes where such background doesn't look good because it creates an unneeded visual noise: theme placeholder, your embedded placeholder, the icon itself. There are also themes without any placeholders, so your white blob makes the icon look like it's a folder that has 3 items inside.

PR #388, opened 2016-08-11
Trim down what is precached.

ebidel:

We were precaching more than needed. Also inlines some resources on other pages.


jeffposnick:
👍

PR #392, opened 2016-08-22
Escape characters with special meaning in regexps

### LINKED ISSUE #391: Escape special characters in search regular expressions
Our documentation (via example queries) suggests keyword search with light label:value and math>=expr matching. It does not suggest regular expressions. In light of this, the current behaviour of constructing a RegExp from the search string is confusing, particularly considering that searching for issues with names containing square brackets will yield no results.

For this reason I think we should escape characters with special meaning in regular expressions. If we wish to support regular rexpressions, I would suggest something like:

    matches:"[a-zA-Z]*[1-3]"

or

    /[a-zA-Z]*[1-3]/

mdittmer:
Closes #391

### DIFF
--- a/static/elements/chromedash-featurelist.html
+++ b/static/elements/chromedash-featurelist.html
@@ -90,15 +90,23 @@
         this.unlisten(window, 'scroll', '_onScrollList');
       },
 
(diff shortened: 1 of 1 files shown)


ebidel:
LGTM

PR #399, opened 2016-08-24
Redirect to feature's page when it is saved/updated

### LINKED ISSUE #371: Usage Metrics dropdown doesnt show in FF
Probably something to do with lazy loading :\

ebidel:
R: @RByers @jeffposnick @beaufortfrancois @johnmellor @dtapuska 

Also Fixes #371.

Also adds toast pop to let users know their feature was saved and it will be a few minutes before it shows up in the main list:

screen shot 2016-08-24 at 3 21 41 pm


jeffposnick:
👍

PR #400, opened 2016-08-27
Fixes #379

### LINKED ISSUE #379: Features without desktop shipping milestone are showing under Chrome 1
example is the network api:
https://www.chromestatus.com/features/6338383617982464

It doenst have a desktop milestone but it does have an android milestone. Need to look into this.

ebidel:

This also adds an additional layer of memcaching  that should have a small perf boost in the response time of features.json

### DIFF
--- a/models.py
+++ b/models.py
@@ -372,7 +372,8 @@ class Feature(DictModel):
       # milestone is defined, that will take default.
       shipping_features = map(getSortingMilestone, shipping_features)
 
-      # TODO: sort on name after sorting on milestone.
+      # First sort by name, then sort by feature milestone (latest first).
+      shipping_features.sort(key=lambda f: f.name, reverse=False)
       shipping_features.sort(key=lambda f: f._sort_by_milestone, reverse=True)
 
       # Constructor the proper ordering.
--- a/models.py
+++ b/models.py
@@ -229,9 +229,14 @@ class Feature(DictModel):
     d['impl_status_chrome'] = IMPLEMENTATION_STATUS[self.impl_status_chrome]
     d['meta'] = {
       'origintrial': self.impl_status_chrome == ORIGIN_TRIAL,
-      'needsflag': self.impl_status_chrome == BEHIND_A_FLAG,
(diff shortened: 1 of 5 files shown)


models.py
```diff
@@ -323,32 +328,58 @@ def get_chronological(self, limit=None, update_cache=False):
...
+        feature._sort_by_milestone = (feature.shipped_milestone or
+                                      feature.shipped_android_milestone)
+        return feature
+
+      # Sort the feature list on either Android shipping milestone or desktop
+      # shipping milestone, depending on which is specified. If a desktop
+      # milestone is defined, that will take default.
+      shipping_features = map(getSortingMilestone, shipping_features)
+
+      # TODO: sort on name after sorting on milestone.
```
jeffposnick: Any reason not to add the code to sort by name into this PR? You might be able to just put together a synthetic sort key that combines the version, a separator string, and then the name, and keep most of the code the same:

```python
shipping_features.sort(key=lambda f: "%s$$$%s" % (f._sort_by_milestone, f.name), reverse=True)
```

jeffposnick:
Just the comment on the sorting by name TODO. If it's not important to add to this PR, 👍

ebidel:
Unfortunately, I don't think it's that easy :\ We need two different sorting orders: first sort on the version (`reverse=True`) so latest is first, then a sort on the name (`reverse=False`) so they're alphabetical.

Example:
```
 [
'52$$$Xsomething',
'50$$$Desktop Ship 50',
'49$$$Deprecated 49',
'1$$$Chrome 1',
'52$$$Android shipping 52',
'40$$$Behind a flag 40',
'33$$$Android shipping  33'
].sort(reverse=True)
```

=>`['52$$$Xsomething', '52$$$Android shipping 52', '50$$$Desktop Ship 50', '49$$$Deprecated 49', '40$$$Behind a flag 40', '33$$$Android shipping  33', '1$$$Chrome 1'`

The first two are out of order.

ebidel:
Found a solution. PTAL.

jeffposnick:
👍

PR #404, opened 2016-09-01
URI-encode/decode search string that appears in URL hash component

### LINKED ISSUE #398: Search hash needs escaping (eg. spaces)
Type a search term that includes a space, eg. "Event Listeners".  It works, but you get a URL like `https://www.chromestatus.com/features#Event Listeners` which I think isn't valid (it works but most things don't expect spaces in URLs - will replace with %20 which then doesn't work).  Eg. see #397

mdittmer:
Closes #398.


ebidel:
LGTM

PR #405, opened 2016-09-01
Cache templates, other UI/perf adjustments

ebidel:
R: @jeffposnick  @mdittmer  @RByers 

- Perf: Adds some caching to server templates
- Perf: adds css containment to several components to reduce layout/paint work
- aligns implementation status text 

screen shot 2016-09-01 at 1 44 05 pm


samccone:
lgtm :+1:

PR #418, opened 2016-10-04
Fix release/impl status links

### LINKED ISSUE #409: Milestone filtering picks up numbers in any field
Clicking on "54 beta" in the current sidebar does a search for `=54`. `=54`, when used as a search term, should only pick up features with a Milestone of 54. But it looks like the filtering is being applied to all fields, so instances of 54 in any property cause the feature to be included.

Also, when you expand the feature that's been included in the list, the "No active development" version is highlighted, rather than either the version corresponding to the actual feature (51, in this case) or the version corresponding to the search (54).

image

mdittmer:
Fix the format of the query injected when clicking sidebar links associated with release numbers and `impl_status_chrome`. Fixes #409.

### DIFF
--- a/templates/features.html
+++ b/templates/features.html
@@ -117,9 +117,10 @@ var featuresUnveilMetric = new Metric('features_unveil');
 var featuresFetchMetric = new Metric('features_loaded');
 
 chromeMetadata.addEventListener('query-changed', function(e) {
-  var propertyName = e.detail.version.match(/^[0-9]+$/) ? 'milestone' :
-      'impl_status_chrome';
-  search.value = propertyName + '=' + e.detail.version;
+  var value = e.detail.version;
+  var isMilestone = value.match(/^[0-9]+$/);
+  search.value = isMilestone ? 'milestone=' + value :
+      'impl_status_chrome:"' + value + '"';
   featureList.filter(search.value);
 });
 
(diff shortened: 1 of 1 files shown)


templates/features.html
```diff
@@ -222,6 +225,11 @@
     resolve();
 
     featuresPromise.then(function(features) {
+      features.map(function(feature) {
+      feature.milestone = feature.shipped_milestone || feature.shipped_android_milestone ||
```
ebidel: indentation

ebidel: Actually what is this for? I don't see where it is used.

templates/features.html
```diff
@@ -117,7 +117,10 @@
 var featuresFetchMetric = new Metric('features_loaded');
 
 chromeMetadata.addEventListener('query-changed', function(e) {
-  search.value = '=' + e.detail.version;
+  var value = e.detail.version;
+  var isMilestone = value.match(/^[0-9]+$/);
+  search.value = isMilestone ? 'milestone=' + value :
```
mdittmer: `milestone` property computed on line `229` used here.

templates/features.html
mdittmer: Fixed.

templates/features.html
ebidel: Ah sorry. I may have been looking at commit view instead of all changes. I see it now.

ebidel:
LGTM

PR #422, opened 2016-10-10
Add survey banner

beaufortfrancois:
Here's a patch that adds a simple survey banner as you can see in the screenshots below.

@jeffposnick  @PaulKinlan 

screenshot 2016-10-10 at 10 07 05 pm

localhost-8080-features nexus 5x

### DIFF
--- a/static/sass/_vars.scss
+++ b/static/sass/_vars.scss
@@ -21,7 +21,8 @@ $content-padding: 16px;
 $default-border-radius: 3px;
 
 $app-drawer-width: 200px;
-$header-height: 202px;
+$header-height: 135px;
+$banner-height: 67px;
 $max-content-width: 768px;
 /* ----- */
 
--- a/static/sass/_vars.scss
+++ b/static/sass/_vars.scss
@@ -21,6 +21,7 @@ $content-padding: 16px;
 $default-border-radius: 3px;
 
 $app-drawer-width: 200px;
+$header-height: 202px;
 $max-content-width: 768px;
 /* ----- */
 
--- a/templates/base.html
+++ b/templates/base.html
@@ -99,6 +99,13 @@ var $$ = function(selector) {
     <div class="main-toolbar">
       <div class="toolbar-content">
         {% include "header.html" %}
+        {% block banner %}
+        <div id="banner">
+          <p>
+            Help us learn about how you want to learn about the latest Chrome features by filling out this very short <a href="https://docs.google.com/a/google.com/forms/d/e/1FAIpQLSd5lL2cRXOnxYZF8qaPykLsU6-1_p4M_1PatwGaIwL3eVcOMA/viewform?c=0&w=1" target="_blank">survey</a>.
+          </p>
+        </div>
+        {% endblock %}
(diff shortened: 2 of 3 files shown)


templates/base.html
```diff
@@ -99,6 +99,13 @@
     <div class="main-toolbar">
       <div class="toolbar-content">
         {% include "header.html" %}
+        {% block banner %}
+        <div id="banner">
+          <p>
+            Help us learn about how you want to learn about the latest Chrome features by filling out this very short <a href="https://docs.google.com/a/google.com/forms/d/e/1FAIpQLSd5lL2cRXOnxYZF8qaPykLsU6-1_p4M_1PatwGaIwL3eVcOMA/viewform?c=0&w=1" target="_blank">survey</a>.
```
ebidel: Help us improve this site by filling out this....

jeffposnick:
I think the `<app-drawer>` (the stuff on the left-hand side) positioning needs to be fiddled with to account for the extra space that the banner is taking up.

image

It's currently using a `top: -120px` and only when it gets down to around `top: -58px` or so is it fully displayed without being cut off, based on experimenting in Chrome's Style DevTools.

image

The `bottom: -120px` is problematic too, though it's less obvious that the bottom is being cut off.

I'll poke around a bit to figure out where those values are coming from, unless you know off the top of your head.

jeffposnick:
Looks like those values [come from](https://github.com/PolymerElements/app-layout/blob/master/app-drawer/app-drawer.html#L58) the `<app-drawer>` element itself.

jeffposnick:
Looks like https://github.com/GoogleChrome/chromium-dashboard/blob/master/static/sass/main.scss has a few places with a hardcoded value of `135px`, representing the height of the header. This is used to calculate the top margin for the `<app-drawer>` content, among other things.

@beaufortfrancois, can you refactor out the references to `135px` in that file into a SASS variable and then update it to take into account the new height of the header?

beaufortfrancois:
Done! Please merge if that looks good for both of you.

static/sass/_vars.scss
```diff
@@ -21,6 +21,7 @@ $content-padding: 16px;
 $default-border-radius: 3px;
 
 $app-drawer-width: 200px;
+$header-height: 202px;
```
ebidel: `135px`

beaufortfrancois: With the banner it is 202px, not 135px.

beaufortfrancois: Do you want me to set it to 135px and add 67px each time?

ebidel: Yea, I think `$header-height: 135px`; is less confusing. Then add another var for the banner? 

Also what happens when we don't have a banner? This will need to account for that.

beaufortfrancois: Done!

beaufortfrancois:
How does that look now?

ebidel:
LGTM

beaufortfrancois:
Can you push this live?

ebidel:
done

PR #424, opened 2016-10-16
Scroll/reveal header for mobile

ebidel:
R: @jeffposnick @beaufortfrancois @GoogleChrome/chromium-dashboard 

We currently have a lot of wasted space. The main content area is the most important thing, and since the last UI refresh, the header takes up much more space than it should. This PR moves the drawer content to the top left corner and makes the header scroll away and reveal itself again on scroll. For mobile, this creates a lot more space to view the features, samples, and other content. 

update: the footer text has been centered, so ignore that in the new screenshots.

BEFORE:

screen shot 2016-10-16 at 11 09 23 am

AFTER:
screen shot 2016-10-16 at 11 08 38 am

screen shot 2016-10-16 at 11 36 34 am

### DIFF
--- a/static/sass/main.scss
+++ b/static/sass/main.scss
@@ -277,6 +277,9 @@ footer {
       max-width: $max-content-width;
     }
   }
+  footer {
+    padding-left: $app-drawer-width;
+  }
 }
 
 @media only screen and (max-width: 700px) {
--- a/static/sass/main.scss
+++ b/static/sass/main.scss
@@ -262,7 +262,7 @@ footer {
   margin: $content-padding;
   position: relative;
   height: 100%;
-  margin-top: $header-height;
+  // margin-top: $header-height;
 }
 
 #panels {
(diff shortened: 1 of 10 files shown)


beaufortfrancois:
If @PaulKinlan is fine with removing survey, I'm good.

static/sass/main.scss
```diff
@@ -241,14 +241,14 @@ footer {
...
 
-#banner {
-  display: flex;
-  margin: $content-padding 0;
-  max-width: $max-content-width;
-  background: #fff3e0;
-  font-weight: 500;
-  padding: $content-padding $content-padding / 2;
-}
+// #banner {
```
beaufortfrancois: Please remove entirely if we don't need it anymore

static/sass/main.scss
```diff
@@ -262,7 +262,7 @@ footer {
   margin: $content-padding;
   position: relative;
   height: 100%;
-  margin-top: $header-height + $banner-height;
```
beaufortfrancois: Ditto.

ebidel: Done

static/sass/main.scss
ebidel: Done

ebidel:
I spoke with Dru about the banner and he said take it down by Monday :)

beaufortfrancois:
We're good then :)

PR #425, opened 2016-10-20
Adding polymer summit video

ebidel:
R: @beaufortfrancois @jeffposnick @mdittmer @RByers 

-Also makes footer span entire page on desktop

screen shot 2016-10-19 at 5 39 14 pm


beaufortfrancois:
When window innerWidth is 653px, footer get in this weird state (see below).
Notice how the "Chrome Platform Status" is also merged into the "Chrome logo" at the top.

screenshot 2016-10-20 at 9 26 22 am

ebidel:
Yep, the inbetween breakpoint weirdness isn't new. I've noticed that before and we should fix it, but this PR is unrelated :)

ebidel:
Done

beaufortfrancois:
LGTM

PR #433, opened 2016-11-16
Faster, slimmer, simplified data model

### LINKED ISSUE #321: Simplify features.json schema
https://www.chromestatus.com/features.json has grown out of control as we've added stuff over the years. Here's a proposal for a new schema that:
1. consolidates browser-specific info under "browsers"
2. uses small keys (smaller response!)
3. This restructuring means downstream clients (ie status page etc) do less fix up and can discard entire sub objects they don't care about. 

Additionally, we could remove `null` fields from the server's response. This will make the response even smaller #perfmatters.

We should also version the URL (e.g. `features_v2.json`) so consumers can migrate at their own pace.

```
{
  "id": 6021353392046080,
  "name": "CSS Flexbox: New intrinsic size algorithm",
  "category": "CSS",
  "summary": "The CSS Flexbox specification changed the algorithm for calculating a flexbox's intrinsic size (see below for the spec link...",
  "created": "2015-08-05 22:40:46.897000",
  "comments": "",
  "standards": {
    "spec": "https://drafts.csswg.org/css-flexbox/#intrinsic-sizes",
    "status": { // was standards_str
      "text": "Editor's draft", "val": 4,
    },
    "visibility": "Only a very small number of web developers will care about",
    "footprint": 4,
  },
  "updated": {
    "when": "2015-08-05 22:40:46.897540", // was updated
    "by": "cbiesinger@chromium.org" // was created_by
  }
  "resources:": {
    "samples": [], // was sample_links
    "docs": [], // was doc_links
  },
  "tags": [ // was search_tags
    "Flexbox", "flex"
  ],
  "browsers": {
    "chrome": {
      "bug": "http://crbug.com/240765",
      "owners": ["cbiesinger@chromium.org"], // was owner
      "experiment": false, // was experimentalframework
      "prefixed": false,
      "flag": false,
      "first_of_milestone": true,
      "status": {"text": "No active development", // was impl_status_chrome, "val": 1},
      "desktop": null, // was shipped_milestone
      "android": null, // was shipped_android_milestone
      "webview": null, // was shipped_webview_milestone
      "ios": null // was shipped_ios_milestone
    },
    "opera": {
      "desktop": null,
      "android": null, // was shipped_opera_android_milestone
    }
    "ff": {
      "view": {"text": "Public support", "val": 3, "url": "https://bugzilla.mozilla.org/show_bug.cgi?id=1055887" /* was ff_views_link */}
    },
    "edge": {
      "view": {"text": "No public signals", "val": 5, "url": null /* was ie_views_link */}
    }
    "safari": {
      "view": {"text": "No public signals", "val": 5, "url": null /* was safari_views_link */}
    },
    "webdev": {
      "view": {"text": "No signals", "val": 5}
    }
  }
}
```

ebidel:
R: @jeffposnick @mdittmer 

This consolidates the data model as discussed in #321 while not making drastic changes.

In this PR,`/features` will now pull from a new endpoint `/features_v2.json` ([try it](https://2016-08-26-newmodel-ssr2-dot-cr-status.appspot.com/features_v2.json)). This reduces the current 722 feature download by 14.5% (1.1MB -> 941KB). As we continue to add more features, the reduced bytes will be even better over time :) The old endpoint, `/features.json` is preserved so downstream clients (MS dashboard, etc.) can continue to use that for the foreseeable future. No changes for them.

FWIW, I had some other SSR experimentation in here. I'd suggest reviewing the PR from the "files changed" view and not by commit.

@mdittmer can you take a look at the searching/filtering stuff? I think I got it all.

Fixes #321

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,5 +1,5 @@
 application: cr-status
-version: 2016-08-26-newmodel-ssr2
+version: newmodel-ssr
 runtime: python27
 threadsafe: true
 api_version: 1
--- a/app.yaml
+++ b/app.yaml
@@ -1,5 +1,5 @@
 application: cr-status
-version: 2016-08-26-newmodel-ssr
+version: 2016-08-26-newmodel-ssr2
 runtime: python27
 threadsafe: true
 api_version: 1
--- a/app.yaml
+++ b/app.yaml
@@ -1,5 +1,5 @@
 application: cr-status
-version: 2016-10-27
+version: 2016-08-26-newmodel-ssr
 runtime: python27
 threadsafe: true
 api_version: 1
--- a/server.py
+++ b/server.py
@@ -74,6 +74,8 @@ def annotate_first_of_milestones(feature_list, version=None):
   try:
     omaha_data = get_omaha_data()
 
+    # JSON data is an array of chrome versions for each platform. Search the
+    # array for the first `version` we find for windows, and break out.
     win_versions = omaha_data[0]['versions']
     for v in win_versions:
       s = v.get('version') or v.get('prev_version')
@@ -257,7 +259,7 @@ class FeaturesAPIHandler(common.JSONHandler):
     if version is None:
       version = 1
     else:
-      version = int(version[2:])
+      version = int(version)
 
     KEY = '%s|v%s|all' % (models.Feature.DEFAULT_MEMCACHE_KEY, version)
     feature_list = memcache.get(KEY)
@@ -269,7 +271,7 @@ class FeaturesAPIHandler(common.JSONHandler):
 
 # Main URL routes.
 routes = [
-  ('/features(_v[0-9]*)?.json', FeaturesAPIHandler),
+  (r'/features(?:_v(\d+))?.json', FeaturesAPIHandler),
   ('/(.*)/([0-9]*)', MainHandler),
   ('/(.*)', MainHandler),
 ]
--- a/server.py
+++ b/server.py
@@ -74,13 +74,11 @@ def annotate_first_of_milestones(feature_list, version=None):
   try:
     omaha_data = get_omaha_data()
 
-    # JSON data is an array of chrome versions for each platform. Search the
-    # array for the first `version` we find for windows, and break out.
     win_versions = omaha_data[0]['versions']
-    for v in win_versions:
-      s = v.get('version') or v.get('prev_version')
-      LATEST_VERSION = int(s.split('.')[0])
-      break
+
+    # Find the latest canary major version from the list of windows versions.
+    canary_versions = [x for x in win_versions if x.get('channel') and x.get('channel').startswith('canary')]
+    LATEST_VERSION = int(canary_versions[0].get('version').split('.')[0])
 
     # TODO(ericbidelman) - memcache this calculation as part of models.py
     milestones = range(1, LATEST_VERSION + 1)
--- a/server.py
+++ b/server.py
@@ -45,53 +45,64 @@ def first_of_milestone(feature_list, milestone, start=0):
       return i
   return -1
 
+def first_of_milestone_v2(feature_list, milestone, start=0):
+  for i in xrange(start, len(feature_list)):
+    f = feature_list[i]
+    desktop_shipped_milestone = f['browsers']['chrome']['desktop']
+    if (str(desktop_shipped_milestone) == str(milestone) or
+        f['browsers']['chrome']['status']['text'] == str(milestone)):
+      return i
+    elif (desktop_shipped_milestone == None and
+          str(f['browsers']['chrome']['android']) == str(milestone)):
+      return i
+  return -1
+
+def get_omaha_data():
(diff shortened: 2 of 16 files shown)


app.yaml
```diff
@@ -1,5 +1,5 @@
 application: cr-status
-version: 2016-10-27
+version: 2016-08-26-newmodel-ssr2
```
jeffposnick: I know this is just an arbitrary identifier, but do you want to update it to a more accurate date?

server.py
```diff
@@ -43,55 +43,69 @@ def first_of_milestone(feature_list, milestone, start=0):
...
+
+def annotate_first_of_milestones(feature_list, version=None):
+  try:
+    omaha_data = get_omaha_data()
+
+    win_versions = omaha_data[0]['versions']
+    for v in win_versions:
+      s = v.get('version') or v.get('prev_version')
+      LATEST_VERSION = int(s.split('.')[0])
+      break
```
jeffposnick: I don't understand the logic of a `for` loop that has an unconditional `break` as its last statement.

I realize that this code isn't new and was present in the earlier revision, but I'm taking a look at it for the first time, so bear with me 😄

server.py
```diff
@@ -246,8 +246,30 @@ def get(self, path, feature_id=None):
...
+  def __get_feature_list(self, version=None):
+    feature_list = models.Feature.get_chronological(version=version) # Memcached
+    annotate_first_of_milestones(feature_list, version=version)
+    return feature_list
+
+  def get(self, version=None):
+    if version is None:
+      version = 1
+    else:
+      version = int(version[2:])
```
jeffposnick: Could you include a comment explaining what the expected value for `version` is going to be? It's not clear why you're unconditionally discarding the first two characters of the string here, so seeing an example would help.

server.py
```diff
@@ -246,8 +246,30 @@ def get(self, path, feature_id=None):
...
+    feature_list = memcache.get(KEY)
+    if feature_list is None:
+      feature_list = self.__get_feature_list(version)
+      memcache.set(KEY, feature_list)
+    return common.JSONHandler.get(self, feature_list, formatted=True)
+
+
 # Main URL routes.
 routes = [
+  ('/features(_v[0-9]*)?.json', FeaturesAPIHandler),
```
jeffposnick: I guess this feeds into the previous comment I had, an explains why you want to discard the first two characters of the `version` value.

How about changing the regular expression to exclude the `_v` from the capture group and ensure that there's always a digit whenever there's a `_v`?

```py
(r'/features(?:_v(\d+))?.json', FeaturesAPIHandler)
```

app.yaml
ebidel: done

server.py
ebidel: Better! Thank you. 

Done

ebidel:
PTAL

server.py
jeffposnick: It seems weird to use a `for` loop with an unconditional break for that purpose. Unless I'm missing something, you only look at the first item and then exit the loop.

Isn't this equivalent?

```py
v = win_versions[0]
s = v.get('version') or v.get('prev_version')
LATEST_VERSION = int(s.split('.')[0])
```

Or is there some Python magic that happens with `for ... in ...` iteration that doesn't happen if you just look at the first item?

ebidel: `win_versions[0]` is an array. 

Looks like they've changed the data format a bit since I last looked at this. I've gone ahead and simplified how the latest version is found by using a list comprehension to find the canary channels in the array, then using the first's `version`.

static/elements/chromedash-legend.html
```diff
@@ -42,11 +42,11 @@
     <section>
       <label>Example search queries</label>
       <ul class="queries">
-        <li><span>"shipped_milestone&lt;30"</span>features that landed before 30</li>
-        <li><span>"shipped_android_milestone&gt;40"</span>features that landed on Anrdoid after 40 (milestone types: android, ios, webview, opera, opera_android)</li>
-        <li><span>"shipped_milestone&lt;=50"</span>features in desktop chrome 50</li>
+        <li><span>"browsers.chrome.desktop&lt;30"</span>features that landed before 30</li>
```
mdittmer: This mechanism makes for very tight query parsing code, but I wonder whether it's the UX we want. Personally, I'm divided on the issue of "browsers.foo.bar":

Pros: Easy parsing, very precise, parsimonious if referring to other parts of the data model.

Cons: Expecting user to know about/use our data model, strange to qualify that the browser is Chrome on chromestatus.com, more verbose.

One option would be to support both; keep a map of "special" key shortcuts, and fallback on data-model-walking.

@ebidel and @jeffposnick, thoughts?

ebidel: Agreed. From the UX perspective, requiring folks to map our data model to search queries is friendly. I would be for supporting a more friendly shorthand. Would you be interested in helping with that after this PR lands?

mdittmer:
Sure. I can help out. Otherwise LGTM.

PR #443, opened 2017-01-14
Distinguished Chrome for iOS from other versions

phistuck:
Some entries creep in that mention that a feature is supported on Chrome for iOS when it does not. Instead of removing it completely, since a few features are apparently polyfilled by Chrome for iOS, the fields were re-ordered a bit to have it as the last Chrome-related field (right behind the Opera section) and the help text was prepended with some wording to make it more noticable that it is Chrome for iOS and that it usually does not include those features.

### DIFF
--- a/models.py
+++ b/models.py
@@ -641,6 +641,9 @@ class FeatureForm(forms.Form):
                       '(either enabled by default, origin trial, intervention, '
                       'or deprecated)')
 
+  SHIPPED_IOS_HELP_TXT = ('Fill this only in case this feature is relevant to Chrome for iOS '
+                         '(most are not)!' + SHIPPED_HELP_TXT)
+
   #name = PlaceholderCharField(required=True, placeholder='Feature name')
   name = forms.CharField(required=True, label='Feature')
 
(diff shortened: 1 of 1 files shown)


ebidel:
I would argue that the current help text is sufficient: "First milestone the feature shipped with this status (either enabled by default, origin trial, intervention, or deprecated)".

If the feature isn't shipping on Chrome iOS, then they shouldn't be entering values. If you're seeing a lot mistakes, perhaps we can keep the current text, but append "(please double check that this feature is relevant to iOS. Some are not)"

phistuck:
The problem is that people often read the beginning instead of the end. Putting this in the beginning (unfortunately, not in the absolute beginning, because the name of the field is prepended) might make it stand out more.

I do not really mind either way, as long as people stop putting false values there.
I am not really sure this field is needed at all. The vast majority of features is irrelevant to Chrome for iOS and we do not update it when Safari gains support for a feature (which makes Chrome for iOS gain support for it, if you have that iOS version).

phistuck:
Do you want me to change the text to your suggestion?

ebidel:
Yea, I'd prefer the same lingo but with the appended "(Note: please make sure this feature is relevant to iOS)."

jrobbins:
This PR seems obsolete or redundant.  Somewhere along the line, the text "(RARE)" was added to the field-level help that we show for iOS milestones.

Issue #439, opened 2016-12-13
Remove the red highlighting for usage <0.03%


foolip:
0.03% isn't mentioned in http://www.chromium.org/blink and it's not a rule that things with higher usage can't be removed. Since we have to judge the usage differently depending on what kind of thing it is (attribute, method, code path in parser, etc.) it's just as well to not make 0.03% special on chromestatus.com.

FYI @chrishtr @dglazkov @jeisinger @RByers @tkent-google

RByers:
Yeah I agree having a special "cut-off" value has caused more harm than good (0.03% of page views is still a LOT of page views, yet people tend to assume that it's fine to break anything below that without further consideration / mitigations.

/cc @mdittmer

PR #447, opened 2017-02-03
Remove the 0.03% "at risk" cut-off

foolip:
0.03% has never been a hard rule, and over the past few years we
haven't given it much weight at all.

No more red!

Fixes https://github.com/GoogleChrome/chromium-dashboard/issues/439

Drive-by: fix typo formmatted -> formatted


ebidel:
LGTM

foolip:
Thank you, Eric! As a memento, here is how it was before:
fireshot capture 14 - chrome platform sta_ - https___www chromestatus com_metrics_feature_popularity

PR #488, opened 2017-08-01
Fixes #487 - http -> https for link rel alternate urls

### LINKED ISSUE #487: link rel alternate for ATOM feeds is an http link

PaulKinlan:
This fix changes the URL that the link rel=alternate points to to be https.... selfishly I needed it so I could build Chromestatus deck.... but I think it would be good.

Question: It doesn't auto redirect existing requests to https, should it?

### DIFF
--- a/templates/feature.html
+++ b/templates/feature.html
@@ -15,7 +15,7 @@
 {% endblock %}
 
 {% block rss %}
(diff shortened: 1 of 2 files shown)


ebidel:
Thanks. I think was a reason for now auto redirecting to https (feed readers didn't like it or something) but can't recall now.

ebidel:
LGTM

PR #515, opened 2017-10-10
Change "pages surveyed" to "page views"

foolip:


foolip:
@ebidel I think the old wording was easy to misread as being something more like a crawler.

@RByers FYI.

Issue #279, opened 2016-04-29
Timelines should show data for only the dominant version


RByers:
@foolip mentioned it would be really nice to see milestone markers on the timeline graphs.  On a related note, I've seen people confused before about an apparent huge spike in usage which was really just showing the metric making it to stable channel (before then I assume the timeline shows the beta/dev usage as a function of the total page view on all channels).

To avoid confusion, I suggest that we restrict the timeline to showing only stable channel (we can use the Google-internal uma dashboard when we need to drill further).  If we do that, then also adding stable-release-date milestone markers should be pretty easy / useful too.

@ebidel don't worry, I'm not asking you to do this - as we discussed, I'm working on finding an engineer (and/or some of my own time) to help with the dashboard.

ebidel:
Is there a way to query UMA to return just stable channel results? I'm not sure if it's currently possible with our current setup.

RByers:
It's definitely possible on the internal dashboard (just have a "Channel" filter), but I don't know how that's exposed down in the data.  Maybe we can just show only data for whatever major version is dominant on any given day (it's conflating data across versions that's the fundamental problem)?

ebidel:
I need to look at stuff again, but IIRC, the json data we get from the GCS buckets is across all channels. It's not separated out.

RByers:
Re-titling to describe the issue: people get confused about what the data means when the UseCounter in question hasn't actually been in stable for awhile.  I think the simplest fix is probably to have the UMA pipeline include only stable data, but I don't know the details of how that works and there are other options (eg. ideally you'd be able to see a separate line for dev channel or something).

I keep seeing people get confused by this, let's mark it high priority for whenever predictability team decides to return to doing a bunch of chromestatus work...

ebidel:
@RByers have you raised these questions with the UMA team? If we decide what we need, I'm sure they can craft a dremel query for it. 

I'm a little worried about storing data for multiple channels. Do you think we can return only stable chrome for now? Worry about the other channel lines later?

RByers:
> @RByers have you raised these questions with the UMA team? If we decide what we need, I'm sure they can craft a dremel query for it.

No.  IIRC the only context I have is that thread you drove back from when things broke.

> I'm a little worried about storing data for multiple channels. Do you think we can return only stable chrome for now? Worry about the other channel lines later?

Yes, showing only stable for now to keep it simple and unconfusing was my suggestion.

ebidel:
[...]

PR #521, opened 2017-10-27
Update for new UMA server

RByers:
Update the API used to talk to the UMA server for the new format.  This won't cause any change in behavior, just get off the legacy endpoints/APIs and make it possible to take advantage of the additional information in a future commit.  Corresponds to google internal changes in cl/173205164.

Part of fixing #279

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -78,7 +78,7 @@ class UmaQuery(object):
     self.property_map_class = property_map_class
 
   def _FetchData(self, date):
-    params = '?end_date=%s&day_count=1' % date.strftime('%Y%m%d')
+    params = '?date=%s' % date.strftime('%Y%m%d')
     url = UMA_QUERY_SERVER + self.query_name + params
     result = _FetchWithCookie(url)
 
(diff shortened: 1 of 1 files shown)


RByers:
@ebidel I don't think I can test this myself, but I think these are the right changes to move to the new server API I landed (after you've verified that the cron job doesn't blow up tonight I guess).

admin.py
```diff
@@ -121,7 +121,9 @@ def _SaveData(self, data, date):
           date=date,
           #hits=num_hits,
           #total_pages=total_pages,
-          day_percentage=rate
+          day_percentage=bucket_dict['rate']
```
ebidel: is a `rate` property guaranteed?

RByers: Yes, if a bucket exists in the results at all, it'll be guaranteed to have `rate`, `milestone` and `low_volume` properties.

RByers:
For the record, the Google internal code that this matches is in cl/173205164

MD-hub-lab:
I want to update the Google Chrome

PR #522, opened 2017-11-01
Chart UI updates for new metrics

### LINKED ISSUE #141: chromedash-featurelist on-scroll listener slows down initial load
When the features list first comes in, `scrollToMilestone` is called by chromedash-featurelist. Looking at a profile of the page load, this call takes ~100ms on my Z620. I suspect this is because of the `on-scroll` listener, which repeatedly tries to calculate which milestone to highlight in the left nav.

This could probably be avoided by waiting to add the on-scroll listener until after page load is finished.

ebidel:
R: @RByers 

Also adds a new trendline to the graphs.

screen shot 2017-11-01 at 10 35 18 am

Fixes #141

### DIFF
--- a/static/elements/chromedash-timeline.html
+++ b/static/elements/chromedash-timeline.html
@@ -1,10 +1,9 @@
 <link rel="import" href="../bower_components/polymer/polymer.html">
-<link rel="import" href="../bower_components/google-apis/google-legacy-loader.html">
+<link rel="import" href="google-charts-lib.html">
 
 <dom-module id="chromedash-feature-timeline">
 <link rel="import" type="css" href="../css/elements/chromedash-timeline.css">
 <template>
-  <google-legacy-loader on-api-load="_chartAPILoaded"></google-legacy-loader>
   <select value="{{selectedBucketId::change}}">
     <option disabled value="1">Select a property</option>
     <template is="dom-repeat" items="{{props}}" as="bucket">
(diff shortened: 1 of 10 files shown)


static/elements/chromedash-timeline.html
```diff
@@ -56,21 +55,40 @@
...
       for (var i = 0, item; item = data[i]; ++i) {
         var dateStr = item.date.split('-');
         var date = new Date();
         date.setFullYear(parseInt(dateStr[0]), parseInt(dateStr[1]) - 1, parseInt(dateStr[2]));
-        rowArray.push([
-          date,
-          parseFloat((item.day_percentage * 100).toFixed(6))
-          // parseFloat(item.day_percentage * 100)
-        ]);
+        var row = [date, parseFloat((item.day_percentage * 100).toFixed(6))];
```
RByers: Is the `toFixed` here left over from the precision truncation stuff?  We don't need that anymore - the full precision exposed by the back-end handler is approved to be made public (see go/chromestatus-usecounter).

templates/metrics/css/animated.html
```diff
@@ -30,7 +30,9 @@
   <div>
     <h3>About this data</h3>
     <p class="description">
-    We've been using Chrome's <a href="https://cs.chromium.org/chromium/src/tools/metrics/histograms/enums.xml" target="_blank">anonymous usage statistics</a> to count the CSS properties which are animated. A property that is animated multiple times during a single page view is counted each time it animates. <b>Percentages on this page indicate the amount of times (as the fraction of all animated properties) each property is used in a Transition or Animation</b>. Data is ~24 hrs old and reflects usage across all channels and platforms.</p>
+    We've been using Chrome's <a href="https://cs.chromium.org/chromium/src/tools/metrics/histograms/enums.xml" target="_blank">anonymous usage statistics</a> to count the CSS properties which are animated.
+    A property that is animated multiple times during a single page view is counted each time it animates.
```
RByers: The above line is no longer true, just delete it?  The statement below accurately reflects how it works.

static/elements/chromedash-timeline.html
ebidel: Yes. Cool, I'll remove all the truncation stuff (backend and frontend parts).

ebidel: Actually, we may still need it for the historical data points. The raw metrics are saved and sent to the client truncated.

templates/metrics/css/animated.html
ebidel: Do you mean just keep this sentence:

"Perrcentages on this page indicate ...

static/elements/chromedash-timeline.html
RByers: Ugh. Maybe rely on some helper that does truncation just for data with old dates?

templates/metrics/css/animated.html
RByers: Yes.  Remove line 34 (as it's contradictory to line 35 IMHO).

ebidel: done

static/elements/chromedash-timeline.html
ebidel: Filed https://github.com/GoogleChrome/chromium-dashboard/issues/523 for now to track this. I don't think it hurts to keep the truncation in for now.

ebidel:
PTAL

maryroseMXXXVI:
Hi guys how it's going

PR #537, opened 2018-03-22
Noted the Android WebView metrics inclusion

phistuck:


RByers:
Sorry I didn't see this sooner, but I believe this comment is incorrect.  I [clarified on the list](https://groups.google.com/a/chromium.org/forum/#!topic/blink-api-owners-discuss/CHEFTQ65WEA) - WebView data is NOT currently exposed in chromestatus.  I suggest just reverting this commit.

PR #543, opened 2018-03-29
Make changes in notification emails easier to read.

jpmedley:
Also, do you know where to look to make the notifier stop escaping HTML so that we can actually see the formatting?


jpmedley:
I think they'll appreciate it more when I start editing summary, which I'll be doing a lot more of.

PR #558, opened 2018-06-08
Add disclaimer to 'Consensus & standardization'.

jpmedley:
Let me know what you think.


jpmedley:
How long before this shows up on the live site?

ebidel:
Should be live now.

PR #564, opened 2018-07-11
Improve help text.

jpmedley:


jpmedley:
@ebidel The cause of the Lighthouse error appears to be (educated guess) the 49 occurrences of '<' in mockdata.json. I don't want to escape them before checking with you. Are they safe to change?

models.py
```diff
@@ -887,8 +887,8 @@ class FeatureForm(forms.Form):
       help_text='Capitalize only the first letter and the beginnings of proper nouns.')
 
   summary = forms.CharField(label='', required=True, max_length=500,
-      widget=forms.Textarea(attrs={'cols': 50, 'placeholder': 'Summary description', 'maxlength': 500}),
-      help_text='Complete sentences only. Provide a one sentence description followed by one or two lines explaining how this feature helps web developers.')
+      widget=forms.Textarea(attrs={'cols': 50, 'placeholder': 'Summary description', 'maxlength': 500}),label="Summary",
```
ebidel: Replace the `label=''` on 889 instead of adding another `label`.

models.py
```diff
@@ -887,8 +887,8 @@ class FeatureForm(forms.Form):
...
+      help_text='Complete sentences only. Provide a one sentence description followed by one or two lines explaining how this feature helps web developers. <br/>Note: line breaks are ignored in some views of this content.')
```
ebidel: Did you run this locally? Not sure if the HTML will render as expected and/or be escaped.

PR #586, opened 2019-02-05
Remove Opera from Dashboard

jpmedley:
Eric,

A while back we discussed removing Opera from Chrome Status. If I remember correctly, we agreed that it should go. 

I've started by commenting what I think should be removed. Let me know if I've made any mistakes or omissions. Then I'll remove it for real.

Thanks,
Joe


static/elements/chromedash-feature.html
```diff
@@ -165,7 +165,7 @@ <h3>Chromium status</h3>
               <span>[[feature.browsers.chrome.webview]]</span>
             </span>
           </template>
-          <template is="dom-if" if="[[!feature.browsers.chrome.origintrial]]">
+          <!-- <template is="dom-if" if="[[!feature.browsers.chrome.origintrial]]">
```
ebidel: Think you still want to keep this outer conditional. Just comment out the inner opera bits.

static/elements/chromedash-legend.html
```diff
@@ -44,7 +44,8 @@ <h3>Search</h3>
       <ul class="queries">
         <li><span>"browsers.chrome.desktop&lt;30"</span>features that landed before 30</li>
         <li><span>"browsers.chrome.desktop&lt;=50"</span>features in desktop chrome 50</li>
-        <li><span>"browsers.chrome.android&gt;40"</span>features that landed on Chrome Android after 40 (milestone types: browsers.chrome.android, browsers.chrome.ios, browsers.chrome.webview, browsers.opera.desktop, browsers.opera.android)</li>
+        <!-- <li><span>"browsers.chrome.android&gt;40"</span>features that landed on Chrome Android after 40 (milestone types: browsers.chrome.android, browsers.chrome.ios, browsers.chrome.webview, browsers.opera.desktop, browsers.opera.android)</li> -->
```
ebidel: Keep this. Just remove the opera one.

ebidel:
There's some client stuff in https://github.com/GoogleChrome/chromium-dashboard/blob/master/static/js/admin/feature_form.js.

static/elements/chromedash-feature.html
jpmedley: The inner Opera is the only thing inside this particular tag which is why I its included. The closing template is on line 192.

jpmedley:
PTAL

Should be change the `switch` to an `if` in feature_form.js?

static/elements/chromedash-feature.html
ebidel: I see now. Wasn't reading the diff correctly.

ebidel:
You can leave it a switch. Can you remove the actual code bits?

jpmedley:
can you [look at this line](https://github.com/GoogleChrome/chromium-dashboard/pull/586/files#diff-26a5cefe6eb37e2e7f4f59470166ad52R130). There should be a closing {% endif %} for it, but I'm having trouble seeing which one it is.

jstenback:
Joe, I'm not clear on exactly which line you're referring to in that diff (the link doesn't point to a specific line), but AFAICT that diff as a whole is well balanced in terms of if and endif statements.

jpmedley:
The link points to 130 when I follow it. 

My changes to that file have 4 'if's and 3 'endif's.

jstenback:
Ah, I see, apologies for not noticing that. The issue, FAICT, is that you're leaving an {% if ... %} block in the file inside of an HTML comment. That still likely gets evaluated and leads to a mismatch (see https://stackoverflow.com/questions/719915/how-to-put-comments-in-django-templates). If you'd remove that whole line instead, or use {% #... } for that line, things ought to work. The matching endif is [here](https://github.com/GoogleChrome/chromium-dashboard/pull/586/files#diff-26a5cefe6eb37e2e7f4f59470166ad52L177) in case that helps.

jpmedley:
I'm not leaving it inside a comment. I commented the code initially because we don't have a staging server for this and I wanted Eric to sanity check the work before I proceeded. (Uncommenting is quicker than restoring deleted lines.) 

I see where I messed up now. I had already included the replacement line (which is where the extra 'if' came from). I had forgotten why that other line was there.

jpmedley:
@ebidel I think I'm ready for final review.

jpmedley:
@ebidel Is that Lighthouse error because of something I touched? There's no link that would let me get more information.

(Is this something I'll be able to investigate when I get this running locally?)

jpmedley:
When this is merged, I should do the deployment to see if anything goes wrong.

jpmedley:
I missed something. I just tested locally. Opera is gone from the edit forms, but still appears in status entries. 

Do we want to continue displaying the data we have?
@jstenback @PaulKinlan

ebidel:
Ignore the Lighthouse bot.

If you're running locally and have made changes to the polymer elements (you have), you'll need to run `gulp vulcanize` or set `VULCANIZE = False` in settings.py to see your updates. I recommend running gulp so you don't accidentally deploy unminified web components.

https://github.com/GoogleChrome/chromium-dashboard#debugging--settings

jpmedley:
I ran `gulp vulcanize`. I still see Opera on my Status entries.

static/js/admin/feature_form.js
```diff
@@ -22,36 +22,21 @@ form.addEventListener('change', function(e) {
...
-    return;
-  }
-  const operaVersion = chromeVersion - 13; // e.g. Chrome 32 ~ Opera 19
-  if (!operaDesktop.classList.contains('interacted')) {
-    operaDesktop.value = operaVersion;
-  }
-  if (!operaAndroid.classList.contains('interacted')) {
-    operaAndroid.value = operaVersion;
-  }
+function toggleSpecLink(stdStage) {
```
ebidel: There's linting errors with this. It's never called?

ebidel: Actually, we removed this in https://github.com/GoogleChrome/chromium-dashboard/commit/c4fc8bdd97cb48c6d5bb0c35d9507185ab6e57c5

jpmedley: Done.

package.json
```diff
@@ -22,8 +22,11 @@
     "url": "https://github.com/GoogleChrome/chromium-dashboard/issues"
   },
   "devDependencies": {
+    "@babel/core": "^7.3.4",
```
ebidel: please undo these unrelated changes.

ebidel: please remove these unrelated changes

static/js/admin/feature_form.js
```diff
@@ -22,37 +22,13 @@ form.addEventListener('change', function(e) {
...
-  }
-  const operaVersion = chromeVersion - 13; // e.g. Chrome 32 ~ Opera 19
-  if (!operaDesktop.classList.contains('interacted')) {
-    operaDesktop.value = operaVersion;
-  }
-  if (!operaAndroid.classList.contains('interacted')) {
-    operaAndroid.value = operaVersion;
-  }
-}
+const specLink = document.querySelector('#id_spec_link');
```
ebidel: this can be removed.

Issue #595, opened 2019-02-28
Unable to edit existing feature


dway123:
While trying to edit https://www.chromestatus.com/features/5120155113750528, I'm able to enter an edit page, but when clicking submit, I get an internal server error 500 error.

"500 Internal Server Error
The server has either erred or is incapable of performing the requested operation."

jstenback:
This appears to be due to missing multiline=True arguments to some of the new db.StringProperty() calls when creating the containers for a few of the new values. I have a fix coming...

jstenback:
I went ahead and deployed the proposed PR to hopefully unblock people who are trying to edit features.

jpmedley:
I just got an email from Yoav Weiss confirming that he can edit entries now.

PR #589, opened 2019-02-14
Initial introduction of Blink Launch Process data into Chromestatus.

jstenback:
Add a set of fields to the entry creation/edit form so that we collect
all the information that is part of the blink launch process. This
also makes chromestatus generate up to date intent templates with the
future intent of eliminating the need to manually create and send
intent emails. This includes making feature entries aware of the
current intent stage they're in as well as holding on to all the
additional data so that creating intent messages for later stages of
the launch process is a matter of updating the entry and generating a
new intent message.

PR #596, opened 2019-02-28
Fix editing of entries with multiline values

jstenback:
This fixes an issue (#595) where several of the form fields added in #589 
were not marked as multiline entries in the database (i.e. the form fields
could only contain a single line of text). This changes that, and also fixes
an issue with the Web Platform Tests checkbox couldn't be left unchecked.


ebidel:
BTW, I no longer have write access to the repo so @jpmedley or someone else needs to review. You could also add me back to the repo if you want :)

PR #597, opened 2019-02-28
Expand the length of the free form text fields

jstenback:
This expands the length of the free form text fields from 500 to 1480.
There's a hard maximum of 1500 in the database, and it seems the front
end text field limiting code is buggy and doesn't correctly account for
newlines in the form fields, so I cut this a bit short of the hard limit
to help users more often know that they've reached the limit. This
change also loosens the restriction for the intent to implement link as
users don't always have one at this point.


models.py
```diff
@@ -946,7 +946,7 @@ class FeatureForm(forms.Form):
   name = forms.CharField(required=True, label='Feature',
       help_text='Capitalize only the first letter and the beginnings of proper nouns.')
 
-  summary = forms.CharField(label='', required=True, max_length=500,
```
jpmedley: Do you know exactly what the `max_length` field did? This seems redundant, but is it?

jstenback: All of the reading I did suggested they're redundant, so AFAIK there's no material difference between the two. I went with `maxlength` in the textarea attrs for consistency with the surrounding code, but I don't have a strong feeling either way here.

PR #600, opened 2019-03-01
Update README.md

jpmedley:
There is an `app_version` field but not an `app` field in the `app.yaml` file. I looked at several random deployments throughout the history of  the site and can find now evidence that Eric ever did this or that it ever needs to be done.


ebidel:
`version` used to be a thing in app.yaml. Now you do it at deploy time. LGTM!

PR #603, opened 2019-03-01
Make motivation visible and clarify external content.

jpmedley:


jpmedley:
I tried to investigate the CI problem, but I don't have access to settings.

jstenback:
I believe I was able to re-trigger the ci for this PR, waiting for results...

jstenback:
Looks like travis was happy with this change this time around. I say we merge this...

jpmedley:
I'm seeing it broken again. 

Sidebar: I've always followed the practice of not merging my own material. Do you not do that? If you had seen travis run successfully I would have expected you to merge (and vice-versa).

jstenback:
Hmm, the second CI run is green, which is what matters. The first run, the one that failed, still shows up, but we can safely ignore that one now that we have a newer one that passed. Lighthouse tests is another issue, but I'm under the impression we don't need to block merging on that one.

WRT merging policy, I'm happy either way. To me it doesn't matter who pushes the merge button assuming (enough) tests pass and the change is reviewed.

jpmedley:
Odd. I wasn't see green on the second one. Also issue #612 is related to this push. It's not deployed yet for that reason.

Issue #612, opened 2019-03-07
I'm unable to see some entries


jpmedley:

jpmedley:
@jstenback I sent you an email on this.

jpmedley:
This is a regression in #603

PR #613, opened 2019-03-07
Fix typo in templates/feature.html

jpmedley:
PR #603 had `eindif` instead of `endif` in one of the lines. Addresses #612.


jpmedley:
@jstenback Can you please merge this. I don't have access to do it.

PR #626, opened 2019-03-21
Example of converting chromedash-color-status to use LitElement

justinfagnani:
This isn't working, but just shows how a Polymer 1 element can be converted to a LitElement.


robdodson:
This looks pretty nice to me. I'm assuming we could get rid of the [html file](https://github.com/GoogleChrome/chromium-dashboard/pull/626/files#diff-455d2880e41a1366be59fc93d1ea34d9R1) and switch to using something like rollup or webpack to do the [module/nomodule pattern](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/) to import our stuff?

justinfagnani:
Yeah, this was me trying to be incremental. If you update the imports you can remove the chromedash-color-status.html file, and at the end there are no more .html files in /elements at all.

Since there aren't that many elements, I think the migration would be pretty easy, with the toughest part being infra around npm and builds. It'd be great if in dev there was no build at all.

ebidel:
:+1

robdodson:
giphy

jrobbins:
This PR seems obsolete now.   liyangguang converted our polymer files to LitElement in 2019 and I merged it in early 2020.

PR #640, opened 2019-04-10
Setup rollup.js and convert chromedash-color-status to lit-element

liyangguang:
The page renders correctly in latest Chrome.
Lit-element and Polymer 1.x live together fine, at least locally. (Updated README on how to run it locally).


jpmedley:
@liyangguang Did someone in Developer Relations ask you to do this?

robdodson:
@jpmedley ah yep that was me :) @liyangguang is a 20%'er helping us migrate off of the deprecated web components APIs.

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
Could you add the following block to the `.eslintrc`'s `parserOptions` block.

```
"parserOptions": {
  "ecmaVersion": 8,
  "sourceType": "module"
},
```

I think that'll make it stop complaining about using `import`.

robdodson:
I noticed a number of templates were deleted but wasn't sure why that was?

jpmedley:
@robdodson You said you were going to ask him to make his changes against another branch. We should probably do that in case we need to fix something I production.

Here's a development branch. https://github.com/GoogleChrome/chromium-dashboard/tree/dev

robdodson:
yep @liyangguang would you be ok doing your pull request against the dev branch instead? We can continue to land your commits over there and then switch over when we're ready.

robdodson:
also cc @justinfagnani for an optional quick look at the lit code if he has time. Otherwise I think we can merge it into the dev branch when the questions are answered.

jpmedley:
Gang,

I intended for `dev` to be a long-running branch for the time being. I don't know how long the polymer conversion will take. John's update to using Chrome Status to generate Intents still hits the occasional edge case. We need to preserve the ability to rush small fixes into production until we're sure the polymer conversion is done.

@jstenback @robdodson @liyangguang

robdodson:
@jpchase yep understood. This change, and any future work we do on the conversion will all go to the `dev` branch. We can rebase against `master` to stay in sync and when the time comes, merge `dev` into `master` once we've verified it all looks good.

jpchase:
> @jpchase yep understood. This change, and any future work we do on the conversion will all go to the `dev` branch. We can rebase against `master` to stay in sync and when the time comes, merge `dev` into `master` once we've verified it all looks good.

I think this comment was intended for @jpmedley.

package.json
```diff
@@ -12,7 +12,9 @@
     "lint": "gulp lint",
     "build": "gulp",
     "deploy": "./scripts/deploy_site.sh",
-    "start": "./scripts/start_server.sh"
+    "start": "./scripts/start_server.sh",
+    "build-components": "rollup -c",
+    "watch-components": "rollup -c -w"
```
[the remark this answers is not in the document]
justinfagnani: Ultimately this should be unnecessary during development, I hope. Is there anything in here that depends on being bundled?

package.json
```diff
@@ -47,8 +49,13 @@
     "http2-push-manifest": "^1.0.0",
     "lighthouse-ci": "https://github.com/ebidel/lighthouse-ci",
     "load-grunt-tasks": "^4.0.0",
+    "rollup": "^1.9.0",
+    "rollup-plugin-commonjs": "^9.3.4",
```
justinfagnani: what CJS modules are being used in the client? If lit-element is the only dependency, there should be any.

rollup.config.js
```diff
@@ -0,0 +1,21 @@
...
+
+export default {
+  input: 'static/js/component-bundle-entry.js',
+  output: {
+    file: 'static/js/component-bundle.js',
+    format: 'iife', // immediately-invoked function expression — suitable for <script> tags
+    sourcemap: true
+  },
+  plugins: [
+    resolve(), // tells Rollup how to find date-fns in node_modules
```
justinfagnani: Is date-fns really used? It's not listed in package.json

justinfagnani: date-fns 2.0 publishes modules to npm. How do you feel about pre-releases?

static/elements/chromedash-color-status.mjs
```diff
@@ -0,0 +1,34 @@
+import { LitElement, html } from 'lit-element';
```
justinfagnani: why .mjs? Browsers don't care about the extension.

static/elements/chromedash-color-status.mjs
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  render() {
+    return html`
+      <link rel="stylesheet" href="/static/css/elements/chromedash-color-status.css">
+      <span id="status" style="background-color: ${this._getColor()}"></span>
+    `;
+  }
+
+  _getColor() {
```
justinfagnani: a couple of options here, though this is fine as-is:

1. If `_getColor` is only used in `render()` you could just compute the value directly in `render()`.
2. If this were a very high-frequency, performance sensitive element, and you wanted to avoid calculating this derived value every time, you could memoize it in `update()`:

```js
update(changedProps) {
  if (changedProps.has('value')) {
    this._color = ....
  }
  super.update(changedProps);
}
```

or use the `guard()` directive:

```js
import {guard} from 'lit-html/directives/guard.js';
//...
render() {
  return html`
    <span id="status" style="background-color: ${guard([this.value], () => this._getColor()}"></span>
  `;
}
```

Of course, this isn't an expensive operation here, but I just wanted to point this out in case there is one in the future.

static/elements/chromedash-feature.html
```diff
@@ -1,7 +1,6 @@
 <link rel="import" href="../bower_components/polymer/polymer.html">
 <link rel="import" href="../bower_components/iron-icon/iron-icon.html">
 <link rel="import" href="../bower_components/paper-styles/color.html">
-<link rel="import" href="chromedash-color-status.html">
```
justinfagnani: You're removing the HTML import, but not adding a JS `import` or `<script type="module">`. How are you getting `chromedash-color-status` into here?

package.json
robdodson: I think `date-fns` is being used. It's called out in the rollup.config.

static/elements/chromedash-color-status.mjs
robdodson: I personally don't mind the mjs extension since it conveys that the js is in a module but I'm fine either way.

liyangguang:
> I noticed a number of templates were deleted but wasn't sure why that was?

They got removed when I refresh .gitignore.
We have a gitignore rule /css, which excludes them. I guess it's not intended, changing it to static/css and adding back those template/metrics/css files.

package.json
liyangguang: CJS & date-fns: my bad, forgot to delete after testing the set up.

scripts: (from what I see) the `start` script only brings up the server, but nothing for frontend code refresh. So if I want to get auto compile when developing, I need to have one command window running start_server.sh, and another one running watch-component. (The gulp and grunt both aren't watched, I guess we used to just run it manually after editing?)
Is this the part you are talking about?

rollup.config.js
liyangguang: My bad, not intended to be here.

static/elements/chromedash-color-status.mjs
liyangguang: I saw this practice from https://philipwalton.com/articles/deploying-es2015-code-in-production-today/. But no strong preference at all. I'll just change back to .js

static/elements/chromedash-feature.html
liyangguang: All lit-element components (currently only chromedash-color-status) are bundled by rollup into one .js file, and added to the page base template (base.html and blank.html).

From this and the script comment, I guess your expectation is: modules are not bundled together, and loaded in each template they are being used. Is it?

package.json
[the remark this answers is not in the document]
liyangguang: Good point!

In the current setup, components are bundles into one .js file, added in base.html and blank.html.

static/elements/chromedash-feature.html
justinfagnani: Every file that depends on another should directly import it. If you don't model the dependency graph accurately, code-splitting bundlers will not be able to work correctly.

I would actually recommend that you add Rollup last in this process. You should be able to get very far without bundling anything, since all the current browsers support modules.

liyangguang:
> Gang,
> 
> I intended for `dev` to be a long-running branch for the time being. I don't know how long the polymer conversion will take. John's update to using Chrome Status to generate Intents still hits the occasional edge case. We need to preserve the ability to rush small fixes into production until we're sure the polymer conversion is done.
> 
> @jstenback @robdodson @liyangguang

Just want to point out - from my experiments, lit-element components and polymer 1.x components lives together fine. So if we want, we can migrate components one-by-one and pushing to prod while migrating. So we don't need to do a mega merge after all components are finished, and have to sync the code with any small fixes in master.

justinfagnani:
It's probably going to require more work to go to prod with a combination of Polymer 1 and LitElement. Under polyfills Polymer 1 elements can cause problems for lit-html because they synchronously create child DOM before lit-html gets a chance to process the template instances and record where the dynamic parts are. There are workarounds, but I think everything will work nicer if it's all converted first.

How long do you think that would take? There aren't many elements. I can help out.

justinfagnani:
Also, the build setup will be simpler when you get rid of Bower and only use npm.

static/elements/chromedash-feature.html
liyangguang: Sounds good, thanks!

Can you give me some advice on the process - "add Rollup last in this process"?
My understanding are:
- The usage of rollup in our current setup are: 1) handle import lit-element. 2) combine all components into one .js file. Now it seems 2) is not a good practice.
- Since the components are lit-element, they need to be bundled with lit-element.

So what I should do now are:
MyComponent (using lit-element) => bundle with lit-element, export MyComponentBundled. Same for all components. But how should I do this bundling? (If I use rollup/webpack, they always do the (2) above, right?)

liyangguang:
> It's probably going to require more work to go to prod with a combination of Polymer 1 and LitElement. Under polyfills Polymer 1 elements can cause problems for lit-html because they synchronously create child DOM before lit-html gets a chance to process the template instances and record where the dynamic parts are. There are workarounds, but I think everything will work nicer if it's all converted first.
> 
> How long do you think that would take? There aren't many elements. I can help out.

Make sense!
I think after this first PR to setup the process, migrating should be very fast - 2-3 days of work. (2-3 week of time, since 20%).

static/elements/chromedash-feature.html
justinfagnani: I'm coming from the perspective that a bundler should only take a working project as input and output an optimized working projects. LitElement doesn't need to be bundled with components that use it - LitElement will load when the component imports it. The only thing we need to do right now is translate bare import specifiers to URLs. Rollup does this with the resolve plugin, but there are dev servers like Polymer CLI's server or [open-wc.org's dev server](https://open-wc.org/developing/owc-dev-server.html) that do this on the fly.

Bundling into a single file is perfectly fine in production, but it shouldn't be required for development, and one reason to structure your imports so that every file imports only the direct dependencies they use is so that the project will load without bundling. (another is so IDEs and other tools can understand a module better)

> MyComponent (using lit-element) => bundle with lit-element, export MyComponentBundled. Same for all components.

I definitely would not bundle LitElement into every component. That will severely bloat the bundles.

What I would do is just get the project working without any bundling. Then analyze whan bundles you need based on the entrypoints. A good bundler will be able to handle multiple entry points and build multiple bundles, with shared bundles for the shared libraries.

liyangguang: Ah, Okay. I'll try this method. Thanks!

liyangguang:
Per justinfagnani's comment, I tried a different approach.
- Import each component on the page that using it.
- Removed Rollup, using unpkg to load lit-element.

robdodson:
LGTM. @justinfagnani wdyt?

liyangguang:
Comments on this topic from #652 

@justinfagnani: I worry about the unpkg URLs. They can sometimes cause module duplication that can break things. If it's working for now, that's fine. I'd rather see named specified and either Rollup's node module resolution plugin, or a dev server that rewrites module specifiers like Koa with https://github.com/Polymer/koa-node-resolve

@liyangguang Agree. I put unpkg here temporary to make the app run without getting blocked by the module resolution decision. I found it a bit tricky on the code base. The app is heavily integrated with Django, so I cannot just run the front end using a devserver like polymer serve. I have to run the python dev_appserver. I'm not sure if it's possible to run both and together locally and have the module resolution in the node devserver pass into the python devserver. The only solution I can come up with is the one I used in #640 - use rollup in both local and prod.

Issue #641, opened 2019-04-16
500 error when adding a new feature


mfalken:
This happened twice, and I lost all form input in an unrecoverable way :(

I don't know if I did something like use multiple lines as mentioned in issue #106 

I will try adding a skeleton feature then slowly updating sections one-by-one.

jpmedley:
The error logs are reporting: 'ValueError: invalid literal for int() with base 10: ' in admin.py at line 443.

jpmedley:
@mattto Did you get your feature submitted?

mattto:
Yep, just by submitting a skeleton first then filling it out in multiple updates.

mugdhalakhani:
Running into this as well. Who's the best person to assign this to?

beverloo:
@jstenback @addyosmani @PaulKinlan fyi

lilles:
I've tried multiple times to edit and generate an intent-to-ship for prefers-color-scheme today and get 500 every time. Related to this?

lilles:
> I've tried multiple times to edit and generate an intent-to-ship for prefers-color-scheme today and get 500 every time. Related to this?

I got it to work to have single paragraphs everywhere.

kevers-google:
I've hit this issue twice in the past week with intent-to-implement-and-ship for web animation features.  Both times the first submit fails, and second submit succeeds.

jstenback:
I merged PR #644 this morning and deployed the change. I would really appreciate if someone can provide additional information here if this issue still reproduces. I'll keep an eye on the logs etc as well.

kevers-google:
Hit this issue again yesterday. Possible that it is not the same bug.  Needed to reenter all of the data.

jstenback:
Interestingly enough I see zero evidence of that in the logs (not saying it didn't happen, just that I have nothing to go on in trying to understand what happened :( ). Any more detail you have on what happened would be greatly appreciated (time of error, ip address if you're willing to share, what you did differently when it worked the second time, if anything, etc).

jpmedley:
@jstenback I just learned that service workers can return 500 errors. Chrome Status has a service worker. You might look there.

jrobbins:
This was fixed years ago.

PR #644, opened 2019-05-08
Make all_platforms_descr a multi-line fileld, and add logging.

jstenback:
This PR changes the all_platforms_descr form field to be a multiline
form field. This fixes some reported 500 errors (#641), and also adds
logging for catching a yet to be understood error with intent_stage not
being an integer string.


jpmedley:
@jstenback It's been a few months, but as I recall, we always get the Lighthouse error and don't need for it to block. Is that what you remember?

jstenback:
That is my recollection as well. I'll merge this, thanks for the review!

Issue #625, opened 2019-03-20
Migrate away from web component v0


liyangguang:
This site is built using Web Component V0, which is being deprecated (https://www.chromestatus.com/features/5144752345317376).

We should migrate it away from it, possibly using https://lit-element.polymer-project.org/.

liyangguang:
In addition to components, some related topics came up in code review:
- Update all the sass (at least the ones for components). If possible, change them all to native css and move into components.
- Change to Typescript (?)

liyangguang:
Related PRs:
- #640: rollup setup
- #652: Features page

liyangguang:
lit-virtualizer is discussed offline with the team.
Currently blocked for some features.

liyangguang:
#656: Schedule, metric, samples pages.

jrobbins:
This was accomplished last year.  Thanks so much for doing all that work.

PR #652, opened 2019-06-16
[Migrate to lit-element - part 1] Features page

liyangguang:
This is part 1 of the migration. It renders the /features page.

**Overview**
- Not using lit-virtualizer yet. The default transform Y of items is incorrect. Debugging. Thus the left sidebar highlighting on scrolling is not working yet.
- Other pages: schedule, metric, admin - are not finished.
- Using unpkg to load modules for now to start working before we settle on #640.
- As discussed with Rob, it's hard to break it down to bites of working PR, thus a huge one.

**Detailed code changes:**

Overall readability / code cleanup
- Removed all Polymer 1, bower related code. Updated gulp. Added `gulp watch`
- Added a very bacis `developer-documentation.md` and updated `readme`
- Component, always prefix “chromedash-”, same in the file name.
- Variables pointing to HTML elements always suffix “El”
- Convert old code to ES6 (const/fetch/etc.)
- Move js code in templates to /static/js
- Move components created in templates into separate files in /static/elements/
- Deleted *-import.js files (used for vulcanize), and import component used in each file.
- Move all js to /static/js-src. And use /static/js only for gulp generated files (similar to the /static/sass/ and /static/css/). Updated gulp and gitignore accordingly.

chromedash-feature.js
- Add tooltip to all the filer buttons - category/owner/component
- Fix get cr number
- Keyup enter listener is removed (what’s the expected behavior? See nothing in prod)
- Avoid using deep object values as component property, when we need to update the property (e.g. this.feature.open => this.open)


liyangguang:
@robdodson, @justinfagnani

static/elements/chromedash-color-status.js
```diff
@@ -0,0 +1,29 @@
+import {LitElement, html} from 'https://unpkg.com/@polymer/lit-element@latest/lit-element.js?module';
+
+const CYAN = 120;
+const DEFAULT_MAX = 7;
+
+class ChromedashColorStatus extends LitElement {
+  static get properties() {
+    return {
+      max: {type: Number}, // From parent
```
justinfagnani: what's the `// From parent` comment for?

static/elements/chromedash-color-status.js
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor() {
+    super();
+    this.max = DEFAULT_MAX;
+  }
+
+  render() {
+    const color = `hsl(${Math.round(CYAN - this.value * CYAN / this.max)}, 100%, 50%)`;
+    return html`
+      <link rel="stylesheet" href="/static/css/elements/chromedash-color-status.css">
```
justinfagnani: Can you use `static get styles()` instead? This will avoid FOUC, and give better performance by using Constructible Stylesheets.

If this is generated by Sass, there's a transform out there that will generate JS modules that export LitElement `css` templates from Sass.

static/elements/chromedash-feature.js
```diff
@@ -0,0 +1,492 @@
...
+const MAX_RISK = MAX_VENDOR_VIEW + MAX_WEBDEV_VIEW + MAX_STANDARDS_VAL;
+
+const IS_PUSH_NOTIFIER_ENABLED = window.PushNotifier.GRANTED_ACCESS;
+const IS_PUSH_NOTIFIER_SUPPORTED =
+    window.PushNotifier && window.PushNotifier.SUPPORTS_NOTIFICATIONS;
+
+class ChromedashFeature extends LitElement {
+  static get properties() {
+    return {
+      feature: {type: Object}, // From attribute
```
justinfagnani: Any change you'd use TypeScript in the future so we can see what type of object this is?

static/elements/chromedash-feature.js
```diff
@@ -0,0 +1,492 @@
...
+     * Url has two formats: "http://crbug.com/111111", and
+     * "https://bugs.chromium.org/p/chromium/issues/detail?id=111111" */
+    const matches = link.match(/(id=|crbug.com\/)([0-9]+)/);
+    if (matches) {
+      return matches[2];
+    }
+    return '';
+  }
+
+  set _crBugNumber(value) {/* Never called */}
```
justinfagnani: If it's never called, you can omit the setter. Then a accidental call will cause an error.

static/elements/chromedash-feature.js
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  set _commentHtml(value) {/* Never called */}
+
+  _fireEvent(eventName, detail) {
+    let event = new CustomEvent(eventName, {detail});
+    this.dispatchEvent(event);
+  }
+
+  toggle(e) {
```
justinfagnani: What does this toggle? Can you add a comment or expand the name?

static/elements/chromedash-feature.js
```diff
@@ -0,0 +1,492 @@
...
+    } else {
+      PushNotifications.unsubscribeFromFeature(featureId);
+    }
+  }
+
+  render() {
+    return html`
+      <link rel="stylesheet" href="/static/css/elements/chromedash-feature.css">
+
+      <div class="main-content-area">
```
justinfagnani: You can use ids instead of singleton classes in shadow DOM.

justinfagnani: You probably don't need this element, since it looks like a wrapper on the entire shadow DOM contents. Most likely you can just style the host instead.

static/elements/chromedash-featurelist.js
```diff
@@ -0,0 +1,387 @@
...
+    this._featuresFetchMetric.start();
+
+    this._loadData();
+  }
+
+  _loadData() {
+    const featureUrl = location.hostname == 'localhost' ?
+      'https://www.chromestatus.com/features_v2.json' : '/features_v2.json';
+
+    fetch(featureUrl).then((res) => res.json()).then((features) => {
```
justinfagnani: Can you use async functions and await? This would be more readable, IMO.

static/elements/chromedash-featurelist.js
```diff
@@ -0,0 +1,387 @@
...
+
+  _computeMilestoneString(str) {
+    return isNaN(parseInt(str)) ? str : 'Chrome ' + str;
+  }
+
+  render() {
+    return html`
+      <link rel="stylesheet" href="/static/css/elements/chromedash-featurelist.css">
+
+      <div id="featurelist">
```
justinfagnani: You might not need this element.

static/elements/chromedash-sample-panel.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    this.filtered = features;
+
+    this._fireEvent('update-length', {lenght: this.filtered.length});
+  }
+
+  render() {
+    return html`
+      <div id="sample-panel">
```
justinfagnani: remove wrapper element?

static/elements/chromedash-sample-panel.js
```diff
@@ -0,0 +1,170 @@
...
+      const category = this.selectedCategory.textContent.trim();
+      const regex = new RegExp(category, 'i');
+      features = features.filter((feature) => {
+        return regex.test(feature.category);
+      });
+    }
+
+    this.filtered = features;
+
+    this._fireEvent('update-length', {lenght: this.filtered.length});
```
justinfagnani: `lenght` is misspelled. Who's accessing it?

static/elements/chromedash-sample-panel.js
```diff
@@ -0,0 +1,170 @@
...
+                <div class="summary">${feature.summary}</div>
+                <div class="sample_links">
+                  <div class="demo-links layout horizontal center">
+                    ${this.feature.sample_links.map((link) => html`
+                      <a href="${link}" target="_blank" class="demo-link">
+                        <iron-icon icon="chromestatus:open-in-browser"></iron-icon> View Demo
+                      </a>
+                      `)}
+                  </div>
+                  <iron-icon icon="${this._computeIconId(feature.category)}"
```
justinfagnani: I have some ways to replace iron-icon. We can chat more more detail...

static/elements/chromedash-sample-panel.js
```diff
@@ -0,0 +1,170 @@
...
+            `)}
+        </ul>
+      </div>
+    `;
+  }
+}
+
+customElements.define('chromedash-sample-panel', ChromedashSamplePanel);
+
+// TODO: move this into a behavior. It's duplicated from chromedash-featurelist.html.
```
justinfagnani: It can just be a module. I don't think it needs to be a mixin.

static/elements/chromedash-schedule.js
```diff
@@ -0,0 +1,257 @@
+import {LitElement, html} from 'https://unpkg.com/@polymer/lit-element@latest/lit-element.js?module';
+import 'https://unpkg.com/@polymer/iron-icon/iron-icon.js?module';
+
+class ChromedashSchedule extends LitElement {
+  static get properties() {
```
justinfagnani: Remove

static/elements/chromedash-schedule.js
```diff
@@ -0,0 +1,257 @@
...
+    return {
+    };
+  }
+
+  constructor() {
```
justinfagnani: remove

static/elements/chromedash-schedule.js
```diff
@@ -0,0 +1,257 @@
...
+      PushNotifications.subscribeToFeature(featureId);
+    } else {
+      PushNotifications.unsubscribeFromFeature(featureId);
+    }
+  }
+
+
+  render() {
+    return html`
+      <div>
```
justinfagnani: remove?

static/elements/chromedash-schedule.js
```diff
@@ -0,0 +1,257 @@
...
+            </div>
+            <div class="milestone_info layout horizontal center-center">
+              <h3><span class="channel_label">Stable</span> [[_computeDaysUntil(channels.stable.stable_date)]]
+                <span class="release-stable">( [[_computeDate(channels.stable.stable_date)]] )</span>
+              </h3>
+            </div>
+            <div class="features_list">
+              <div class="features_header">Features in this release:</div>
+
+              <template is="dom-repeat" items="[[_objKeys(channels.stable.components)]]" as="componentName">
```
justinfagnani: replace with lit-html

static/elements/chromedash-schedule.js
```diff
@@ -0,0 +1,257 @@
...
+                <h3 class="feature_components">{{componentName}}</h3>
+                <ul>
+                  <template is="dom-repeat" items="[[_featuresFor(channels.stable.components, componentName)]]" as="f">
+                    <li data-feature-id$="[[f.id]]">
+                      <a href$="/feature/[[f.id]]">[[f.name]]</a>
+                      <span class="icon_row">
+                        <template is="dom-if" if="[[f.browsers.chrome.origintrial]]">
```
justinfagnani: replace with lit-html here and below

static/sass/elements/chromedash-toast.scss
```diff
@@ -0,0 +1,39 @@
+:host {
```
justinfagnani: This looks like plain CSS. Can you move it into `chromedash-toast.js`?

static/sass/elements/chromedash-metrics.scss
```diff
@@ -57,14 +57,19 @@
         }
       }
 
-      x-meter {
+      chromedash-x-meter {
```
justinfagnani: The only Sass in here is the nesting and variable reference below. If you manually expanded the nested and made the variable a CSS variable, this file could be inlined into `chromedash-metrics.js`

static/elements/chromedash-color-status.js
liyangguang: I'm noting where the value is coming from. Here it means the max value is set in the attribute of the component, in the parent component template.

static/elements/chromedash-color-status.js
liyangguang: Yeah. I haven't looked into the sass folder at all. I'll take a look.

static/elements/chromedash-feature.js
liyangguang: If I don't have this line, the getter won't work.

reduced test case: This component renders text "foo". But when I comment out the setter, it does not. Is it a bug in lit-element?

class MyGetter extends LitElement {
  static get properties() {
    return {
      test: {type: String},
    };
  }

  get test() {
    return 'foo';
  }

  // set test(foo) {
  //   return;
  // }

  render() {
    return html`${this.test}`;
  }
}

customElements.define('my-getter', MyGetter);

liyangguang:
> Looks really nice.
> 
> I worry about the unpkg URLs. They can sometimes cause module duplication that can break things. If it's working for now, that's fine. I'd rather see named specified and either Rollup's node module resolution plugin, or a dev server that rewrites module specifiers like Koa with https://github.com/Polymer/koa-node-resolve
> 
> Looking at the CSS, is seems like Sass isn't really required for most of it. LitElement works really nicely when you just use the `static get styles()` method of providing CSS. It's a performance enhancement too.

Great! Thanks for reviewing!

Overall:

1. unpkg & module resolution: Agree. I put unpkg here temporary to make the app run without getting blocked by the module resolution decision. I found it a bit tricky on the code base. The app is heavily integrated with Django, so I cannot just run the front end using a devserver like `polymer serve`. I have to run the python dev_appserver. I'm not sure if it's possible to run both and together locally and have the module resolution in the node devserver pass into the python devserver. The only solution I can come up with is the one I used in #640 - use rollup in both local and prod.
2. CSS: Good suggestion! I just blindly imported existing sass output for now. Will take a closer look after finishing all the js part.
3. chromedash-schedule and more: noted in the PR description. Only /features part is finished. Other files are WIP. I'll send followup PR for them.

static/elements/chromedash-color-status.js
liyangguang: Changed to "From attribute" to be consistent with others.

static/elements/chromedash-color-status.js
liyangguang: I'll work on all the css after first pass of js and setup.

static/elements/chromedash-feature.js
liyangguang: Good suggestion!

static/elements/chromedash-feature.js
liyangguang: Changed to a better name.

static/elements/chromedash-feature.js
liyangguang: Done.

liyangguang: Done

static/elements/chromedash-featurelist.js
liyangguang: Done

static/elements/chromedash-featurelist.js
liyangguang: done

static/elements/chromedash-sample-panel.js
liyangguang: (this and schedule.js are WIP.)

static/elements/chromedash-timeline.js
```diff
@@ -0,0 +1,218 @@
...
+    if (!this.props.length) {
+      return;
+    }
+
+    const feature = this.props.find((el) => el[0] === parseInt(this.selectedBucketId));
+    if (feature) {
+      const featureName = feature[1];
+      const REPORT_ID = '1M8kXOqPkwYNKjJhtag_nvDNJCpvmw_ri';
+      const dsEmbedUrl = `https://datastudio.google.com/embed/reporting/${REPORT_ID}/page/tc5b?config=%7B"df3":"include%25EE%2580%25800%25EE%2580%2580IN%25EE%2580%2580${featureName}"%7D`;
+      this.$.httparchivedata.src = dsEmbedUrl;
```
robdodson: I think `this.$` was a Polymer API to access an element by id. But I don't think it exists on lit-element. https://github.com/Polymer/lit-element/issues/57

@justinfagnani is that correct?

static/elements/chromedash-timeline.js
```diff
@@ -0,0 +1,218 @@
...
+
+      this.$.bigquery.textContent = `#standardSQL
```
robdodson: Same comment as above. Double check that `this.$` works in lit

static/elements/chromedash-timeline.js
robdodson: I noticed there are a number of elements using `this.$` so it might be worth doing a pass to catch all of those.

robdodson: Justin suggested caching a reference to the elements using the `firstUpdated` callback.

static/elements/chromedash-userlist.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  removeUser(idx) {
+    return this.splice('users', idx, 1);
+  }
+
+  ajaxSubmit(e) {
+    e.preventDefault();
+
+    if (this.$.form.checkValidity()) {
```
robdodson: See previous comment on `this.$`

static/elements/chromedash-userlist.js
```diff
@@ -0,0 +1,105 @@
...
+      // TODO(ericbidelman): move back to this.$.form.email.value when SD
+      // polyfill merges the commit that wraps element.form.
+      const email = this.$.form.querySelector('input[name="email"]').value;
```
robdodson: See previous comment on `this.$`

static/elements/chromedash-metrics.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    this.maxPercentage = this.viewList.reduce((accum, currVal) => {
+      return Math.max(accum, currVal.percentage);
+    }, 0);
+
+    if (location.hash) {
+      const hash = decodeURIComponent(location.hash);
+      if (hash) {
+        const el = this.$['stack-rank-list'].querySelector(hash);
```
robdodson: See previous comment on `this.$`

static/elements/chromedash-timeline.js
```diff
@@ -0,0 +1,218 @@
...
+
+    let view = table;
+
+    if (!showAllHistoricalData) {
+      const startYear = (new Date()).getFullYear() - 2; // Show only 2 years of data by default.
+      view = new google.visualization.DataView(table);
+      view.setRows(view.getFilteredRows([{column: 0, minValue: new Date(startYear, 0, 1)}]));
+    }
+
+    const chart = new google.visualization.LineChart(this.$.chart);
```
robdodson: See previous comment on `this.$`

static/elements/chromedash-timeline.js
liyangguang: Sorry for the noise. This file is WIP.
I just removed all WIP files. This PR is only the features page.

robdodson: oops! my mistake :)

liyangguang:
Hey @robdodson @justinfagnani , I have another batch of pages ready. I guess it would be good if we can merge this PR then I send the next one, so you can see a better diff.

Unresolved issues in this PR:
- Clean up CSS. Noted in #625. Will update after all the js.
- Rollup/unpkg: Copied comments to #640. It's intended to use a temporary solution in this PR. Currently I feel using rollup for both local and prod is feasible. We can continue discussion there.

static/elements/chromedash-feature.js
liyangguang: Never mind. I updated them all to one time assignment instead of getters. (since the values do not change).

Regarding the getter/setter issue on lit-element: After I upgrade lit-element from rc to latest, it works as expected now.

robdodson:
Hey @liyangguang, it's on my to do list to take another look at this today. thank you for your patience!

robdodson:
This seems good to land so we can continue. Thanks @liyangguang!

robdodson:
@jstenback or @jpmedley would one of y'all mind merging this when you get a chance. It should only affect the `dev` branch.

robdodson:
Thank you @jstenback !

Issue #529, opened 2018-01-18
Add more validation to feature submission form


ebidel:
Many times iOS/android/webview versions are not filled in by chromium devs.

Before the feature gets added, pop up a dialog asking if everything looks right.
Important fields that are blank should be called out.

jalajcodes:
I'd like to work on this issue.

yuktibagrodia:
@ebidel can I work on it? if it's not done yet?

435akashjha:
can I work on this issue and add form validation in the form

FarhanAliRaza:
@jrobbins hello sir I want to work on this for GSoC 2021. Can I work on this if it is not done?

GouravSingal-code:
can i work on it

impact-maker:
@jpmedley So this issue doesn't need more help as it is currently handled by the team. This is what you wanted to communicate?

siddhiguptas:
Hi @ebidel !!

I’d love to work on this issue as my first contribution. I believe it would be a great opportunity for me to contribute and learn more about form validation and UI improvements. Would it be okay for me to take it up?

Looking forward to your guidance!

Mukta-Sawant:
Hi, @jpmedley @jrobbins @ebidel I would like to work on this issue. Can you please assign this to me? 
Thanks, Mukta

suprith-hub:
Hi, @ebidel @jrobbins enquiring the fields to be shown in the pop-up.
1. What are the important fields mentioned in the context?
Right now, the mandatory fields are scrolled with red border highlighting the textfield.
2. Are mandatory fields same as important fields?
3. Should the pop up appear after mandatory fields are filled?
4. Should the pop-up be for editing the form or just adding new feature?

suprith-hub:
@jcscottiii if I get assistance on this I can move ahead...

pradyut4u:
hi i would like to work on this issue if its available

dhanushkumar-amk:
Hi, @jpmedley @jrobbins @ebidel I would like to work on this issue. Can you please assign this to me?

js0414030:
Hi, I’m interested in working on this issue. Can you please assign it to me?

14-Prasanna:
Hi, @ebidel @jpmedley I would like to work on this issue. Can you please assign it to me ?

itsAyush5:
Can u please assign this issue to me @ebidel @jpmedley @jrobbins

[...]

PR #662, opened 2019-08-04
Adding popup in feature submission form

tapaswenipathak:
Fixes https://github.com/GoogleChrome/chromium-dashboard/issues/529.

Creating a draft PR.

- [ ] Implement popup template
- [ ] Show popup template with the missing fields

### DIFF
--- a/templates/popup.html
+++ b/templates/popup.html
@@ -1,0 +1,33 @@
+d>
+<meta charset="utf-8">
+<title>{% block page_title %}{% endblock %}{{ APP_TITLE }}</title>
+<meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
+
+<link rel="manifest" href="/static/manifest.json">
+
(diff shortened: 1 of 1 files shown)


jpmedley:
I only see the popup. I don't see the code that triggers it, nor the text that's displayed. 

Also, this issue needs to do some things that aren't specifically called out. 

* If iOS is included, the contributor needs to be prompted to confirm that it should be **present**.
* If WebView or Android are missing, the contributor needs to be prompted to confirm that they should **not be present**.
* If the Status in Chromium field is set to Origin Trial, no value is allowed for WebView and the contributor needs to be informed that the value will be removed. Then the value needs to be removed.

I apologize that these aren't listed. At the time the tickets were opened I thought we were going to do this soon, and Eric and I both have these in our heads.

tapaswenipathak:
Hey: thanks for the inputs and sorry for me being late in responding, closing the pr, I will see if I like few different tickets.

Issue #372, opened 2016-07-28
Status Item Contains Link to Itself Labeled 'Sample'.


jpmedley:
The link is circled below. It only shows up in the [/features](https://www.chromestatus.com/features/5668635896971264) view. It's not in the [/feature](https://www.chromestatus.com/feature/5668635896971264) view.
![image](https://cloud.githubusercontent.com/assets/8204171/17224598/779a9de4-54b6-11e6-8c0b-1500ba9ad712.png)

ebidel:
Looks like https://github.com/GoogleChrome/chromium-dashboard/blob/master/templates/feature.html#L68-L77 needs to loop over the list of samples.

However, for this example, it looks like Sample 2 is bogus. It links to the feature itself (https://www.chromestatus.com/features/5668635896971264).

PR #663, opened 2019-08-04
Expose bogus sample 2 item appearing

tapaswenipathak:
Fixes https://github.com/GoogleChrome/chromium-dashboard/issues/372.

- [x] Iterate over the list
- [x] Use the fields from iterated list and show as per the list

### DIFF
--- a/templates/feature.html
+++ b/templates/feature.html
@@ -73,10 +73,12 @@
     </div>
   </section>
 
-  {% if feature.summary %}
-  <section id="summary">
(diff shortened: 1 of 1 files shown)


jpmedley:
Did you investigate this. Eric's comment was off the top of his head. It may or may not have been correct. I'm not sure how looping solves the problem in the ticket I opened.

@jstenback

tapaswenipathak:
Hey: thanks for the inputs and sorry for me being late in responding, closing the pr, I will see if I like picking few different tickets.

Issue #654, opened 2019-06-21
Add 'Deprecated' link to list of items in navigation pane. 


jpmedley:
Current deprecations are always listed in the [beta release post](https://blog.chromium.org/2019/06/chrome-76-beta-dark-mode-payments-new.html) and in a [column on /Updates](https://developers.google.com/web/updates/tags/deprecations).

Enterprise has requested a running list of current deprecations. Fortunately we're already able to do this. Chrome Status has functionality for displaying lists of features with a specific status. The statuses we make explicit with links (see the image) do not currently include deprecations. 

This proposal is to add a link making deprecations easy to find.

![image](https://user-images.githubusercontent.com/8204171/59934078-9ce97580-93ff-11e9-9b96-b8cdd9b28a78.png)

jpmedley:
@jstenback I went to deploy this just now and found that this didn't add anything to the side panel. I can't find anything else that might affect it. Can you please take a look.

jpmedley:
Fixed.

![image](https://user-images.githubusercontent.com/8204171/76436449-785e7080-6375-11ea-9a2d-b9b75c63d8fc.png)

PR #667, opened 2019-08-14
Followup fix for #654.

jstenback:
This fixes the chromedash-metadata template to include the deprecated
features entry in the versions list. This also fixes a problem that was
uncovered where an ordered list that was in the incorrect order which
resulted in the wrong implementation status being shown in the versions
list.


models.py
```diff
@@ -411,7 +413,7 @@ def _annotate_first_of_milestones(self, feature_list, version=None):
         IMPLEMENTATION_STATUS[NO_ACTIVE_DEV],
         IMPLEMENTATION_STATUS[PROPOSED],
         IMPLEMENTATION_STATUS[IN_DEVELOPMENT],
-        IMPLEMENTATION_STATUS[DEPRECATED]
+        IMPLEMENTATION_STATUS[DEPRECATED],
```
jpmedley: Did you mean to add a comma to the last item in an array?

jstenback: Yep, for consistency with the surrounding code. In JS code the linter even requires commas at the end of lists like this (not something I personally agree with, but consistency outweighs my personal feelings here :).

PR #677, opened 2019-09-25
Add lit-virtualizer

liyangguang:
- Use lit-virtualizer.
- Sidebar highlighting updates when scrolling.
- Removed temporary code handling scroll behavior for a normal HTML list.


liyangguang:
@robdodson

static/elements/chromedash-featurelist.js
```diff
@@ -279,59 +274,46 @@ class ChromedashFeaturelist extends LitElement {
...
-    let selector;
-    if (!isNaN(Number(targetId))) {
-      /* number is not valid id. so we added prefix 'id-' in the template. */
-      selector = `#id-${targetId}`;
-    } else {
-      selector = `#${targetId}`;
+    for (let i = 0, f; f = this.filtered[i]; ++i) {
+      if (f.id === targetId) {
+        this.virtualizerEl.scrollToIndex(i);
+        const featureEl = document.querySelector('chromedash-featurelist').shadowRoot.querySelector(`chromedash-feature[id=id-${f.id}]`);
```
robdodson: I think this is fine to get things working, but it does defeat the encapsulation of shadow dom.

Perhaps a cleaner model would be to set the id on chromedash-featurelist and have it locate the correct child and set the open property on it. Ideally this should all be done through the public interfaces of these components without needing to query into their shadow roots.

robdodson:
I noticed the virtualizer seems to have large gaps between the items:

image

robdodson:
cc @straversi and @justinfagnani for their reviews as well.

static/elements/chromedash-feature.js
```diff
@@ -29,23 +29,18 @@ class ChromedashFeature extends LitElement {
...
-    super();
-    this.open = false;
-  }
-
   firstUpdated() {
     this._initializeValues();
   }
 
-  updated(props) {
+  shouldUpdate(props) {
```
justinfagnani: You shouldn't do any side-effects in `shouldUpdate`. Why did this change from `updated`? If it's to go before the first render, then override `update()` and do the work before the `super.update(changedProperties)` call.

static/elements/chromedash-featurelist.js
```diff
@@ -30,6 +31,10 @@ class ChromedashFeaturelist extends LitElement {
     this._loadData();
   }
 
+  firstUpdated() {
+    this.virtualizerEl = this.shadowRoot.querySelector('lit-virtualizer');
```
justinfagnani: style nit: is `virtualizerEl` public? Maybe prefix with a `_`.

static/elements/chromedash-featurelist.js
justinfagnani: +1

At least add a comment before this line saying why you're doing this and a TODO to fix it.

static/elements/chromedash-featurelist.js
```diff
@@ -369,16 +351,24 @@ class ChromedashFeaturelist extends LitElement {
...
+      </style>
+
+      <lit-virtualizer
+        .scrollTarget=${window}
+        .items=${this.filtered}
+        @rangechange=${this._onScrollList}
+        .renderItem=${(feature) => html`
+          <div class="item">
+            <div ?hidden="${this._computeMilestoneHidden.call(this, feature, this.features, this.filtered)}"
+                 class="milestone-marker">${this._computeMilestoneString.call(this, feature.browsers.chrome.status.milestone_str)}</div>
```
justinfagnani: Why `this._computeMilestoneString.call(this, ...`? That's the same as `this._computeMilestoneString(...`

static/elements/chromedash-featurelist.js
```diff
@@ -369,16 +351,24 @@ class ChromedashFeaturelist extends LitElement {
...
+            <chromedash-feature id="id-${feature.id}" tabindex="0"
+                 @feature-toggled="${this._onFeatureToggled.bind(this)}"
```
justinfagnani: Rather than bind a new closure each render, write `_onFeatureToggled` as a class property with an arrow function

```js
constructor() {
  this._onFeatureToggled = (e) => { ... };
}
```

and

```
@feature-toggled="${this._onFeatureToggled}"
```

static/elements/chromedash-feature.js
```diff
@@ -438,6 +434,7 @@ class ChromedashFeature extends LitElement {
                 <div class="doc_links">
                   <chromedash-multi-links
                       .links="${this.feature.resources.docs}"
+                      title="Doc"
```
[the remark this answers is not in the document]
liyangguang: Oops, didn't read closely.

static/elements/chromedash-featurelist.js
liyangguang: Thanks for the suggestion. Applied the new method.

static/elements/chromedash-feature.js
liyangguang: >  it's to go before the first render, then override `update()` and do the work before the `super.update(changedProperties)` call.

Exactly. Thanks for the suggestion. Applied.

static/elements/chromedash-featurelist.js
liyangguang: The context for function calls inside `renderItem` is `lit-virtualizer` instead of this component (`chromedash-featurelist`). So I have to bind/call them with this.

Not sure if it's designed behavior of lit-virtualizer.

static/elements/chromedash-featurelist.js
liyangguang: Done

static/elements/chromedash-featurelist.js
liyangguang: Oh, after some more experiments, it turns out only the event handler need this binding (and changed to a static arrow function).

liyangguang:
> I noticed the virtualizer seems to have large gaps between the items:
> 
> image

Ah, yes, it's expected in this PR. https://github.com/PolymerLabs/uni-virtualizer/issues/15. In the rollup PR (#679), I moved the CSS into components, which solves this issue.

static/elements/chromedash-featurelist.js
robdodson: would it be possible to bind the function in the constructor?

```
constructor() {
  super();
  this._computeMilestoneString = this._computeMilestoneString.bind(this);
}
```

If that doesn't work, it might be good to add a comment explaining why we used `call()`.

liyangguang:
> LGTM. Added one comment to fix.

_onFeatureToggled

liyangguang:
Good suggestion!

`_computeMilestoneString` is not needed to be bound any more. Bound `_onFeatureToggled` in the constructor.

robdodson:
@jpmedley are you ok merging this into the dev branch?

liyangguang:
> @jpmedley are you ok merging this into the dev branch?

Hey~ Friendly ping.

jpmedley:
I have no objections.

330badboi:
denial of pull requests

330badboi:
denial of merge requests

PR #683, opened 2019-10-28
Fix refs to django template constants that moved

jrobbins:
TOKEN_VAR and others got moved into template.base somewhere between django 1.4 and 1.11.


jrobbins:
@googlebot I signed it!

PR #693, opened 2019-11-08
Rename "Intent to Implement" as "Intent to Prototype"

cwilso:
This PR renames the labels and sets up the proper naming for the generated intent template, but does not rename the internal definitions or macros.

### DIFF
--- a/models.py
+++ b/models.py
@@ -76,7 +76,7 @@ INTENT_REMOVE = 6
 
 INTENT_STAGES = {
   INTENT_NONE: 'None',
-  INTENT_IMPLEMENT: 'Implement',
+  INTENT_IMPLEMENT: 'Prototype',
   INTENT_EXPERIMENT: 'Experiment',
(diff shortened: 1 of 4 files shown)


models.py
```diff
@@ -76,7 +76,7 @@
 
 INTENT_STAGES = {
   INTENT_NONE: 'None',
-  INTENT_IMPLEMENT: 'Implement',
+  INTENT_IMPLEMENT: 'Prototype',
   INTENT_EXPERIMENT: 'Experiment',
   INTENT_EXTEND_TRIAL: 'Extend Origin Trial',
   INTENT_IMPLEMENT_SHIP: 'Implement and Ship',
```
jrobbins: Should "Implement and Ship" also be changed to "Prototype and Ship"?

cwilso: I don't think so.  We've kind of deprecated this stage in the process anyway, but if you're proceeding straight to ship, I think "Implement" is more appropriate.

PR #694, opened 2019-11-18
Add DevRel field to Feature

beaufortfrancois:
This PR simply adds a DevRel field (Developer relations emails) to the Feature table so that web developers can query through ChromeStatus API the appropriate DevRel folks for each feature.

```js
const response = await fetch('https://chromestatus.com/features_v2.json');
const features = await response.json();
features.forEach(feature => {
  const devrel = feature.browsers.chrome.devrel;
  const featureName = feature.name;
  console.log(`Web Dev POC for "${featureName}": ${devrel}`);
});
``` 

FYI @petele @jeffposnick @tomayac @jpmedley

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -363,6 +363,10 @@ class FeatureHandler(common.ContentHandler):
     if blink_components:
       blink_components = filter(bool, [x.strip() for x in blink_components.split(',')])
 
+    devrel = self.request.get('devrel') or []
+    if devrel:
+      devrel = [db.Email(x.strip()) for x in devrel.split(',')]
+
     try:
       intent_stage = int(self.request.get('intent_stage'))
     except:
@@ -397,6 +401,7 @@ class FeatureHandler(common.ContentHandler):
       feature.owner = owners
       feature.bug_url = bug_url
       feature.blink_components = blink_components
+      feature.devrel = devrel
       feature.impl_status_chrome = int(self.request.get('impl_status_chrome'))
       feature.shipped_milestone = shipped_milestone
       feature.shipped_android_milestone = shipped_android_milestone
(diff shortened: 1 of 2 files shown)


jpmedley:
@beaufortfrancois Before we merge this, we should update the [devrel_team.yaml](https://github.com/GoogleChrome/chromium-dashboard/blob/master/data/devrel_team.yaml) so that it returns the correct people. (I think that's the file.) We maybe need some help from the leads team to sort this. @PaulKinlan

jpmedley:
@jrobbins Is probably who should code review this.

beaufortfrancois:
As far as I understand, `devrel_team.yaml` is only used to populate blink components. So it's not related to this PR. See https://github.com/GoogleChrome/chromium-dashboard/commit/ecd5550d1f77cd4f567cbfcffb6e7188d54ddb8a#diff-b407e6436cf8d0285ac990756db3c1c7

jpmedley:
My point is that somewhere this system already has functionality for matching features with folks in DevRel. It's the basis for update emails. Is your change new functionality or does it connect somehow to what we already have?

image

beaufortfrancois:
A Blink>Component is different than a feature right?

beaufortfrancois:
For instance, I'm an owner of Blink>Bluetooth but there are several chrome features related to this component: 
image

jpmedley:
Correct. When Eric set up the notification system he organized it by blink component.

beaufortfrancois:
It doesn't connect them. It simply adds a new column that is only visible from the API.

jpmedley:
Which would be a demoralized data set. Also, how would be add, edit, or remove the data?

beaufortfrancois:
From admin form, as usual.

beaufortfrancois:
Here's some screenshot.

image

PR #718, opened 2020-01-11
Several small fixes to lit-element code

jrobbins:
With these changes, I am able to run on prod and use all tool features.


liyangguang:
LGTM

PR #722, opened 2020-01-16
Use lit-virtualizer with UI state in the items rather than mutating.

jrobbins:
Instead of mutating a chromedash-feature element when the user clicks to expand it, the code now *only* fires an event with the updated state.  That state is then used by render() in chromedash-feature-list.

This seems to work on FF, safari, and chrome.

Demo link: https://2020-01-15-dot-cr-status.appspot.com

I realize that my JS and web-component skills are a little rough.  If you see a cleaner way to do this, please don't hesitate to guide me.

### DIFF
--- a/static/elements/chromedash-featurelist.js
+++ b/static/elements/chromedash-featurelist.js
@@ -14,6 +14,7 @@ class ChromedashFeaturelist extends LitElement {
       metadataEl: {attribute: false}, // The metadata component element. Directly edited in template/features.html
       searchEl: {attribute: false}, // The search input element. Directly edited in template/features.html
       filtered: {attribute: false},
+      filteredWithState: {attribute: false},
     };
   }
 
@@ -21,6 +22,7 @@ class ChromedashFeaturelist extends LitElement {
     super();
     this.features = [];
     this.filtered = [];
+    this.filteredWithState = [];
     this.metadataEl = document.querySelector('chromedash-metadata');
     this.searchEl = document.querySelector('.search input');
     this.whitelisted = false;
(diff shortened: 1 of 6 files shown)


static/elements/chromedash-featurelist.js
```diff
@@ -14,20 +14,23 @@ class ChromedashFeaturelist extends LitElement {
...
     this.filtered = [];
+    this.filteredWithState = [];
     this.metadataEl = document.querySelector('chromedash-metadata');
     this.searchEl = document.querySelector('.search input');
     this.whitelisted = false;
     this._hasInitialized = false; // Used to check initialization code.
     this._hasScrolledByUser = false; // Used to set the app header state.
     /* When scrollTo(), we also expand the feature. This is the id of the feature. */
     this._scrollOpenFeatureId = undefined;
+    this.openFeatures = new Map();
```
liyangguang: nit: seems it's a id/boolean Map. use Set instead? `this.openedFeatures = new Set()`

static/elements/chromedash-featurelist.js
```diff
@@ -137,9 +140,20 @@ class ChromedashFeaturelist extends LitElement {
     }
   }
 
+  _calcFilteredWithState() {
```
liyangguang: Looks like `this.filtered` is no longer used, and `_calcFilteredWithState` is called every `this.filtered` gets updated. Is it safe to just change the structure of `filtered` to be  `filteredWithState`, and use just one property?

static/elements/chromedash-featurelist.js
jrobbins: Done.

static/elements/chromedash-featurelist.js
jrobbins: I was able to eliminate that as a property and just have openFeatures.

PR #726, opened 2020-01-18
Handle clicks on sorting arrow icons.

jrobbins:
This avoids a JS error when the user clicks on the sort triangle icons in the headers of the metrics pages.  The data-order attribute is defined on the <span> but if the user clicks on the icon, e.target will be the icon which has no dataset.order.

### DIFF
--- a/static/elements/chromedash-metrics.js
+++ b/static/elements/chromedash-metrics.js
@@ -74,7 +74,7 @@ class ChromedashMetrics extends LitElement {
   sort(e) {
     e.preventDefault();
 
-    const order = e.target.dataset.order || e.target.parentNode.dataset.order;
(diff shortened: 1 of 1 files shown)


static/elements/chromedash-metrics.js
```diff
@@ -74,7 +74,7 @@ class ChromedashMetrics extends LitElement {
   sort(e) {
     e.preventDefault();
 
-    const order = e.target.dataset.order;
+    const order = e.target.dataset.order || e.target.parentNode.dataset.order;
```
liyangguang: `e.currentTarget` should cover both cases.

jrobbins: Much better.  Thank you.

PR #728, opened 2020-01-22
Remove lit-virtualizer until it is fully supported

jrobbins:
While investigating issue #737, I realized that it will be pretty difficult for me to get lit-virtualizer to work perfectly and to debug it.  While anything is possible with enough time and effort, I want to switch gears to work on collaborative functionality.  So, I'd like to try taking lit-virtualizer out until I get a chance to re-implement the feature list view to look more like a Monorail table and/or lit-virtualizer graduates from pre-release status.

This change removes lit-virtualizer and simply renders each feature element.  Rendering 1600 elements takes about 3 seconds, which feels like it is too slow, especially for the search-as-you-type feature.  So, I am capping the number of features shown at 500.   I don't think that many users would scroll down through 500 features.  If they want to get to features in earlier releases, they can click on a release number.

### DIFF
--- a/static/elements/chromedash-featurelist.js
+++ b/static/elements/chromedash-featurelist.js
@@ -1,6 +1,5 @@
 import {LitElement, html} from 'lit-element';
 // eslint-disable-next-line no-unused-vars
-import {LitVirtualizer} from 'lit-virtualizer';
 import './chromedash-feature';
 import style from '../css/elements/chromedash-featurelist.css';
 
@@ -47,7 +46,7 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   firstUpdated() {
-    this._virtualizerEl = this.shadowRoot.querySelector('lit-virtualizer');
+    // @@@ this._virtualizerEl = this.shadowRoot.querySelector('lit-virtualizer');
   }
 
   async _loadData() {
@@ -335,7 +334,7 @@ class ChromedashFeaturelist extends LitElement {
 
     for (let i = 0, f; f = this.filtered[i]; ++i) {
       if (f.id === targetId) {
-        this._virtualizerEl.scrollToIndex(i);
+        // @@@ this._virtualizerEl.scrollToIndex(i);
         return;
       }
     }
@@ -388,6 +387,7 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   render() {
+    console.log('num features = ' + this.filtered.length);
     const filteredWithState = this.filtered.map((feature) => {
       return {
         feature: feature,
@@ -401,11 +401,7 @@ class ChromedashFeaturelist extends LitElement {
         .item {width: 100%}
       </style>
 
-      <lit-virtualizer
-        .scrollTarget=${window}
-        .items=${filteredWithState}
-        @rangechange=${this._onScrollList}
-        .renderItem=${(item) => html`
+      ${filteredWithState.map((item) => html`
           <div class="item">
             <div ?hidden="${this._computeMilestoneHidden(item.feature, this.features, this.filtered)}"
                  class="milestone-marker">${this._computeMilestoneString(item.feature.browsers.chrome.status.milestone_str)}</div>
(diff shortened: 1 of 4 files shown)


jrobbins:
Demo link: https://2020-01-21-dot-cr-status.appspot.com/features/

static/elements/chromedash-featurelist.js
```diff
@@ -388,24 +383,26 @@ class ChromedashFeaturelist extends LitElement {
...
     return html`
       <link rel="stylesheet" href="/static/css/elements/chromedash-featurelist.css">
       <style>
         .item {width: 100%}
       </style>
 
-      <lit-virtualizer
-        .scrollTarget=${window}
-        .items=${filteredWithState}
-        @rangechange=${this._onScrollList}
```
liyangguang: Do we want to recreated this feature? - highlight the left sidebar item when scrolling.

(If not, remove _onScrollList in this component.)

static/elements/chromedash-featurelist.js
```diff
@@ -333,11 +330,9 @@ class ChromedashFeaturelist extends LitElement {
...
     this._setOpenFeatures(targetId, true);
 
-    for (let i = 0, f; f = this.filtered[i]; ++i) {
-      if (f.id === targetId) {
-        this._virtualizerEl.scrollToIndex(i);
-        return;
-      }
+    const targetEl = this.shadowRoot.querySelector('#id-' + targetId);
+    if (targetEl) {
+      targetEl.scrollIntoView();
```
liyangguang: This would only scroll the item at the top of the screen, which is covered by the header.

I think we can add these below this line to scroll pass the header.
```
const heightOfHeader = $('.main-toolbar').getBoundingClientRect().height;
window.scrollBy(0, heightOfHeader * -1);
```

static/elements/chromedash-featurelist.js
```diff
@@ -388,24 +383,26 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   render() {
```
liyangguang: Nit: I believe it's not a good practice to have computation and new variables inside the render function. Move it out if possible.
Not blocking this PR. Maybe add a TODO.

static/elements/chromedash-featurelist.js
```diff
@@ -388,24 +383,26 @@ class ChromedashFeaturelist extends LitElement {
...
-    const filteredWithState = this.filtered.map((feature) => {
+    console.log('num features = ' + this.filtered.length);
```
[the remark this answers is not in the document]
jrobbins: Gone.

static/elements/chromedash-featurelist.js
jrobbins: Let's try removing it and see if anyone complains.

static/elements/chromedash-featurelist.js
jrobbins: Added TODO.

static/elements/chromedash-featurelist.js
jrobbins: Thank you! That's exactly the kind of help that I needed.

jrobbins:
Rob, Yangguang suggested that I run this past you.  Basically, lit-virtualizer had some odd behavior that I spent a day investigating without figuring out: scrollbar appearing on items that changed size, and some items being hidden when they should not be.  The current app UI has a very long scrolling list with each feature being a complex item, so lit-virtualizer helped make that fast.  However, I don't think that we are tied to that UI choice.  Limiting the list length to 500 gives acceptable performance without needing to depend on lit-virtualizer.  I'd be open to trying to use lit-virtualizer again once it reaches 1.0.

jrobbins:
Updated demo link: https://2020-01-22-dot-cr-status.appspot.com

robdodson:
That sounds reasonable to me. @straversi may be interested in any feedback you have on lit-virtualizer.

PR #732, opened 2020-01-24
Use more memcache chunks to stay under 1M size limit.

jrobbins:
There was existing logic to split the list of feature objects into chunks in memcache to stay under a 1MB per item size limit.  However, apparently 500 features can add up to more than 1MB, so I reduced it to 300.

This is already live on the site because the problem was causing 500s.  See:
https://pantheon.corp.google.com/errors/CJaB2c2LzZLPeA?time=PT1H&project=cr-status&organizationId=433637338589


jrobbins:
Friendly ping.  I need to land this change before I can do any other deployment.

PR #734, opened 2020-01-28
Set up unit tests locally and in travis.

jrobbins:


jrobbins:
Johnny, please take a look.  In this CL:
+ Set up a test directory for python unit tests
+ Add a script to package.json to make 'nom run test' run the python scripts and document that in README.md
+ Add the mock library for python mock objects
+ Implement a few simple unit tests, including one that uses mock
+ Add a test/testing_config.py file that finds the needed GAE libraries and adds them to the python system path so that they can be imported in code that is being tested
+ Hack on the travis config to get it to install gcloud with python libraries without running out of disk space, and make it run the tests

This is all pretty new to me, so please let me know if you know of a better way to do any of this.

I have the unit tests passing under travis.  However, if you click through to travis you will see that there is also a step that is supposed to deploy the app to a specific version name and then hit that with lighthouse tests, and that seems to have been broken for a while already.

jrobbins:
To clarify, the lighthouse checks are running, but they are testing the "lighthouse-ci-staging" version of the app that was deployed in 2017.

travis/install_google_cloud_sdk.sh
```diff
@@ -8,10 +8,12 @@ openssl aes-256-cbc -K $encrypted_aee7e38c959c_key -iv $encrypted_aee7e38c959c_i
     -in gcloud-client-secret.json.enc -out gcloud-client-secret.json -d
 
 # Download & install the Google Cloud SDK
-curl https://sdk.cloud.google.com | bash
+#curl https://sdk.cloud.google.com | bash
```
jstenback: Given that this happens in the before_install hook now, maybe call that out here and remove the commented out code here (and below)?

PR #758, opened 2020-02-13
Refactor current email notification code to use django template and add tests

jrobbins:
This is the existing email notification functionality, just implemented with more readable and testable code.

I believe I also fixed at least one defect, where new feature emails had the text "{component_name}" in the intro instead of the component name being inserted.


jrobbins:
I changed it to use the django "date" filter instead of having any python code.

PR #761, opened 2020-02-19
Remove date from email body

jrobbins:
Dates are tricky because of timezones.  The current date generation is actually done in GMT, which means that it shows tomorrow's date starting at 4 or 5pm Pacific time.  We don't know what timezone any particular user is in at any particular moment.

Email messages have a Date: header that should be displayed by the email client using the preferred timezone for that user.  So, we can just rely on that instead of putting the date in the email body.  

There is a lag between the feature record being changed and the email being sent, but it should only be seconds or minutes, not enough to make the date look wrong.

### DIFF
--- a/templates/new-feature-email.html
+++ b/templates/new-feature-email.html
@@ -15,7 +15,7 @@
 
 <p><b><a href="https://www.chromestatus.com/feature/{{id}}"
          >{{feature.name}}</a></b>
-  (added {{feature.created|date}})</p>
+</p>
 <p><b>Milestone</b>: {{milestone}}</p>
 <p><b>Implementation status</b>: {{status}}</p>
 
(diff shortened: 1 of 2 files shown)


templates/new-feature-email.html
```diff
@@ -15,7 +15,7 @@
 
 <p><b><a href="https://www.chromestatus.com/feature/{{id}}"
          >{{feature.name}}</a></b>
-  (added {{feature.created|date}})</p>
```
aglaforge: Should this also be removed in the code that injects the date string as a field into the object?

jrobbins: The python code does not specifically add the date field to the template data dictionary.  It used to do that but I already removed it in a previous CL.  The current code accesses a field of the Feature object, and that Feature object is needed for other parts of the template.

PR #771, opened 2020-02-28
Fix email link to updated feature.

jrobbins:
I mistakenly had {id} where {{id}} was needed.   This fixes the problem, and I added a test.  Grepping for '{' shows no other cases of this syntax error.


jrobbins:
Oh, I also made a cosmetic change to the email formatting by removing some of the <br/> elements.

PR #772, opened 2020-03-03
Send emails to starrers.

jrobbins:
In this PR:
* Add a datastore entity type UserPref to store user preferences, including prefs for email notifications.
* When generating email notifications, look up users who starred the feature, get their prefs, and notify them with an appropriate reason footer.

### DIFF
--- a/models.py
+++ b/models.py
@@ -1193,6 +1193,30 @@ class FeatureForm(forms.Form):
        self.fields[field].widget.attrs['required'] = 'required'
 
 
+class UserPref(DictModel):
+  """Describes a user's application preferences."""
+
+  email = db.EmailProperty(required=True)
+  notify_as_starrer = db.BooleanProperty(default=True)
+
+  @classmethod
+  def get_prefs_for_emails(self, emails):
+    """Return a list of UserPrefs for each of the given emails."""
+    q = UserPref.all()
+    q.filter('email IN', emails)
+    user_prefs = q.fetch(None)
+    found_set = set(up.email for up in user_prefs)
+
+    # Make default prefs for any user that does not already have an entity.
+    new_prefs = [UserPref(email=e) for e in emails
+                 if e not in found_set]
+    for np in new_prefs:
+      np.put()
+      user_prefs.append(np)
+
+    return user_prefs
+
+
 class AppUser(DictModel):
   """Describes a user for whitelisting."""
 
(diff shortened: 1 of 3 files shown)


jpmedley:
@jrobbins I didn't think about this before now. I do a lot of copy editing of summaries. I fix spelling and punctuation. I ask feature owners to clarify things. Am I now going to be spamming people who starred features?

jrobbins:
It will only send emails to users who have starred the feature or who own the component or who subscribe to all changes.   So, for a fairly new feature, there are probably not too many people who have starred it, and they might mostly be people who are involved in or very interested in the feature, so I would not consider those emails to be undesired.  

Also, users have ways to control the amount of email that they get, including: unstarring a particular feature that they might no longer be interested in, disabling email notifications for all starred features, using gmail filters, and using a dedicated account (like @chromium.org) with a separate inbox.

However, we could add a "[X]Send Email" checkbox to the editing form.  You could uncheck it for those edits that you are pretty sure that no one really would want to know about.  E.g., if you are making a pass over a bunch of features and updating http to https in some link or something like that.  I'll add a new GitHub issue to track that as an enhancement request.

models.py
```diff
@@ -1193,6 +1193,30 @@ def __init__(self, *args, **keyargs):
        self.fields[field].widget.attrs['required'] = 'required'
 
 
+class UserPref(DictModel):
+  """Describes a user's application preferences."""
+
+  email = db.EmailProperty(required=True)
+  notify_as_starrer = db.BooleanProperty(default=True)
```
aglaforge: I can guess, based on the field name what behavior notify_as_starrer implies.  However, it may be worth explicitly adding a comment to remove any potential for ambiguity.

notifier.py
```diff
@@ -196,6 +198,21 @@ def get_user_stars(self, email):
     logging.info('returning %r', feature_ids)
     return feature_ids
 
+  @classmethod
+  def get_feature_starrers(self, feature_id):
+    """Return a list of UserPref objects of all users that starred a feature."""
```
aglaforge: If I understand the logic correctly, this comment may be slightly misleading.  The return is the set of users who have starred a feature and also opted in to notifications.

jrobbins: OK, since I already landed this CL, I am following-up with a new CL:

PR #780, opened 2020-03-11
Implement star icons in lit-element code.

jrobbins:
This implements star icons instead of bell icons for users interest in individual features.

In this CL:
+ Changed the icons themselves
+ Now making XHR calls to our own server rather than firebase for these stars
+ Don't render star icons when the user is signed out
+ Provide an empty but valid JSON response for /features/star/set.

TODO for upcoming CLs:
+ Show disabled star icon when user is signed out.
+ Convert "bell" subscriptions to stars when users visit.


static/elements/chromedash-feature.js
```diff
@@ -178,7 +179,7 @@ class ChromedashFeature extends LitElement {
     this._fireEvent('filter-component', {val: e.currentTarget.innerText});
   }
 
-  subscribeToFeature(e) {
+  starFeature(e) {
```
liyangguang: nit: since it's staring and un-staring. rename it to "toggleStar"?

static/js-src/notifications.js
```diff
@@ -218,9 +218,45 @@ class PushNotifier {
   }
 }
 
+
+class StarService {
+  static getStars() {
+    return new Promise((resolve, reject) => {
```
liyangguang: This two methods looks a bit verbose. How about:
```
getStars() {
  const url = location.hostname == 'localhost' ?
    'https://www.chromestatus.com/features/star/list' :
    '/features/star/list';
  return fetch(url, {
    method: 'POST',
    headers: {'Content-Type': 'application/json'},
    body: JSON.stringify({}),
  })
  .then((res) => res.json())
  .then((res) => res.featureIds)
  .catch((error) => /* Do something if applicable? */);
}
```

static/elements/chromedash-feature.js
jrobbins: Done.

Issue #783, opened 2020-03-12
Need Escalation Path for Email address of Chromium Owners


jpmedley:
A recent change to the code means that bouncing email address are automatically removed from recipient list. Unfortunately, not all emails can be treated this way. For the purposes of explanation, I'll divide email addresses into three groups.

* Those for people who have subscribed to receive updates.
* Owners of features that have already shipped.
* Owners of features that have not shipped.

An escalation path is needed for email address in the last category that bounce. Someone needs to determine whether the feature in question needs a new owner or whether its status entry should be marked, "No longer pursuing" or "No active development."

**This feature request is to set the mechanism for notifying a particular person or persons that a new owner needs to be found.** The mechanism should allow for multiple persons to be notified. The question of who should be notified is larger than this group. (Since I've already been doing this for the last several years, I'm happy to keep doing this until a decision is made.)

![image](https://user-images.githubusercontent.com/8204171/76566895-f5b7dd00-646a-11ea-8dc3-e86b4837cad7.png)

jpmedley:
cc: @carolrivero

PR #788, opened 2020-03-13
Escalate bounces to an admin mailing list.

jrobbins:
This addresses issue #783 by sending an email to notify a mailing list whenever there is a bounce, because bouncing outbound email indicates that we might need a new owner for a component or feature.

### DIFF
--- a/notifier.py
+++ b/notifier.py
@@ -108,7 +108,8 @@ def convert_reasons_to_task(addr, reasons, email_html, subject):
 
 def make_email_tasks(feature, is_update=False, changes=[]):
   """Return a list of task dicts to notify users of feature changes."""
-  feature_watchers = models.FeatureOwner.all().filter('watching_all_features = ', True).fetch(None)
+  feature_watchers = models.FeatureOwner.all().filter(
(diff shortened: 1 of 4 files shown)


notifier.py
```diff
@@ -465,12 +465,25 @@ def post(self):
...
+    message = mail.EmailMessage(
+        sender='Chromestatus <admin@%s.appspotmail.com>' % settings.APP_ID,
+        to=settings.BOUNCE_ESCALATION_ADDR,
+        subject=msg,
+        html='See subject.  Check logs.')
+    message.check_initialized()
+    if settings.SEND_EMAIL:
+      message.send()
+
+
```
jstenback: Nit: maybe cut down on the empty lines here?

jrobbins: Fixed.  And, I also fixed some line-length style problems in this file.

jrobbins:
Joe,  I added the entire copy of the original outbound email's text to the escalation email.  It will not have a link to the feature that was modified because it is just the plain text rather than HTML, however, it will have the full name of the feature.

jpmedley:
Thanks!

PR #796, opened 2020-03-26
Implement settings page.

jrobbins:
This addresses issue #754.

### DIFF
--- a/models.py
+++ b/models.py
@@ -1252,6 +1252,13 @@ class UserPref(DictModel):
     return user_prefs
 
 
+class UserPrefForm(forms.Form):
+  notify_as_starrer = forms.BooleanField(
+      required=False,
+      label='Notify as starrer',
+      help_text='Send you notification emails for features that you starred?')
+
+
 class AppUser(DictModel):
   """Describes a user for whitelisting."""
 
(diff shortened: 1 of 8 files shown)


users.py
```diff
@@ -73,8 +73,49 @@ def _delete(self, user_id):
         found_user.delete()
 
 
+class SettingsHandler(common.ContentHandler):
+
+  def load_user_pref(self):
```
aglaforge: This seems like a pretty common case.  Would it make sense to create some type of base handler class that includes this logic that folks like SettingsHandler could depend on?

models.py
```diff
@@ -1252,6 +1252,13 @@ def get_prefs_for_emails(self, emails):
     return user_prefs
 
 
+class UserPrefForm(forms.Form):
+  notify_as_starrer = forms.BooleanField(
+      required=False,
```
aglaforge: I'm curious.  It seems like one potential option here would be to make this field required and set a default to False.  Is there an advantage to making it optional vs required w/ a default?

users.py
jrobbins: I moved it into models.UserPref so that it can be called from any handler that needs it.
I don't like to build up too much logic in base classes when there is another good place to put it.

models.py
jrobbins: I don't know if the concept of required really makes sense for a checkbox because it is always either checked or unchecked, it cannot be undefined.   If I remove required=False then django adds a line of red text that says "This field is required." which seems awkward on that tiny page that has only one checkbox on it.

PR #798, opened 2020-03-26
Add 'On hold' to Chromium statuses.

jpmedley:


jpmedley:
@jrobbins Can you look at this today. I actually need to set this on a feature.

jrobbins:
Sorry to be late reviewing that change.  I went ahead and landed it and I can deploy later today.

jpmedley:
Did you deploy this? I don't see it in the interface even after a hard
reload.
Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195

jpmedley:
Thank you.
Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195

jrobbins:
I hit problems with 'npm' on my cloudtop instance that day. That is resolved now.  I will wait until your other CL lands and then deploy today.

jrobbins:
This is live now.

PR #803, opened 2020-03-31
Switch to new release info URL and show dates again.

jrobbins:
The new backend for schedule info accepts the same query string parameters as the old backend, so the change here is minimal.

I also added pretty-printing to the JSON to aid debugging.  It increases the size of the HTML page a small amount but whitespace compresses well so I think it is worth it.

### DIFF
--- a/schedule.py
+++ b/schedule.py
@@ -39,11 +39,25 @@ def fetch_chrome_release_info(version):
            'mstone=%s' % version)
     result = urlfetch.fetch(url, deadline=60)
     if result.status_code == 200:
-      data = json.loads(result.content)['mstones'][0]
-      del data['owners']
-      del data['feature_freeze']
-      del data['ldaps']
-      memcache.set(key, data)
+      try:
+        data = json.loads(result.content)['mstones'][0]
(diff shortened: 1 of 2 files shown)


jrobbins:
The app doesn't seem to do arithmetic on the milestone numbers.  It just gets a list of milestones from omaha and then fetches release info for each of those.

Demo link:
https://2020-03-31-dot-cr-status-staging.appspot.com/features/schedule

aglaforge:
Got it, there are currently no 82 releases currently on OmahaProxy, so all the look-ups succeed.  Does the current code handle look-up failures gracefully?

jrobbins:
It caches on success, so that should paper over some transient errors.  Failures would show up in the logs as 500s, so I would see them in Google Cloud Console.

jrobbins:
As per our discussion, I will make it fail gracefully in the situation where chromiumdash does not successfully give release info for some milestone.

jrobbins:
OK, this version detects a non-200 response from the backend, or any failure in parsing the JSON from the backend, and simply does not display dates in that case.

PR #825, opened 2020-05-07
Partial implementation of feature editing guide UX.

jrobbins:
This is the first CL in this line of work. The form processing is not working yet, I just wanted to pin down some of the pieces before implementing more: source file organization, form representation, process representation, near-term URL space design, and new process overview element.

Since these pages deal with one feature at a time, I am taking a very simple approach of including the needed data as JSON in the page rather than building an API.  Also I chose to define the Blink Intents process in python and pass that to the client-side code because I am expecting that we will add more server-side logic and validation.

### DIFF
--- a/guide.py
+++ b/guide.py
@@ -1,0 +1,434 @@
+from __future__ import division
+from __future__ import print_function
+
+# -*- coding: utf-8 -*-
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import datetime
+import json
+import logging
+import os
+import re
+import sys
+import webapp2
+from bs4 import BeautifulSoup
+from HTMLParser import HTMLParser
+from xml.dom import minidom
+from django import forms
+
+# Appengine imports.
+from google.appengine.api import memcache
+from google.appengine.api import users
+from google.appengine.ext import db
+from google.appengine.api import taskqueue
+
+# File imports.
+import common
+import guideforms
+import models
+import processes
+import settings
+
+
+# Forms to be used for each stage of the process.
+STAGE_FORMS = {
+    models.INTENT_NONE: guideforms.Incubate,
+    models.INTENT_IMPLEMENT: guideforms.Prototype,
+    models.INTENT_EXPERIMENT: guideforms.DevTrial,
+    models.INTENT_EXTEND_TRIAL: guideforms.OriginTrial,
+}
+
+class FeatureNew(common.ContentHandler):
+
+  def get(self, path):
+    user = users.get_current_user()
+    if user is None:
+      return self.redirect(users.create_login_url(self.request.uri))
+
+    # TODO(ericbidelman): This creates a additional call to
+    # _is_user_whitelisted() (also called in common.py), resulting in another
+    # db query.
+    if not self._is_user_whitelisted(user):
+      #TODO(ericbidelman): Use render(status=401) instead.
+      #self.render(data={}, template_path=os.path.join(path + '.html'), status=401)
+      common.handle_401(self.request, self.response, Exception)
+      return
+
+    template_data = {
+        'overview_form': guideforms.OverviewForm(),
+        }
+
+    self._add_common_template_values(template_data)
+
+    self.render(data=template_data, template_path=os.path.join(path + '.html'))
+
+  def post(self, path):
+    user = users.get_current_user()
+    if user is None or (user and not self._is_user_whitelisted(user)):
+      common.handle_401(self.request, self.response, Exception)
+      return
+
+    feature = models.Feature(
+        category=int(self.request.get('category')),
+        name=self.request.get('name'),
+        intent_stage=models.INTENT_NONE,
+        summary=self.request.get('summary'),
+        impl_status_chrome=models.NO_ACTIVE_DEV,
+        visibility=models.WARRANTS_ARTICLE,
+        standardization=models.EDITORS_DRAFT,
+        web_dev_views=models.DEV_NO_SIGNALS)
+    key = feature.put()
+
+    # TODO(ericbidelman): enumerate and remove only the relevant keys.
+    memcache.flush_all()
+
+    redirect_url = '/guide/edit/' + str(key.id())
+    return self.redirect(redirect_url)
+
+
+class ProcessOverview(common.ContentHandler):
+
+  def get(self, path, feature_id):
+    user = users.get_current_user()
+    if user is None:
+      # Redirect to public URL for unauthenticated users.
+      return self.redirect(self.DEFAULT_URL + '/' + feature_id)
+
+    # TODO(ericbidelman): This creates a additional call to
+    # _is_user_whitelisted() (also called in common.py), resulting in another
+    # db query.
+    if not self._is_user_whitelisted(user):
+      common.handle_401(self.request, self.response, Exception)
+      return
+
+    template_data = {
+        'overview_form': guideforms.OverviewForm(),
+        'process_json': json.dumps(
+            [stage._asdict() for stage in processes.BLINK_PROCESS]),
+        'progress_so_far': [],
+        }
+
+    f = models.Feature.get_by_id(long(feature_id))
+    if f is None:
+      return self.redirect(self.ADD_NEW_URL)
+
+    progress_so_far = []  # An unordered list of progress item strings.
+    progress_so_far = ['Explainer', 'API design']
+
+    # Provide new or populated form to template.
+    template_data.update({
+        'feature': f,
+        'feature_id': f.key().id,
+        'feature_json': json.dumps(f.format_for_template()),
+        'progress_so_far': json.dumps(progress_so_far),
+    })
+
+    self._add_common_template_values(template_data)
+    self.render(data=template_data, template_path=os.path.join(path + '.html'))
+
+
+class FeatureEditStage(common.ContentHandler):
+
+  LAUNCH_URL = '/admin/features/launch'
+
+  INTENT_PARAM = 'intent'
+  LAUNCH_PARAM = 'launch'
+
+  def __FullQualifyLink(self, param_name):
+    link = self.request.get(param_name) or None
+    if link:
+      if not link.startswith('http'):
+        link = db.Link('http://' + link)
+      else:
+        link = db.Link(link)
+    return link
+
+  def __ToInt(self, param_name):
+    param = self.request.get(param_name) or None
+    if param:
+      param = int(param)
+    return param
+
+  def __get_blink_component_from_bug(self, blink_components, bug_url):
+    if blink_components[0] == models.BlinkComponent.DEFAULT_COMPONENT and bug_url:
+      result = urlfetch.fetch(bug_url)
+      if result.status_code == 200:
+        soup = BeautifulSoup(result.content, 'html.parser')
+        components = soup.find_all(string=re.compile('^Blink'))
+
+        h = HTMLParser()
+        return [h.unescape(unicode(c)) for c in components]
+    return blink_components
+
+  def get(self, path, feature_id, stage_id):
+    user = users.get_current_user()
+    if user is None:
+      # Redirect to public URL for unauthenticated users.
+      return self.redirect(self.DEFAULT_URL + '/' + feature_id)
+
+    if not self._is_user_whitelisted(user):
+      common.handle_401(self.request, self.response, Exception)
+      return
+
+    template_data = {
+        }
+
+    f = models.Feature.get_by_id(long(feature_id))
+    if f is None:
+      return self.redirect(self.ADD_NEW_URL)
+
+    detail_form_class = STAGE_FORMS.get(int(stage_id), models.FeatureForm)
+
+    # Provide new or populated form to template.
+    template_data.update({
+        'feature': f,
+        'feature_id': f.key().id,
+        'feature_form': detail_form_class(f.format_for_edit()),
+    })
+
+    if self.LAUNCH_PARAM in self.request.params:
+      template_data[self.LAUNCH_PARAM] = True
+    if self.INTENT_PARAM in self.request.params:
+      template_data[self.INTENT_PARAM] = True
+
+    self._add_common_template_values(template_data)
+
+    self.render(data=template_data, template_path=os.path.join(path + '.html'))
+
+  def post(self, path, feature_id, stage_id):
+    user = users.get_current_user()
+    if user is None or (user and not self._is_user_whitelisted(user)):
+      common.handle_401(self.request, self.response, Exception)
+      return
+
+    spec_link = self.__FullQualifyLink('spec_link')
+
+    explainer_links = self.request.get('explainer_links') or []
+    if explainer_links:
+      explainer_links = filter(bool, [x.strip() for x in re.split('\\r?\\n', explainer_links)])
+
+    bug_url = self.__FullQualifyLink('bug_url')
+    intent_to_implement_url = self.__FullQualifyLink('intent_to_implement_url')
+    origin_trial_feedback_url = self.__FullQualifyLink('origin_trial_feedback_url')
+
+    ff_views_link = self.__FullQualifyLink('ff_views_link')
+    ie_views_link = self.__FullQualifyLink('ie_views_link')
+    safari_views_link = self.__FullQualifyLink('safari_views_link')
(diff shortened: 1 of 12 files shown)


jrobbins:
Demo:
Sign in with your @google.com account.
https://2020-05-07-dot-cr-status-staging.appspot.com/guide/edit/5637476211228672
and
https://2020-05-07-dot-cr-status-staging.appspot.com/guide/new

Submitting the "new" form will create a feature entity on the staging server.  But, the other editing forms are not working yet and might not have the right set of fields yet.

static/elements/chromedash-process-overview.js
```diff
@@ -0,0 +1,71 @@
...
+    this.start = 'Start';
+    this.preview = 'Preview';
+  }
+
+  render() {
+    let featureId = this.feature.id;
+    return html`
+     <table cellspacing=0>
+       <tr>
+         <th width="100">Stage</th>
```
liyangguang, line 37: Nit: Use css here too. `style="width: 100px"`

static/elements/chromedash-process-overview.js
```diff
@@ -0,0 +1,71 @@
...
+  static styles = style;
+
+  static get properties() {
+    return {
+      feature: {type: Object},
+      process: {type: Array},
+      progress: {type: Array},
+      revisit: {type: String},
+      start: {type: String},
+      preview: {type: String},
```
liyangguang, line 18: Looks like these 3 are just a const, and used only once.
1. Maybe just use the string in the template instead of have a variable for it?
2. If we want to keep string consts as variables, put them before the `class` - `const REVISIT_STRING = 'Revisit';` etc.

templates/guide/stage.html
```diff
@@ -0,0 +1,46 @@
...
+{% block content %}
+<section>
+  <form name="feature_form" method="POST" action="{{current_path}}">
+    <table>
+      {{ feature_form }}
+    </table>
+    <ul>
+      <li>
+        <input type="checkbox" name="intent_to_implement" id="id_intent_to_implement">
+        <label for="id_intent_to_implement">Generate "Intent to <span id="id_intent_stage_name"></span>" email on next page (you will copy &amp; paste it to send)</label>
```
liyangguang, line 30: Seems not used? Can it be removed?

jrobbins: Yes, but I didn't want to get into changing that part in this CL.  I added a TODO comment to redo that part and I will follow up with a separate CL.

static/elements/chromedash-process-overview.js
jrobbins: I removed these.  I had planned to use this element in on the new.html page with different words, but then I decided to leave it off.

static/elements/chromedash-process-overview.js
jrobbins: Done.

static/elements/chromedash-process-overview.js
```diff
@@ -0,0 +1,71 @@
...
+    this.progress = [];
+    this.revisit = 'Revisit';
+    this.start = 'Start';
+    this.preview = 'Preview';
+  }
+
+  render() {
+    let featureId = this.feature.id;
+    return html`
+     <table cellspacing=0>
```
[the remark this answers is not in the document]
jrobbins: Done.

jpmedley:
We need to discuss how to do user assistance. Is this part of the UI work we've contracted? This PR continues the current practice of adding instruction as [help text under each field](https://github.com/GoogleChrome/chromium-dashboard/pull/825/files#diff-d33e4dca2717b9c80afb9764ae206affR31).  I don't think many read this text. Whatever help is provided, I need to know when to review and how much time I should spend on it. I've added an item to this week's agenda.

guideforms.py
```diff
@@ -0,0 +1,308 @@
...
+        ('Which pieces of the API surface are you looking to gain insight on? '
+         'What metrics/measurement/feedback will you be using to validate '
+         'designs? Double check that your experiment makes sense given that '
+         'a large developer (e.g. a Google product or Facebook) likely '
+         'can\'t use it in production due to the limits enforced by origin '
+         'trials.\n\nIf Intent to Extend Origin Trial, highlight new/different '
+         'areas for experimentation. Should not be an exact copy of goals '
+         'from the first Intent to Experiment.')),
+
+    'experiment_timeline': forms.CharField(
```
jpmedley, line 215: Can we break this into separate begin and end fields?

guideforms.py
```diff
@@ -0,0 +1,308 @@
...
+
+  doc_links = ALL_FIELDS['doc_links']
+  standardization = ALL_FIELDS['standardization']
+  spec_link = ALL_FIELDS['spec_link']
+  tag_review = ALL_FIELDS['tag_review']
+  intent_to_implement_url = ALL_FIELDS['intent_to_implement_url']
+
+
+class DevTrial(forms.Form):
+
```
jpmedley, line 279: Why are there no begin and milestone files?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+            [stage._asdict() for stage in processes.BLINK_PROCESS]),
+        'progress_so_far': [],
+        }
+
+    f = models.Feature.get_by_id(long(feature_id))
+    if f is None:
+      return self.redirect(self.ADD_NEW_URL)
+
+    progress_so_far = []  # An unordered list of progress item strings.
+    progress_so_far = ['Explainer', 'API design']
```
aglaforge, line 129: Can remove the line instantiating the empty list (it's immediately replaced w/ another object).

guide.py
```diff
@@ -0,0 +1,434 @@
...
+}
+
+class FeatureNew(common.ContentHandler):
+
+  def get(self, path):
+    user = users.get_current_user()
+    if user is None:
+      return self.redirect(users.create_login_url(self.request.uri))
+
+    # TODO(ericbidelman): This creates a additional call to
```
aglaforge, line 61: There are lots of TODO's in the code pinned to ericbidelman.  If we want to retain them, they should get updated to reflect a current owner.

Are there opportunities for code reuse to improve maintainability?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+
+  LAUNCH_URL = '/admin/features/launch'
+
+  INTENT_PARAM = 'intent'
+  LAUNCH_PARAM = 'launch'
+
+  def __FullQualifyLink(self, param_name):
+    link = self.request.get(param_name) or None
+    if link:
+      if not link.startswith('http'):
```
aglaforge, line 153: Should this be https?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+    if param:
+      param = int(param)
+    return param
+
+  def __get_blink_component_from_bug(self, blink_components, bug_url):
+    if blink_components[0] == models.BlinkComponent.DEFAULT_COMPONENT and bug_url:
+      result = urlfetch.fetch(bug_url)
+      if result.status_code == 200:
+        soup = BeautifulSoup(result.content, 'html.parser')
+        components = soup.find_all(string=re.compile('^Blink'))
```
aglaforge, line 170: This seems a touch fragile and also likely won't work w/ Restrict issues.  Are there alternatives here?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+    safari_views_link = self.__FullQualifyLink('safari_views_link')
+    web_dev_views_link = self.__FullQualifyLink('web_dev_views_link')
+
+    # Cast incoming milestones to ints.
+    shipped_milestone = self.__ToInt('shipped_milestone')
+    shipped_android_milestone = self.__ToInt('shipped_android_milestone')
+    shipped_ios_milestone = self.__ToInt('shipped_ios_milestone')
+    shipped_webview_milestone = self.__ToInt('shipped_webview_milestone')
+    shipped_opera_milestone = self.__ToInt('shipped_opera_milestone')
+    shipped_opera_android_milestone = self.__ToInt('shipped_opera_android_milestone')
```
aglaforge, line 238: Do all milestones map to integers (e.g., Android letter releases)?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+
+    owners = self.request.get('owner') or []
+    if owners:
+      owners = [db.Email(x.strip()) for x in owners.split(',')]
+
+    doc_links = self.request.get('doc_links') or []
+    if doc_links:
+      doc_links = filter(bool, [x.strip() for x in re.split('\\r?\\n', doc_links)])
```
aglaforge, line 246: Nit: I'm surprised to see only new line based delimiters.

guide.py
```diff
@@ -0,0 +1,434 @@
...
+
+    devrel = self.request.get('devrel') or []
+    if devrel:
+      devrel = [db.Email(x.strip()) for x in devrel.split(',')]
+
+    try:
+      intent_stage = int(self.request.get('intent_stage'))
+    except:
+      logging.error('Invalid intent_stage \'{}\'' \
+                    .format(self.request.get('intent_stage')))
```
aglaforge, line 268: A little surprised that we throw in the middle of processing.  Seems like we'd want to aggregate errors to the end.

guide.py
```diff
@@ -0,0 +1,434 @@
...
+      feature.experiment_risks = self.request.get('experiment_risks')
+      feature.experiment_extension_reason = self.request.get('experiment_extension_reason')
+      feature.ongoing_constraints = self.request.get('ongoing_constraints')
+    else:
+      # Check bug for existing blink component(s) used to label the bug. If
+      # found, use the first component name instead of the generic "Blink" name.
+      try:
+        blink_components = self.__get_blink_component_from_bug(blink_components, bug_url)
+      except Exception:
+        pass
```
aglaforge, line 348: Should there at least be logging here (for this exception)?

guide.py
```diff
@@ -0,0 +1,434 @@
...
+      feature.experiment_goals = self.request.get('experiment_goals')
+      feature.experiment_timeline = self.request.get('experiment_timeline')
+      feature.experiment_risks = self.request.get('experiment_risks')
+      feature.experiment_extension_reason = self.request.get('experiment_extension_reason')
+      feature.ongoing_constraints = self.request.get('ongoing_constraints')
+    else:
+      # Check bug for existing blink component(s) used to label the bug. If
+      # found, use the first component name instead of the generic "Blink" name.
+      try:
+        blink_components = self.__get_blink_component_from_bug(blink_components, bug_url)
```
aglaforge, line 346: We likely don't do it very often, but this looks like an expensive call (that could fail, since it depends on a network fetch).  Any need for caching?

guide.py
jpmedley: I believe they do for all Chromium forks. In any case, MDN has compiled [a pretty complete list](https://github.com/mdn/browser-compat-data/tree/master/browsers) of known version numbers for major browsers.

aglaforge: Thanks for the MDN reference, that clarified my Android question.

This comment then becomes more of a nit.  I see a recent example where Safari did use a minor release (13.1).  I know it's come up in Chrome at least once (4.1 was the last that I recall). 
From that perspective, there really isn't anything stopping the various release teams from changing their versioning/ naming conventions.  

I suspect it's not worth a lot of code complexity, but moving away from int (to something like string) could afford value for future proofing.

processes.py
```diff
@@ -0,0 +1,90 @@
...
+ProcessStage = collections.namedtuple(
+    'ProcessStage',
+    'name, description, progress_items, incoming_stage, outgoing_stage')
+
+
+BLINK_PROCESS = [
+  ProcessStage(
+      'Start incubation',
+      'Create an initial WebStatus feature entry and kick off standards '
+      'incubation (WICG) to share ideas',
```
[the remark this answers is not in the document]
jrobbins: Let's discuss at our meeting this week.  If we can't be specific, then maybe we need a choice of a couple different processes where each one could have specific details, even if some are N/a for certain features.

processes.py
```diff
@@ -0,0 +1,90 @@
...
+      ['Feature name, description, owners, and other details',
+       'WICG discourse post',
+       'Create repo',
```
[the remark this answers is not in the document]
jrobbins: Done.

processes.py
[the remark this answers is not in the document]
jrobbins: Done.

processes.py
```diff
@@ -0,0 +1,90 @@
...
+      ],
+      models.INTENT_NONE, models.INTENT_IMPLEMENT),
+  ProcessStage(
+      'Start prototyping',
+      'Share an explainer doc and API. '
```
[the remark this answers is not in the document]
jrobbins: They were two bullet items in the "Detailed steps" column of Chris's go/chromium-process-deets spreadsheet.  I haven't seen the term "API explainer".  I just made up a one-line description of each stage myself, so I am open to whatever description Chris and everyone thinks makes sense.

processes.py
```diff
@@ -0,0 +1,90 @@
...
+      ['OT request',
+       'OT available',
+       'OT results',
+      ],
+      models.INTENT_EXTEND_TRIAL, models.INTENT_EXTEND_TRIAL),
+  ProcessStage(
+      'Prepare to ship',
+      'Lock in shipping milestone. Finalize docs and announcements. '
+      'Further standardization.',
+      ['Intent to Ship mail',
```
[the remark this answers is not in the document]
jrobbins: Changed in 3 places.  Chris's spreadsheet says "mail" but I agree that "email" is more precise.

processes.py
```diff
@@ -0,0 +1,90 @@
...
+       'Request to migrate incubation',
+       'TAG issues addressed',
+       '3 LGTMs',
```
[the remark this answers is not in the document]
jrobbins: Done.

templates/guide/edit.html
```diff
@@ -0,0 +1,44 @@
...
+<style>
+.delete-button {
+  margin-left: 20px;
+}
+</style>
+{% endblock %}
+
+{% block subheader %}
+<div id="subheader">
+  <h2>Edit feature: {{ feature.name }}</h2>
```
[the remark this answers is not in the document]
jrobbins: I'll add that as part of an upcoming CL.   For now, it is indicated by the location of the "Start" button, but I'll have it be explicitly stated also.

guideforms.py
jrobbins: I added a TODO for that change.  For this CL, I am just breaking up the existing form into smaller forms.  experiment_timeline was a single field on the old form and in our data.  So, splitting it would require migrating old data or coping with both formats.

guideforms.py
jrobbins: This CL only has 4 process phases out of the 9 in Chris's spreadsheet.  I'm planning on getting that much working before going for completeness.  I have added a TODO to make sure that every field in the old form is included.

guide.py
jrobbins: That second line is a placeholder for where I will need to add some logic to detect which progress items have been done so far.  I added a TODO comment to make that explicit.  I'll resolve that TODO in an upcoming CL.

guide.py
jrobbins: I deleted these two TODOs because I don't think the they are helpful.  But, as I touch other code, I will update the TODO owner as I go.

guide.py
jrobbins: This is existing code that I just copied and have not reworked yet.  I'll get to it in an upcoming CL.  We do want to encourage everyone to use https, however for this validation, startswith('http') will allow either http or https.  Our database probably has a lot of old entries with http that would fail to validate and prevent someone from editing any other field in that form for the feature.

guide.py
jrobbins: I added a TODO to use the Monorail API rather than scraping.

guide.py
jrobbins: Good points.  This is existing code that relates to the existing database schema, so it would take some work to change.   I added a TODO for that.

guide.py
jrobbins: I'm guessing that that was done that way because URLs can contain commas.   I'd prefer to leave that as-is unless there are indications that our users are having trouble with it.

guide.py
jrobbins: That's just logging that line to the GAE logs.   In the old form, I believe that is a <select> field, so users would need to go out of their way to submit a value that does not parse, in which case it defaults to stage 1 (Prototype).  When I rework the form processing, I'll make a helper function so that all the int fields are parsed and validated consistently.

guide.py
jrobbins: That's some sloppy existing code.  And, as mentioned above, I don't think that it can work these days.  That whole else-clause is actually for the feature creation code path, which I have implemented in the FeatureNew class now.  So, I just deleted the whole thing.

guide.py
jrobbins: Yeah, it would be challenging to do this right and make it reliable.   This is existing code that I don't think has been working for awhile.  I don't know if users even complained that it stopped working.  So, let's try leaving it out for a while, we can always add it back in if users say it should be a priority.

processes.py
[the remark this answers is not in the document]
jpmedley: I didn't give good feedback here. I was trying to ask for clarity on 'share an API'. Is that a design? Is that a prototype? It doesn't seem like either could be ready at this stage. That's why explainer is a requirement. @cwilso

guide.py
jpmedley: I have seen no evidence of that. But I have seen people assuming that the explainer and tracking bug fields can take multiple URLs. They cannot.

PR #838, opened 2020-06-03
Implement unlisted features

jrobbins:
In this CL:
+ Add a boolean field named unlisted
+ Include that field in the existing editing forms and the new guide forms and form handlers
+ Add a condition to get_chronological to exclude any unlisted features
+ Unit tests
+ Reduce HTTP cache time on JSON feeds 

I believe that a 600 second cache time was previously used because the feature list page was slow to load.  The number of features in the list was capped a few months ago, so that page loads in reasonable time even on a cache miss.  A shorter cache time makes setting the unlisted boolean take effect more quickly.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -394,6 +394,7 @@ class FeatureHandler(common.ContentHandler):
       feature.name = self.request.get('name')
       feature.intent_stage = intent_stage
       feature.summary = self.request.get('summary')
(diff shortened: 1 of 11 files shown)


jpmedley:
The help text looks good. 

I have a nit. Despite what your typing teacher may have told you, current practice is to have only one space after a period.

PR #840, opened 2020-06-05
Show unlisted features to users who could edit them

jrobbins:
As discussed today, this CL allows any user who is qualified to edit features to also see unlisted features.  Otherwise, it would be possible for user to lose the URL for a feature and not be able to get back to it.

In this CL:
+ Pass a show_unlisted parameter to get_chronological() to include unlisted features.  That parameter is determined by _is_user_whitelisted().
+ Move _is_user_whitelisted() into the base class so that it can be called from either a page handler or a JSON handler.
+ Make these feeds private so that one user does not see a feed cached for another user
+ Add indication that an unlisted feature is being shown.

### DIFF
--- a/common.py
+++ b/common.py
@@ -71,6 +71,26 @@ class BaseHandler(webapp2.RequestHandler):
     # Settings can't be global in python 2.7 env.
     logging.getLogger().setLevel(logging.DEBUG)
 
+  def _is_user_whitelisted(self, user):
+    if not user:
+      return False
+
+    is_whitelisted = False
+
+    if users.is_current_user_admin():
+      is_whitelisted = True
+    elif user.email().endswith('@chromium.org') or user.email().endswith('@google.com'):
+      is_whitelisted = True
+    else:
+      # TODO(ericbidelman): memcache user lookup.
+      query = models.AppUser.all(keys_only=True).filter('email =', user.email())
+      found_user = query.get()
+
+      if found_user is not None:
+        is_whitelisted = True
+
+    return is_whitelisted
+
 
 class JSONHandler(BaseHandler):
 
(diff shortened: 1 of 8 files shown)


common.py
```diff
@@ -71,6 +71,26 @@ def __init__(self, request, response):
...
 
+  def _is_user_whitelisted(self, user):
+    if not user:
+      return False
+
+    is_whitelisted = False
+
+    if users.is_current_user_admin():
+      is_whitelisted = True
+    elif user.email().endswith('@chromium.org') or user.email().endswith('@google.com'):
```
jpmedley, line 82: There are emails from other domains. I don't know if there are any unaffiliated contributors using @gmail.com. I do know there are email addresses from Intel, Samsung, Microsoft, and others.

models.py
```diff
@@ -784,7 +784,11 @@ def getSortingMilestone(feature):
         temp_feature_list.extend(feature_list[key])
       feature_list = temp_feature_list
 
-    return feature_list
+    allowed_feature_list = [
+        f for f in feature_list
```
jpmedley, line 788: "f for f in"?

static/elements/chromedash-feature.js
```diff
@@ -295,6 +295,10 @@ class ChromedashFeature extends LitElement {
       </hgroup>
       <section class="desc" @click="${this._togglePanelExpansion}">
         <summary>
+          ${this.feature.unlisted ?
+             html`<p><b>This feature is only shown in the feature list
+                        to users who could edit it.</b></p>
```
jpmedley, line 300:
```suggestion
                        to users with edit access.</b></p>
```

templates/feature.html
```diff
@@ -78,7 +78,8 @@ <h2>
 
     {% if feature.unlisted %}
     <section id="access">
-      <p><b>Not shown in feature list</b></p>
+      <p><b>This feature is only shown in the feature list to users who
+          could edit it.</b></p>
```
jpmedley, line 82:
```suggestion
      <p><b>This feature is only shown in the feature list to users with
          edit access.</b></p>
```

templates/guide/metadata.html
```diff
@@ -13,7 +13,8 @@
     </div>
     <div>Category: {{ feature.category }}</div>
     {% if feature.unlisted %}
-      <div>Not shown in feature list</div>
+    <div>This feature is only shown in the feature list
+         to users who could edit it.</div>
```
jpmedley, line 17:
```suggestion
         to users who with edit access.</div>
```

tests/models_test.py
```diff
@@ -118,6 +118,16 @@ def test_get_chronological__unlisted(self):
         ['feature one'],
         names)
 
+  def test_get_chronological__unlisted_shown(self):
+    """Unlisted features are still included for users allowed to edit."""
```
jpmedley, line 122:
```suggestion
    """Unlisted features are included for users with edit access."""
```

models.py
jstenback: Gotta love python :)

common.py
jpmedley: Right. I see it.

jrobbins:
I accepted all the suggested wording changes.

PR #842, opened 2020-06-05
Flesh out fast track, PSA, and deprecation processes.

jrobbins:
In this CL:
+ Fast track: split dev trials from origin trials
+ PSA: Redo all steps
+ Deprecation: Initial draft, added to new.html custom form element
+ Revised all names, descriptions, and progress items
+ Sort feature_type choices by ID so they appear in logical order rather than alphabetical in the metadata editing form

### DIFF
--- a/processes.py
+++ b/processes.py
@@ -72,10 +72,11 @@ BLINK_PROCESS_STAGES = [
 
   ProcessStage(
       'Dev trials and iterate on design',
-      'Publicize test-readiness.  Share sample code. '
+      'Publicize availablity for developers to try. '
+      'Provide sample code. '
       'Request feedback from browser vendors.',
-      ['DevRel publicity request',
-       'Samples',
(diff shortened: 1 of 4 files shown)


processes.py
```diff
@@ -72,10 +72,11 @@ def process_to_dict(process):
...
   ProcessStage(
       'Dev trials and iterate on design',
-      'Publicize test-readiness.  Share sample code. '
+      'Publicize availablity for developers to try. '
+      'Provide sample code. '
       'Request feedback from browser vendors.',
-      ['DevRel publicity request',
-       'Samples',
+      ['Samples',
+       'Draft API overview (may be on MDN)',
```
cwilso, line 79: Strike "(may be on MDN)" here, as it's misleading.  (Leave it in the existing feature track.)

jrobbins: Done.

PR #843, opened 2020-06-09
Define all stage-specific forms and their fields.

jrobbins:
In this CL:
+ Define Django form fields for all remaining Feature attributes in the old UI.
+ Define forms for each stage of each process, and put fields into them
+ Change the way that the Django form is looked up to make it more clear
+ Change the logic to decide which stage gets the "Revisit", "Start", and "Preview" options.

Discussed today but not in this CL: 
+ Making the same field optional on one stage-specific form and then required on a later stage-specific form.
+ Actions and advice
+ New feature fields, e.g., initial public proposal URL.

### DIFF
--- a/guideforms.py
+++ b/guideforms.py
@@ -326,7 +326,7 @@ ALL_FIELDS = {
         required=False, label='Tracking bug URL',
         help_text=
         ('Tracking bug url (https://bugs.chromium.org/...). This bug '
-         'should have "Type=Feature" set and be world readable.'
+         'should have "Type=Feature" set and be world readable. '
          'Note: This field only accepts one URL.')),
 
     'blink_components': forms.ChoiceField(
--- a/guideforms.py
+++ b/guideforms.py
@@ -293,7 +293,7 @@ ALL_FIELDS = {
         label='Platform Support Explanation', required=False,
         widget=forms.Textarea(attrs={'rows': 2, 'cols': 50, 'maxlength': 2000}),
         help_text=
-        ('Explanation for why this feature is, or is not, '
+        ('Explain why this feature is, or is not, '
          'supported on all platforms.')),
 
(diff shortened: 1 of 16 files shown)


guideforms.py
```diff
@@ -256,25 +262,147 @@
...
+    'visibility': forms.ChoiceField(
+      label='Developer visibility',
+      choices=models.VISIBILITY_CHOICES.items(),
+      initial=models.WARRANTS_ARTICLE,
+      help_text=('How much press / media / web developer buzz will this '
+                 'feature generate?')),
+
+    'search_tags': forms.CharField(
+        label='Search tags', required=False,
+        help_text='Comma separated keywords used only in search'),
```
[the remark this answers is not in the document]
jrobbins: For now, it doesn't seem to work at all.  I can look into fixing it for the search box on this site.  Also, eventually we will add these search tags to the feature detail page so that search engines will see them too.

models.py
```diff
@@ -88,22 +108,25 @@
...
-}
+INTENT_STAGES = collections.OrderedDict([
+  (INTENT_NONE, 'None'),
+  (INTENT_INCUBATE, 'Start incubating'),
+  (INTENT_IMPLEMENT, 'Start prototyping'),
+  (INTENT_EXPERIMENT, 'Dev trials'),
+  (INTENT_IMPLEMENT_SHIP, 'Evaluate readiness to ship'),
+  (INTENT_EXTEND_TRIAL, 'Origin trials'),
+  (INTENT_SHIP, 'Prepare to ship'),
+  (INTENT_REMOVE, 'Remove'),
```
cwilso, line 127: Need a "SHIPPED" stage

jrobbins: Added.

PR #848, opened 2020-06-16
Updated process for deprecation.

jrobbins:
In this CL:
+ Use the term "Deprecation trial"
+ Revise the steps for the deprecation process
+ Rename INTENT_REMOVE to INTENT_REMOVED because we are using INTENT_SHIP and then INTENT_REMOVED afterwards
+ Add links to launching features page for process documentation
+ Also, add progress item for vendor signals

Let me know if the "vendor signals" progress item is a bad idea.  I just felt that it helps clarify what is happening in that stage.


jrobbins:
Demo is live on staging.
Screenshot: https://screenshot.googleplex.com/4tzwpOwDm7P

PR #851, opened 2020-06-17
Implement detection of progress items

jrobbins:
In this CL:
+ Dictionary of progress detectors
+ Code to populate progress_so_far by calling progress detector functions
+ JS code change to use hyperlinks if they are safe
+ CSS tweaks to account for differences in width between "o" and checkmark
+ Turn off autocomplete for new feature name

### DIFF
--- a/processes.py
+++ b/processes.py
@@ -99,6 +99,7 @@ BLINK_PROCESS_STAGES = [
       'Work through a TAG review and gather vendor signals.',
       ['TAG review request',
        'Vendor signals',
+       'Doc links',
        'Documentation signoff',
        'Estimated target milestone',
       ],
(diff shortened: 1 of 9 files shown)


processes.py
```diff
@@ -318,3 +319,41 @@ def process_to_dict(process):
...
+    'Request signals':
+    lambda f: bool(
+        f.ff_views != models.NO_PUBLIC_SIGNALS or
+        f.safari_views != models.NO_PUBLIC_SIGNALS or
+        f.ie_views != models.NO_PUBLIC_SIGNALS),
+
+    'Estimated target milestone':
+    lambda f: bool(f.shipped_milestone),
+
+    'Code in Chromium':
```
aglaforge, line 353: Is there also a "Code in repo"

jrobbins: Yes, but I believe that refers to a GitHub repo or something for the WICG.  I don't think we have a field for the user to specify that yet.

PR #855, opened 2020-06-19
Minor text corrections

jpmedley:


jrobbins:
Joe, is anything more needed for this change?  
Do you want to "squash and merge" this change?  Or, I could do it for you, if you like.

jpmedley:
@jrobbins I'll go ahead and merge it. We should talk about procedure. I hadn't because I generally follow the rule not to merge my own stuff, ensuring that every change gets a second pair of eyes.

Issue #862, opened 2020-06-22
'On hold' status in Chromium is not displaying on rendered page. 


jpmedley:
**Describe the bug**
This shows what's in the status record:
![image](https://user-images.githubusercontent.com/8204171/85325036-c9efbe80-b47f-11ea-959f-8a932c67de80.png)
This shows what renders on the page:
![image](https://user-images.githubusercontent.com/8204171/85325069-db38cb00-b47f-11ea-9b87-04d45d7cf474.png)

**To Reproduce**
Click the edit icon on [this status entry](https://www.chromestatus.com/feature/5685965186138112)

'On hold' is the most recently added option for the 'Status in Chromium' field. I don't know whether we missed something or whether this is a regression affecting the feature in general.

I've already tried cache clearing as a possible solution.

jpmedley:
@jrobbins @jstenback

jpmedley:
I've downgraded this to P1 because changing the status to 'In development' was a reasonable fix for this feature. Incidentally, that also confirms that this bug is tied to the 'On hold' status.

jrobbins:
Looking at the code, there seems to be some if-then logic to display a small set of messages rather than displaying the chrome implementation status string directly.
https://github.com/GoogleChrome/chromium-dashboard/blob/master/templates/feature.html#L148

The git blame Information for those lines does not seem too helpful.

Do you know of any reason why we should not just display the chrome implementation string at that spot in the UI?

jpmedley:
I think I've worked out what the original implementation was trying to do. Bug #202 hints at it, but doesn't come out and say it.

Some users were selecting 'Enabled by default' as the status for deprecations and removals on the grounds that the code that deprecates or removes is enabled by default. 

This is a kind of confusion between the code point of view and the platform user point of view. This kind of POV error happens enough that I suspect I'll be correcting Chrome status entries regularly (though certainly not often) if the logic is removed. There may be a better way to do it though.

jrobbins:
Would it work to clarify it in the help text or field label to add the word "feature"?  

Old label: Status in Chromium

Proposed label: Feature implementation status in Chromium

There currently is no help text.  What do you think would be most helpful?

jpmedley:
I have little confidence in the utility of the help text. I could live with "Availability in Chromium". Make sure it gets changed in both the editable and display versions of the form. Do we need to run this by the committee? It certainly seems like something we should announce on blink-dev@ to avoid confusion.

[...]

PR #863, opened 2020-06-22
Directly display the Status in Chromium string.

jrobbins:
Let me know if you think this resolves issue #862 correctly.


jpmedley:
Please look at my comments on #862 before I proceed.

jrobbins:
I think this already landed as part of another PR.

PR #877, opened 2020-07-09
Fix migration of ie_views.

jrobbins:
This was mistakenly left off an earlier CL.


jpmedley:
Guys,

The syntax looks right, but I don't know enough about this part of the system to do anything like a real code review.

Issue #880, opened 2020-07-09
Intent email previews do not include fields that are conditioned on feature stage


jrobbins:
**Describe the bug**
The chromestatus app had logic in the intent_to_implement.html template to include various fields based on feature.intent_stage.   That stopped working when the stages were renamed.

**To Reproduce**
Sign in and visit, e.g., https://chromestatus.com/admin/features/launch/5191745052606464?intent

**Expected behavior**
It should have some sections for experiment risks, experiment timeline, etc.

**Additional context**
The conditions in the template should be using feature.intent_stage_int now so that they do not depend on the exact name of each stage.

jrobbins:
The immediate fix is live.  But, I prefer to keep this issue open until I rewrite that template to make it more maintainable by moving logic into python and adding a python unit test.

PR #881, opened 2020-07-10
Fix failing template conditions in intent emails.

jrobbins:
This should fix issue #880.
 
This is a very straightforward CL that substitutes each existing condition with updated code that should work.  I have also added a comment to refactor most of the decision logic into python where it would be more testable.

### DIFF
--- a/templates/blink/intent_to_implement.html
+++ b/templates/blink/intent_to_implement.html
@@ -24,6 +24,11 @@
   {{feature.name}}
 </div>
 
+{% comment %}
+  TODO(jrobbins): rather than checking intent_stage_int, that logic should be
+  in python code and passed to this template as a list of field names to show.
+{% endcomment %}
+
 <p>Body</p>
 <div class="email"><label>Contact emails</label>
 {% if not feature.owner %}None{% endif %}{% for owner in feature.owner %}<a href="mailto:{{owner}}">{% if forloop.last %}{{owner}}</a>{% else %}{{owner}}</a>,{% endif %}{% endfor %}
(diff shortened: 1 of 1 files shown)


jrobbins:
This change is already live on prod, so viewing a URL like
https://chromestatus.com/admin/features/launch/5191745052606464?intent
shows the result of this CL.

templates/blink/intent_to_implement.html
```diff
@@ -24,14 +24,19 @@
...
+
 <p>Body</p>
 <div class="email"><label>Contact emails</label>
 {% if not feature.owner %}None{% endif %}{% for owner in feature.owner %}<a href="mailto:{{owner}}">{% if forloop.last %}{{owner}}</a>{% else %}{{owner}}</a>,{% endif %}{% endfor %}
 
 <label>Explainer</label>
 {% for link in feature.explainer_links %}<a href="{{link}}">{{link}}</a>
 {% endfor %}
-{% if feature.intent_stage == "Ship" %}<label>Spec</label>
+{% if feature.intent_stage_int == 5 %}<label>Spec</label>
```
jpmedley, line 39: I get why you got rid of the string. Could you do this with constants, though? That would make it easier to work out what code is doing?

jrobbins:
This CL is just a direct translation of the old string constants to the
same underlying numeric constants so that it is clear that this is not a
logic change.

After this lands, I'll follow up with a rewrite of this template.  Using
constants in the template would be better than numbers, making it roughly
as clear as the original code, however that code was not very clear, and we
have no unit tests for template logic. I plan to do better by moving most
of the conditional logic into python and passing a flat list of sections to
show in the template.  That will allow me to test it with python unit tests.

templates/blink/intent_to_implement.html
aglaforge: Referencing the INTENT_SHIP var (or a value derived from it) would be ideal for the sake of maintainability (understand that plumbing those into the Django content/ model object may be tricky).

Barring that, having an explanation of what 5 means in the form of a comment, since that info is lost w/ this edit.

aglaforge: Perhaps the models.INTENT_STAGES object could be injected into the context map so that those values can be referenced.

https://github.com/GoogleChrome/chromium-dashboard/blob/aacad37c9269df336de7a5b966cffdfa52212c71/guide.py

Issue #874, opened 2020-06-25
Make 'Edit' look clickable


hwikyounglee:
**Guide UX Feedback**
Either we use link for 'Edit' or match color to #4580c0

![image](https://user-images.githubusercontent.com/7561395/85777942-ef2a3a00-b6d6-11ea-84ee-58e6698e3d21.png)

jpmedley:
Addressed by #884.

Issue #871, opened 2020-06-25
Consider moving the "Delete" link.


jpmedley:
It seems like it might be safer to move the "Delete" link away from the "Edit" link even if it does ask, "Are you sure?".

![image](https://user-images.githubusercontent.com/8204171/85743144-a8c5e280-b6b8-11ea-8895-fc7795f51c75.png)

jpmedley:
Addressed by #884.

PR #884, opened 2020-07-10
Make edit button look clickable, and separate delete button.

jrobbins:
This should resolve issue #874 and issue #871.


jrobbins:
Screen shot (when signed in as a site admin):
https://screenshot.googleplex.com/Bz7gQ3iJtvF

PR #885, opened 2020-07-17
Correct WebView instructions.

jpmedley:
@jrobbins Can you jump on this real quick. I just notice yesterday that the WebView instructions were wrong. We need this text until we can fix this for real.


jrobbins:
Thanks for noticing this and fixing it.

It is live on the site now.

jpmedley:
Thanks!

Issue #873, opened 2020-06-25
Add focus indicators on form controls


hwikyounglee:
**Guide UX Feedback**

https://chromestatus.com/guide/new
Use 'Tab' key to navigate the controls, and observe the focus indicator

Currently only the <select> controls show the focus indicator. 
Recommending we add focus indicators for other controls to help form-filling using keyboard. 

Thanks for looking into this!

jrobbins:
To clarify, the checkbox and radio buttons on that form are missing the indication of focus.

After investigating, I found that the normal browser focus indication had been set to "none" throughout the site.

jpmedley:
@jrobbins I was tempted to close this. We should probably wait for deployment and verification when closing bugs.

PR #888, opened 2020-07-22
Allow inputs to have defaul outline when focused.

jrobbins:
This should resolve issue #873. 

For some reason, the existing CSS had outline:none for all inputs.  That disabled the focus indicator that is normally provided by the browser.  Doing that is discouraged, so I removed it.
https://developer.mozilla.org/en-US/docs/Web/CSS/outline#Accessibility_concerns

Screenshots:
Radio button with focus:
https://screenshot.googleplex.com/gZcBjQJU696.png

Text field with focus:
https://screenshot.googleplex.com/WBtvEhKHgTJ.png

### DIFF
--- a/static/sass/forms.scss
+++ b/static/sass/forms.scss
@@ -61,7 +61,6 @@ table {
 }
 
 input, textarea {
-  outline: none;
   padding: 7px;
 }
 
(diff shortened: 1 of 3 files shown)


jrobbins:
Oh, I also removed a duplicate "Use old UI" link that was shown below the feature entry form.  We only need one link for that, it was added at the top of the form before our soft launch.

Issue #868, opened 2020-06-24
Version Number Boxes Need to Appear with "Status in Chromium" Field


jpmedley:
**Describe the bug**
In the new design, the version number box are in separate subforms. This is fine in most instances. Some status require that version numbers must be entered. 

**To Reproduce**
Edit a feature and open "Prepare to ship".
![image](https://user-images.githubusercontent.com/8204171/85612460-c1c58980-b60d-11ea-9999-7d7999816ad4.png)

**Expected behavior**
The version number boxes need to appear for the following statuses:
* Behind a flag
* Enabled by default
* Deprecated
* Removed
* Origin trial

**Additional context**
Leaving the new interface as is will lead to an increase in status entries being advanced to one of the above status without version numbers being added. This will:
* Provide incomplete information to web developers.
* Require me to write more emails asking developers to fill in this information.

jpmedley:
Here's an illustration from an update that come through recently. Notice that the intent stage has been advanced to "Prepare to ship". DevRel needs to know whether this is in 85 or later, but the form doesn't prompt him to fill that in.

![image](https://user-images.githubusercontent.com/8204171/85743916-4e795180-b6b9-11ea-911c-55e934eb8bf9.png)

jpmedley:
I got this back from the feature owner. 
![image](https://user-images.githubusercontent.com/8204171/85776856-f1d85f80-b6d5-11ea-8559-790e657806ae.png)

jpmedley:
@jrobbins Is there any chance you can prioritize this soon?

jrobbins:
I'll look into it today.

jrobbins:
This is being addressed in PR #1011.

jrobbins:
This was deployed.

jpmedley:
Yeah!

PR #890, opened 2020-07-22
Add shipped milestones to prepare-to-ship form

jrobbins:
This should resolve issue #868.
The problem was that the shipped milestone fields were not part of the "prepare to ship" form.

The JS code that shows or hides them based on the implementation status selector seems to work when those fields are present.  However, that JS code also shows the milestones for implementation statuses of "Browser intervention" and "On hold".  Do I need to hide the milestone fields for those values?

### DIFF
--- a/guideforms.py
+++ b/guideforms.py
@@ -650,10 +650,16 @@ class NewFeature_OriginTrial(forms.Form):
 class Any_PrepareToShip(forms.Form):
 
   field_order = (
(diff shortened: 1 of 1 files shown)


jpmedley:
I can code, but I'm not an engineer, so I'm never comfortable merging on this repo. I found no issues with this.

Thank you for jumping on this quickly.

Issue #892, opened 2020-07-23
Phase out use of service worker


jrobbins:
While service worker is awesome and appropriate for many web applications, this application is pretty small so it loads quickly even with service worker.  We don't need it any longer for push notifications, and we don't want the added complexity of that layer of caching.

PR #936, opened 2020-07-28
Remove service worker

jrobbins:
While service worker is awesome, it is not needed for this app.  Removing it will simplify debugging of JS and CSS issues.

This will resolve issue #892.


jrobbins:
Demo link:
https://2020-07-28-dot-cr-status-staging.appspot.com/

jpmedley:
What happens to all the people have installed this on desktops and home screens?

jrobbins:
> What happens to all the people have installed this on desktops and home screens?

I think that the icon disappears or it becomes a plain bookmark icon.  The install button was only ever offered on mobile, so I don't think many users have it.

I added to my phone's home screen and then updated the staging server.  The icon is still there, but it might disappear after 24 hours.  I looked through various specs and docs and failed to find what the expected behavior is.

Issue #939, opened 2020-07-29
Don't show red "This field is required" for <select> form fields


jrobbins:
**Describe the bug**
In the guide UX, after creating a feature than moving to the incubation form, the <select> for technical footprint shows a red warning message "this field is required" even though the user will alway be posting some value for that <select>.

**Expected behavior**
Don't clutter the page with unnecessary warnings.

PR #940, opened 2020-07-29
Make all ChoiceFields non-required.

jrobbins:
All Django form field widgets default to being required and display a red warning message if the underlying object has no value defined at the time that the form is populated based on the underlying Feature object.  However, that is moot for ChoiceFields because the widget always has some value selected anyway.

This should resolve issue #939.


jrobbins:
Ping.

Issue #935, opened 2020-07-24
Add a "shipped" stage at the end of the process to give a feeling of completion


jrobbins:
In our user study the first user said that they would like to see a "Shipped" stage in the process after all the other stages were completed.

This might differ from other stages in that:

1. The button for it might say "Set stage" or something like that instead of saying "Start" because there is no work to start working on in this stage.
1. All the fields associated with that stage would be review of values that should have already been filled in.  E.g., milestones.

PR #941, opened 2020-07-29
Add Shipped stage.

jrobbins:
We got user feedback that a final "Shipped" stage would make the process feel more complete.

This should resolve issue #935.

### DIFF
--- a/processes.py
+++ b/processes.py
@@ -131,10 +131,20 @@ BLINK_PROCESS_STAGES = [
        'TAG issues addressed',
        'Three LGTMs',
        'Updated vendor signals',
-       'Finalized target milestone',
+       'Updated target milestone',
       ],
       [('Draft Intent to Ship email', INTENT_EMAIL_URL)],
       models.INTENT_IMPLEMENT_SHIP, models.INTENT_SHIP),
+
+  ProcessStage(
+      'Ship',
+      'Update milestones and other information when the feature '
+      'actually ships.',
+      ['Finalized vendor signals',
+       'Finalized target milestone',
+      ],
+      [],
+      models.INTENT_SHIP, models.INTENT_SHIPPED),
   ]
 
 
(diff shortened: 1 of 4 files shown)


jrobbins:
Screenshots for a feature in the Shipped stage for
+ New feature incubation
  https://screenshot.googleplex.com/xeu475Rr0qR
+ Implementing an existing feature spec
  https://screenshot.googleplex.com/BV8OgspNVCd
+ Change to existing code
  https://screenshot.googleplex.com/M3QaoGgZkkd
+ Deprecation (process is unchanged)
  https://screenshot.googleplex.com/6X6TwsakWdk

processes.py
```diff
@@ -131,10 +131,20 @@ def process_to_dict(process):
...
-       'Finalized target milestone',
+       'Updated target milestone',
       ],
       [('Draft Intent to Ship email', INTENT_EMAIL_URL)],
       models.INTENT_IMPLEMENT_SHIP, models.INTENT_SHIP),
+
+  ProcessStage(
+      'Ship',
+      'Update milestones and other information when the feature '
+      'actually ships.',
```
jpmedley, line 142:
```suggestion
      'ships.',
```

processes.py
```diff
@@ -195,10 +205,20 @@ def process_to_dict(process):
...
-       'Finalized target milestone',
+       'Updated target milestone',
       ],
       [('Draft Intent to Ship email', INTENT_EMAIL_URL)],
       models.INTENT_EXPERIMENT, models.INTENT_SHIP),
+
+  ProcessStage(
+      'Ship',
+      'Update milestones and other information when the feature '
+      'actually ships.',
```
jpmedley, line 216:
```suggestion
      'ships.',
```

processes.py
```diff
@@ -243,10 +263,20 @@ def process_to_dict(process):
...
-       'Finalize target milestone',
+       'Updated target milestone',
       ],
       [('Draft Intent to Ship email', INTENT_EMAIL_URL)],
       models.INTENT_EXPERIMENT, models.INTENT_SHIP),
+
+  ProcessStage(
+      'Ship',
+      'Update milestones and other information when the feature '
+      'actually ships.',
```
jpmedley, line 274:
```suggestion
      'ships.',
```

jpmedley:
@jrobbins Do you have this staged anywhere?

jrobbins:
Ping.  I'd like to include this in the next round of the user study, which is today.

processes.py
cwilso: Maybe "when the feature has actually shipped" is the right phrase.  I get why Joe doesn't like "actually" here, but I worry we're not being directive enough without it?

jrobbins: I'll keep what I have for this CL just so that I can land this for a UX test today, but I am happy to make whatever wording change we agree on in an upcoming CL.

PR #943, opened 2020-07-31
 CODE_OF_CONDUCT

gaushikmr:


jpmedley:
@gaushikmr, thank you for pointing that we are missing a code of conduct. We are actually required to use the Chromium code of conduct. #944

gaushikmr:
@jpmedley  shall i do that

PR #958, opened 2020-08-07
Label the summary field, reduce comment rows, and reorder fields.

jrobbins:
This addresses some of the items in your "WebStatus - Notes" doc.

### DIFF
--- a/guideforms.py
+++ b/guideforms.py
@@ -482,7 +482,8 @@ class NewFeatureForm(forms.Form):
 
   field_order = (
       'name', 'summary', 'owner',
-      'unlisted', 'blink_components', 'category',
+      'blink_components', 'category',
+      'unlisted',
       'feature_type')
   name = ALL_FIELDS['name']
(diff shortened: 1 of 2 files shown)


jrobbins:
Screenshots
https://screenshot.googleplex.com/wwGieoKsN8m
and
https://screenshot.googleplex.com/obHHZ6453A1

aglaforge:
It seems odd to have Unlisted be in the middle of the form, since it doesn't describe the features meta data.  Perhaps it should be closer to the submit button (which feels more like state of the data object).

jrobbins:
> It seems odd to have Unlisted be in the middle of the form, since it doesn't describe the features meta data. Perhaps it should be closer to the submit button (which feels more like state of the data object).

I have moved it lower, however it cannot be below the feature type radio buttons because that is hard-coded HTML in the template rather than in the form definition in python.  I kind of liked it at the top because people need to know about access restrictions on the data that they are working with as they read down the page.

PR #959, opened 2020-08-07
Use side-by-side striped tables for the feature metadata

jrobbins:
This addresses one of the items in your "WebStatus - Notes" doc about using a mutli-column layout to make more efficient use of space on the metadata panel in the Guide UX.  Also, the fields are reordered to match the ordering of my other CL that changed the field order.

### DIFF
--- a/static/sass/guide.scss
+++ b/static/sass/guide.scss
@@ -9,6 +9,18 @@
   margin-top: 20px;
 }
 
+.flex-cols {
+  display: flex;
+  flex-flow: row wrap;
+  align-items: flex-start;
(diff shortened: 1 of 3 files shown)


jrobbins:
Screenshot
https://screenshot.googleplex.com/JdfqqxTrdcN

jpmedley:
Looks good.

jpmedley:
This doesn't look like it can reflow on mobile.

jrobbins:
> This doesn't look like it can reflow on mobile.

Good point.  I'll fix that before submitting.

PR #966, opened 2020-08-14
Add fields to capture email addresses of API owners that LGTM things.

jrobbins:
Add fields for i2e and i2s LGTMs to capture the email addresses of the API owners who approve these threads.

Eventually these might be automatically captured by having the API owners sign in and click a "LGTM" button in the tool themselves.  But, for now, we will ask feature developers to type in these email addresses from the thread.

Add progress detectors to set the checkmarks for these fields.

Also, for the existing spec process, there was no field to capture the URL of the intent to ship thread, so I added that.


jrobbins:
@googlebot I consent.

jrobbins-at-chromium:
@googlebot I consent

guideforms.py
```diff
@@ -335,6 +335,20 @@
...
+        widget=forms.EmailInput(
+            attrs={'multiple': True, 'placeholder': 'email'}),
+        help_text=('Full email address of API owner who LGTM\'d the '
+                   'Intent to Experiment email thread.')),
+
+    'i2s_lgtms': forms.EmailField(
+        required=False, label='i2s LGTMs by',
+        widget=forms.EmailInput(
+            attrs={'multiple': True, 'placeholder': 'email, email, email'}),
+        help_text=('Full email addresses of API owner who LGTM\'d '
```
cwilso, line 349: ownerS.

(Also, does this need to say comma-delimited?)

models.py
```diff
@@ -1188,14 +1192,27 @@ class FeatureForm(forms.Form):
...
+      widget=forms.EmailInput(
+          attrs={'multiple': True, 'placeholder': 'email'}),
+      help_text=('Full email address of API owner who LGTM\'d the '
+                 'Intent to Experiment email thread.'))
+
+  i2s_lgtms = forms.EmailField(
+      required=False, label='i2s LGTMs by',
+      widget=forms.EmailInput(
+          attrs={'multiple': True, 'placeholder': 'email, email, email'}),
+      help_text=('Full email addresses of API owner who LGTM\'d '
```
cwilso, line 1213: owners.

jrobbins:
I was thinking that users would need to type in email addresses for now.  I want to support an "LGTM" button in the tool, but I don't think that that can be done in a fast-follow.  It could end up in Q4.  I'll file an issue for it now.

guideforms.py
```diff
@@ -335,6 +335,20 @@
         ('If your feature was available as an Origin Trial, link to a summary '
          'of usage and developer feedback. If not, leave this empty.')),
 
+    'i2e_lgtms': forms.EmailField(
+        required=False, label='i2e LGTM by',
+        widget=forms.EmailInput(
+            attrs={'multiple': True, 'placeholder': 'email'}),
+        help_text=('Full email address of API owner who LGTM\'d the '
```
jpmedley, line 342: Who's supposed to enter them? The feature owner or the API owner who gave the LGTM?

processes.py
```diff
@@ -203,10 +204,10 @@ def process_to_dict(process):
       'Prepare to ship',
       'Lock in shipping milestone. Finalize docs and announcements. '
       'Further standardization.',
-      ['Intent to Ship email',
-       'Three LGTMs',
-       'Documentation signoff',
+      ['Documentation signoff',
        'Updated target milestone',
+       'Intent to Ship email',
+       'Three i2s LGTMs',
```
jpmedley, line 210:
```suggestion
       'Three LGTMs for the intent to ship',
```

jpmedley: Do we have room to spell this out? Abbreviations of any kind are barriers to new participants, even if they seem obvious to us.

jpmedley: If we have the room, I can change the others, but won't object if you get there first.

guideforms.py
jrobbins: Done.

models.py
jrobbins: Done.

guideforms.py
jrobbins: The feature owners for now.  Eventually we will have API owners click an LGTM button, but that will come later.

jrobbins:
OK, I think I have addressed all the feedback.  Please take another look.

guideforms.py
aglaforge: My strong suspicion is that feature owners will not come back to the tool to do this.  It's manual work (copying and pasting from the email thread) that won't help them move the feature along (i.e., there's no enticement).

If we pick this path, over a checkbox, an LGTM button (even in a light weight form) will likely have to come pretty quickly.

processes.py
jrobbins: Done.

PR #970, opened 2020-08-17
Remove footprint field from the UI.

jrobbins:
As discussed on an email thread, remove the the technical footprint field from the UI.

Existing features that have this value set in the datastore should retain the existing value, out of sight.  We might want to use this existing value, e.g., to guess at a feature_type for existing features that don't have a feature_type specified.

The technical footprint value is not mentioned in the current feature launch process:
https://www.chromium.org/blink/launching-features


jrobbins-at-chromium:
@googlebot I consent

models.py
```diff
@@ -584,10 +567,6 @@ def format_for_template(self, version=None):
           'text': STANDARDIZATION[self.standardization],
           'val': d.pop('standardization', None),
         },
-        'footprint': {
```
jpmedley, line 587: Does the comma on the previous line need to go away?

jrobbins: Nope. Trailing commas are considered good in Python because it allows for addition and removal of other dictionary items with a cleaner-looking diff (like the one in this CL).
https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas
and
https://google.github.io/styleguide/pyguide.html

PR #971, opened 2020-08-18
Don't default new feature owner to the wrong address.

jrobbins:
This problem was previously reported by aglaforge.

The existing code initialized a class variable with the user who was signed in at the time that that python code was first loaded in that GAE server instance.  So, for example, if jrobbins tested the app after deploying it, the feature owner would always default to jrobbins until that GAE instance was shut down.

The new code provides no default in the form field definition, but it provides an initial value when the form instance is created.  Only the new UI has this logic.  The old UI will have that field blank.

### DIFF
--- a/guide.py
+++ b/guide.py
@@ -97,8 +97,10 @@ class FeatureNew(common.ContentHandler):
       common.handle_401(self.request, self.response, Exception)
       return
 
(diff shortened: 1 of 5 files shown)


jrobbins-at-chromium:
@googlebot I consent

PR #977, opened 2020-08-25
Callout elements for on-page help

jrobbins:
This should resolve issue #978.

This implements a simple on-page help UI functionality that displays a message in a word-bubble near a UI element on a page.  The message has an "X" icon to dismiss the message, and it will not appear for that user again.  This feature is called "cue cards" because that is the name of a similar feature in Monorail.

The first use of a callout cue card in ChromeStatus is one that tells users the meaning of progress checkmarks.

### DIFF
--- a/static/elements/chromedash-callout.js
+++ b/static/elements/chromedash-callout.js
@@ -1,0 +1,81 @@
+import {LitElement, css, html} from 'lit-element';
+
+class ChromedashCallout extends LitElement {
+  static get properties() {
+    return {
+      targetEl: {type: Element},
+      // TODO(jrobbins): Support sides other than "south".
+      side: {type: String}, // "north", "south", "east", or "west"
+      hidden: {type: Boolean},
+    };
+  }
+
+  constructor() {
+    super();
+    this.side = 'south';
+    this.hidden = false;
+  }
+
+  static get styles() {
+    return css`
+      :host {
+        /* TODO(jrobins): Put this into a global theme.css file. */
+        --blue-900: #174EA6;
+
+        --callout-bg-color: var(--blue-900);
+        --callout-text-color: white;
+      }
+      #bubble {
+        position: relative;
+        display: inline-block;
+        background: var(--callout-bg-color);
+        border-radius: 8px;
+        padding: 8px;
+        max-width: 24em;
+        color: var(--callout-text-color);
+        font-weight: bold;
+      }
+      #bubble[hidden] {
+        display: none;
+      }
+      #bubble:after {
+        content: '';
+        position: absolute;
+        border: 20px solid transparent;
+      }
+      /* Bubble will be located on the south side of targetEl. */
+      #bubble.south:after {
+        top: 0;
+        left: 50%;
+        width: 0;
+        height: 0;
+        border-bottom-color: var(--callout-bg-color);
+        border-top: 0;
+        margin-left: -20px;
+        margin-top: -20px;
+     }
+     #closebox {
+       float: right;
+       margin-left: 8px;
+     }
+    `;
+  }
+
+  close() {
+    this.hidden = true;
+  }
+
+  render() {
+    return html`
+      <div id="bubble" class="${this.side}" ?hidden=${this.hidden}>
+        <iron-icon id="closebox"
+          icon="chromestatus:close" @click=${this.close}></iron-icon>
+        <div style="margin: 4px">
+          <slot></slot>
+        </div>
+      </div>
+    `;
+  }
+}
+
+customElements.define('chromedash-callout', ChromedashCallout);
--- a/static/elements/chromedash-callout.js
+++ b/static/elements/chromedash-callout.js
@@ -11,7 +11,7 @@ class ChromedashCallout extends LitElement {
       top: {type: Number},
       left: {type: Number},
       cue: {type: String}, // String to send to the server when dismissed.
-      dismissedCues: {attribute: false, type: Array},
+      dismissedCues: {attribute: false, type: Object},
     };
   }
 
@@ -22,7 +22,7 @@ class ChromedashCallout extends LitElement {
     this.top = 0;
     this.left = 0;
     this.signedIn = false;
-    this.dismissedCues = [];
+    this.dismissedCues = {};
   }
 
   connectedCallback() {
--- a/static/elements/chromedash-callout.js
+++ b/static/elements/chromedash-callout.js
@@ -11,7 +11,7 @@ class ChromedashCallout extends LitElement {
       top: {type: Number},
       left: {type: Number},
       cue: {type: String}, // String to send to the server when dismissed.
-      dismissedCues: {attribute: false},
+      dismissedCues: {attribute: false, type: Array},
     };
   }
 
--- a/static/elements/chromedash-callout.js
+++ b/static/elements/chromedash-callout.js
@@ -3,17 +3,40 @@ import {LitElement, css, html} from 'lit-element';
 class ChromedashCallout extends LitElement {
   static get properties() {
     return {
-      targetEl: {type: Element},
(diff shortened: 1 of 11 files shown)


static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
+import {LitElement, css, html} from 'lit-element';
+
+class ChromedashCallout extends LitElement {
+  static get properties() {
+    return {
+      signedin: {type: Boolean},
```
liyangguang, line 6: nit: `isSignedIn`. and `targetId` the next line.

Please ignore if following Google styleguide is not necessary in this repo.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+    this.dismissedcues = [];
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    if (this.dismissedcues.includes(this.cue)) {
+      // Don't show the cue because the user has already dismissed it.
+      return;
+    }
+    let targetEl = this.parentNode.querySelector('#' + this.targetid);
```
liyangguang, line 34:
```
try {
  this.attach(this.parentNode.querySelector('#' + this.targetid));
} catch (error) {
   console.error('Failed to attach target', error);
}
```

1. querySelector can throw an error when this.target is invalid
2. nit: use console.error.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+        margin-top: -20px;
+     }
+     #closebox {
+       float: right;
+       margin-left: 8px;
+     }
+    `;
+  }
+
+  contentOffsetTop(el) {
```
liyangguang, line 87: It looks like getting the Y position of an element on the page?

`el.getBoundingClientRect().top + window.scrollY` might be better.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+    let offset = 0;
+    while (el.offsetParent && el != this.offsetParent) {
+      offset += el.offsetTop;
+      el = el.offsetParent;
+    }
+    return offset;
+  }
+
+  contentOffsetLeft(el) {
```
liyangguang, line 96: Same here, using `getBoundingClientRect()` can simplify it.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+      /* Bubble will be located on the south side of target element. */
+      #bubble.south:after {
+        top: 0;
+        left: 40px;
+        width: 0;
+        height: 0;
+        border-bottom-color: var(--callout-bg-color);
+        border-top: 0;
+        margin-left: -20px;
```
liyangguang, line 77: I think margins of this absolute positioned element can be achieved by updating `left: 20px; top: -20px` above.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+    while (el.offsetParent && el != this.offsetParent) {
+      offset += el.offsetLeft;
+      el = el.offsetParent;
+    }
+    return offset;
+  }
+
+  attach(el) {
+    if (this.side == 'south') {
+      let targetBot = this.contentOffsetTop(el) + el.offsetHeight;
```
liyangguang, line 107: (optional) nit: Use more meaningful name for `targetBot` and `attach()`.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+    /* TODO(jrobbins): Implement support for other sides. */
+    this.hidden = false;
+  }
+
+  dismiss() {
+    this.hidden = true;
+    if (this.signedin) {
+      window.CuesService.dismissCue(this.cue);
+    }
```
liyangguang, line 119: Does it do anything if not signed in? Might worth adding a comment explaining, or log an error if it shouldn't happen.

static/elements/chromedash-callout.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  render() {
+    return html`
+      <div id="bubble" class="${this.side}" ?hidden=${this.hidden}
+          style="top:${this.top}px; left:${this.left}px;">
+        <iron-icon id="closebox"
+          icon="chromestatus:close" @click=${this.dismiss}></iron-icon>
+        <div style="margin: 4px">
```
liyangguang, line 128: Nit: Add a class for the style.

static/elements/chromedash-process-overview.js
```diff
@@ -138,6 +142,18 @@ class ChromedashProcessOverview extends LitElement {
          </tr>
        `)}
      </table>
+
+    ${Object.keys(this.progress).length ? html`
+      <chromedash-callout
+        cue="progress-checkmarks" targetid="progress-header" signedin
```
liyangguang, line 148: Is the value for `signedin` missing intentionally?

static/elements/chromedash-process-overview.js
```diff
@@ -138,6 +142,18 @@ class ChromedashProcessOverview extends LitElement {
...
+        .dismissedcues=${this.dismissedcues}>
```
liyangguang, line 149: Is the `.` in front of this line intentional?

static/elements/chromedash-callout.js
jrobbins: Done.  I thought that they needed to match the attribute name.  But, they can be camel case here.

static/elements/chromedash-callout.js
jrobbins: Done. Comment added to explain that that case can happen, but it is a no-op.

static/elements/chromedash-process-overview.js
jrobbins: That's a shorthand for a boolean property being true, but I added ="true" to make it more explicit.

static/elements/chromedash-process-overview.js
jrobbins: Yes, that sets the property of the chromedash-callout element without going through an attribute.

static/elements/chromedash-callout.js
jrobbins: Done.  Actually an ID.

static/elements/chromedash-callout.js
jrobbins: Done.

static/elements/chromedash-callout.js
jrobbins: Done.  Thanks.

static/elements/chromedash-callout.js
jrobbins: That calculation was the part of the CL that I struggled with the most.  I considered using getBoundingClientRect, but then I wasn't sure about the coordinate system reference point.  The callout itself is position:absolute, which means that its position is relative to its lowest positioned ancestor.  Whereas, getBoundingClientRect and window.scrollY would be relative to the document.  So, wouldn't I need to dynamically create the chromedash-callout element as a child of the document rather than somewhere deep in the DOM?

For now, I am sticking with what I have because it works.  I added a TODO to come back to try a simpler design.

static/elements/chromedash-callout.js
jrobbins: That is simpler.  Thanks!

jrobbins-at-chromium:
@googlebot I consent.

static/elements/chromedash-callout.js
liyangguang: getBoundingClientRect is relative to the viewport, so it shouldn't be affected by position property on itself, or its parent/ancestors.

SGTM to keep it as is, and revisit it in the future.

Issue #982, opened 2020-08-26
There is no clear way to exit the Guide UX edit page or stage forms


jrobbins:
**Guide UX Feedback**
During a user test, a user told us that it was unclear how to exit the Guide UX pages.  There is a "cancel" button, but the user preferred to navigate out rather than cancel. 

For the overall edit UX, there is currently no way to navigate back to the feature detail page.  Instead the user must click the "All features" tab at the top of the page, which may not be exactly what they wanted.

PR #983, opened 2020-08-26
Add simple "back" navigation out of guide pages.

jrobbins:
This should address issue #982.


jrobbins:
Note, that .delete-button rule was not being used because it was on a page that does not have the feature delete button.  I think it was there due to a copy and paste error that was not caught in an earlier review.

Issue #990, opened 2020-08-28
Offer a flat view of all fields in the guide UX


jrobbins:
**Guide UX Feedback**
During user testing, a user requested the ability to see and edit the complete list of fields on one page.   While the main point of the Guide UX is to teach and reinforce user awareness of the development process, it is also reasonable that users sometimes want to just dive in to edit one field regardless of the process.

PR #991, opened 2020-08-28
Implement edit-all page

jrobbins:
This should resolve issue #990.

In this CL:
+ Go meta on the creation of the forms so that we have far fewer lines of code in guideforms.py.
+ Define new forms for Flat_* that cover all fields.
+ Add a new page handler and template that shows the concatenated list of flat forms with section headers.
+ Style tweaks.
+ Make all fields non-required because we had previously planned that, and required fields make editing a single field on the edit-all page very tedious.
+ Add a "Final review" row in the process overview with a link to the edit-all page.

### DIFF
--- a/guide.py
+++ b/guide.py
@@ -80,6 +80,18 @@ STAGE_FORMS = {
         },
 }
 
+# Forms to be used on the "Edit all" page that shows a flat list of fields.
+# [('Section name': form_class)].
+FLAT_FORMS = [
+    ('Feature metadata', guideforms.Flat_Metadata),
+    ('Indentify the need', guideforms.Flat_Identify),
+    ('Prototype a solution', guideforms.Flat_Implement),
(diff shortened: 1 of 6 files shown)


jrobbins:
Screenshots:

Link to "edit all" page:
https://screenshot.googleplex.com/7nUAGnToAsBgoT4

Edit all page (zoomed out):
https://screenshot.googleplex.com/4Wk2S4P7zfEPWXh

jpmedley:
This looks good. I'm glad it's broken down by stage. 

Maybe change the link name to 'Edit any field" instead of 'Edit all fields'.

jrobbins:
> Maybe change the link name to 'Edit any field" instead of 'Edit all fields'.

I like the idea, but I just tried it, and "Edit any field" sounds like the user is going to need to select which one field to edit.  So, let's stick with "Edit all fields" for now and see if users do OK with that.

Issue #993, opened 2020-09-01
Make the Guide UX the default


jrobbins:
The Guide UX will become the default, but the old UI will still be available.
We will track usage of the old UI and shut it down when usage is very low.

PR #994, opened 2020-09-02
Make guide UX the default

jrobbins:
This should resolve issue #993.

In this CL:
+ Change the new feature and editing icon links from /admin/feature to /guide.
+ Fix: Add the initial_public_proposal field to the edit-all page.
+ Fix: Make the various Cancel buttons actually cancel the edit and reset the form or navigate back to the overview.


jrobbins:
Note: This CL is already live.

Issue #866, opened 2020-06-24
"Experiment Timeline" should be two numeric fields.


jpmedley:
**Is your feature request related to a problem? Please describe.**
Expectecd experimental timelines are something we may need to sort on or perform operations on. For example, we could autodisplay an origin trial's end date based on its end version number. Currently, this information is entered as a text field which ensures inconsistency in how timeline versions are record. This means that any operations based on this field would be unreliable, inconsistently applied, or both.

![image](https://user-images.githubusercontent.com/8204171/85568173-51574200-b5e6-11ea-9013-d93b7a43a042.png)

**Describe the solution you'd like**
Two boxes that only allow numbers.

PR #1011, opened 2020-09-11
Split timeline into numeric fields and separate impl fields

jrobbins:
This is to address parts of Joe's proposal for version numbers
https://docs.google.com/document/d/1PuUsfK0wbMqen0R6EBOuo_H33TA2ks3nQ8WdR0oVlCU/edit?ts=5f5a2ba7#

In this CL:
+ Deprecate the Experiment Timeline text field and replace it with numeric start and end milestone number fields
 - (The Experiment Timeline field is still shown, but disabled so that people can see it for existing features)
+ Reorganize the stage-specific forms into two parts: standardization of the feature definition, and implementation in chromium.
 - Use a checkbox to advance impl_status_chrome to a new value when that is expected
 - Importantly: The Dev Trials stage that sets impl_status_chrome to BEHIND_A_FLAG does not ask for a webview milestone, but the Ship stage does have it.
+ Add a flag_name field for features behind a flag
+ Adjust help text on existing shipped_milestone fields to remove references
+ Organize the flat field list to have standardization and implementation fields in the same order as the stage forms, but don't double the number of cards.

### DIFF
--- a/models.py
+++ b/models.py
@@ -150,7 +150,7 @@ IMPLEMENTATION_STATUS = OrderedDict()
 IMPLEMENTATION_STATUS[NO_ACTIVE_DEV] = 'No active development'
 IMPLEMENTATION_STATUS[PROPOSED] = 'Proposed'
 IMPLEMENTATION_STATUS[IN_DEVELOPMENT] = 'In development'
-IMPLEMENTATION_STATUS[BEHIND_A_FLAG] = 'Behind a flag'
+IMPLEMENTATION_STATUS[BEHIND_A_FLAG] = 'In developer trial (Behind a flag)'
 IMPLEMENTATION_STATUS[ENABLED_BY_DEFAULT] = 'Enabled by default'
 IMPLEMENTATION_STATUS[DEPRECATED] = 'Deprecated'
 IMPLEMENTATION_STATUS[REMOVED] = 'Removed'
@@ -1013,6 +1013,7 @@ class Feature(DictModel):
   shipped_android_milestone = db.IntegerProperty()
   shipped_ios_milestone = db.IntegerProperty()
   shipped_webview_milestone = db.IntegerProperty()
+  flag_name = db.StringProperty()
 
   owner = db.ListProperty(db.Email)
   footprint = db.IntegerProperty()  # Deprecated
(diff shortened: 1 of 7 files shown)


jrobbins:
Demo at  https://661b1e4-dot-cr-status-staging.appspot.com

Screenshots:
https://screenshot.googleplex.com/7ufLobnyTaym66Y
https://screenshot.googleplex.com/8VnHUgG5aTXqsqx
https://screenshot.googleplex.com/3QHLHtgB27wFhNM
https://screenshot.googleplex.com/4t4HUcDipN9tSoa

models.py
```diff
@@ -1109,15 +1114,15 @@ def __init__(self, *args, **kwargs):
 class FeatureForm(forms.Form):
 
   SHIPPED_HELP_TXT = ('First milestone to ship with this '
-                      'status. Applies to: Enabled by default, Behind a flag, '
-                      'Origin trial, Browser Intervention, and Deprecated. If '
+                      'status. Applies to: Enabled by default, In developer trial, '
+                      'Browser Intervention, and Deprecated. If '
```
jpmedley, line 1118: How are we dealing with removals?

models.py
```diff
@@ -1109,15 +1114,15 @@ def __init__(self, *args, **kwargs):
...
                       'the flag is \'test\' rather than \'experimental\' set '
                       'status to In development.')
 
   SHIPPED_WEBVIEW_HELP_TXT = ('First milestone to ship with this status. '
                               'Applies to Enabled by default, Browser '
                               'Intervention, and Deprecated.\n\n NOTE: for '
-                              'statuses Behind a flag and Origin trial this '
+                              'statuses In developer trial and Origin trial this '
```
jpmedley, line 1125: Wouldn't this just go away?

jpmedley (changes requested):

models.py
jrobbins: We still need that constant for the old UI.  However, since you pointed this out, I changed the help text for the new UI to just use SHIPPED_HELP_TXT for webview.  There is no need to tell the user to leave it blank because the field is not offered until they get past dev trials.

models.py
jrobbins: For a deprecation:
+ At the "Dev trial for deprecation" stage we prompt for all milestones other than webview
https://screenshot.googleplex.com/C8qBt8JbujayiCw
+ At the "Prepare for deprecation trail" stage we ask for OT start and end milestones
https://screenshot.googleplex.com/AkiauAUbLE56Q8k
+ At the "Prepare to ship" and "Remove code" stages we prompt for all "shipped" milestones.
https://screenshot.googleplex.com/5ZdArhYDDkbfHto
https://screenshot.googleplex.com/JxTPRamWUwYGVQv

Does that sound OK?

jpmedley (approved):

jrobbins:
@cwilso ping.  Please let me know if you think this is a good idea.

cwilso (approved):
I think this looks good.

jpmedley:
Implements #866

Issue #984, opened 2020-08-27
Platform version numbers shown on inappropriate status.


jpmedley:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
"In development" status shows platform version numbers.

![image](https://user-images.githubusercontent.com/8204171/91484016-660dbb80-e85d-11ea-9b4a-0e7ac82d4797.png)

**To Reproduce**
1. In any issue, change the **Status in Chromium** field to "Enabled by default", "Behind a flag", or "Origin trial". Add version numbers to the provided boxes. Save the entry.
1. Edit the issue. Change the **Status in Chromium** field to "In development". Save the entry.

**Expected behavior**
The version number fields should not appear at all. 

**Additional context**
The current behavior assumes that if the `this.feature.browsers.chrome.*` have values that they should be displayed. It would probably be better to base the display on the value of `this.feature.browsers.chrome.status.text`.

PR #1012, opened 2020-09-11
Hide milestones in feature list entries for prelaunch features.

jrobbins:
This should resolve issue #984.

In the feature list, a feature entry no longer show milestone values when the feature is in a prelaunch stage.

My understanding is that displaying these milestones is confusing because they are not relevant before launch, even if a developer entered them.

### DIFF
--- a/static/elements/chromedash-feature.js
+++ b/static/elements/chromedash-feature.js
@@ -106,7 +106,8 @@ class ChromedashFeature extends LitElement {
       'No active development',
       'Proposed',
       'In development',
-      'Behind a flag',  // TODO(jrobbins): Update when PR #1011 lands.
+      // TODO(jrobbins): Update when we change value in models.py.
+      'In developer trial (Behind a flag)',
       'Origin trial',
       'On hold',
     ];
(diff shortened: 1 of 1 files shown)


static/elements/chromedash-feature.js
```diff
@@ -111,6 +101,19 @@ class ChromedashFeature extends LitElement {
     return vendors + webdevs + standards;
   }
 
+  _getIsPreLaunch() {
+    const PRE_LAUNCH_STATUSES = [
+      'No active development',
+      'Proposed',
+      'In development',
+      'Behind a flag',
```
jpmedley, line 109: Shouldn't this be 'Developer trial.'

jpmedley (changes requested):

static/elements/chromedash-feature.js
jrobbins: That change to models.py is still pending in a different CL.  However, thank you for pointing out that this will need to change after that lands.  I'll add a TODO here.

jpmedley:
You should probably also test whether these items show up when navigating via the sidebar. To state another way, if a future version number is hidden, will I still see the feature in the list/search for that version number? 

image

cwilso (approved):
LGTM, pending answer to Joe's question about if prelaunch features show up in milestone nav when hidden.

jpmedley (approved):

cwilso (commented):
LGTM.

cwilso:
(Sorry, getting review notifications I'm having trouble getting rid of - don't mind me.  :)

PR #1014, opened 2020-09-14
Increase instance size for more RAM.

jrobbins:
One of our metrics cron jobs has been running out of RAM, so this CL increases it.

We normally only have a handful of instances running, so the increased cost of the new instance size is not signifiant.


jpchase (approved):
LGTM

PR #1016, opened 2020-09-14
Allow generation of any intent email at any time.

jrobbins:
Previously, intent email generation was always done based on the current state of the feature entry.  That meant that, e.g., a user could not preview the Intent to Ship email before the feature reached that stage, and users could not go back to look at the Intent to Prototype email template after the feature had progressed.  Now we allow viewing of any intent email for a selected stage at any time.

In this CL:
+ Add the stage_id placeholder to the URLs for the actions that link to intent email templates
+ JS code to insert the appropriate stage_id value into the action URL
+ Add stage_id as an optional parameter to the intent email template handler
+ Add logic to use the supplied stage_id, or default to feature.intent_stage for users following old URLs
+ Updated unit tests

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -231,7 +231,7 @@ VIEW_FEATURE_URL = '/feature'
 class IntentEmailPreviewHandler(common.ContentHandler):
   """Show a preview of an intent email, as appropriate to the feature stage."""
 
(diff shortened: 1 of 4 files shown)


aglaforge (approved):
Nice, thank you!

Issue #989, opened 2020-08-27
Next steps for the Blink launch process don't say what email address to send the email to


sky2:
I created a new feature and got to the stage about sending an email. It doesn't actually say what email address to use.

jpmedley:
We'll address this. In the meantime the email address is blink-dev@chromium.org.

jrobbins:
This is live now.  The template now says to send the email to blink-dev.

PR #1018, opened 2020-09-15
Add fields to intent emails

jrobbins:
In this CL:
+ Add fields for blink component, launch bug URL, search tags, initial public proposal,  and TAG review status.
+ Move existing links to previous intent email thread to the bottom and add missing fields.
+ Reorganize fields to put summary and related fields in a more logical order at the top.
+ Add a line to explicitly say that these emails go to blink-dev. (issue #989)
+ Add a warning when viewing the intent email template for an unlisted feature.
+ Add whitespace at bottom to work-around floating footer (that footer will be redone soon)


templates/blink/intent_to_implement.html
```diff
@@ -35,21 +54,17 @@ <h4>Contact emails</h4>
     {% endfor %}
   {% endif %}
 
+  {% if feature.search_tags %}
+    <br><br><h4>Search tags</h4>
+    {{feature.search_tags |join:", "}}
+  {% endif %}
+
```
aglaforge, line 61: I'm curious why we'd include search tags in the intents mail.

aglaforge: Should we link to a query in chromestatus?

aglaforge (approved):
Approved, but two questions re: search terms.

templates/blink/intent_to_implement.html
jrobbins: I made them links to queries.  I am thinking that the value of including this field is so that people can find related features for comparison.

PR #1022, opened 2020-09-17
Implement soft delete for features.

jrobbins:
Instead of immediately deleting data from datastore, we will now mark feature entries as deleted.
Entries marked as deleted do not show up in queries or when the user attempts to view an individual feature.

There is no UI for undelete yet, but a feature that was accidentally marked as deleted could be unmarked via the Datastore pages on the Google Cloud Console.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -460,8 +460,10 @@ class FeatureHandler(common.ContentHandler):
       if feature is None:
         return self.redirect(self.request.path)
 
+      # TODO(jrobbins): implement undelete UI.  For now, use cloud console.
       if 'delete' in path:
-        feature.delete()
+        feature.deleted = True
+        feature.put()
         memcache.flush_all()
         return # Bomb out early for AJAX delete. No need to redirect.
 
--- a/models.py
+++ b/models.py
@@ -999,6 +999,8 @@ class Feature(DictModel):
   summary = db.StringProperty(required=True, multiline=True)
   origin_trial_feedback_url = db.LinkProperty()
   unlisted = db.BooleanProperty(default=False)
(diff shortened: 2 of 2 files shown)


jstenback (approved):
LGTM

admin.py
```diff
@@ -460,8 +460,10 @@ def post(self, path, feature_id=None):
       if feature is None:
         return self.redirect(self.request.path)
 
+      # TODO(jrobbins): implement undelete UI.  For now, use cloud console.
       if 'delete' in path:
-        feature.delete()
+        feature.deleted = True
```
aglaforge, line 465: Could we make this into a more generic status attribute?  That would open the door for some future tool changes (e.g., proposals, drafts, duping, etc.).

jrobbins: I came back to implement this, but then I remembered a good reason to use a separate boolean for deleted even if we use an enum for some other future state values:  When you undelete, you want the object to go back to the state that it was in before it was deleted.  Using a state enum would overwrite the previous state enum value when deleting, so the information is no longer there at the time of undeleting.

aglaforge: Got it.  Taking the example of duping, transitioning the a deleted state would be meaningfully lossy and we'd want to restore the original pre-deleted state.  To preserve that we'd need to introduce some type of historic field and synchronize, which is more complex than just having a separate state + a boolean for deleted.

aglaforge: In that case, let's call adding a formal document/ feature state as a separate, independent workitem.

jrobbins: OK, I will keep the deleted boolean in this CL and add a TODO to add an entry_state enum in the future.

Issue #969, opened 2020-08-17
Summary and other text fields should show variable-width fonts during editing.


jpmedley:
**Describe the bug**
The use of a fixed-width font in free-form text fields leads some users to think they have press enter every time their cursor approaches the right edge of the input box. I've already edited two summaries to remove the errant crlf characters.

jrobbins:
This might have been fixed by other changes.  I don't see fixed-width fonts being used now.

Is there any field where you still see fixed-width fonts?

jpmedley:
I'm still seeing this:

![image](https://user-images.githubusercontent.com/8204171/93783425-858bd000-fbe0-11ea-82ba-4596a591062d.png)

jrobbins:
I must have been looking in the wrong place when I wrote that.
I see the fixed-width fonts during editing now.
I will make this change now.

PR #1026, opened 2020-09-21
Use proportional font "system-ui" in textarea elements.

jrobbins:
This should resolve issue #969.

Rather than specify roboto, I am using "system-ui" which is the browser default for most inputs and <select> elements.  It is a proportional font on mac and chromeos.


jrobbins:
screenshots:
chrome on macOS: https://screenshot.googleplex.com/4UDZK6NRqerx6xH
chrome on chromeos: https://screenshot.googleplex.com/APKp7NkAKqSg9Sy

jpmedley (approved):

PR #1033, opened 2020-09-23
Refactor sass into lit-css: userlist, timeline, color-status.

jrobbins:
More refactoring from sass into lit-css.


static/elements/chromedash-color-status.js
```diff
@@ -20,6 +18,42 @@ class ChromedashColorStatus extends LitElement {
...
 
+  static get styles() {
+    return [
+      SHARED_STYLES,
+      css`
+      span {
+        background-color: rgb(255,0,0);
+        border-radius: 50%;
+        display: inline-block;
+        height: 10px; // Default.
```
liyangguang, line 29: Should the comment be `/* default */`? (`// comment` is a sass syntax, not css syntax. Does lit-element css follow the syntax of css?)

liyangguang (approved):

static/elements/chromedash-color-status.js
jrobbins: Fixed.

Thanks for catching this.  I looked for any other instances of the same problem and found none.

lit-css does follow the syntax of css.  The use of // was a syntax error, but it just didn't result in any visual differences in this case because the properties that were ignored were overridden by other values anyway.

Issue #806, opened 2020-04-07
Comments box is escaping markup


jpmedley:
Here's what I'm seeing:

![image](https://user-images.githubusercontent.com/8204171/78694200-2ac01f80-78b1-11ea-85cc-aa088f42ccc7.png)
 
Here's what's in the editing view:

![image](https://user-images.githubusercontent.com/8204171/78694260-3dd2ef80-78b1-11ea-9626-9633971d00b6.png)

mustaqahmed:
Thanks for flagging this.  Done clarifying the comment, copied the decisive comment from the bug.

Issue #999, opened 2020-09-03
Comments field is incorrectly escaped


jrobbins:
**Describe the bug**
In this feature 
https://chromestatus.com/features/5078049450098688
the comments include a link that is shown with visible markup.

The user only typed the bare URL.  Our UI component converted it to an <a> tag, but then rendered that as plain text.

**Expected behavior**
If we urlize text, we should keep track of the fact that it is markup and not let it be escaped afterwards.

The defect seems to be in chromedash-feature.js which calls urlize() but then renders the result as a regular string rather than as an html-string.  Probably the function _getCommentHtml() should construct and return a lit-html object.

jpmedley:
It seems like this should apply to any free-form entry field, summary and motivation specifically.

PR #1038, opened 2020-09-24
Fix urlize() usage for feature comments by defining autolink().

jrobbins:
This should resolve issues #806 and #999.


jrobbins:
I will redo this CL to clean up the diff.

PR #1040, opened 2020-09-25
Add help text as placeholder.

jpmedley:


models.py
```diff
@@ -1136,6 +1136,12 @@ class FeatureForm(forms.Form):
                               'statuses In developer trial and Origin trial this '
                               'MUST be blank.')
 
+SUMMARY_PLACEHOLDER_TXT = (
+    'NOTE: This text describes this feature in the eventual beta release post '
+    'as well as possibly in other external places externally.&#10;&#10;'
```
jrobbins, line 1141: I don't think that &#10; works in this context.  Some of our existing code using \n to make a line break.  That works in Chrome, but not safari, which is probably still OK so long as the text is readable when it runs together.  A combination of \n and a space is probably better.

jrobbins (changes requested):

cwilso (approved):
LGTM (with Jason's suggested changes)

models.py
jpmedley: When I was playing around with this in DevTools, the page wouldn't display \n, but it would display `&#10;.

guideforms.py
```diff
@@ -23,14 +23,18 @@
...
 
-
 SHIPPED_HELP_TXT = (
     'First milestone to ship with this status. Applies to: Enabled by '
     'default, Behind a flag, Origin trial, Browser Intervention, and '
     'Deprecated. If the flag is \'test\' rather than \'experimental\' set '
     'status to In development. If the flag is for an origin trial set status '
     'to Origin trial.')
 
+SUMMARY_PLACEHOLDER_TXT = (
```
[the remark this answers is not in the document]
jpmedley: Done. After this is merged, I'll submit a separate PR to deal with SHIPPED_HELP_TEXT.

jrobbins (changes requested):

jpmedley:
Everything's done. PTAL

models.py
```diff
@@ -1136,6 +1136,12 @@ class FeatureForm(forms.Form):
                               'statuses In developer trial and Origin trial this '
                               'MUST be blank.')
 
+SUMMARY_PLACEHOLDER_TXT = (
```
jrobbins, line 1139: The travis build is failing because this line needs to be indented more to align with the other constants.

jrobbins (changes requested):

models.py
jpmedley: I fixed it. I hope.

guideforms.py
```diff
@@ -53,14 +51,12 @@
 
     'summary': forms.CharField(
         required=True,
-        widget=forms.Textarea(attrs={'cols': 50, 'maxlength': 500}),
+        widget=forms.Textarea(
+            attrs={'cols': 50, 'maxlength': 500, 'placeholder': models.SUMMARY_PLACEHOLDER_TXT}),
```
jrobbins, line 55: This needs to be models.FeatureForm.SUMMARY_PLACEHOLDER_TXT.
Sorry to give the wrong example before.

jrobbins (changes requested):

jpmedley:
PTAL.

jrobbins (approved):

jrobbins:
Perfect: "All checks have passed".  Thanks for working through these details.

jpmedley:
Should I merge now? (I'm never sure. On the other projects I submit to, we never merge our own PRs.)

jrobbins:
Sure, go ahead.

I think that the workflow for project members should be to merge the PR themselves after the button is green (meaning that it has changed approved, tests passing, and no conflicts).  If we get a PR from someone outside the project, then we would merge for them since they would lack the needed permission.

cwilso (approved):
LGTM!

jpmedley:
Woohoo! 🎉

PR #1041, opened 2020-09-25
Make milestone query not match unrelased features.

jrobbins:
This is a follow up to your question on this recent CL
https://github.com/GoogleChrome/chromium-dashboard/pull/1012

Clicking on a milestone link in the left does a query like "milestone=84" which looks for features that have a milestone field with value 84.  There is no milestone field in the JSON feed, it is set by JS code.  This CL adds a JS if-statement that only sets milestone when the feature has is_released set, and that is set based on the impl_status_chrome value.

### DIFF
--- a/models.py
+++ b/models.py
@@ -144,6 +144,10 @@ INTERVENTION = 9
 ON_HOLD = 10
 NO_LONGER_PURSUING = 1000 # insure bottom of list
 
+RELEASE_IMPL_STATES = {
+  ENABLED_BY_DEFAULT, DEPRECATED, REMOVED, ORIGIN_TRIAL, INTERVENTION,
+}
+
 # Ordered dictionary, make sure the order of this dictionary matches that of
 # the sorted list above!
 IMPLEMENTATION_STATUS = OrderedDict()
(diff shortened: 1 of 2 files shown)


models.py
```diff
@@ -144,6 +144,10 @@
 ON_HOLD = 10
 NO_LONGER_PURSUING = 1000 # insure bottom of list
 
+RELEASE_IMPL_STATES = {
+  ENABLED_BY_DEFAULT, DEPRECATED, REMOVED, ORIGIN_TRIAL, INTERVENTION,
```
jpmedley, line 148: Should `BEHIND_A_FLAG` be part of this list?

jpmedley (changes requested):

models.py
jrobbins: I thought that you didn't want it to be considered released.  If I add BEHIND_A_FLAG here, then a feature that has that implementation status and also has a milestone set would show up in a milestone query.

Should I add it or not?

jpmedley: I guess I've been a bit inconsistent about that. If we want engineers to experiment, then they need to be able to find out what they can experiment with. I think it should be added.

jrobbins: ok will do

jrobbins:
Any other comments?

jpmedley (approved):

PR #1062, opened 2020-10-09
Refactor memcache chunk logic and reuse it for datapoints.

jrobbins:
This should resolve the problem that we were seeing with some graphs failing to load because cached data points were great than the memcache size limit.

In this CL:
+ Refactor and de-duplicate memcache chunk logic that was already in our code
+ Use the lower number of entities per chunk in call cases (300 rather than sometimes 500)
 - This may cause some extra cache misses when the app is first deployed, but that data will simply be re-cached.
+ Update callers to use refactored functions
+ Fix logic to detect a partial cache miss in PopularityTimelineHandler class.
+ Add unit tests

### DIFF
--- a/metrics.py
+++ b/metrics.py
@@ -29,6 +29,17 @@ CACHE_AGE = 86400 # 24hrs
 
 class TimelineHandler(common.JSONHandler):
 
+  def make_query(self, bucket_id):
+    query = self.MODEL_CLASS.all()
+    query.filter('bucket_id =', bucket_id)
+    # The switch to new UMA data changed the semantics of the CSS animated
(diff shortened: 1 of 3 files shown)


jrobbins:
Demo link:
https://239d877-dot-cr-status.appspot.com/metrics/css/timeline/popularity/89

Note: this is a non-default version on the prod server.  It shares the same memcache as the live version.  Testing out this version can cause items to be put into memcache.  In some cases, that will make the original defect non-reproducible for a given chart on the live version because data will have been successfully cached by this version.

jstenback (approved):
This looks great to me, good to see consolidation of the memcache chunk management code.

PR #1063, opened 2020-10-09
Work-around inability to check stackrank expected cache chunks.

jrobbins:
This is a follow-up to my previous CL.

The old code for this check was definitely wrong because part of the if-condition was always true:
len(a_dict.keys()) == len(a_dict) will always be true.  I fixed that defect, but that unmasked a bigger defect.

Most of the cache chunking logic checks that it loaded all needed chunks by counting all items of a given entity type.  However, stack rank items don't consider all entities of a given type, so that approach to counting does not work.

To work around this, I will make the cron work-around more aggressive by having it always recalculate stack rank items.  it runs every 30 minutes.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -9,5 +9,5 @@ cron:
   url: /cron/update_blink_components
   schedule: every day 04:30
 - description: sort metrics and load into memcache without request time limits
-  url: /data/featurepopularity
+  url: /data/featurepopularity?refresh=1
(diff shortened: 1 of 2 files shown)


jrobbins:
This CL is already live on the site.

jstenback (approved):
Looks good, let's get this into production and see how it holds up.

PR #1085, opened 2020-11-03
Implement chromedash-feature-detail component

jrobbins:
This is a big step toward resolving issue #925. 

In this CL:
+ Implement new web component chromedash-feature-detail that is basically a multi-section property sheet
+ Add a JSON representation of field definitions (with field name, type, and help text) and pass that to the new web component
  - The details of these field definitions are extracted from the django field objects
+ Implement a specific request handler for the feature detail page requests and add a specific app route to map requests to that handler.  This avoids using the confusing MainHandler which is full of undocumented, overlapping cases.
+ Add a read-only mode to the feature editing metadata section.  This won't be used in this CL, but it might be useful in the future.

Not in this CL:
+ There will be a change coming to templates/feature.html to use <chromedash-feature-detail> in a user-visible way.  However, I am leaving that part out of this CL to limit the scope of the CL.


jrobbins:
Demo link:
https://fbfa171-dot-cr-status-staging.appspot.com/feature/5751558595674112

This is before I reverted the change to feature.html, so you can see the new chromedash-feature-detail element taking up the majority of the page.   When this CL lands, that element will be latent. But then after making some more changes to the top of the page, in another CL, I will add it again in an expandable section.

liyangguang (approved):

liyangguang (approved):

liyangguang:
Hmm, it seems that I may have lost write access to this repo.  It says "At least 1 approving review is required by reviewers with write access."

guideforms.py
```diff
@@ -820,3 +822,80 @@ def define_form_class_using_shared_fields(class_name, field_spec_list):
...
+        'ie_views', 'ie_views_link', 'ie_views_notes',
+        'web_dev_views', 'web_dev_views_link', 'web_dev_views_notes',
+        'security_review_status', 'privacy_review_status',
+        'ergonomics_risks', 'activation_risks', 'security_risks',
+        'debuggability',
+        'all_platforms', 'all_platforms_descr', 'wpt', 'wpt_descr',
+        'sample_links', 'devrel', 'ready_for_trial_url',
+        'flag_name'),
+    models.INTENT_IMPLEMENT_SHIP: make_display_specs(
+        'tag_review', 'tag_review_status',
```
jpmedley, line 882: Should this include shipping milestones?

guideforms.py
```diff
@@ -820,3 +822,80 @@ def define_form_class_using_shared_fields(class_name, field_spec_list):
...
+        'i2e_lgtms', 'r4dt_lgtms',
+        'experiment_timeline',  # Deprecated
+        'ot_milestone_desktop_start', 'ot_milestone_desktop_end',
+        'ot_milestone_android_start', 'ot_milestone_android_end',
+        ),
+    models.INTENT_SHIP: make_display_specs(
+        'shipped_milestone', 'shipped_android_milestone',
+        'shipped_ios_milestone', 'shipped_webview_milestone',
+        'intent_to_ship_url', 'i2s_lgtms'),
+    models.INTENT_SHIPPED: make_display_specs(
```
jpmedley, line 897: This stage should include shipping milestones.

jpmedley (changes requested):

guideforms.py
jrobbins: Each field can only appear once, and I thought INTENT_SHIP was a better match to the editing experience.  The stages appear right next to each other, so I think users will spot the field that they are looking for.

PR #1088, opened 2020-11-06
Add feature details property sheet to collapsed section of feature detail page

jrobbins:
This resolves issue #925.

In this CL:
+ Add the <chromedash-feature-detail> element to the feature page inside a collapsed section (accordion)
+ Implement a new element <chromedash-accordion> that can be used as a collapsable section or one part of an accordion widget.
+ Improve chromedash-feature-detail to show the active stage, and allow both "Metadata" and "Misc" cards.
+ Change DISPLAY_FIELDS_IN_STAGES to not duplicate data contained in the legacy feature detail page

I realize that one expandable section looks a little lonely at the bottom of the page all by itself.  Eventually, it might be joined by other sections for metrics, reviews / approvals.

### DIFF
--- a/guideforms.py
+++ b/guideforms.py
@@ -572,7 +572,7 @@ METADATA_FIELDS = [
      'category',
      'feature_type', 'intent_stage',
      'search_tags',
-     # Implementtion
+     # Implemention
      'impl_status_chrome',
      'blink_components',
      'bug_url', 'launch_bug_url',
--- a/guideforms.py
+++ b/guideforms.py
@@ -860,9 +860,24 @@ def make_display_specs(*shared_field_names):
 
 DEPRECATED_FIELDS = ['standardization']
 
+DISPLAY_IN_FEATURE_HIGHLIGHTS = [
+    'name', 'summary',
+    'motivation',
+    'unlisted', 'owner',
+    'search_tags',
+    # Implementtion
+    'impl_status_chrome',
+    'blink_components',
+    'bug_url',
+    'comments',
+]
+
 DISPLAY_FIELDS_IN_STAGES = {
+    'Metadata': make_display_specs(
+        'category', 'feature_type', 'intent_stage',
+        ),
     models.INTENT_INCUBATE: make_display_specs(
-        'motivation', 'initial_public_proposal_url', 'explainer_links'),
+        'initial_public_proposal_url', 'explainer_links'),
     models.INTENT_IMPLEMENT: make_display_specs(
         'spec_link', 'api_spec', 'intent_to_implement_url'),
     models.INTENT_EXPERIMENT: make_display_specs(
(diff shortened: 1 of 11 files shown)


jrobbins:
Demo link:
https://67b0877-dot-cr-status-staging.appspot.com/feature/5751558595674112

guideforms.py
```diff
@@ -860,9 +860,24 @@ def make_display_specs(*shared_field_names):
 
 DEPRECATED_FIELDS = ['standardization']
 
+DISPLAY_IN_FEATURE_HIGHLIGHTS = [
+    'name', 'summary',
+    'motivation',
+    'unlisted', 'owner',
+    'search_tags',
+    # Implementtion
```
aglaforge, line 868: Nit: Typo in comment.

aglaforge (commented):
Experience wise the feature looks fine (as does the technical implementation).  The data grouping is a little strange to read (i.e., the subject headings don't quite make sense for what's presented underneath w/ out being a Blink Dev).

I'd suggest to following to clean-up the headings to make the groupings a bit more logical.

If we can make changes, I'd suggest the following:

1.) Metadata is fine.
2.) Combine Start incubating + Start prototyping into something like "Spec links"
3.) Dev trials and integrate on design (rename to "API Evaluation"), might merge with "Evaluate readiness to ship."
4.) Origin trial - Looks ok.  Is there no link to the OT?
5.) Prepare to ship - "Targets"

guideforms.py
jrobbins: Fixed.

jrobbins:
Regarding making the grouping more logical for a non-blink developer:  That will be what we try to do on the upper part of the page.   In a future CL, when we come back to add more to the highlights that are always shown at the top of the page, we might try to find a logical grouping that works for the subset of fields that get shown there.

I don't think that there will be a logical grouping for all fields that makes sense to all users and that fits all feature types.  Even if we work through that one time, it will be hard to maintain as we add fields.  So, to achieve the goal of having a comprehensive page that is maintainable, this CL adds a "details" part of the page to show every field in an organization that should make sense to developers because it aligns with the process that that particular feature is following (which keeps it consistent with the guide UX).  To clarify this two-pronged strategy, I changed the title of the expandable section to "Feature details by stage".

aglaforge:
If I might suggest two small tweaks then:

1.) Change the expand title to "Additional fields by process phase" (i.e., to make expectations clear)
2.) Add a url param that can expand the details by default (e.g., to link to an intent thread for reviewers, so that they don't have to do an extra click).

jrobbins:
> If I might suggest two small tweaks then:
> 
> 1.) Change the expand title to "Additional fields by process phase" (i.e., to make expectations clear)
> 2.) Add a url param that can expand the details by default (e.g., to link to an intent thread for reviewers, so that they don't have to do an extra click).

These are both done.

aglaforge (approved):
LGTM.  Thanks for making those changes.

Issue #1081, opened 2020-10-22
Setting "Will this feature be supported on all six Blink platforms" for intent prototype


wanderview:
Hello, I recently created a new chromestatus entry:

https://www.chromestatus.com/feature/5731920199548928

When generating the intent to prototype the email seemed to always include "no" for "Will this feature be supported on all six Blink platforms?"

I couldn't figure out how to get chromestatus to understand that I plan to ship this on all platforms.  For now I just changed the email I actually sent.

jrobbins:
Sorry if that is unclear.   The checkbox for that is in the form for the Dev Trails stage:
https://www.chromestatus.com/guide/stage/5731920199548928/2

I realize that your feature is still on the "Start prototyping" stage.  So, naturally you have not filled that out yet.  In fact, that field should not appear in the I2P email.  I'll change that.

PR #1090, opened 2020-11-09
Only show all-platforms boolean in the r4t email.

jrobbins:
This should resolve issue #1081.

For most of the fields in the intent emails, I basically show the field if it has a non-empty value.  This means that as the process progresses, fields get filled in, and the later intent emails have more information because everything is shown.  However, booleans have no non-empty values so for this one, I had shown it only when showing fields relevant to dev trials.


cwilso (approved):
LGTM

PR #1092, opened 2020-11-10
Remove custom tooltips because brosers provide tooltips.

jrobbins:
This removes the custom tooltips that were implemented with CSS ::after rules.
They are not needed because browsers already display tooltips for elements that have  a title="" attribute.
In fact, hovering over any of these elements for a few seconds produced two tooltips that might even overlap each other.

I seem to recall you pointing this out as a UI awkwardness.


jrobbins:
Before:
double-tooltips

After:
single-tooltip

aglaforge (approved):
Thanks for fixing that.

PR #1095, opened 2020-11-12
Add a sentence clarifying meaning of recently-added UseCounters

### LINKED ISSUE #1094: Add text clarifying meaning of recently-added UseCounters
UseCounter data only takes into account release channels a UseCounter has already been shipped to. For example, a brand-new UseCounter would have both the numerator and denominator of the usage fraction coming from the number of page loads in the Canary channel. As the UseCoutner is deployed to more channels, both automatically expand to all such channels.

To clarify that this is the behavior, let's add the following text to all UseCounter pages:

*Note: Newly added use counters that are not on Chrome stable yet will only have data from the Chrome channels they're on.*

jrobbins:
Resolves #1094 

In a recent email thread it was discovered that UseCounter metrics are calculated with a denominator for the channels that include that feature, which can be significantly different for features that have reached the stable channel vs not reached stable yet.

### DIFF
--- a/static/elements/chromedash-timeline.js
+++ b/static/elements/chromedash-timeline.js
@@ -249,16 +249,16 @@ class ChromedashTimeline extends LitElement {
       hadEl.src = dsEmbedUrl;
 
       const bigqueryEl = this.shadowRoot.getElementById('bigquery');
-      bigqueryEl.textContent = `#standardSQL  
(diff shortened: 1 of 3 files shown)


jpmedley:
Text looks good.

jrobbins:
Ping.

jstenback (approved):
LGTM

PR #1098, opened 2020-11-19
pushState when clicking on sidenav links

jrobbins:
When changing the search query by clicking on sidebar links, add that to the browser history so that the user can navigate back to previous states without immediately leaving the page.

### DIFF
--- a/static/elements/chromedash-featurelist.js
+++ b/static/elements/chromedash-featurelist.js
@@ -209,19 +209,25 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   // Directly called from template/features.html
-  filter(val) {
+  filter(val, shouldPushState) {
+    this.searchEl.value = val;
+    const pushOrReplaceState = (
+      history &&
+          shouldPushState ? history.pushState.bind(history) :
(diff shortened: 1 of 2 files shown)


static/elements/chromedash-featurelist.js
```diff
@@ -209,19 +209,25 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   // Directly called from template/features.html
-  filter(val) {
+  filter(val, shouldPushState) {
+    this.searchEl.value = val;
+    const pushOrReplaceState = (
+      history &&
+          shouldPushState ? history.pushState.bind(history) :
+        history.replaceState.bind(history));
```
liyangguang, line 217: Should this be this?
```
history && (shouldPushState ? history.pushState.bind(history) : history.replaceState).bind(history)
```

liyangguang (approved):

static/elements/chromedash-featurelist.js
jrobbins: Good catch. That is a logic error.  It would show up on any browser that did not have a history object.  I will fix.

PR #1099, opened 2020-11-20
Remove unused cloudstorage library.

jrobbins:
This library was unused.  There was only one mention of it in the current code which was just an import line.

The library was added on July 10, 2015
https://github.com/GoogleChrome/chromium-dashboard/commit/b494874ae156fad1cf1a029467ae9ac40b57ffcb
It looks like metrics data was being loaded from GCS at that time.

It was changed to fetch data from uma-export on September 16, 2016.
https://github.com/GoogleChrome/chromium-dashboard/commit/91a3c1b0ad13aa2f385562fc72c37424e38c87da
But, the cloudstorage import line and directory was not removed.


jstenback (approved):
LGTM, nice find!

aglaforge (approved):
\o/ Clean-up.

PR #1106, opened 2020-12-05
Replace memcache with a RAM-based cache.

jrobbins:
This resolves issue #1071 and simplifies the overall app while doing that.

I looked into what it would take to upgrade from appengine's memcache to Cloud MemoryStore, and I decided to try avoid that because of the increased complexity of adding another backend dependency.  It would add a step for developers running locally and it would complicate our unit tests.

This new RAM cache is based on a similar approach that has been used in monorail since 2016 (in addition to memcache).  Basically, each AppEngine instance maintains its own python dictionary to cache objects locally.  We lose out on the potential speedup of one instance computing a value that another instance could then use, but if such a case made the app to slow, then we should solve that problem rather than relying on caching.  Local caches need a distributed invalidation scheme so that no instance holds onto data after some other instance has edited that data.  In this code, I use a simple datastore entry with a timestamp.  Monorail uses a more complex SQL table with rows that can specify individual cache entries to invalidate, but since we have such a low frequency of writes, I don't think that complexity is worth it.  We could always add it.

In this CL:
+ Add new ramcache module and its unit tests
+ Search and replace memcache -> ramcache. 
+ Add some TODO comments about chunking code that could be removed.
+ Add calls to check for cache invalidation at the start of each request handler.

In a future CL I will replace webapp2 with flask, and at that time I will make the call to check cache invalidation more systemic in a base class rather than in each handler.

### DIFF
--- a/models.py
+++ b/models.py
@@ -26,7 +26,7 @@ import time
 from google.appengine.ext import db
 # from google.appengine.ext.db import djangoforms
 from google.appengine.api import mail
-from google.appengine.api import memcache
+import ramcache
 from google.appengine.api import urlfetch
 from google.appengine.api import taskqueue
 from google.appengine.api import users
@@ -383,14 +383,14 @@ class BlinkComponent(DictModel):
     """Returns the list of blink components from live endpoint if unavailable in the cache."""
     key = '%s|blinkcomponents' % (settings.MEMCACHE_KEY_PREFIX)
 
-    components = memcache.get(key)
+    components = ramcache.get(key)
     if components is None or update_cache:
       components = []
       url = self.COMPONENTS_ENDPOINT + '?cache-buster=%s' % time.time()
       result = urlfetch.fetch(url, deadline=60)
       if result.status_code == 200:
         components = sorted(json.loads(result.content))
-        memcache.set(key, components)
+        ramcache.set(key, components)
       else:
         logging.error('Fetching blink components returned: %s' % result.status_code)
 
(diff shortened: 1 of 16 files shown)


models.py
```diff
@@ -791,11 +791,13 @@ def get_chronological(
                            'cronorder', limit, version)
 
     keys = get_chunk_memcache_keys(Feature.all(), KEY)
-    feature_list = memcache.get_multi(keys)
+    feature_list = ramcache.get_multi(keys)
+    logging.info('getting chronological feature list')
```
jstenback, line 795: I have no objections to this logging statement, but wanted to make sure you intended to leave that there.

ramcache.py
```diff
@@ -0,0 +1,168 @@
...
+
+# Whenever the cache would have more than this many items, some
+# random item is dropped, or the entire cache is cleared.
+# If our instances are killed by appengine for exceeding memory limits,
+# we can configure larger instances and/or reduce this value.
+MAX_CACHE_SIZE = 10000
+
+
+def set(key, value, time=None):
+  """Emulate the memcache.get() method using a RAM cache."""
```
jstenback, line 66: s/get/set/

ramcache.py
```diff
@@ -0,0 +1,168 @@
...
+def _check_expired(keys):
+  now = int(time_module.time())
+  for key in keys:
+    if key in expires and expires[key] < now:
+      del expires[key]
+      del global_cache[key]
+
+
+def get(key):
+  """Emulate the memcache.set() method using a RAM cache."""
```
jstenback, line 85: s/set/get/

ramcache.py
```diff
@@ -0,0 +1,168 @@
...
+  return global_cache.get(key)
+
+
+def get_multi(keys):
+  """Emulate the memcache.get_multi() method using a RAM cache."""
+  _check_expired(keys)
+  return {
+      key: global_cache[key]
+      for key in keys
+      if key in global_cache}
```
jstenback, line 98: That closing bracket being on a line of its own would make this look less like something's got left out here IMO.

jstenback (approved):

models.py
jrobbins: Yeah, I'd like to leave it in for now.  If something goes wrong, having more stuff logged will make investigation easier.

jrobbins:
Thanks for the review.  Before I merge it, I just want to test it on staging a bit and consider some tweaks to improve thread safety.

jrobbins:
I completed testing on staging after finding and fixing one problem.

The main thread safety issue that I was concerned about was the fact that setting the cached value and cached expiration time are done in two separate steps.  However, our rate of writes are so low that it seems unlikely to ever matter, and the view places where we set an expiration time, we always set the same expiration time, so I think that even the scenario I had in mind would turn out OK.

Issue #1114, opened 2020-12-14
Edge views not included in intent emails


jrobbins:
**Describe the bug**
The intent email templates include FF and safari views, but leave out the views for Edge.

**Expected behavior**
When we have data for a UA signal, we should display it.

**Additional context**
Edge signals was removed in
https://github.com/GoogleChrome/chromium-dashboard/commit/6af1b1366623a28bd65e9a9863a9824f7b9b82c2#diff-ec8cd8ecec9f640e8a0209f66301563e8a00d7941ddfe23d364fa858257b2b6e
based on following https://bit.ly/blink-signals very literally, but I think that it makes sense to include Edge signals if there is a value for them.

mrego:
I believe the idea was not to have Edge signals in the form directly, that would mean they could only be present in old features; for the new ones they won't have Edge signals, and then they won't appear on the intent mails (see #1051).

jrobbins:
This change was reverted.

PR #1116, opened 2020-12-15
Add Edge signals back to intent emails.

jrobbins:
This resolves issue #1114 

Edge is not listed in the doc that I was looking at for the expected contents of intents emails, however as user asked for it, so I have added it back.


jrobbins:
Chris, is this the right idea?  Should Edge be listed in the intents emails or is it somehow not desired?

For reference, bit.ly/blink-signals lists Gecko and WebKit only.

cwilso (approved):
LGTM

PR #1121, opened 2020-12-19
Upgrade Guide pages to use flask web application framework

jrobbins:
This is progress toward resolving issue #1070.

In this CL:
+ Convert guide.py to use the flask-related base class and app object defined in common.py
+ Add a few methods to FlaskHandler in common.py to help streamline this and future conversions
+ The existing code returned 401 when the signed in user did not have permission to do something.  I think 403 is more correct.
+ Unit tests


aglaforge (approved):
Larger CL, but the changes are logical and straightforward.

PR #1122, opened 2020-12-22
Upgrade admin.py to use flask.

jrobbins:
This is the next step in resolving issue #1070.

In this CL:
+ Convert admin.py to use the new flask base classes in common.py.
+ Refactor some of the code in common.py from having one huge handler with lots of if-statements that look at the path into separate handlers for each path for the old UI new feature, edit feature, and delete feature handlers.
+ Change a helper property in FlaskHandler from request_uri to instead return the whole request.
+ Refactor calls to invalidate the RAM cache on every GET and POST request sent to any FlaskHandler, this removes a requirement that individual handlers make this call before accessing cached data.
+ Update existing unit tests in admin_test.py

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -23,19 +23,13 @@ import json
 import logging
 import os
 import re
-import sys
-import webapp2
-from HTMLParser import HTMLParser
 from xml.dom import minidom
 
 # Appengine imports.
 import ramcache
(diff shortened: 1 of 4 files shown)


aglaforge (approved):
LGTM

PR #1124, opened 2020-12-22
Upgrade metrics.py to use flask.

jrobbins:
This is the next step in resolving issue #1070.

In this CL:
+ Upgrade request handlers in metrics.py
+ Refactor some functions from common.JSONHandler to metrics.py because they were only used there.
+ Improve common.FlaskHandler to add per-class options for HTTP cache control and allow JSON handlers to return a list (Flask automatically JSONifies python dicts, but not lists).
+ Added some unit tests for metrics.py

### DIFF
--- a/common.py
+++ b/common.py
@@ -154,22 +154,6 @@ class BaseHandler(webapp2.RequestHandler):
 # TODO(jrobbins): phase out this class and have all calls use FlaskHandler.
 class JSONHandler(BaseHandler):
 
-  def __truncate_day_percentage(self, data):
-    # Need 8 decimals b/c num will by multiplied by 100 to get a percentage and
-    # we want 6 decimals.
-    data.day_percentage = float("%.*f" % (8, data.day_percentage))
-    return data
-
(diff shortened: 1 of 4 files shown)


jrobbins:
Demo links:
https://796f868-tainted-dot-cr-status-staging.appspot.com/metrics/feature/popularity
and
https://796f868-tainted-dot-cr-status-staging.appspot.com/metrics/css/popularity
and
https://796f868-tainted-dot-cr-status-staging.appspot.com/metrics/css/timeline/popularity/89
  (Note that the top graph does not render on staging due to an unrelated limitation)

aglaforge (approved):
This all looks good.

Minor note: When looking over the demo, I noticed something visually odd (not related to this change).  Because we don't truncate or elide the text of the metric names the formatting of the graphs can vary for longer names.  This is mostly noticeable when scrolling, which likely isn't the most common navigation.  I wouldn't prioritize any work here, but would note it in case we make any future changes/ visual refreshes to this page.

PR #1125, opened 2020-12-23
Upgrade blink_handlers.py and schedule.py to flask

jrobbins:
This is more progress on issue #1070.

In this CL:
+ Upgrade blink_handler.py (which offers an admin UI for component subscribers) and schedule.py (which shows the chrome release page) to use flask
+ Simplified the app.yaml entry for the schedule page because the wildcard is not needed.
+ Tweaks to settings.py and models.py to get subscriber population working.  This is only used when testing locally.
+ No new tests for blink_handler.py because it seems like the wrong approach to subscriptions and I plan to replace it eventually

### DIFF
--- a/blink_handler.py
+++ b/blink_handler.py
@@ -26,10 +26,41 @@ import yaml
 
 import common
 import models
-import ramcache
 import settings
 import util
 from schedule import construct_chrome_channels_details
+
+
+class PopulateSubscribersHandler(common.FlaskHandler):
+
+  def __populate_subscribers(self):
+    """Seeds the database with the team in devrel_team.yaml and adds the team
+      member to the specified blink components in that file. Should only be ran
+      if the FeatureOwner database entries have been cleared"""
+    f = file('%s/data/devrel_team.yaml' % settings.ROOT_DIR, 'r')
+    for profile in yaml.load_all(f):
+      blink_components = profile.get('blink_components', [])
+      blink_components = [models.BlinkComponent.get_by_name(name).key() for name in blink_components]
+      blink_components = filter(None, blink_components) # Filter out None values
+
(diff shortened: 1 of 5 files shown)


jrobbins:
Demo pages:
https://d47a6f4-tainted-dot-cr-status-staging.appspot.com/features/schedule
and
https://d47a6f4-tainted-dot-cr-status-staging.appspot.com/admin/subscribers
and
https://d47a6f4-tainted-dot-cr-status-staging.appspot.com/admin/blink

blink_handler.py
```diff
@@ -104,47 +103,47 @@ def get(self, path):
...
                                    primary=params.get('primary'))
-    self.response.set_status(200, message='User added to subscribers')
-    return self.response.write(json.dumps(params))
+    return params
 
 
-class SubscribersHandler(common.ContentHandler):
+class SubscribersHandler(common.FlaskHandler):
+
+  TEMPLATE_PATH = 'admin/subscribers.html'
```
aglaforge, line 134: Curious about this style choice.  Since this is a class bound attribute, was there a reason not to add an overloadable accessor method (versus using a public attribute with CONST style nomenclature)?

aglaforge (approved):
Logic makes sense.  Had one question re: use of the const to specify the template name (i.e., why not use a method instead?).

blink_handler.py
jrobbins: It just emphasizes that it is supposed to be a simple value with no computation or side effects and no need to do anything analogous to calling super().  The syntax is simpler than a function would be.  For comparison, for the HTTP headers stuff uses functions so that a subclass can override it to call super() and then add another header.

PR #1126, opened 2020-12-23
Upgrade notifier.py to use flask.

jrobbins:
This makes more progress on resolving issue #1070.

In this CL:
+ Delete some request handlers that were related to push notifications.  They were not being used.
+ Upgrade remaining handlers to use flask.
  - Flask handlers cannot just fall off the end of the handler function, they need to return something, so I returned a JSON constant
  - Note that tasks that fail due to bad input still return a 200 so that the GAE task system does not keep retrying
+ Update unit tests.

Note that the bounce handler still imports something under "webapp", but that is different than "webapp2".  That dependency will also need to be changed in the future, but that will come later.


aglaforge (approved):
LGTM

PR #1127, opened 2020-12-30
20201228 upgrade server to flask

jrobbins:
This should resolve issue #1070.

In this CL:
+ Break if-else logic of server.py into distinct handlers for each URL.
+ Upgrade each handler to use FlaskHandler
+ List all the routings explicitly, including redirects and templates that had no python logic.
+ Add utility classes to common.py for simple handlers for redirects and templates that need no python logic.
+ Add some long-missing unit tests now that this code is more testable

### DIFF
--- a/server.py
+++ b/server.py
@@ -52,7 +52,7 @@ class FeatureDetailHandler(common.FlaskHandler):
         'process_json': json.dumps(processes.process_to_dict(feature_process)),
         'field_defs_json': json.dumps(field_defs),
         'feature': f.format_for_template(),
-        'feature_id': f.key().id,
+        'feature_id': f.key().id(),
         'feature_json': json.dumps(f.format_for_template()),
         'updated_display': f.updated.strftime("%Y-%m-%d"),
     }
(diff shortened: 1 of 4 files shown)


jrobbins:
Demo server:
https://cr-status-staging.appspot.com/metrics/css/popularity
and all the other metrics pages

Note that the staging server does not fill in the top graph on the metric detail pages.  That is an existing limitation of the staging server, not affected by this CL.

server.py
```diff
@@ -55,169 +52,142 @@ def get(self, feature_id):
...
+    }
+    return template_data
+
+
+class OmahaDataHandler(common.FlaskHandler):
+
+  JSONIFY = True
+
+  def get_template_data(self):
+    omaha_data = util.get_omaha_data()
```
aglaforge, line 167: Does this hit memcache?

aglaforge (approved):
LGTM, one minor questions about the OmahaData fetch, but otherwise the CL looked logical.

server.py
jrobbins: Yes, it is cached.  I removed memcache and we now use ramcache.py.

PR #1128, opened 2020-12-30
Remove remaining references to webapp2 in unused code.

jrobbins:
Resolves issue #1070.

In this CL:
+ Delete webapp2 handlers in common.py that were no longer used.
+ Move some unit tests for a BaseHandler method over to test the same method now that it lives in FlaskHandler.
+ Delete bare-bones webapp2 error handlers in favor of just using flask.abort().

### DIFF
--- a/common.py
+++ b/common.py
@@ -25,7 +25,6 @@ import os
 import re
 import time
 import traceback
-import webapp2
 
 import flask
 import flask.views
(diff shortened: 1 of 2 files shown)


aglaforge (approved):
LGTM, thanks for the clean-up.

PR #1129, opened 2020-12-30
Remove old memcache chunking work-around because ramcache does not need it

jrobbins:
Previously I removed our dependency on memcache and implemented ramcache.py as a simpler alternative.

This CL follows up on that by removing chunking logic that had been put in place to work around memcache's limit of 1M per item.  It also adds simple logging of the cache hit ratio.

ramcache.py does not impose any limit on the size of each cached item.  It uses RAM in the GAE instances up to the limit that we set on the number of cached items.  If we set that limit too high, we would see GAE instances being terminated for exceeding their memory limit, and we do not see that in production.


metrics.py
```diff
@@ -182,25 +178,12 @@ def get_template_data(self):
...
-          self.PROPERTY_CLASS.all(), self.MEMCACHE_KEY)
-      logging.info('looking for keys %r' % keys)
-      properties = ramcache.get_multi(keys)
-      logging.info('found chunk keys %r' % (properties and properties.keys()))
-
-      # TODO(jrobbins): We are at risk of displaying a partial result if
-      # memcache loses some but not all chunks.  We can't estimate the number of
-      # expected cached items efficiently.  To counter that, we refresh
-      # every 30 minutes via a cron.
+      properties = ramcache.get(self.MEMCACHE_KEY)
```
aglaforge, line 181: This is pretty minor, but for consistency sake, we might want to refactor this CONST to be RAMCACHE_KEY to avoid any confusion about the backing service.

aglaforge (approved):
LGTM.  Only nit would be to consider removing references to MEMCACHE in the code (e.g., MEMCACHE_KEY).

PR #1137, opened 2021-01-07
Remove unneeded MEMCACHE_KEY_PREFIX and simplify other names.

jrobbins:
This is a follow-up CL to last week's removal of some unneeded logic that was used with memcache.  In this CL:

+ Remove settings.MEMCACHE_KEY_PREFIX.  The purpose of that constant was to separate memcache entries used by different versions of the app.  That is not needed with ramcache because each app version runs in separate GAE instances that have their own RAM.

+ Simplify some names and comments from "memcache" to "cache" to avoid misleading comments.

+ Delete some TODOs that were done in last week's CL.

### DIFF
--- a/common.py
+++ b/common.py
@@ -272,7 +272,7 @@ class FlaskHandler(flask.views.MethodView):
     elif user.email().endswith(('@chromium.org', '@google.com')):
       can_edit = True
     else:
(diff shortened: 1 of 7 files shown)


aglaforge (approved):
LGTM!  Nice, thank you.

PR #1141, opened 2021-01-14
Upgrade from GAE taskqueue to Google Cloud Tasks

jrobbins:
Resolves issue #1072.

In this CL:
+ Add entries to requirements.txt for Google Cloud Tasks and specific version of indirect dependencies with versions that work with python 2.7.
+ Add requirements.dev.txt and a libraries section in app.yaml and notifier.yaml to allow GCT to work
+ Various changes to appengine_config to get the new libraries to work
+ cloud_tasks_helpers.py with new functions to enqueue a task.  This is based on the corresponding file in Monorail.
  - Include a local version of the GCT client that simply hits the task handler synchronously since we have no GCT server locally.
+ Change the auth for task handlers from GAE admin check to checking GCT headers in common.py.  This allows them to be run locally and also aligns with how this auth would need to happen in python 3.
+ Add dispatch.yaml and change the deploy_site.sh script to include that file.  taskqueue could target the notifier service of our app, but GCT just makes a general incoming request that needs to be routed to the app service that we want.
+  Update models.py and notifier.py code that created tasks to call our new code.
+ Add unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
 python:
   - "2.7"
 dist: trusty
-sudo: false
+sudo: true
 git:
   depth: 1
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,7 @@ before_install:
 install:
   - npm install -g gulp
   - npm install
-  - echo "About to run travis-deps"
   - npm run travis-deps
-  - echo "Done with travis-deps"
 script:
   - npm run lint
   - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,9 @@ before_install:
 install:
   - npm install -g gulp
   - npm install
+  - echo "About to run travis-deps"
   - npm run travis-deps
+  - echo "Done with travis-deps"
 script:
   - npm run lint
   - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
   - npm install -g gulp
   - npm install
   - npm run deps
+  - python -m pip install --no-deps -r requirements.dev.txt
 script:
   - npm run lint
   - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
 python:
   - "2.7"
 dist: trusty
-sudo: true
+sudo: false
 git:
   depth: 1
 env:
@@ -27,8 +27,7 @@ before_install:
 install:
   - npm install -g gulp
   - npm install
-  - python -m pip install --no-deps -r requirements.dev.txt
-  - npm run deps
+  - npm run travis-deps
 script:
   - npm run lint
   - npm run test
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,8 @@ before_install:
 install:
   - npm install -g gulp
   - npm install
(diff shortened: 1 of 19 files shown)


jrobbins:
This is currently live on https://cr-status-staging.appspot.com.

The main way that we use tasks is for sending email notifications.  The staging server does send email notifications for changes, but rather than send them to users, it sends them all to 
https://groups.google.com/a/google.com/g/cr-status-staging-emails
It can take a minute or two for emails to show up there.

aglaforge (approved):
The change itself looks fine.   Couple random thoughts:

- If there are any files that have an original source that fall's outside of our project, it would be good to put some comment annotation noting where they came from (i.e., if someone needed to investigate in the future).
- The files with Apache licenses make me wonder if we need to do something similar in terms of expressing licensing and copyright in our files.
- The referenced group alias has restricted visibility/ membership.  ACLs might need to be updated.

jrobbins:
> * If there are any files that have an original source that fall's outside of our project, it would be good to put some comment annotation noting where they came from (i.e., if someone needed to investigate in the future).

Good point.  I have added a URL to that file to link to the code that inspired it.

> * The files with Apache licenses make me wonder if we need to do something similar in terms of expressing licensing and copyright in our files.

All our python source files have header comments that include copyright information.  I normally don't include that for config files and templates, but we could.

Are you questioning the choice of the Apache license?  Other chromium stuff uses a BSD-style license and attributes ownership to "The Chromium Authors" rather than "Google."  I am willing to change all the copyright headers in our codebase.  Technically, changing a license is supposed to be done with consultation with past contributors, but I think going ahead would be OK because (a) Apache and BSD licenses are pretty similar, (b) the old code under the old license would still be accessible by checking it out from GitHub, and (c) more and more of this code is getting rewritten as I go along.

> * The referenced group alias has restricted visibility/ membership.  ACLs might need to be updated.

I could create a public mailing list archive and use that instead of this one in the google.com domain.  However, that will cause information leaks.  

I think of the staging server as a scratch area and the generated emails are really just for debugging rather than having any meaningful content.  Since I use the staging server to test out new app versions, if there was ever an information leak that generated outbound emails due to a defect in a version that I was testing, I'd prefer that it leak to an archive that is not widely viewable. 

Also, even without any defects, the staging server sends to cr-status-staging-emails+email-address-of-user@googlegroups.com   That address information is useful for debugging email generation.  However, if someone were to view the archive, they would be seeing all the email addresses and email contents for other people who used the staging server.  That is very different from the production server, which sends emails to the users, and each user only sees their own email address and content that was intended for them.

aglaforge:
Re: License, no strong preference between BSD and Apache.  Seeing the Apache notices made me wonder what we were doing in our other source files and I couldn't remember seeing any (in fairness most of what I look at are diffs).  Mostly just wanted to make sure that we were following consistent practices w/ our source code.  Following Chromium standards makes sense here.

Re: Group.  No need for a public alias.  Was just trying to follow the testing instructions w/ my Google account and couldn't access the group/ contents.   If we can adjust the Group so that Googlers can see the messages/ join that should make testing easier.  I suspect that we won't published anything to that group that's particularly confidential to internal stakeholders.

jrobbins:
>  If we can adjust the Group so that Googlers can see the messages/ join that should make testing easier.

Done.

PR #1145, opened 2021-01-21
Make metrics fetch more robust and debuggable.

jrobbins:
Make fetching metrics more robust based on a new understanding of how uma-export works.
Specifically, uma-export may take as long as 48 hours to produce metrics, so we want to keep trying.

In this CL:
+ Add ability to specify a date in the query string so that that we can manually fill in data for any missed dates.
 - This helps avoid the problem where running the cron uses a value of "yesterday" that does not match what was used in a recent failure.
+ When run normally, it tries each of the past 5 days rather than just yesterday
+ To avoid requesting data that it already has, bucket_id 0 is used as a "capstone" entry to indicate that all datapoints for a given day were already saved
+ Add logging to aid in any future debugging
+ Eliminate the "cookie" mode for working with metrics in a development environment because uma-export does not support that.  (In a separate uma-export CL, I have added cr-status-staging to its allowlist).
+ Run the cron job more often.  It will be a no-op if we already have all the data that we need.
+ Deploy cron.yaml every time that we deploy the app so that it is not forgotten
+ Add unit tests

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -42,25 +42,23 @@ UMA_QUERY_SERVER = 'https://uma-export.appspot.com/chromestatus/'
 HISTOGRAMS_URL = 'https://chromium.googlesource.com/chromium/src/+/master/' \
     'tools/metrics/histograms/enums.xml?format=TEXT'
 
-# Path to a local cookie to supply when running dev_appserver
-# Go to https://uma-export.appspot.com/login and login then copy the
-# value of SACSID from browser and write this to a cookie file.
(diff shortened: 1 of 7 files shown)


jrobbins:
A demo version of this is live on cr-status-staging.  However, my change to uma-export has not gone live yet, so the demo is not working yet.  Please don't merge this CL for me, I will merge it once I know it has worked in staging.

aglaforge (approved):
This all looks reasonable.

PR #1150, opened 2021-01-26
Continue requesting metrics if one is missing.

jrobbins:
After testing in staging I realized that there was a logical error in the code.  When the first day is not yet ready, our app should continue to request previous days to fill in any data that was not previously available but might now be available.


aglaforge (approved):
LGTM

PR #1097, opened 2020-11-19
Move "No active development" features to near bottom of feature list

jrobbins:
Resolves issue #1035.

The top of the feature list page will now start with Proposed features.

PR #1154, opened 2021-01-28
Fix an off-by-one error in feature list left-nav.

jrobbins:
This fixes an error that was pointed out on chat today.

The error was introduced in
https://github.com/GoogleChrome/chromium-dashboard/pull/1097
where I removed one of the status nav elements but did not update the index constants on these lines.

Basically, four nav elements are added for dev, canary, beta, and stable.  If some of those channels as the same version number of as the neighboring channel, then the entries are merged rather than keeping both.   Since these indexes were off by one, the wrong ones were being merged.

The defect was probably not noticed sooner because the channels have distinct versions on most days.

### DIFF
--- a/static/elements/chromedash-metadata.js
+++ b/static/elements/chromedash-metadata.js
@@ -175,10 +175,10 @@ class ChromedashMetadata extends LitElement {
 
     // Consolidate channels if they're the same.
     if (this._channels.canary == this._channels.dev) {
-      this._versions.splice(5, 1);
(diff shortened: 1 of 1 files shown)


jrobbins:
FYI, I already pushed this change live because it is a very localized fix, it was the only change since my last deployment, and I wanted to see if it actually resolved the problem so that I could respond to the user who reported it.

aglaforge (approved):
LGTM

PR #1157, opened 2021-02-05
Revert "Add Edge signals back to intent emails."

jrobbins:
Reverts GoogleChrome/chromium-dashboard#1116

This revert removes Edge signals from intent emails.


cwilso (approved):
LGTM

PR #1158, opened 2021-02-09
Save metric data  using fewer queries.

jrobbins:
This fixes a problem with the ability to specify a previous date via a query-string parameter.  The logic for that was correct, but end-user requests are limited to 60 seconds so the request would time out.  Retrying the request would take just without making additional progress, so it was hopeless.

This replaces ~3000 queries with one query plus ~3000 python set membership checks, which should run much faster.  Even if it times out, making the request again will skip forward over the work that was previously done more efficiently, so it will make progress and eventually complete.

The normal cron execution of this handler worked because those requests have a 10 minute time limit.

Also in this CL, add logging with a unique string so that I can create a log-based alert to notify us if we consistently fail to fill in metrics data.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -103,7 +103,8 @@ class UmaQuery(object):
     j = json.loads(json_content)
     if not j.has_key('r'):
       logging.info(
-          '%s results do not have an "r" key in the response' % self.query_name)
+          '%s results do not have an "r" key in the response: %r' %
+          (self.query_name, j))
       logging.info('Note: uma-export can take 2 days to produce metrics')
       return (None, 404)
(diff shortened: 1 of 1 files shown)


aglaforge (commented):
LGTM.  Thank you for also adding that final >2 day check. 

Out of curiosity, do ERROR logs trigger mail?  If not, do we need a more active notification?

aglaforge (approved):

jrobbins:
The error logging info will not trigger an email by itself.  But, I can set up a logs-based alert in pantheon that will look for that string and send an email.

PR #1177, opened 2021-02-22
Add hard-coded rule to notify on missing webview milestone.

jrobbins:
This addresses a request from the webview team to get notifications when someone forgets to set that milestone.

Eventually, we should have user-configurable subscriptions to various kinds of changes, but for now this is hard-coded.


aglaforge (approved):
LGTM.  Hopefully someday we can get a more generic configuration for firing notifications and dispatching subscriptions.  In the meantime, this is prefect for what it's doing.  Thank you.

Issue #170, opened 2014-10-03
Send HTTP Strict Transport Security header


ghost:
www.chromestatus.com is already https only, but of course there are many websites that still link to http://www.chromestatus.com. By enabling HSTS, it saves the time to make the first http connection.

beaufortfrancois:
Hey @ebidel, why don't we simply submit "www.chromestatus.com" for inclusion in Chrome's HTTP Strict Transport Security (HSTS) preload list at https://hstspreload.appspot.com?

To see how it would look like in Chrome, you can simply go to `chrome://net-internals/#hsts` and enter "www.chromestatus.com" as Domain name and click the "Add" button. Then, check out `chrome://net-internals/#events` and notice there are no HTTP connections.

jpmedley:
@jrobbins Do you know if this has been done?

jrobbins:
It has not been done.  It seems like a good idea, and probably not too difficult.  I think we need to add a response header before we can submit the site to https://hstspreload.appspot.com.

jrobbins:
There's a little more work to do here.  I need to deploy this change and then submit our site to https://hstspreload.appspot.com.

jpmedley:
Sorry about that.

jrobbins:
No worries.

I have deployed and submitted the site now. 

IIUC, it might be be awhile before chrome starts automatically using https.  When it does take effect, It is supposed to be transparent, so we might not notice unless we test it out.

PR #1187, opened 2021-02-25
Add Strict-Transport-Security header since we are all https.

jrobbins:
Partly resolves issue #170.


aglaforge (approved):
Thank you for adding that!

jpmedley:
>Partly resolves issue #170.

I closed #170. Should I not have done so?

jrobbins:
I reopened the issue for now because I still need to submit our site to https://hstspreload.appspot.com.

PR #1188, opened 2021-02-25
Remove a few more references to unused service worker.

jrobbins:
In this CL:
+ Remove app.yaml and docs references to service-worker stuff that is no longer used.  And, a JS file that is no longer used.
+ Delete some app.yaml entries that allowed access to a developer console which is not needed and could be a risk if the restrictions to accessing it were ever removed


aglaforge (approved):
LGTM

PR #1190, opened 2021-03-03
Start offering a simple REST API.

jrobbins:
In prepration for needing more of an API for upcoming UI functionality, this change refactors existing code for stars and cue cards into a new simple REST API.  The proposed UI changes need an API rather than form submission because we don't want to do a complete page load for, e.g., opening the approvals pane or autosaving draft text.

In this CL:
+ Define a new common.APIHandler to do overhead needed for API requests and to use more distinct names (get -> do_get) so that we can grep through code later.
+ Set up a new source code directory for the API, and fix some unit test path stuff
+ Create api/register.py that lists out all API entry points
 - app.yaml entry for the API
+ Create api/cues_api.py and api/stars_api.py as a new home for these existing handler.  Old handler code will be removed a while after the next deployment so that we do not break web pages that users have already loaded.
 - Also, refactored unit tests for that code

### DIFF
--- a/api/cues_api.py
+++ b/api/cues_api.py
@@ -1,8 +1,5 @@
-from __future__ import division
-from __future__ import print_function
-
 # -*- coding: utf-8 -*-
-# Copyright 2020 Google Inc.
+# Copyright 2021 Google Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


api/cues_api.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+
+import common
+import models
+
+# We only accept known cue name strings.
+ALLOWED_CUES = ['progress-checkmarks']
+
+
+class CuesAPI(common.APIHandler):
```
aglaforge, line 28: It's not clear what this API does in practice.  I'd recommend adding a comment here.

aglaforge: Example:  It's unclear why always returning an empty map would be appropriate for non-400 cases.

api/stars_api.py
```diff
@@ -0,0 +1,59 @@
...
+      feature_ids = notifier.FeatureStar.get_user_stars(user.email())
+    else:
+      feature_ids = []  # Anon users cannot star features.
+
+    data = {
+        'featureIds': feature_ids,
+        }
+    return data
+
+  def do_post(self):
```
aglaforge, line 39: Please add a comment to explain the logic of this method.  Inferring from the code it's getting the stars for an individual user.  I'd just call it out.

api/cues_api.py
jrobbins: I added docstrings.

It was returning an empty dict simply because that call did not need any return value.  I changed it to send {'message': 'Done'} and added a comment.

api/stars_api.py
jrobbins: Done.

aglaforge (approved):
Thank you!

PR #1193, opened 2021-03-04
Fix new crbug link on feature detail page.

### LINKED ISSUE #1192: The bug icon on the feature detail page doesn't work
**To Reproduce**
1. Go to any feature detail page
2. Click on the bug icon

**Expected behavior**
Opens a page on bugs.chromium.org.

Instead, it opens the same page that you are on because that link has no href attribute.

jrobbins:
Fixes #1192.

I think this defect arose because of an unclear mix of info in the python dict returned from Feature.get_feature().  It would be better to consistently do like I am doing here in server.py where the python dict for the feature has *only* info directly from the Feature DB entity, whereas other info needed for the UI is passed to the template engine separately by the handler.

### DIFF
--- a/server.py
+++ b/server.py
@@ -59,6 +59,7 @@ class FeatureDetailHandler(common.FlaskHandler):
         'feature_id': f.key().id(),
         'feature_json': json.dumps(f.format_for_template()),
         'updated_display': f.updated.strftime("%Y-%m-%d"),
+        'new_crbug_url': f.new_crbug_url(),
     }
     return template_data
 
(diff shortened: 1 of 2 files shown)


jpmedley (approved):

jpmedley:
@jrobbins I'm leaving this for you to merge in case you didn't intend for me to be the reviewer.

jrobbins:
Thanks.  I think it is always best to wait for the PR author to merge in the change.  Just approving it is the only thing that the reviewer needs to do.  It was not a mistake, I selected you as the reviewer because this was a simple, user-visible change.

PR #1199, opened 2021-03-11
Remove links to use old feature editing UI.

jrobbins:
I reviewed access logs for the old editing pages and found that only 2 users had done a POST to them in the past few months.  I emailed each user and got back a response from one of them.  He said that he just wanted a flat list of all fields, so he would be willing to use the new flat list rather than the old UI.

Rather than completely remove the "Use old UI" link from the editing page, I changed it to "Edit all fields" so that users who are looking for the "Use old UI" link will see that in the same position.  Eventually we can actually remove it to simplify the page content.

### DIFF
--- a/templates/guide/edit.html
+++ b/templates/guide/edit.html
@@ -8,8 +8,8 @@
 {% block subheader %}
 <div id="subheader" style="display:block">
   <a style="float:right"
-     href="/admin/features/edit/{{ feature_id }}">Use old UI</a>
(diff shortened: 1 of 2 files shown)


jstenback (approved):
LGTM!

jpmedley:
I'm just curious. Have you received any complaints about this?
Joe Medley | Technical Writer, Chrome DevRel | ***@***.*** |
 816-678-7195

PR #1202, opened 2021-03-14
Added a link in superscript to access timeline and stackrank from nar…

### LINKED ISSUE #302: Ability to choose between 'stack rank' and 'timeline' disappears on narrow viewports
When the viewport narrows enough, the buttons to switch between timeline view and stack rank for feature popularity disappear and there is no way to change views.

image

shivamag00:
…row view ports

Merging this will close #302


shivamag00:
@googlebot I signed it!

shivamag00:
This PR is for fixing the following issue.
#302

shivamag00:
This is how the link appears on narrow view ports

a
B

jrobbins:
Thanks for your work on this feature, but I don't think that this is the right approach.   I should have spelled out the kind of solution that I was expecting in the issue.  I will add more info there now.  We don't use "Switch to other mode" links very frequently, although there have been some examples (e.g., "Try new UI").  And, that is not a common pattern in other Google UIs.

I think what is needed instead is a responsive UI approach.   We have conditional CSS based on with width of the viewport, but in the case where the viewport is narrow, we currently hide the side navigation buttons.  The missing part of that implementation is that we should still offer those side-nav elements in a way that does not use up horizontal space on the page.  Two common approaches to that are (1) introducing a hamburger menu to the site or some button that would show the side nav temporarily overlapping other content on the page, or (2) displaying the side-nav items in a a horizontal layout across the top of the main content area.

shivamag00:
UPDATE: I have raised a new PR #1220 in which I have followed the second approach suggested by @jrobbins

Issue #431, opened 2016-11-04
Search terms don't actually impact built-in search


RByers:
I'm guessing "search terms" is for SEO, right?
@bokand was surprised that his [top controls feature](https://www.chromestatus.com/feature/6241601772847104) couldn't be found when [searching](https://www.chromestatus.com/features#top%20controls) for "top controls" even though that string was in the search terms field.

Perhaps it's easy to extend the search to the contents of those fields?

vschs007:
can you provide a snapshot of it ?

jpmedley:
@jrobbins FYI

arjunrkaushik:
@RByers I would like to work on this issue

jrobbins:
@arjunrkaushik go for it.

I believe that the solution would be to add the feature keywords to the temporary string that is built up in this function
https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/elements/chromedash-featurelist.js#L201
In fact, motivation and some other fields of the feature entry should probably be in there too.

jpmedley:
@arjunrkaushik Thank you for volunteering. Please link to this issue in your pull request.

impact-maker:
@jrobbins  Hi, what is the status of this bug. Is it ready to merge or needs any alternate solution.

GouravSingal-code:
can i work on it

suprith-hub:
Hi, @RByers. Is this issue resolved? 
Because in the All features search the above behavior works well, i.e, searching for `top controls` does give Inert URL Bar in the result.
Or does it mean similar issue still exist? for other search keywords...?

RByers:
Looks like it's working now to me! I don't know if any other similar issues exist, but I'm fine calling this fixed. Thanks!

PR #1206, opened 2021-03-16
Tried fixing error #431

Ritarshi-Chattaraj:
added feature.keyword to temporary string regExp at #L201 function

### DIFF
--- a/static/elements/chromedash-featurelist.js
+++ b/static/elements/chromedash-featurelist.js
@@ -199,7 +199,7 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   _filterKeyword(regExp, feature) {
-    return (feature.name + '\n' + feature.summary + '\n' + feature.comments)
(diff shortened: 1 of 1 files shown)


static/elements/chromedash-featurelist.js
```diff
@@ -199,7 +199,7 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   _filterKeyword(regExp, feature) {
-    return (feature.name + '\n' + feature.summary + '\n' + feature.comments)
+    return (feature.name + '\n' + feature.summary + '\n' + feature.comments+ '\n' + feature.keyword)
```
jrobbins, line 202: In models.py this search tag information is added to the JSON representation of the feature under the name "tags", so you need to change "keyword" to "tags".

Also, almost all our code is formatted to wrap at 80 columns, so please word-wrap this line.

Ritarshi-Chattaraj: Done.

jrobbins (approved):

Issue #1073, opened 2020-10-19
Convert from using app identity library to env. vars


jrobbins:

vishwjeet-singh:
hi @jrobbins, I would like to work on this issue can you please explain how to start on this issue.

MissMwendwa:
hi @jrobbins, in this particular problem, is there a bug or it is being developed from scratch?

MissMwendwa:
Though in my line of thought, I am thinking that this can be implemented using javascript.

MissMwendwa:
can this be done using the python programming language?

shivamag00:
@jrobbins Please close this issue also as it is solved by CL #1207 (Merged)

PR #1207, opened 2021-03-16
Replaced app identity lib with env vars

CrypticWizard:
This PR is corresponding to the issue #1073 (created by @jrobbins)

- Removed the import of app identity library, thereby, reducing dependence on GAE SDK for smooth transition to python3 later on.
- Used the value of key 'APPLICATION_ID' from os.environ as the new APP_ID

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -37,12 +37,15 @@ DEV_MODE = os.environ['SERVER_SOFTWARE'].startswith('Development')
 UNIT_TEST_MODE = os.environ['SERVER_SOFTWARE'].startswith('test')
 
 #setting GOOGLE_CLOUD_PROJECT manually in dev mode
-APP_ID = os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev') if DEV_MODE else os.environ['GOOGLE_CLOUD_PROJECT']
+if DEV_MODE or UNIT_TEST_MODE:
+  APP_ID = os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev')  
+else: 
+  APP_ID = os.environ['GOOGLE_CLOUD_PROJECT']
 
 SITE_URL = 'http://%s.appspot.com/' % APP_ID
 CLOUD_TASKS_REGION = 'us-central1'
 
-if UNIT_TEST_MODE == 'testbed-test':
+if UNIT_TEST_MODE:
   APP_TITLE = 'Local testing'
   SITE_URL = 'http://127.0.0.1:8888/'
 elif DEV_MODE:
--- a/settings.py
+++ b/settings.py
@@ -4,8 +4,6 @@ from __future__ import print_function
 import logging
 import os
 
-from google.appengine.api import app_identity
-
 
 #Hack to get custom tags working django 1.3 + python27.
 INSTALLED_APPS = (
@@ -39,7 +37,7 @@ DEV_MODE = os.environ['SERVER_SOFTWARE'].startswith('Development')
 UNIT_TEST_MODE = os.environ['SERVER_SOFTWARE'].startswith('test')
 
 
-APP_ID = app_identity.get_application_id()
+APP_ID = os.environ['APPLICATION_ID']
 SITE_URL = 'http://%s.appspot.com/' % APP_ID
 CLOUD_TASKS_REGION = 'us-central1'
 
--- a/settings.py
+++ b/settings.py
@@ -36,12 +36,13 @@ SEND_EMAIL = False  # Just log email
 DEV_MODE = os.environ['SERVER_SOFTWARE'].startswith('Development')
 UNIT_TEST_MODE = os.environ['SERVER_SOFTWARE'].startswith('test')
 
+#setting GOOGLE_CLOUD_PROJECT manually in dev mode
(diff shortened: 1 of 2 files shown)


CrypticWizard:
@googlebot I signed it!

settings.py
```diff
@@ -39,7 +37,7 @@
 UNIT_TEST_MODE = os.environ['SERVER_SOFTWARE'].startswith('test')
 
 
-APP_ID = app_identity.get_application_id()
+APP_ID = os.environ['APPLICATION_ID']
```
jrobbins, line 40: I don't think that this quite works because
os.environ['APPLICATION_ID'] has a value with a "~" in it.
E.g., "s~cr-status-staging"

Also, while APPLICATION_ID seems to be present for python2 on appengine, it is not in the list of environment variables that are supposed to be there.  See:
https://cloud.google.com/appengine/docs/standard/python/runtime
and
https://cloud.google.com/appengine/docs/standard/python3/runtime

Both of those have GOOGLE_CLOUD_PROJECT, which seems to have just exactly the string that we need.  However, there is no GOOGLE_CLOUD_PROJECT when running locally.  So, you could use os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev') which will return 'dev', which should be OK.

The condition on line 44 will fail because GOOGLE_COULD_PROJECT will never be 'testbed-test', so you could instead use UNIT_TEST_MODE for that condition.

jrobbins (changes requested):

settings.py
CrypticWizard: Done!

settings.py
```diff
@@ -38,12 +36,13 @@
...
 
+#setting GOOGLE_CLOUD_PROJECT manually in dev mode
+APP_ID = os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev') if DEV_MODE else os.environ['GOOGLE_CLOUD_PROJECT']
 
-APP_ID = app_identity.get_application_id()
 SITE_URL = 'http://%s.appspot.com/' % APP_ID
 CLOUD_TASKS_REGION = 'us-central1'
 
-if APP_ID == 'testbed-test':
+if UNIT_TEST_MODE == 'testbed-test':
```
shivamag00, line 45: `UNIT_TEST_MODE` will be either `TRUE` or `FALSE`, depending on, if the value of  `os.environ['SERVER_SOFTWARE']` starts with `test` (See line 39). Hence, I would recommed that you change this line to following:-
`if UNIT_TEST_MODE:`

shivamag00 (commented):
Travis CI is failing currently. The following changes should prevent Travis CI from failing.

settings.py
```diff
@@ -38,12 +36,13 @@
 DEV_MODE = os.environ['SERVER_SOFTWARE'].startswith('Development')
 UNIT_TEST_MODE = os.environ['SERVER_SOFTWARE'].startswith('test')
 
+#setting GOOGLE_CLOUD_PROJECT manually in dev mode
+APP_ID = os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev') if DEV_MODE else os.environ['GOOGLE_CLOUD_PROJECT']
```
shivamag00, line 40: @CrypticWizard, Tests on Travis CI are failing due to the following issue - if `DEV_MODE` is `False` and `UNIT_TEST_MODE` is` True`, `APP_ID` will be assigned using `os.environ['GOOGLE_CLOUD_PROJECT']` which will raise a `key error` since `GOOGLE_CLOUD_PROJECT` does not exist while running tests in Travis CI. Hence, I would recommend that you add `UNIT_TEST_MODE` in the condition using `OR` as shown below

`APP_ID = os.environ.get('GOOGLE_CLOUD_PROJECT', 'dev') if DEV_MODE or UNIT_TEST_MODE else os.environ['GOOGLE_CLOUD_PROJECT']`

settings.py
shivamag00, line 45: `UNIT_TEST_MODE` will be either `TRUE` or `FALSE`, depending on, if the value of `os.environ['SERVER_SOFTWARE']` starts with `test` (See line 39). Hence, I would recommed that you change this line to following:-
`if UNIT_TEST_MODE:`

shivamag00 (changes requested):

settings.py
CrypticWizard: Done!

settings.py
CrypticWizard: Done! Thanks for your valuable input.

static/elements/chromedash-metadata.js
```diff
@@ -154,6 +154,7 @@ class ChromedashMetadata extends LitElement {
 
   _processResponse(response) {
     // TODO(ericbidelman): Share this data across instances.
+    console.log(response);
```
jrobbins, line 157: I don't think this belongs in this PR.  Please revert this file to have a more focused PR.

jrobbins (changes requested):
The settings.py changes look good.

static/elements/chromedash-metadata.js
CrypticWizard: My bad. Its done btw. Sorry for the inconvenience.

jrobbins (approved):

PR #1214, opened 2021-03-17
Fixed the footer to the bottom for easy accessibility

### LINKED ISSUE #1096: Links at bottom of page are hard to get to

**Describe the bug**
The "Discuss   File an issue       Privacy      About" links at the very bottom of https://chromestatus.com/features are below the very long list of features, which makes them hard to discover and hard to get to even when you know they're there.

**To Reproduce**
1. Go to https://chromestatus.com/features
2. Scroll down.
3. Keep scrolling.
4. Some more scrolling.
5. Find the links you were looking for.

**Expected behavior**
These links are available from some sort of menu without scrolling past all of the individual features.

CrypticWizard:
Merging this will close #1096 

Since the features list isn't paginated yet, one had to scroll to the bottom to reach the footer links. To solve this I have fixed the position of the footer to the bottom of the viewport. This can be reverted back once we are done paginating the features list.

- **Earlier**

> Screenshot from 2021-03-17 21-25-18

- **Now** 

> Screenshot from 2021-03-17 21-24-35

### DIFF
--- a/static/sass/main.scss
+++ b/static/sass/main.scss
@@ -195,7 +195,11 @@ header {
 
 footer {
   background: var(--footer-background);
+  background-color: #fff;
   box-shadow: 0 -2px 5px var(--shadow-color);
+  width: 100%;
+  position: fixed;
+  bottom: 0;
   display: flex;
   flex-direction: column;
   justify-content: center;


jrobbins:
Thanks for the effort that you put into this change.  If there's a way to give you credit for it in GSOC stats or something, I will be happy to do that, even though I don't think that we should land this PR.

I should have been more proactive in laying out my preferred resolution approach on the issue.   This site used to have a bunch of floating elements on every page and it turned out to be a big pain to use and it complicated every UI change.  So, I want to keep a very simple, flat page layout and address the root causes of current UI decisions that don't scale up to handle the content that we have.

CrypticWizard:
Not a problem at all, sir. I understand the underlying issues. In fact, while working on this, I went a bit further and tried to solve other UI issues as well related to some elements' positioning and responsiveness but soon realised what I had stepped into. So, I retracted back.  
I would love to be a part of the UI revamp process if possible since I love doing it. Also, I'm currently working on pagination. I hope that would help in resolving some of the UI issues.

PR #1217, opened 2021-03-18
Remove lighthouse-ci travis target.

jrobbins:
This code was supposed to run lighthouse-ci in travis on every pull request and display the result as a comment on the pull request discussion.  I think that is kind of neat, however I don't think that it has worked in the last couple of years.  Specifically, the step that deploys a test version to our production site just exits without actually deploying.   I could potentially get that working, but I don't like the idea of a travis service account having the ability to deploy software to our servers, so I removed that IAM role.

Since these deployments have not been happening, there is no "lighthouse-ci-staging" version of our app to test.  AppEngine routes those requests to the default version of the app.  So, the lighthouse scores that we are seeing are the ones for the currently live version of the app, not for the change that is under review.  Anyone could run lighthouse against the live site just by using chrome dev tools.

I'd prefer to remove this complexity from our travis config and the clutter from our pull request conversations.


jstenback (approved):
Agreed, while this would be nice, it hasn't worked in a long time. Let's get rid of it and consider what else makes sense along these lines down the road as needed.

PR #1218, opened 2021-03-18
Hide components by default on the schedule page.

jrobbins:
This is a baby step toward the UI changes I presented to the team recently.

In this CL:
+ Change the "Hide blink components" checkbox to "Show blink components" and reverse the logic.
+ Streamline the styling by removing the list item numbering.

### DIFF
--- a/static/elements/chromedash-schedule.js
+++ b/static/elements/chromedash-schedule.js
@@ -44,8 +44,9 @@ class ChromedashSchedule extends LitElement {
 
   static get properties() {
     return {
-      channels: {attribute: false}, // Assigned in schedule.js, value from Django
-      showBlink: {attribute: false}, // Edited in schedule.js
+      // Assigned in schedule-apge.js, value from Django
+      channels: {attribute: false},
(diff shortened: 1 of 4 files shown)


static/elements/chromedash-schedule.js
```diff
@@ -45,7 +45,7 @@ class ChromedashSchedule extends LitElement {
   static get properties() {
     return {
       channels: {attribute: false}, // Assigned in schedule.js, value from Django
-      hideBlink: {attribute: false}, // Edited in schedule.js
+      showBlink: {attribute: false}, // Edited in schedule.js
```
jstenback, line 48: Is the comment inaccurate there, or do we need to update shedule.js as well?

jstenback (approved):
LGTM with the comment cleaned up or otherwise resolved!

static/elements/chromedash-schedule.js
jrobbins: That comment must be out of date because there is no schedule.js.  The file is named schedule-page.js and that is included in this PR.  I updated the comment.

PR #1219, opened 2021-03-18
Remove obsolete stars and cues handlers.

jrobbins:
These XHR handlers were rewritten as part of a more organized JSON REST API in https://github.com/GoogleChrome/chromium-dashboard/commit/6bba409a33232a8642bf30334e446eccae81dd3a.  That change went live on March 4, 2021, which was 14 days ago.  So, anyone who has loaded a chromestatus page since then has been hitting the new URLs and the new code.

After landing this PR, I will most likely deploy it next week, which means that users who have had a browser tab open since before March 4 (three weeks by then) will find that clicking the star icon no longer works.   There will be some such users, but probably not too many.  And, losing a star click is easily worked-around by reloading the page and clicking the star again.


jstenback (approved):
LGTM

PR #1220, opened 2021-03-18
20210319 fix timeline narrrow ports

### LINKED ISSUE #302: Ability to choose between 'stack rank' and 'timeline' disappears on narrow viewports

shivamag00:
Merging this will close #302 
As @jrobbins suggested [here](https://github.com/GoogleChrome/chromium-dashboard/issues/302#issuecomment-802165384), there are two common approaches to fix the issue:-
(1) introducing a hamburger menu to the site or some button that would show the side nav temporarily overlapping other content on the page, or
(2) displaying the side-nav items in a a horizontal layout across the top of the main content area.

I have followed the second approach. 

Snapshots:-

On wide viewport
w

On narrow viewport
n

**Note**

1. The All Properties horizontal row will only appear for the routes `metrics/css/popularity` and `/metrics/css/timeline/popularity`


static/sass/main.scss
```diff
@@ -427,6 +427,11 @@ footer {
 
 }
 
+//Horizontal Properties Row should be visible only in narrow view ports
+@media only screen and (min-width: 700px){
+  #properties-row{display: none}
```
jrobbins, line 432: This could be useful on other pages too.  Let's generalize the name of this element ID.  Let's try something like #horizontal-sub-nav.

templates/metrics/css/popularity.html
```diff
@@ -9,6 +9,14 @@
   {% include "metrics/css/_css_metric_nav.html" %}
 {% endblock %}
 
+{% block properties %}
+<nav class="data-panel" style="font-size: smaller;">
```
jrobbins, line 13: I don't think that we need "smaller".  Does it look OK with regular-sized text?

templates/metrics/css/popularity.html
```diff
@@ -9,6 +9,14 @@
...
+  All Properties:<a href="/metrics/css/timeline/popularity" style="padding-left: 2%; padding-right: 2%;">Timeline</a> |
```
jrobbins, line 14: The phrase "All properties" is redundant because it already appears in the heading line just below this nav.  Try "View as:" instead.

templates/metrics/css/popularity.html
```diff
@@ -9,6 +9,14 @@
   {% include "metrics/css/_css_metric_nav.html" %}
 {% endblock %}
 
+{% block properties %}
+<nav class="data-panel" style="font-size: smaller;">
+  All Properties:<a href="/metrics/css/timeline/popularity" style="padding-left: 2%; padding-right: 2%;">Timeline</a> | 
+  <a href="/metrics/css/popularity"  style="padding-left: 2%; padding-right: 2%;">Stack Rank</a>
```
jrobbins, line 15: For this link and the others, please remove the style attributes and add CSS rules instead.  You could put it in metrics.scss after .description.    Also, rather than padding as a %, try using var(--content-padding-quarter) or just 4px.

jrobbins (changes requested):
This looks like it is on the right track, but I have some requested changes.

templates/metrics/css/popularity.html
shivamag00: I made it smaller so that it looks  different than the other text elements on the same page. However, I will remove it as you have suggested.

static/sass/main.scss
shivamag00: Done

templates/metrics/css/popularity.html
shivamag00: Done

templates/metrics/css/popularity.html
shivamag00: Done

shivamag00:
@jrobbins I have made all the changes that were requested. Following snapshot illustrates how the horizontal nav bar appears now
n

jrobbins (approved):

Issue #1013, opened 2020-09-13
Percentage values order is not changed on the page "usage > all features > stack rank" when clicking on "Percentage" column title


d-01:
**Describe the bug**
The percentage values order is not changed on the page *usage > all features > stack rank* (<https://www.chromestatus.com/metrics/feature/popularity>) when clicking on the *Percentage* column title.

**To Reproduce**
1. Go to <https://www.chromestatus.com/metrics/feature/popularity>
   1. Notice the default sort order is "by *Percentage* descending".
2. Click on the *Percentage* column title.
   1. Notice the sort order has changed from descending to ascending.
   1. Feature names order has changed properly.
   1. Bars order has changed properly.
   1. Percentage values order remains unchanged, thus not in sync with feature names.
3. Screenshots.

   Descending sort order (default) by *Percentage*:

   ![1_desc_order_](https://user-images.githubusercontent.com/45796739/93007510-4f13dc80-f572-11ea-9261-88098567fd26.png)

   Ascending order by *Percentage*:

   ![2_asc_order_](https://user-images.githubusercontent.com/45796739/93007554-ab76fc00-f572-11ea-8191-f4119681bce3.png)

**Expected behavior**
When clicking on the *Percentage* column title, the percentage values order should get changed accordingly.

**Additional context**
This behavior reproduced with:
1. Firefox 80.0 (64-bit), Windows 7 x64
2. Chromium 72.0.3626.121 (32-bit), Windows 7 x64

CrypticWizard:
@jrobbins This issue seems resolved now.

d-01:
The issue resolved. Confirmed just now in:
- [x] Firefox 88.0
- [x] Chromium 84.0.4147.135

PR #1224, opened 2021-03-19
Fix sorting of percentage values

CrypticWizard:
This PR is in response to issue #1013 

The percentage value was being calculated in the firstUpdated() method, hence didn't update after the first insert. **I wasn't able to test this code though because properties data is not available locally**.

### DIFF
--- a/data/devrel_team.yaml
+++ b/data/devrel_team.yaml
@@ -1,8 +1,3 @@
-name: Adam Argyle
-email: argyle@google.com
-blink_components:
-  - Blink>CSS
----
 name: Addy Osmani
 twitter: addyosmani
 email: addyo@google.com
--- a/data/devrel_team.yaml
+++ b/data/devrel_team.yaml
@@ -1,3 +1,8 @@
+name: Adam Argyle
+email: argyle@google.com
+blink_components:
+  - Blink>CSS
+---
 name: Addy Osmani
 twitter: addyosmani
 email: addyo@google.com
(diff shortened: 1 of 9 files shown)


data/devrel_team.yaml
```diff
@@ -1,3 +1,8 @@
+name: Adam Argyle
```
jrobbins, line 1: This doesn't belong in this PR.

static/elements/chromedash-x-meter.js
```diff
@@ -42,19 +42,15 @@ class ChromedashXMeter extends LitElement {
...
   showTimeline() {
     window.location.href = this.href;
   }
 
   render() {
     return html`  
-      <div style="width: ${(this.value / this.max * 100)}%">  
-        <span>${this.valueFormatted}</span> 
+      <div @click="${this.showTimeline}" style="width: ${(this.value / this.max * 100)}%">  
+        <span>${this.value <= 0.000001 ? '<=0.000001%' : this.value + '%'}
```
jrobbins, line 52: Since this code no longer uses this.valueFormatted, you should delete line 9.

jrobbins (changes requested):

data/devrel_team.yaml
CrypticWizard: Done!

static/elements/chromedash-x-meter.js
CrypticWizard: Done!

jrobbins (approved):

PR #1232, opened 2021-03-23
Fixes chromedash-feature-collapse 

### LINKED ISSUE #636: accessibility: cannot click to select text without causing chromedash-feature to close
The chromedash-feature widgets contain text which cannot be selected without closing the widget. This seems to only happen for clicks on text within the description (class="desc").

Many of our feature descriptions contain links-as-text or code snippets in their descriptions which cannot be selected.

A work around is to click the 'view on a standalone page' button, as the standalone page version does not have this same bug.

Reproduction:
1) open https://www.chromestatus.com/features/5144752345317376
2) notice that the text includes a link as text to a google groups page
3) attempt to click that link, feature widget closes
4) re-open feature
5) attempt to highlight that link text, feature widget closes

Expected:
3) link to google group should be a link
5) selecting text should not close widget

CrypticWizard:
The old conditions to check whether the clicked element is a link weren't working because _e.currentTarget_ wasn't returning the required element. 

I used the _nodeName_ property to determine if it's a link and the _type_ property of _window.getSelection()_ object if text has been selected. Some browsers don't return the _type_ property, so I added another condition for that.

Fixes #636
@jrobbins

### DIFF
--- a/static/elements/chromedash-feature.js
+++ b/static/elements/chromedash-feature.js
@@ -140,10 +140,14 @@ class ChromedashFeature extends LitElement {
   }
 
   _togglePanelExpansion(e) {
-    // Don't toggle panel if tooltip or link is being clicked.
+    // Don't toggle panel if tooltip or link is being clicked
+    // or if text is being selected.
     const target = e.currentTarget;
+    const textSelection = window.getSelection();
+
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
This fix looks good.  Sorry to take so long to review it.

PR #1237, opened 2021-03-24
Flesh out permissions.py and simplify page logic.

jrobbins:
This is part of the work needed to clear the way for us to upgrade to Google Sign-In, and also to have more precise permissions needed for API Owner reviews.

In this PR:
+ permissions.py: Implement some permission checking functions and a decorator for each one
+ permissions_test.py unit tests for those
+ Remove some old permission checking functions in permissions.py and basehandlers.py 
+ Edit various page classes to use the decorator version
+ The new permission functions consistently redirect anon GETs to the login page, so some unit tests needed to specify POST to trigger the 403 response.

### DIFF
--- a/framework/permissions_test.py
+++ b/framework/permissions_test.py
@@ -21,10 +21,13 @@ import testing_config  # Must be imported before the module under test.
 import mock
 import werkzeug.exceptions  # Flask HTTP stuff.
 
+from google.appengine.api import users
+
+from framework import basehandlers
 from framework import permissions
 
 
-class MockHandler(object):
+class MockHandler(basehandlers.BaseHandler):
 
   def __init__(self):
     self.called_with = None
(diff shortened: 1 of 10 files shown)


framework/permissions_test.py
```diff
@@ -21,35 +21,87 @@
...
+      self, func, additional_args,
+      normal='missing', special='missing', admin='missing', anon='missing'):
+    """Test func under four conditions and check expected results."""
+    # Test normal users
+    testing_config.sign_in('user@example.com', 123)
+    user = users.get_current_user()
+    self.assertEqual(normal, func(user, *additional_args))
+
+    # Test special users
+    testing_config.sign_in('user@google.com', 123)
```
cwilso, line 64: A TODO here (and two lines down) to replace @google/@chromium references with scalable solution.

cwilso (approved):
This all looks good.  I would suggest a TODO to capture that sooner or later we need to come up with a better system than explicitly referring to @google.com and @chromium.org users.

Issue #1209, opened 2021-03-17
"Ready for Trial" email doesn't include flag/instructions to try out feature


cwilso:
**Guide UX Feedback**

When drafting a "ready for trial" email, the template does not include the "Flag name" field from the implementation section; it should, because we need to give instructions (in the RFT email) to HOW to try out the feature.  Perhaps we need more hooks for samples, etc. in the RFT as well, and we could probably drop some of the fields (like, TAG review, goals for experiment).

PR #1239, opened 2021-03-24
Add flag_name to intent emails.

jrobbins:
I think this addresses issue #1209.

This just always shows the flag name in every intent email.  If it has not been filled in yet, it will just be a blank space in the intent email.

I let it appear in all intents because I am thinking that it might be useful in all intent emails, even for later stages.  Let me know if you think it is worth focusing these emails more.


cwilso (approved):
LGTM!

Issue #1111, opened 2020-12-11
Process model fell out of step with launching-features doc


jrobbins:
**Describe the bug**
When tracking a feature on the "Implementations of already-defined consensus-based standards" process, the Guide UX shows the first step as "Identify" whereas the launching-features doc says it is "Start Prototyping" and requires an I2P email.

**To Reproduce**
1. Go to the editing page process overview for a feature that is an implementation of an existing standard.
2. Try to follow the exact instructions on the launching-features doc.

**Expected behavior**
There should be a "Start Prototyping" step for that process, but it says "Identify".

**Additional context**
As a work-around, people can generate an I2P email by clicking one of the other intent links and changing the last number in the URL to a 1.

jrobbins:
@cwilso Is the launching-features page under version control?
I'm just wondering if I missed something that was there originally or if I missed a change to it.

PR #1250, opened 2021-04-01
Add finch_url to the ship phase of all proceses except deprecation.

jrobbins:
This addresses issue #1111 and the doc/emails that we have been discussing.

In this PR:
+ Add finch_url field to DB
+ Add finch_url form field and processing of POST data.
+ Locate the finch_url in the "Ship" stage of any process except deprecation.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -1075,6 +1075,8 @@ class Feature(DictModel):
   experiment_extension_reason = db.StringProperty(multiline=True)
   ongoing_constraints = db.StringProperty(multiline=True)
 
+  finch_url = db.LinkProperty()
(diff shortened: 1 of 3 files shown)


jrobbins:
Locating this field in the "Ship" phase is different than what the launching-features doc says.  I'm proposing it here because I am guessing that it would be useful for other feature types as well.  And, a single optional field doesn't seem like enough info to justify a separate "(Optional) Finch" stage, especially since shipping with finch is a way to ship a feature.   However this is just a stab in the dark.  What do you think?

cwilso (approved):

Issue #1242, opened 2021-03-29
Proposed, In development, and Deprecated features are not showing on mobile view.


khubaibalam2000:
**Describe the bug**
On desktop, there is a side panel in which proposed, in development, deprecated features, and milestones can be seen. But this functionality is not showing on the mobile. The attached screenshot is of the desktop view of site, in mobile view, one cannot see this side panel.

![image](https://user-images.githubusercontent.com/74249716/112805058-8d946300-908e-11eb-9dbd-d1e387e657ce.png)

**To Reproduce**
On mobile
1. Go to https://chromestatus.com/features
2. Try to find the filter of 
    a. Proposed 
    b. In development 
    c. Deprecated
    d. Milestones

**Expected behavior**
I think there should be a filter for the proposed, in development, and deprecated features. Also there should be a milestone page or functionality.

shubhammittal588:
@media only screen and (max-width: 700px)
#drawer-column {
    display: none;
}
the display with (max-width:700) is set to none that's why it disappear in mobile mode. We can create a dropdown taking the id="drawer-column" setting min-width=699px
![Screenshot 2021-04-03 at 4 59 13 PM](https://user-images.githubusercontent.com/77587202/113477286-6c70af80-949e-11eb-8922-1d975dee27ba.png)

shubhammittal588:
correction:
max-width:700px with display: block; in a dropdown

Issue #1240, opened 2021-03-27
Milestone page list without scrollable sidebar


khubaibalam2000:
**Is your feature request related to a problem? Please describe.**
While opening milestone pages on chromestatus.com, if you open web page of milestones below 50 or 60, then you have to again scroll to the top of the page. Sometimes, it is annoying, as you are continuously opening multiple milestone pages and again and again scroll to the top to see features list.

**Describe the solution you'd like**
If we just put the milestone pages in a scrollable sidebar, then this problem will resolve, and web page will also look more better.

**Additional context**
https://user-images.githubusercontent.com/74249716/112718842-6e26fa00-8f17-11eb-988e-56f37a02f599.mp4

khubaibalam2000:
I am willing to work on this issue.

jrobbins:
I'm planning on eventually removing the side nav altogether and instead using a set of filter widgets located above the list.  But, I am not ready to spec out that design yet.

khubaibalam2000:
I worked on UIs and designs in my some of my past university projects. I have a design in my mind. Can I share that design with you?

shivamag00:
Just For the record, #1113 contains a proposal for replacing left hand navigation.

PR #1254, opened 2021-04-05
Bugfix #1242 - Proposed, In development, and Deprecated features are not showing on mobile view

shubhammittal588:
## What?
This addresses issue #1242 and made changes in CSS (main.css).

## Why?
```
@media only screen and (max-width: 700px)

#drawer-column {
    display: none;
}
```

the display with (max-width:700) is set to none that's why it disappear in mobile mode. We can create a dropdown taking the 
```
id="drawer-column" max-width:700px with display: block; 
```

## How?
Edited the main.css file and made drawer-column to display: block which was previously set to none.

### DIFF
--- a/static/sass/main.scss
+++ b/static/sass/main.scss
@@ -365,7 +365,7 @@ footer {
 
 
   #drawer-column {
-    display: none;
+    display: block;
   }
   #content-column {
     padding-left: var(--content-padding-half);


shivamag00:
@shubhammittal588, thank you for putting in your efforts to solve this bug. However, this may not be the right approach as simply changing the CSS display property without creating a dropdown, will result in a UI that is difficult to use. I have created a dropdown in #1325 while solving #1240 . I will now try to place this above or below the search box in narrow viewports as suggested by @jrobbins.

shivamag00:
UPDATE:- I have raised PR #1334 to display the dropdown in narrow viewports.

jrobbins:
PR #1334 was merged.

PR #1255, opened 2021-04-05
Fix module name in notifier config file.

jrobbins:
AppEngine gave an error for a very old task that had been stuck in our task queue since Dec 2020.  Surprisingly, recently created tasks seem to work with either the old or new version of this config file, and I verified that they are being routed through the correct GAE service.  Even if both work, this seems more correct.


jstenback (approved):
LGTM

PR #1259, opened 2021-04-09
Add entities, perms, and API handlers for approvals.

jrobbins:
In this CL:
+ Model class for approvals
+ API handlers to get and set approvals
+ Permission checking to see who can set approvals
+ Fetching and parsing the API_OWNERS file
+ Added known approvals to the various processes.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -1,0 +1,91 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import division
+from __future__ import print_function
+
+import collections
+import logging
+import requests
+
+
+ONE_LGTM = 'One LGTM'
+THREE_LGTM = 'Three LGTMs'
+API_OWNERS_URL = (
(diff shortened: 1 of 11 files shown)


jrobbins:
The Travis CI failures are due to an SSL error in the CI service.  I don't think that they are related to this PR, and this code passes all unit tests locally.

internals/approval_defs.py
```diff
@@ -0,0 +1,91 @@
...
+  """Load a list of email addresses from an OWNERS file."""
+  owners = []
+  response = requests.get(url)
+  if response.status_code != 200:
+    logging.error('Could not fetch %r', url)
+    logging.error('Got response %r', response)
+    raise ValueError('Could not get OWNERS file')
+
+  for line in response.iter_lines():
+    print('line is %r' % line)
```
jstenback, line 74: Guessing you don't want to leave this print() in...

jstenback (approved):
Approved, but please have a look at the comments I left.

internals/approval_defs.py
```diff
@@ -0,0 +1,91 @@
...
+  return owners
+
+
+def get_approvers(field_id):
+  """Return a list of email addresses of users allowed to approve."""
+  afd = APPROVAL_FIELDS_BY_ID[field_id]
+
+  if isinstance(afd.approvers, str):
+    owners = fetch_owners(afd.approvers)
+    return owners
```
jstenback, line 91: Did you mean to set afd.approvers here as well? If not, I don't see how it's ever not a string?

jrobbins: That's correct.  In the only examples that we have right now for API Owners the value of afd.approvers will be a URL to fetch.  We keep that URL so that we can fetch it again periodically.    In the future, we may have some other approval types that just have a hard-coded list of approvers.  I'll add a comment to explain that.

cwilso:
Is this for API owner LGTMs, or horizontal reviews like Security/Privacy?

jrobbins:
> Is this for API owner LGTMs, or horizontal reviews like Security/Privacy?

API owners at first, but I am laying the foundation to generalize to horizontal reviews eventually.

jstenback (approved):

PR #1260, opened 2021-04-10
Refactor code for accessing JSON parameters.

jrobbins:
Now that I am starting to write more API handlers, I want to reduce the amount of redundant code for simply accessing JSON parameters in the request and raising validation errors.

In this PR:
+ Add a required= option when getting the current user so that the error response can be generated in just one place.
+ Add methods to get JSON parameters of any type, int, or bool.
+ Add get_specified_feature() to go through the steps of getting the ID and retrieving the associated feature or returning an error.
+ Filled in some missing unit tests.


jstenback (approved):
LGTM

PR #1264, opened 2021-04-13
Give 404 or 400 instead of 500 for unexpected HTTP methods.

jrobbins:
This avoids some noise on the GAE Error Reporting page.  

Some spammers or fuzzers try to do a POST to our site home page.  With our existing code, that actually caused a python exception and gave the requester a 500 response.  We don't really care about those requests, but giving a non-500 response makes it easier to spot true defects in our logs and on the GAE Error Reporting page.

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -46,16 +46,16 @@ class BaseHandler(flask.views.MethodView):
   def request(self):
     return flask.request
 
-  def abort(self, status, msg=None):
+  def abort(self, status, msg=None, **kwargs):
     """Support webapp2-style, e.g., self.abort(400)."""
     if msg:
       if status == 500:
         logging.error('ISE: %s' % msg)
       else:
         logging.info('Abort %r: %s' % (status, msg))
-      flask.abort(status, msg)
+      flask.abort(status, description=msg, **kwargs)
     else:
-      flask.abort(status)
+      flask.abort(status, **kwargs)
 
   def redirect(self, url):
     """Support webapp2-style, e.g., return self.redirect(url)."""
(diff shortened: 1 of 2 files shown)


framework/basehandlers_test.py
```diff
@@ -362,9 +400,9 @@ def test_get_template_path__specalized_by_template_data(self):
     self.assertEqual('special.html', actual)
 
   def test_process_post_data__missing(self):
-    """Every subsclass should overide process_post_data()."""
+    """Subsclass that don't overide process_post_data() give a 400."""
```
jpchase, line 403:
```suggestion
    """Subclasses that don't override process_post_data() give a 400."""
```

Nit: Spelling.

framework/basehandlers.py
```diff
@@ -153,21 +153,22 @@ def delete(self, *args, **kwargs):
...
+  def do_get(self, **kwargs):
     """Subclasses should implement this method to handle a GET request."""
+    # Every API handler must handle GET.
     raise NotImplementedError()
 
-  def do_post(self):
+  def do_post(self, **kwargs):
     """Subclasses should implement this method to handle a POST request."""
-    raise NotImplementedError()
+    self.abort(400, msg='Unexpected HTTP method')
```
jpchase, line 163: I had to remind myself of the various HTTP status codes. Looking at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses), it seems like a 405 would be more appropriate than 400. Although, using 405 seems to require also responding with an [Allow header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow).

That said, if this is to primarily handle malicious requests, then maybe it doesn't matter if we use the "best" status code.

jpchase (commented):
One question about the appropriate status code to use.

framework/basehandlers_test.py
jrobbins: Fixed.

framework/basehandlers.py
jrobbins: You are right that 405 is a better response status.  I went ahead and used that.  Flask already has some support for setting the allow header, I just needed to tell it which methods are defined in the API handler subclass.

jpchase (approved):
LGTM.

PR #1266, opened 2021-04-15
Implement XSRF tokens for HTML forms (not enforced yet).

jrobbins:
As a prerequisite to doing more with approvals, we need to improve security.

In this PR:
+ Generate XSRF tokens for HTML forms the same way it is done in Monorail
+ Check those tokens when the forms are submitted, but rejections are just logged for now

Have not started giving error responses yet because doing so would reject form submission from users who already have browser tabs open before this is deployed.   So, we just log invalid tokens for now.  I'll wait a while until we no longer see rejections logged, and then enable checking.

### DIFF
--- a/framework/basehandlers_test.py
+++ b/framework/basehandlers_test.py
@@ -536,7 +536,7 @@ class FlaskHandlerTests(unittest.TestCase):
   @unittest.skip('TODO(jrobbins): enable after next release')
   @mock.patch('settings.UNIT_TEST_MODE', False)
   def test_require_xsrf_token__missing(self):
-    """We reject a POST with a valid token."""
+    """We reject a POST with a missing token."""
     testing_config.sign_in('user1@example.com', 111)
     form_data = {}
     with test_app.test_request_context('/test', data=form_data):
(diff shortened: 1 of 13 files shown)


framework/basehandlers_test.py
```diff
@@ -521,3 +524,31 @@ def test_require_task_header__missing(self):
...
+    """We accept a POST with a valid token."""
+    testing_config.sign_in('user1@example.com', 111)
+    form_data = {'token': xsrf.generate_token('user1@example.com')}
+    with test_app.test_request_context('/test', data=form_data):
+      self.handler.require_xsrf_token()
+
+  @unittest.skip('TODO(jrobbins): enable after next release')
+  @mock.patch('settings.UNIT_TEST_MODE', False)
+  def test_require_xsrf_token__missing(self):
+    """We reject a POST with a valid token."""
```
jstenback, line 539: That should be "missing", not "valid"?

jstenback (approved):
LGTM with the one comment looked into.

jstenback (approved):

framework/basehandlers_test.py
jrobbins: Fixed.

PR #1275, opened 2021-04-18
20210418 google sign in

### LINKED ISSUE #1201: Use Google Sign-In for authentication
Rather than continue to use the GAE 'users' library that is only available for python2, we need to use some form of oauth for python3.   Google Sign-In seems like a good choice that is available for both.
https://developers.google.com/identity/sign-in/web

**Additional context**
We have a bunch of handlers that use web 1.0 style web forms and we have a multi-page app that loads complete HTML pages when the user navigates.  While that may change in the long term, for the short term we need some way to know the user who is making a request for an HTML page.  

The Google Sign-In library gives an id_token that can be used with XHR requests, but it does not set a cookie, so we would not get anything on page navigation.   To generate personalized (or even permission-checked) pages, we would need a cookie set.  Flask has cookie-based sessions, so maybe we could store the id_token there after it was provided in an XHR request.

As a prerequisite to switching to Google Sign-In, we need to do all permission checking in python code rather than the "login: required" or "login: admin" lines in app.yaml.  Those lines are not available for py3, but even in py2 they will fail to work if we do not use GAE 'users' library cookie-based sessions.

shivamag00:
Merging this should close #1201 

This PR is not intended to be merged at this stage!!!

Also, the code added in this PR may need refactoring. Any feedbacks regarding the same are highly appreciated.

As @jrobbins had mentioned in the issue,  GAE 'users' library has to be replaced with Google Sign-In which is compatible with Python 3 Environment. 
As of now, the following has been done in this PR, 

- I have added the Google Sign-In button in the `header.html` and the necessary JS scripts in `_base.html`
- The Google Sign-In button is working and I am able to fetch the id_token.
- The id_token is being passed to the backend
- The backend is able to check the integrity of this token and fetch email id from the token generated using `google.oauth2.id_token` and `google.auth.transport.requests`

To eliminate the uses of GAE User's library without affecting the existing code, I have done the following
- I have created a new API endpoint - `/auth/v0/login` which is handled by a `API Handler` in `login_api.py` in `api` module. The purpose of this endpoint is to verify the token and then, store it in a session variable.
- Similarly, I have created a new API endpoint - `/auth/v0/logout` which is handled by a `API Handler` in the in `logout_api.py` in `api` module. The purpose of this endpoint is to clear the session.
- There were many methods that were using the following from GAE users' library
      - `User` Class
      - `get_current_user()`
   Hence, in `frameworks/users.py` I have created a new class called `User` which contains the same fields and methods as the GAE users' `User` class. I have also created a method called `get_current_user()`. This method returns an instance of the `User` class by fetching token from the session, verifying and decoding it to get the user details

Left to be done:-
- Removing uses of GAE User's library from *_tests.py
- Checking for users with admin privileges
- Handling the case when the token expires
- Refactor the code

Review Requested:- @jrobbins


shivamag00:
@jrobbins, what is the criteria for classifying a user as admin??

framework/users.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Because administrator status is not persisted in the datastore,
+        `is_current_user_admin()` is a separate function rather than a member function
+        of the `User` class. The status only exists for the user making the current
+        request.
+
+        Returns:
+        `True` if the user is an administrator; all other user types return `False`.
+        """
+        return (os.environ.get('USER_IS_ADMIN', '0')) == '1'
```
jrobbins, line 172: I imagine that this environment variable is set by GAE based on a GAE session cookie.  With Google Sign-In, it will probably never be present.  You can leave this as-is or even just return the constant False.  I figure out another way to represent system administrators and land that change separately.

api/login_api.py
```diff
@@ -0,0 +1,50 @@
...
+  def do_post(self):
+    token = self.get_param('id_token')
+    message = "Unable to Authenticate"
+
+    try:
+      CLIENT_ID = '77756740465-e5r4o15qg4hkdfiucjpl231o79k3ipjv.apps.googleusercontent.com'
+      idinfo = id_token.verify_oauth2_token(token, requests.Request(), CLIENT_ID)
+      # userid = idinfo['sub']
+      # email = idinfo['email']
+      session["id_token"] = token
```
jrobbins, line 43: This makes sense, but it raises the question of how a developer who is working on chromestatus can log in a test users when running locally.  Do you have an idea of how to support that?

framework/users.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def get_current_user():
```
jrobbins, line 175: This also looks pretty good.  However, we could not deploy this to the production server in a single step.  Existing users who have a browser tab open on our site and information typed into a form before the deployment would get an error if they submit the form after the deployment.   We need to check if the user is signed in with the old GAE users library.  I think that case should be supported for POST requests only.  If we continued to honor old GAE users sessions for GETs, then users would never realize that they need to sign in using the new method.

framework/basehandlers.py
```diff
@@ -380,6 +388,7 @@ def FlaskApplication(routes, pattern_base='', debug=False):
   """Make a Flask app and add routes and handlers that work like webapp2."""
 
   app = flask.Flask(__name__)
+  app.secret_key = "abc"
```
jrobbins, line 391: I have a recent CL that securely generates and stores a different server-side secret.  I will write a CL to populate this flask session secret key as well.

jrobbins (commented):
Thanks for doing this! 
This looks like great progress so far.
The general approach seems like the right one.

jrobbins:
@LukeZielinski just FYI that a student is working on this part.

framework/basehandlers.py
jrobbins: I now have this part waiting for review.
https://github.com/GoogleChrome/chromium-dashboard/pull/1285

framework/users.py
shivamag00: Okay, I will add a comment stating the same and leave it as-is.

api/login_api.py
shivamag00: I read on [StackOverFlow](https://stackoverflow.com/questions/17657879/does-google-provide-test-users-for-integration-testing) that Google Sign-In does not support test users. So, the developer will have to log in atleast once by using a Google Account. 

One idea ( but only if we HAVE to support test users) is we can create a separate login system only for the DEV environment.

framework/users.py
shivamag00: Okay... Just to confirm that I understood correctly... We will allow POSTS requests sent by users logged in using the old GAE users library but for GET requests, they will have to login using the newer Google Sign-In

framework/basehandlers.py
shivamag00: Thank you for adding this... I will merge it in this PR

shivamag00:
After the first review, the following tasks are yet to be implemented: -

- Removing uses of GAE User's library from *_tests.py
- Checking for users with admin privileges (@jrobbins will add it separately)
- Handling the case when the token expires
- Refactor the code
- Allow POST requests from users logged in using old GAE Users Library

framework/users.py
shivamag00: Done

framework/basehandlers.py
shivamag00: Done

framework/users.py
shivamag00: @jrobbins, I have implemented it. See `framework/basehandlers.py: line 75` and `framework/permissions.py: Line 33`. Please review.

shivamag00:
@jrobbins I have made some changes so that forms submitted by users logged in using old GAE User's Library are handled properly. I have tested it also according to [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1275#discussion_r619858401) procedure

framework/users.py
jrobbins: Your testing procedure sounds right to me.

The query string parameter to open a dialog also seems like a good solution.  Thanks for thinking through that.  It would be better to initiate the log-in process for the user rather than call alert(), but that can be improved in a later CL.

api/login_api.py
jrobbins: Let's see how it works out to use real accounts in the development environment.  We might not need the complexity of adding a dev-only login system.  And, there is always a risk that a dev-only login system might somehow be enabled in prod by some mistake in the future.

jrobbins:
I'm still working on getting the PR for admin users specified by is_admin.  After that, I will re-review this and test it myself to see if we can get it merged.

api/login_api.py
shivamag00: Okay

shivamag00:
@jrobbins FYI, I will not be able to contribute much this week because of end-semester exams in college.

jrobbins:
@shivamag00 good luck with your exams, and stay safe.

shivamag00:
@jrobbins, I have added your recent changes in the main branch to this PR.

framework/basehandlers.py
```diff
@@ -148,7 +160,10 @@ def get(self, *args, **kwargs):
 
   def post(self, *args, **kwargs):
     """Handle an incoming HTTP POST request."""
-    self.require_signed_in_and_xsrf_token()
+    is_login_request = str(self.request.url_rule) == '/api/v0/login'
+
+    if not is_login_request:
+      self.require_signed_in_and_xsrf_token()
```
shivamag00, line 166: @jrobbins, when a user will login using the Google Sign In, the user will send a POST request to backend to verify the Google Sign In token and store it in a session variable. Since, there is no logged in user when this POST request is made, `self.require_signed_in_and_xsrf_token()` will return a 403 response and user will not be able to login. Hence, I have bypassed this method of POST request is made to  `/api/v0/login` which handles login requests.

shivamag00 (commented):
@jrobbins, executing `npm run test`  after these changes will result in alot of errors. One of the reasons for getting the errors is the following:-

After implementing Google Sign In, we get the current user by decoding the `id_token` (stored in the session variable with the same name - `id_token`). However, while running tests, we initialize environment variables with the details of the test user. These environment variables were used by GAE Users Library to get the current user. Due to differences in the method for getting the current user, `npm run test` finishes with multiple errors. 

As a work-around, I have come up with the following solution:-
While setting up the environment for test, we can initialise the same environment variables that we are currently initialising. In addition to initialising these environment variables, we can set the flask session variable `id_token` with the value `test`. Since the `id_token` generated by the Google Sign In library is a very long and complex string, setting it as `test` should not create any problem. Then, in the method `get_current_user()` of `Framework/users.py`, we can get the current user using the environment variables (initiliased earlier), only when the following condition is met:-
1. The environment is test environment

This will also solve the problem for determining the admin status of the current user while running tests as we will be able to set the admin status using the environment variables.

Please let me know if this solution sounds good so that I can proceed with implementing this solution.

shivamag00:
@jrobbins, Just to verify if it is a typo, shouldn't month be 5 for May here ? 

https://github.com/GoogleChrome/chromium-dashboard/blob/ca1d31b0eac426aba24a0fd4f6017ce048aafc8d/settings.py#L39

jrobbins:
> @jrobbins, Just to verify if it is a typo, shouldn't month be 5 for May here ?
> 
> https://github.com/GoogleChrome/chromium-dashboard/blob/ca1d31b0eac426aba24a0fd4f6017ce048aafc8d/settings.py#L39

Thanks for pointing that out.  My mistake.  I had fixed that as part of the review process, but I forgot to push my changes.  I have pushed and merged it now.

jrobbins (approved):
This all looks good.  Thank you for your work on this.

One detail that I will address in a follow-up CL is the fact that you used user Google Sign-In client ID, whereas I will set up my own.

PR #1283, opened 2021-04-21
Remove unused field intent_template_use_count.

jrobbins:
I don't believe that Feature field intent_template_use_count was ever used.  It was added in your 2019 PR
https://github.com/GoogleChrome/chromium-dashboard/commit/18b81b4fbcf463959783ef126af9d47d5b3b7578
but I don't see where its value was ever output anywhere.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -974,8 +974,6 @@ class Feature(DictModel):
   updated_by = db.UserProperty(auto_current_user=True)
   created_by = db.UserProperty(auto_current_user_add=True)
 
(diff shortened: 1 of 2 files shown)


jstenback (approved):
LGTM

jstenback:
And to clarify, this was added just to have a sense of how often intent templates were used in the tool as we started transitioning developers towards using the tool to generate the intent emails. My plan was to simply use that data, not actually show it, but we never followed through on that. I think it's totally fine to remove this now.

PR #1285, opened 2021-04-22
Refactor Secrets and add session_secret.

jrobbins:
This is step toward solving issue #1201.

As part of the py3 upgrade, we need to replace the GAE users library with Google Sign-In.  However, Google Sign-In does not set a cookie, it only provides an ID token in javascript.  So, we will use flask.session to save the user's ID token.  See
https://github.com/GoogleChrome/chromium-dashboard/pull/1275

Before that can work, we need a session secret.  This PR refactors and expands on the existing code for XSRF secrets.

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -27,11 +27,11 @@ import flask.views
 from google.appengine.api import users
 from google.appengine.ext import db
 
+import settings
 from framework import permissions
 from framework import ramcache
-from framework import xsrf
 from framework import secrets
-import settings
+from framework import xsrf
 from internals import models
 
 from django.template.loader import render_to_string
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -30,6 +30,7 @@ from google.appengine.ext import db
 from framework import permissions
 from framework import ramcache
 from framework import xsrf
+from framework import secrets
 import settings
 from internals import models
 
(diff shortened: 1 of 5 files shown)


framework/basehandlers.py
```diff
@@ -30,6 +30,7 @@
 from framework import permissions
 from framework import ramcache
 from framework import xsrf
+from framework import secrets
```
LukeZielinski, line 33: nit: move up one line to keep alphabetical?

framework/secrets.py
```diff
@@ -0,0 +1,85 @@
...
+      singleton = existing[0]
+
+    if not singleton:
+      logging.info('Creating new secrets')
+      singleton = Secrets()
+      needs_save = True
+
+    if not singleton.xsrf_secret:
+      singleton.xsrf_secret = make_random_key()
+      logging.info('Added XSRF secret: %r' % singleton.xsrf_secret)
```
LukeZielinski, line 61: Maybe a naive question but why is it ok to log the secret? :)

LukeZielinski (approved):

framework/basehandlers.py
jrobbins: Done

framework/secrets.py
jrobbins: Well, only members of the GAE project can view the logs and those same users can view the datastore contents.  But, you are right that it would be better not to log it.   I wanted to have something logged to help debug any future scenario where we somehow change secrets too often (which would then cause us to reject user form submissions and sessions).   I changed it to log only the first 8 characters of the secret's 128, which should be useful for debugging but not enough for someone to forge tokens.

PR #1287, opened 2021-04-26
Implement XSRF token refresh handler.

jrobbins:
This will allow our frontend to request a new XSRF token if the user has a form page open for more than the token expiration time.  This is just the server-side part of it, I will implement the client-side logic in separate PRs.


jstenback (approved):
LGTM

LukeZielinski (approved):

PR #1288, opened 2021-04-27
Allow site admins to be specified via AppUser entities.

jrobbins:
In GAE's python 2 environment, we can call users.is_current_user_admin() to check if the user signed in with a GAE cookie is the same as one of the GAE project admins.   However, GAE cookies are not available in GAE's python 3 environment, in fact, the whole users library is not available.  So, instead we will now keep track of our list users who are administrators of data in our app.

In this CL:
+ Add an is_admin field to AppUser.  This line has been commented out since it was first added in 2013
https://github.com/GoogleChrome/chromium-dashboard/commit/6bff1fbf2a36ebb8407e95689df78fbb6783cfe4
+ Add AppUser methods to return a cached AppUser instance, because we will need one for each permission check.
+ Implement a new case in the logic in permissions.py can_admin_site() to check for is_admin.
+ Add UI form field and form handler logic to create AppUser records with is_admin set, and display which users are admins.

### DIFF
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -24,10 +24,20 @@ from google.appengine.api import users
 from internals import models
 
 
-def can_admin_site(unused_user):
+def can_admin_site(user):
   """Return True if the current user is allowed to administer the site."""
-  # TODO(jrobbins): replace this with user.is_admin.
-  return users.is_current_user_admin()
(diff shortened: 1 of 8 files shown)


jstenback (approved):
LGTM

PR #1291, opened 2021-04-28
Remove serviceworker unregister.

jrobbins:
After we stopped using service worker, I added this code to unregister the worker if it was previously installed for any users.

It was added 10 months do, so I think we can delete this code now.
https://github.com/GoogleChrome/chromium-dashboard/commit/53a92edd2b65485259777341d855ca243ef62053#diff-2fdf9cac2396a7b0dfd648b1512b61bd87eaf37ed45678555eca3d0943e29a78

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -128,17 +128,6 @@ limitations under the License.
     (function() {
       'use strict';
 
-      // Unregister any existing service workers.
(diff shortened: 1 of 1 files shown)


cwilso (approved):
LGTM

PR #1292, opened 2021-04-28
Stop hitting prod when testing locally.

jrobbins:
This removes logic to sometimes hit our prod server (chromestaus.com) when testing locally on localhost (rather than 127.0.0.1). 

This functionality always seemed a little odd to me and I have not made use of it.   There are three reasons to remove this now:
+ Simplify the app so that there is less to explain to new potential contributors
+ Avoid the risk of API requests that might try to alter data to prod
 - This risk will increase because with Google Sign-In we will be using real when testing locally rather than test@example.com.
+ Our OAuth client ID is configured in the Google Cloud console, and it might need a redirect URL which is not allowed to be numeric so we need to be able to use localhost.  I am not sure about that aspect yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ To start the main server and the notifier backend, run:
 ```bash
 npm start
 ```
+Then visit `http://localhost:8080/`.
 
 To start front end code watching (sass, js lint check, babel, minify files), run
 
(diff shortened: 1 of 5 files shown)


cwilso (approved):
LGTM

PR #1304, opened 2021-05-03
Implement site banner.  Set banner for Google Sign-In.

jrobbins:
Implement site banner functionality to display a banner based on info in settings.py.  And, set it for our anticipated deployment of Google Sign-In.

This is based on the way that banners are displayed in Monorail.  The date and time are shown in the user's local format and timezone.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -27,6 +27,17 @@ SEND_ALL_EMAIL_TO = 'cr-status-staging-emails+%(user)s+%(domain)s@google.com'
 
 BOUNCE_ESCALATION_ADDR = 'cr-status-bounces@google.com'
 
+
+# Display a site maintenance banner on every page.  Or, an empty string.
+BANNER_MESSAGE = ('This site will have a new method of signing in. '
+                  'Users will need to sign in again after ')
+
+# Timestamp used to notify users when the read only mode or other status
+# described in the banner message takes effect.  Or, None.  It is
+# expressed as a tuple of ints: (year, month, day[, hour[, minute[, second]]])
+# e.g. (2009, 3, 20, 21, 45) represents March 20 2009 9:45PM UTC.
+BANNER_TIME = (2021, 3, 11, 20, 00)  # May 11, noon pacific
+
 ################################################################################
 
 PROD = False
--- a/static/elements/chromedash-banner.js
+++ b/static/elements/chromedash-banner.js
@@ -1,0 +1,59 @@
+import {LitElement, css, html} from 'lit-element';
+import {nothing} from 'lit-html';
+import SHARED_STYLES from '../css/shared.css';
+
+
+class ChromedashBanner extends LitElement {
+  static get properties() {
+    return {
+      timestamp: {type: Number},
+      message: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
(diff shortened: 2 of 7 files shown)


jrobbins:
Here's what it looks like when running locally.

image

static/elements/chromedash-banner.js
```diff
@@ -0,0 +1,59 @@
...
+  static get properties() {
+    return {
+      timestamp: {type: Number},
+      message: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
+    this.timestamp = (new Date()).getTime() / 1000;
```
liyangguang, line 16: Do we need this initial value? (since there's already `if (!timestamp) return nothing`)

static/elements/chromedash-banner.js
```diff
@@ -0,0 +1,59 @@
+import {LitElement, css, html} from 'lit-element';
+import {nothing} from 'lit-html';
+import SHARED_STYLES from '../css/shared.css';
+
+
+class ChromedashBanner extends LitElement {
+  static get properties() {
+    return {
+      timestamp: {type: Number},
```
liyangguang, line 9: Tiny nit: maybe `timeInSeconds` for better readability.

static/elements/chromedash-banner.js
```diff
@@ -0,0 +1,59 @@
...
+      message: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
+    this.timestamp = (new Date()).getTime() / 1000;
+    this.message = null;
```
liyangguang, line 17: nit: `this.message = ''` or `nothing`? (to match the property type defined above)

liyangguang (approved):

static/elements/chromedash-banner.js
jrobbins: Good point.  I set it to null.

static/elements/chromedash-banner.js
jrobbins: Done.

static/elements/chromedash-banner.js
jrobbins: I think of "timestamp" is being more precise.  Timestamp is a unix timestamp, whereas "timeInSeconds" raises the question of "is that a unix timestamp or a some other measure of time?".   I added a comment to clarify that it is a unix timestamp.

settings.py
```diff
@@ -27,6 +27,17 @@
...
+
+# Display a site maintenance banner on every page.  Or, an empty string.
+BANNER_MESSAGE = ('This site will have a new method of signing in. '
+                  'Users will need to sign in again after ')
+
+# Timestamp used to notify users when the read only mode or other status
+# described in the banner message takes effect.  Or, None.  It is
+# expressed as a tuple of ints: (year, month, day[, hour[, minute[, second]]])
+# e.g. (2009, 3, 20, 21, 45) represents March 20 2009 9:45PM UTC.
+BANNER_TIME = (2021, 3, 11, 20, 00)  # May 11, noon pacific
```
jstenback, line 39: That should be month 5 not 3?

jstenback (approved):
LGTM

settings.py
jrobbins: whoops.  fixed.

PR #1307, opened 2021-05-05
Re-enabled experiment timeline if it is not empty.

jrobbins:
This is based on an email thread from earlier today.

The field remains disabled in the python code because we have a nice declarative way to do that, and that does not allow for conditions.  However, in JS we can just re-enable it for this case.

### DIFF
--- a/static/js-src/admin/feature_form.js
+++ b/static/js-src/admin/feature_form.js
@@ -1,22 +1,28 @@
 (function() {
-'use strict';
+  'use strict';
 
-const fields = document.querySelectorAll('input, textarea');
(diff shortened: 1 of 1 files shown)


jpmedley (approved):
I'm curious, though. Why did you indent all this code?

jpmedley:
Getting Travis errors, so I can't merge.

jrobbins:
> I'm curious, though. Why did you indent all this code?

It was not properly indented in the enclosing function to start with.  Since I touched it, my editor wanted to automatically indent the lines that I touched, and they did not line up with the old code.  So, I corrected the indentation for all of it.

Issue #1059, opened 2020-10-06
KR: We have upgraded to python 3 and replaced GAE services (auth, datastore, mail, etc.)


carolrivero:

jrobbins:
This got done 100% earlier in 2022.

PR #1313, opened 2021-05-07
Add separate prod client_id and registered release candidate URLs.

jrobbins:
This is a step toward resolving issue #1059 by allowing us to deploy and use Google Sign-In on our staging and prod servers.

In this CL:
+ Use a separate client_id just for the prod servers, so that there is no temptation to add more JS origins to that oauth client.
+ When we deploy a new version, also push to the `rc` version.  We need a constant version name so that we have a constant URL for testing because oauth clients can only have a list of specific JS origins with no wildcards.
+ Update README to indicate that testing should be done on rc-dot-cr-status-staging.appspot.com.


cwilso (approved):
LGTM

PR #1314, opened 2021-05-07
Remove banner for Google Sign-In.

jrobbins:
When we deploy on Tuesday, the new sign-in method will be live, so the banner will not longer be needed.


jstenback (commented):
LGTM FWIW

PR #1317, opened 2021-05-10
Prompt the user to sign in for starring.

jrobbins:
In this CL:
+ Register event listeners for the star icons in the case where the user is signed out
+ Implement those event handlers to initiate a sign-in flow
+ Fix the line that was supposed to reload the page after the user successful signs in to make it always work


jstenback (commented):
LGTM FWIW.

PR #1322, opened 2021-05-17
Consolidate some ajax calls into cs-client.js.

jrobbins:
Javascript code that made XHR calls to our GAE app used to have custom logic for each different kind of call.   Now we have all that overhead logic sorted out in cs-client.js, so the previous files were reduced to just a few lines.  I think it would be better to have the entire set of API entry points laid out in one file so that it is easier to find them and easier to keep them consistent going forward.

Also in this PR:
+ Add --enable_console=1 when running locally so that a brand-new developer has a way to create a local admin AppUser entry.


jstenback (approved):
LGTM

jstenback (approved):

PR #1323, opened 2021-05-18
Add CSP headers, nonces, and report handler

jrobbins:
This adds a CSP policy based on the one referenced from go/csp.
Of the two policies defined there, we use the stricter nonce-only version.

For now, we will only report violations to the server and log them.  After reviewing logs in production, we can enable enforcement.


cwilso (approved):
LGTM.

PR #1324, opened 2021-05-18
Fix signed-out star icons on the schedule page.

jrobbins:
This initiates the sign-in flow if the user clicks a star icon when then are signed out.

I had made this work with Google Sign-In on the feature list and feature detail page and apparently I failed to do the same for the star icons on the schedule page.

Also in this PR:  Remove a few obsolete TODOs.


cwilso (approved):
LGTM

Issue #1113, opened 2020-12-14
Proposal for Replacing Left-Hand Navigation


jpmedley:
**Is your feature request related to a problem? Please describe.**
Chrome features have more Chromium statuses than are directly accessible. Meanwhile, the features of all versions of Chrome are listed, regardless of the likelihood of anyone needing to consult them, and the feature state is now shown at all.

![image](https://user-images.githubusercontent.com/8204171/102119325-c79a6b80-3df5-11eb-8227-701d189db6bf.png)

**Describe the solution you'd like**
And expandable custom search that allows returning the intersection of the following criteria:

- Status in Chromium
- Feature state
- Release number
- Feature owner
- *Whatever other criteria we see fit to add*

(The logic of this would seem to be pretty simple: if a criteria is chosen, concatenate it to the query string.)

**Additional context**
Another possible solution is to add drop downs for version, Chrome status, and feature state.

jpmedley:
Sorry for the late reply. We currently have two sets of state values: process stage and implementation status (status in Chromium).   There's currently design work occurring around these. Jason and Panos may feel differently, but I'm not sure this is the best time to jump on this issue.

PR #1325, opened 2021-05-19
Replaced Left Hand Navigation with a dropdown menu

### LINKED ISSUE #1240: Milestone page list without scrollable sidebar

shivamag00:
This PR should close #1240. 
Review Requested: @jrobbins 

We will have to use JS to add the strings `canary` `dev` `beta` `stable` and others instead of CSS as `content` property will not work with option tag. This is still left to be done. 

In this PR, 

- the left-hand navigation is replaced by a drop-down menu
- all chromium statuses which were not directly accessible by the left hand navigation, have also been added to the drop-down list. This was requested in #1113 

The following image illustrates how left hand navigation now looks:-

The first option is disabled because it is used as a placeholder.


shivamag00:
I am using JS (ternary operator) to display the strings `'canary'` `'dev'` `'beta'` `'stable'` after version numbers based on the `class` of the '<select>' tag and value of the loop counter. It now looks like the following:-



I will now add a option `All` in the dropdown which will replace the `Select Chrome Version` option. This `All` option will display the all the versions in same order as it is displayed when the page loads up for the first time. I will also try to link the filter input field with the dropdown such that clicking on `X` should select `All` from the dropdown.

jrobbins:
This is promising.  I would like to eventually replace those side links with a set of filtering widgets for a few different fields of the Feature entries.  This seems like a good step in that direction.  

The label "Chrome versions" is not quite right because there is a mix of implementation status values and version numbers.  The existing UI also mixed those concepts, but putting it in a menu seems to make the problem more noticeable.  Tomorrow, I'll test out this change and try to come up with a better label.

shivamag00:
UPDATE: Clicking on `X` in the Filter Search Box will now change the selected dropdown option in `Chrome Versions` to `Select Chrome Version`. 

Also, I now think that adding an `All` option to the list will not make much sense as the filter can be cleared using the `X` option. Also, when we separate out the chrome versions and statuses (eventually), the field `All` will not make sense.

shivamag00:
SUGGESTION: I have changed the label to `Filter By` and the placeholder for Dropdown to `Select Chrome Version/Status`. This is how it looks now:-

Screenshot 2021-05-21 202520

jrobbins (approved):
Well done.  Thanks very much.

jrobbins:
As a follow-up, you might get this filter menu to show on mobile (with narrow viewports).  The old links simply disappeared.  With the filter widgets approach, the widgets could be shown above or below the search box on small screens.  

Don't worry about making similar changes to the /samples page because I am planning on removing that page in the future.

shivamag00:
@jrobbins, #1334 contains the changes to display the filter menu on narrow view ports.

PR #1326, opened 2021-05-19
Convert old UI-style user admin requests to JSON API.

jrobbins:
Simplify the part of our code that adds and deletes registered users.  The old implementation used javascript to construct and submit an HTML form that was then processed by one of our UI-style request handlers.   This version goes through our JSON API.  

The reason for doing this is to reduce the amount of random JS logic, especially duplication of logic to deal with XSRF tokens for form submission.  And, to set the example of how JS components should focus on interacting with the user while csClient manages the protocol of how to make requests to our python code.

### DIFF
--- a/api/accounts_api.py
+++ b/api/accounts_api.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc.
+#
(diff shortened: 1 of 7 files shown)


cwilso (approved):
LGTM

PR #1332, opened 2021-05-21
Simplify the on-page JS for signIn and signOut

jrobbins:
This uses logic from cs-client.js rather than creating an XHR object directly in the signIn and signOut functions defined in _base.html.

Also, change the sign out link's onClick="" handler to use addEventListener.  This is the preferred style and is needed to avoid a CSP warning because handlers specified in attributes cannot have the required nonce value.

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -65,45 +65,18 @@ limitations under the License.
 
   <script nonce="{{nonce}}">
     function onSignIn(googleUser) {
-      var profile = googleUser.getBasicProfile();
-      var id_token = googleUser.getAuthResponse().id_token;
-        var xhr = new XMLHttpRequest();
-      xhr.open('POST', '/api/v0/login');
-      xhr.setRequestHeader('Content-Type', 'application/json');
-      xhr.onload = function() {
-        if (xhr.status == 200) {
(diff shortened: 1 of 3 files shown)


templates/_base.html
```diff
@@ -65,45 +65,18 @@
...
-        else {
-          // signout if cookie not set
-          signOut();
-        }
-      };
-      let data = JSON.stringify({ "id_token": id_token})
-      xhr.send(data);
-
+        csClient.signIn(googleUser).then(response => {
+            console.log(response);
```
shivamag00, line 69: In case of error (i.e. cookie is not set due to some reason in the backend), `signOut()` must be called so that user is logged out using the Google Sign-In library. Otherwise, in case of the error, Google Sign-In would consider user has logged in but our backend would consider the user has not logged in.

shivamag00 (changes requested):
Requested one change

templates/_base.html
jrobbins: Good point.  That will help avoid user confusion in the case where something goes wrong.   Added.

shivamag00 (approved):

Issue #1180, opened 2021-02-23
Implement API handlers for review comments


jrobbins:

jrobbins:
This was implemented back in 2021 and refined since then.

PR #1333, opened 2021-05-22
Implement Comment API and storage.

jrobbins:
This should resolve issue #1180.

In this PR:
+ Add Comment class to models.py.
+ Implement api/comments_api.py to handle getting all review comments for an approval, and posting a new comment that may have comment text and/or an approval value.  Also register the new API class and implement unit tests.
+ Implement some JS methods in csClient to call the python API.
+ Fix get_int_param() in basehandlers.py to allow an int parameter to return None when optional, and update the unit tests.

### DIFF
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc.
+#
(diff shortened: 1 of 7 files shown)


jstenback (approved):
LGTM

PR #1335, opened 2021-05-24
Fix and simplify project installation.

jrobbins:
I tried a fresh install today and hit some problems due to libraries that no longer support py2.7 in their default version.  So, I am specifying specific versions for more things in requirements.txt.

Also, there was a git sub-module used to check out code from a different project.  This is not currently used.  I believe that it was needed for push notifications.  But, we now use email instead of push notifications.


KyleJu (approved):
Thanks Jason!

PR #1337, opened 2021-05-27
Update fetching of API OWNERS from new URL.

jrobbins:
I previously had an old URL for fetching the API_OWNERS file.  
Also, the gitiles server now seems to serve format=TEXT as a base64 encode file, so we need to decode it.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -70,7 +70,7 @@ def fetch_owners(url):
     logging.error('Got response %r', response)
     raise ValueError('Could not get OWNERS file')
 
-  decoded = base64.b64decode(response.content)
+  decoded = base64.b64decode(response.content).decode()
   for line in decoded.split('\n'):
     logging.info('got line: '  + line)
     if '#' in line:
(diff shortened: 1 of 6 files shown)


internals/approval_defs.py
```diff
@@ -70,7 +70,9 @@ def fetch_owners(url):
     logging.error('Got response %r', response)
     raise ValueError('Could not get OWNERS file')
 
-  for line in response.iter_lines():
+  decoded = base64.b64decode(response.content)
+  for line in decoded.split('\n'):
```
KyleJu, line 74: Does it have to be python3 compatible? in Python 3 `base64.b64decode` return bytesstring and it splits by `(b'\n')`, IIUC.

jrobbins: Good catch.  I added a .decode() to convert to unicode.

I also fixed another case of the same py3 incompatibility in the fetchmetrics.py file.  And, in adding unit tests for that I realized that some existing unit tests needed more mocks to avoid actually hitting the OWNERS file (which make those tests flakey).

KyleJu (approved):
LGTM

cwilso (approved):

PR #1341, opened 2021-05-31
20210423 xsrf token refresh

jimbojr55:
"GIT"


.travis.yml
```diff
@@ -27,15 +27,10 @@ before_install:
 install:
```
jimbojr55, line 27: "GET"

jimbojr55: GET:

jimbojr55 (commented):
GET/.GET\.

.travis.yml
jimbojr55: @jimbojr55

jrobbins:
It looks like you are just experimenting with how GitHub pull requests work.

PR #1349, opened 2021-06-02
Enforce xsrf protections

jrobbins:
For several weeks now, we've been sending out XSRF tokens and checking them without raising errors.  A review of the logs shows on valid requests that logged missing or invalid XSRF tokens.  So, I think we are ready to start enforcing them as requirements on requests.

In this PR:
+ Reject requests that have missing or bad tokens rathe than just logging them
+ Update unit tests
+ Update the admin page for editing blink component owners.  It was not sending tokens in XHR requests.


jstenback (approved):
LGTM

Issue #1076, opened 2020-10-19
Convert application code to use new storage backend


jrobbins:

shivamag00:
@jrobbins, please close this issue.

Issue #1075, opened 2020-10-19
Research which data storage backend to use


jrobbins:

shivamag00:
Because of my limited experience, I am unaware with the term `data storage backend`. Hence, I assume that it means database. If this assumption is wrong, please let me know.

Just for the record, I am aware that as a pre-requisite to Python 3 upgrade, we need to replace our use of DB Client Library for accessing Cloud DataStore as mentioned in this Google Cloud [guide](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-to-cloud-ndb)
 
May we work on this issue, @jrobbins ?

shivamag00:
This [link](https://cloud.google.com/datastore/docs/upgrade-to-firestore) says that all `Datastore` instances will be gradually updated to `Firestore in Datastore` Mode starting this year. 

Also, this [link](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-to-cloud-ndb#key_differences) suggests that there are only minor syntax changes between App Engine NDB (we are using DB Client Library.  App Engine NDB was an improvement over DB Client Library) and Cloud NDB (for accessing Firestore in Datastore Mode). The link also lists the major differences between the two.

shivamag00:
@jrobbins, please let me know what other options for storage backend we can consider.

jrobbins:
I think that this is the most practical upgrade path and that your PRs are on the right track.  Let's consider the research part of this work finished.

PR #1351, opened 2021-06-04
Migration to Cloud NDB

shivamag00:
Merging this will close #1075 and #1076 

This PR is not supposed to be merged at this stage !!!

The PR will replace DB Client Library with Cloud NDB library which is required for migration to Python 3.

Some Important points to note:-
- Cloud NDB seems to be working fine in the interactive console of Cloud Datastore Emulator and is able to return the existing data created using DB Client Library.
- The order of release of the storage client libraries is - `DB Client Library` (oldest), `App Engine NDB Library` and `Cloud NDB library` (latest). We are migrating, directly from `DB Client Library` to `Cloud NDB Library`. 
- There are many changes that were introduced in `App Engine NDB Library`. E.g. Many `property types`, that were present in `DB Client Library` are not present in `App Engine NDB Library`. As `Cloud NDB library` is successor of `App Engine NDB library`, these changes are also present in `Cloud NDB Library`. 
- Deploying and running this PR on local machine, at this stage, will result in `BadArgumentError: Could not import googledatastore. This library must be installed with version >= 6.0.0 to use the Cloud Datastore API.` It may seem counter-intuitive but this error is raised because of the syntax changes introduced in `App Engine NDB Library` ( and as a result, `Cloud NDB library` also). Over the next few days, I will try to remove these syntax errors.

Note:

- This [link](https://cloud.google.com/appengine/docs/standard/python/ndb/db_to_ndb) contains the differences between `DB Client Library` and `App Engine NDB Library`

PR #1352, opened 2021-06-08
GAE DB Client to GAE NDB

shivamag00:
READY FOR MERGE !!!

This PR is a step in solving  #1075 and #1076
Merging this PR will replace DB Client Library with App Engine NDB.


internals/models.py
```diff
@@ -319,20 +320,34 @@ def del_none(d):
...
+    obj = future.get_result()
+    if obj is not None:
+      # test needed because post_get_hook is called even if get() fails!
+      obj.saved = True
+
+  def _post_put_hook(self, future):
+    self.saved = True
+
+  def is_saved():
+    return self.saved
```
shivamag00, line 339: `is_saved()` has been deprecated. These methods may be used to replace as mentioned in [this](https://cloud.google.com/appengine/docs/standard/python/ndb/db_to_ndb) article. 

> model_instance.is_saved() | No direct equivalent; see&nbsp;[Stack Overflow](https://stackoverflow.com/questions/12083254/is-it-possible-to-determine-with-ndb-if-model-is-persistent-in-the-datastore-or/12096066#12096066)&nbsp;for a possible solution.

internals/models.py
```diff
@@ -732,12 +747,12 @@ def get_all(self, limit=None, order='-updated', filterby=None,
...
     if feature_list is None or update_cache:
-      query = Feature.all().order(order) #.order('name')
-      query.filter('deleted =', False)
+      query = Feature.query().order(Feature.order) #.order('name')
+      query.filter(Feature.deleted == False)
 
       # TODO(ericbidelman): Support more than one filter.
       if filterby:
-        query.filter(filterby[0], filterby[1])
+        query.filter(Feature.filterby[0] == filterby[1])
```
shivamag00, line 755: The method signature of `query.filter()` has changed. Just to verify, I hope I have migrated this line correctly.

framework/ramcache.py
```diff
@@ -176,7 +176,7 @@ def invalidate(cls):
   @classmethod
   def check_for_distributed_invalidation(cls):
     """Check if any appengine instance has invlidated the cache."""
-    singleton = cls.get(cls.SINGLETON_KEY, read_policy=db.STRONG_CONSISTENCY)
+    singleton = cls.SINGLETON_KEY.get()
```
shivamag00, line 179: `read_policy` has been deprecated is, now, a synonymn of `read_consistency`. Although the [document](https://googleapis.dev/python/python-ndb/latest/query.html?highlight=strong#google.cloud.ndb.query.Query.__iter__) suggests that we may use `read_consistency=ndb.STRONG` for strong consistency, I get the following error indicating that this library does not have any such constant.

> Traceback (most recent call last):
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/flask/views.py", line 163, in dispatch_request
    return meth(*args, **kwargs)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/framework/basehandlers.py", line 325, in get
    ramcache.check_for_distributed_invalidation()
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/framework/ramcache.py", line 191, in check_for_distributed_invalidation
    SharedInvalidate.check_for_distributed_invalidation()
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/framework/ramcache.py", line 179, in check_for_distributed_invalidation
    singleton = cls.SINGLETON_KEY.get(read_consistence=ndb.STRONG)
AttributeError: 'module' object has no attribute 'STRONG'

shivamag00 (commented):
@jrobbins, please look at these specific changes.

internals/models.py
jrobbins: OK that looks fine.

internals/models.py
jrobbins: I don't think that that can work because there is no 'filterby' field defined, so Feature.filterby will be an error.  The part before the operator needs to be a field defined in the entity class.

There is only one caller that uses the filterby= keyword parameter.  That is in featurelist.py and it only ever passes in ('category =', k).   So, you could take out the filterby= keyword and instead implement a category= keyword that does query.filter(Feature.category == category).

framework/ramcache.py
jrobbins: I think that a strongly consistent read can be achieved by using an ancestor query instead of get().  It would look something like:
singleton = cls.query(ancestor=PARENT_KEY).fetch(1)[0]
I have not tried that in practice, so we will need to do some manual testing.

internals/models.py
shivamag00: Thanks. I have committed this change.

framework/ramcache.py
shivamag00: okay. For the record, I see that you have implemented this in #1356

shivamag00 (commented):
@jrobbins, please look at the following

internals/models.py
```diff
@@ -938,8 +953,8 @@ def __notify_feature_subscribers_of_changes(self, is_update):
        posting to a task queue."""
     # Diff values to see what properties have changed.
     changed_props = []
-    for prop_name, prop in self.properties().iteritems():
-      if prop_name in ('created_by', 'updated_by'):
+    for prop_name, prop in self._properties.iteritems():
+      if prop_name in ('created_by', 'updated_by', 'updated', 'created'):
         continue
       new_val = getattr(self, prop_name, None)
       old_val = getattr(self, '_old_' + prop_name, None)
```
shivamag00, line 960: The properties `updated` and `created` are essentially of type `datetime.datetime`. Hence, if ever (although impossible while using old DB Client Library) there is change between `_old_updated` and `updated`, the following error will be raised if changed in line 957 (above) are ignored.

>  TypeError: datetime.datetime(2021, 6, 9, 10, 46, 8, 833823) is not JSON serializable

Although the above case is impossible to happen if use DB Client Library, the same is not true for the not-so-old NDB Client Library. This is because the arguments `auto_add` and `auto_add_now` only assign the values once the entity is written to datastore. As this [link](https://cloud.google.com/appengine/docs/standard/python/ndb/entity-property-reference#Date_and_Time) says, 

>  The automatic value is not generated until the entity is written; that is, these options don't provide dynamic defaults. (These details differ from the old db API.)

Hence, in case of the not-so-old NDB Client Library, `_old_updated` and `updated` have the same value `None` initially but once the entity is written, the value of updated changes to the required value.

framework/ramcache.py
jrobbins: Yep.  That has merged.   I will try deploying a version of the app with that change later today to tomorrow.  If any problems show up due to that specific change, we will know it independently of the larger change.

pages/guide.py
```diff
@@ -30,7 +30,7 @@
 # TODO(jrobbins): phase out gae_users
 from google.appengine.api import users as gae_users
 from framework import users
-from google.appengine.ext import db
+from google.appengine.ext import ndb
```
jrobbins, line 33: I think this line could be deleted.  It does not seem to be referenced in the file.

internals/models.py
jrobbins: OK. I think I understand.  Thanks for working through that.

pages/guide.py
shivamag00: Okay... I have removed it. Thank you

internals/models.py
```diff
@@ -323,20 +323,11 @@ def del_none(d):
...
-      obj.saved = True
-
-  def _post_put_hook(self, future):
-    self.saved = True
 
   def is_saved(self):
-    return self.saved
+    if self.key:
+      return True 
+    return False
```
shivamag00: The hook `_post_get_hook` only gets called when using `Key.get()` to fetch the entries. Since, we are using `query()` at many places (e.g. to fetch the list of all features), this hook will not be helpful to us. Hence, I have come up with a work-around. 

In old DB Client Library `is_saved()` returns `True` if the model instance has been written (put) to the Datastore at least once. Also the following is mentioned [here](https://cloud.google.com/appengine/docs/standard/python/datastore/modelclass), 
> This method only checks that the instance has been written to the Datastore at least once since it was created. It does not check whether the instance's properties have been updated since the last time it was written.

Hence, we need someway to figure out if the data has been written to the datastore. Since we do not specify the key while creating an entity (except in the case of `SharedInvalidate`), the key is `None` until the entity is written (using `put()` ). Hence, checking if `key` is not `None` could mean that the entity has been written to the datastore.

shivamag00 (commented):
@jrobbins, I have change the implementation of the deprecated `is_saved()` method. I hope it should now work for us.

internals/models.py
jrobbins: Yes, that sounds like it would work.

internals/models.py
```diff
@@ -924,12 +930,11 @@ def new_crbug_url(self):
...
-  def __init__(self, *args, **kwargs):
-    super(Feature, self).__init__(*args, **kwargs)
+  def _pre_put_hook(self):
 
     # Stash existing values when entity is created so we can diff property
     # values later in put() to know what's changed. https://stackoverflow.com/a/41344898
-    for prop_name, prop in self.properties().iteritems():
+    for prop_name, prop in self._properties.iteritems():
       old_val = getattr(self, prop_name, None)
       setattr(self, '_old_' + prop_name, old_val)
```
shivamag00, line 940: In `__init__`, we create properties with suffix `_old_`, so that we could know if they have changed when the user submitted a feature edit form from the frontend.

This won't work in the not-so-old ndb library as unlike the old DB Client Library, `__init__` creates instances with properties set to `None`. Hence, we could set up the old properties (`_old_prop_name`) in `_pre_put_hook()`. It seems to work properly.

testing_config.py
```diff
@@ -67,6 +67,7 @@ def setUpOurTestbed():
   ourTestbed.init_urlfetch_stub()
   ourTestbed.init_user_stub()
   ourTestbed.init_xmpp_stub()
+  ourTestbed.init_memcache_stub()
```
shivamag00, line 70: The not-so-old NDB client library uses memcache by default for caching. Hence, we need it while running tests. However, the Cloud NDB library would not use memcache.

shivamag00 (commented):
@jrobbins, please review the following changes. 

FIY, I have migrated almost all of the tests. However, some tests still show error. I will try to resolve those errors over the weekend.

framework/ramcache.py
jrobbins: FYI, my change was deployed yesterday.  There are no obvious problems with it so far.

jrobbins: So, you just need to merge main into your branch.

testing_config.py
jrobbins: OK that makes sense.  I look forward to being able to remove this line again.

internals/models.py
jrobbins: Hmm, do you see diffs being generated when notification emails are logged to the console after you edit a feature?  You will need to start the feature with one account and then edit it with a different account.

I might not be understanding when this hook is run, but I would think that the properties of the Feature object would have already been updated before put() is called, and thus before _pre_put_hook() is called.  That would cause the differences to always appear to be no difference.

I am thinking that _post_get_hook() would be the right hook.  But, that thought is just based on the names of the hook.  Only testing it will let us know for sure.

internals/models.py
shivamag00: @jrobbins, can we mark this as resolved ?

internals/models.py
```diff
@@ -732,12 +745,12 @@ def get_all(self, limit=None, order='-updated', filterby=None,
     feature_list = ramcache.get(KEY)
 
     if feature_list is None or update_cache:
-      query = Feature.all().order(order) #.order('name')
-      query.filter('deleted =', False)
+      query = Feature.query().order(-Feature.updated) #.order('name')
```
shivamag00, line 748: I migrated line 748 incorrectly earlier. `order` is not a property of Feature. Instead, it is a parameter of the function which contain these lines. On investigating, I found that `order` will always have the value `'-updated'`. Hence, I have changed it to `-Feature.updated`

internals/models.py
```diff
@@ -910,7 +923,10 @@ def crbug_number(self):
 
   def new_crbug_url(self):
     url = 'https://bugs.chromium.org/p/chromium/issues/entry'
-    params = ['components=' + self.blink_components[0] or BlinkComponent.DEFAULT_COMPONENT]
+    if len(self.blink_components) > 0:
+      params = ['components=' + self.blink_components[0]]
+    else:
+      params = ['components=' + BlinkComponent.DEFAULT_COMPONENT]
```
shivamag00, line 929: If `self.blink_components` is an empty list, the line show `Index Out of Range` Error. Hence, I have changed it to this syntax. 

UPDATE: We may not need to check for the empty list case as I have figured out a way to assign default value to `Feature.blink_components` using the constructor. See  [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1352/commits/d67262c4b566b7cce2f6235ddb069ccb005fd57c) commit.

internals/models.py
```diff
@@ -474,6 +480,13 @@ class Feature(DictModel):
...
   DEFAULT_CACHE_KEY = 'features'
 
+  def __init__(self, *args, **kwargs):
+    # Initialise Feature.blink_components with a default value
+    if 'blink_components' not in kwargs:
+      kwargs['blink_components'] = [BlinkComponent.DEFAULT_COMPONENT]
+
+    super(Feature, self).__init__(*args, **kwargs)
+
   @classmethod
```
shivamag00, line 490: As mentioned [here](https://github.com/GoogleChrome/chromium-dashboard/pull/1351#discussion_r645485566), we could not pass the named argument `default=` while declaring the property `blink_components`. This is a workaround for that. It seems to work properly.

shivamag00 (commented):
@jrobbins, I have migrated all the tests. Please review the following changes.

shivamag00 (commented):
@jrobbins, these changes are regarding the deprecated parameter `read_policy`. Please review

framework/ramcache.py
```diff
@@ -176,7 +176,10 @@ def invalidate(cls):
   @classmethod
   def check_for_distributed_invalidation(cls):
     """Check if any appengine instance has invlidated the cache."""
-    singleton = cls.get(cls.SINGLETON_KEY, read_policy=db.STRONG_CONSISTENCY)
+    singleton = None
+    entities = cls.query(ancestor=cls.PARENT_KEY).fetch(1)
+    if entities:
+      singleton = entities[0]
     if not singleton:
```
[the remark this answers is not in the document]
shivamag00: Also, in PR #1356, the change was made for replacing `singleton = cls.get(cls.SINGLETON_KEY)` in `invalidate` function which does not pass the argument `read_policy`. I missed this while reviewing the PR. Should not the change have been made to replce the line 179 in the above snippet ?

framework/ramcache.py
shivamag00: Update:- See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1352#discussion_r650489594) link

internals/models.py
shivamag00: I can see the following line showing ALL the properties of the  feature
> Starting to notify subscribers for feature {..._properties and values_..}

Is it the same diff that you were referring to ? 

I also tried to log to console the properties with and without the `_old_` suffix. It turns out that they have the same values. Hence, `_pre_put_hook()` will not work. Thank you for pointing this out. I did not notice it as there was one more error that was being caused by initializing properties in `__init()__` and the line `Starting to notify subscribers for feature` was same as the line generated by old DB Client Library. 

`_post_get_hook()` will not work as it is a class method unlike `_pre_put_hook()` which is an instance method. Hence, I have used the method `_from_pb()` which is responsible for assigning the values to the instance. See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1352/commits/3bfa050a8b582dfa79fa6c1a95e2263fa8aa8366) commit

shivamag00:
@jrobbins, FIY, I have finished the migration process. Once all the conversations are resolved, I will resolve all the merge conflicts and the PR will be ready to merge (I will update the first line of PR description once it is ready for merging). 

Thank You very much for helping with this PR !!

framework/ramcache.py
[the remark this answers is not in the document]
jrobbins: You are right.  I should have been done in both places.  Thanks for catching that.

framework/ramcache_test.py
```diff
@@ -110,9 +110,12 @@ def testFlushAll(self, mock_invalidate):
...
   def assertTimestampWasUpdated(self, start_time):
-    singleton = ramcache.SharedInvalidate.get(
-        ramcache.SharedInvalidate.SINGLETON_KEY,
-        read_policy=db.STRONG_CONSISTENCY)
+
+    singleton = None
+    entities = ramcache.SharedInvalidate.query(ancestor=ramcache.SharedInvalidate.PARENT_KEY).fetch(1)
+    if entities:
+      singleton = entities[0]
+
```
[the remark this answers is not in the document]
jrobbins: Yes, here too.

framework/ramcache.py
jrobbins: The linked conversation was resolved.

internals/models.py
jrobbins: Yes.

internals/models.py
jrobbins: OK, that looks like it will work.

internals/models.py
```diff
@@ -474,6 +480,15 @@ class Feature(DictModel):
...
 
+  def __init__(self, *args, **kwargs):
+    # Initialise Feature.blink_components with a default value
+    if 'name' in kwargs:        # If name is present in kwargs then it would mean constructor is being called for creating a 
+                                # new feature rather than for fetching an existing feature. 
+      if 'blink_components' not in kwargs:
+        kwargs['blink_components'] = [BlinkComponent.DEFAULT_COMPONENT]
+
+    super(Feature, self).__init__(*args, **kwargs)
+
```
shivamag00, line 491: This change is for assigning the default value to the property `blink_components`

shivamag00 (commented):
@jrobbins, please review this change.

shivamag00:
@jrobbins, FIY, I have resolved all the merge conflicts. If all the conversations can be marked as resolved, this PR is ready to be merged.

jrobbins:
I tested this locally and it does seem like almost everything is working.

One area that I tested was the generation of the diff emails.  I see that if I have a feature that was created under DB, the first time that it is edited using NDB, there are some extra diffs being generated that are not meaningful to the user (example below).   Once the feature has been edited using NDB, subsequent edits using NDB seem to produce the right diffs.

Example of extra diffs.  I had only edited the interop_compat_risks field:
```
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>interop_compat_risks:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> asoiudas oiuas daosiud<br/>
</li><br/>

<li><b>all_platforms:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> False<br/>
</li><br/>

<li><b>web_dev_views_notes:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>all_platforms_descr:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>security_risks:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>flag_name:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>debuggability:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>ergonomics_risks:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>activation_risks:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>ie_views_notes:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>

<li><b>safari_views_notes:</b> <br/>
<b>old:</b> None <br/>
<b>new:</b> <br/>
</li><br/>
</ul>
```

So, I am thinking that either the _from_pb() method needs to call getattr() using a default value that is appropriate for the field type rather than using None.  Or, we could change __notify_feature_subscribers_of_changes() to compare old_val and new_val in a way that skips over a None vs. empty-string case.

That was the only problem that I have found so far.  I'll test on our staging server tomorrow.

jrobbins:
In testing on our staging server with existing data, I see that this problem does occur there.   It does not happen with as many fields as it does when running locally, but I did see it happen with the comment field.  The comment field reported a diff from "None" to an empty string when I made a change to the spec_link field only.  So, please try the fix I suggested in my previous comment.

I also hit another problem, which should also be fairly easy to fix.    Our staging (and production) data has some old entities with old fields that are no longer part of the app, for example the intent_template_use_counter field.  NDB generally ignores these obsolete fields, which is good.  However when we loop over self._properties in DictModel, it picks up these old properties and causes an exception.  I'll suggest the fix on that line in my next comment.

internals/models.py
```diff
@@ -319,20 +320,25 @@ def del_none(d):
...
     if add_id:
-      d['id'] = self.key().id()
+      d['id'] = self.key.integer_id()
     return d
 
   def to_dict(self):
     output = {}
 
-    for key, prop in self.properties().iteritems():
+    for key, prop in self._properties.iteritems():
```
jrobbins, line 341: This includes obsolete fields from old entities in our staging datastore.  Trying to do getattr() with an obsolete field causes an exception.  As a fix, please add
```
    # Skip obsolete values that are still in our datastore
    if not hasattr(self, key):
      continue
```

internals/models.py
jrobbins: That seems good.

internals/models.py
jrobbins: It's probably best to keep this here.  While your linked commit should prevent the error, it's still safer to use defensive coding here.

What you have here is fine, or you could try using `and` and `or` for a more compact expression:
params = ['components=' + (self.blink_components and blink_components[0] or  BlinkComponent.DEFAULT_COMPONENT)]

internals/models.py
jrobbins: This looks good.

internals/models.py
```diff
@@ -970,6 +970,8 @@ def __notify_feature_subscribers_of_changes(self, is_update):
       new_val = getattr(self, prop_name, None)
       old_val = getattr(self, '_old_' + prop_name, None)
       if new_val != old_val:
+        if new_val == '' and old_val is None:
+          continue
         new_val = convert_enum_int_to_string(prop_name, new_val)
         old_val = convert_enum_int_to_string(prop_name, old_val)
```
shivamag00, line 976: @jrobbins,as suggested by you, I have modified the function  `__notify_feature_subscribers_of_changes()`. I tested it also. It seems to fix the problem of None vs Empty String in diffs.

shivamag00:
@jrobbins, FIY, I have fixed both the issues mentioned[ here](https://github.com/GoogleChrome/chromium-dashboard/pull/1352#issuecomment-861937609) and[ here](https://github.com/GoogleChrome/chromium-dashboard/pull/1352#issuecomment-862818725).

internals/models.py
shivamag00: However, there are some fields in the diff with the following old and new values:
`old_val` : None
`new_val`: False

Are such fields supposed to be NOT present in diffs ?

internals/models.py
```diff
@@ -970,6 +970,8 @@ def __notify_feature_subscribers_of_changes(self, is_update):
       new_val = getattr(self, prop_name, None)
       old_val = getattr(self, '_old_' + prop_name, None)
       if new_val != old_val:
+        if new_val == '' and old_val is None:
+          continue
         new_val = convert_enum_int_to_string(prop_name, new_val)
```
shivamag00, line 975: However, there are some fields in the diff with the following old and new values:

```
<li>
     <b>all_platforms:</b> <br/>
     <b>old:</b> None <br/>
     <b>new:</b> False<br/>
</li><br/>
<li>
     <b>wpt:</b> <br/>
     <b>old:</b> None <br/>
     <b>new:</b> False<br/>
</li><br/>
<li>
     <b>prefixed:</b> <br/>
     <b>old:</b> None <br/>
     <b>new:</b> False<br/>
</li>
```
Do we have to skip over None vs False case also?

shivamag00 (commented):
@jrobbins, please look at this

internals/models.py
jrobbins: Yes I think that would be best.  I did not see that happening on our staging server, but I don't see any downside to handling that case because the diffs that you pasted are not helpful to users.

jrobbins: Since this is the last detail, I'll go ahead and merge this PR and send you a new tiny PR to make this last change.  That should allow you to resolve conflicts and dig into the Cloud NDB upgrade.

jrobbins (approved):

internals/models.py
shivamag00: Done. See PR #1364

PR #1355, opened 2021-06-09
Reduce usage of GAE users.

jrobbins:
As part of the Google Sign-in roll out, we still accepted GAE user cookies so that users who had an existing browser tab open would be able to submit a form that they had typed into.

Now that it has been a few weeks, we can remove support for GAE users entirely from basehandlers.py.

Also in this PR, another step toward phasing out GAE users: stop using the is_admin boolean in our test users.  Instead, consider a user to be an admin iff there is a corresponding AppUser datastore entity that has is_admin set.   And, more fully test permissions for registered users (who have an AppUser entity with with is_admin false) and for admin users (who have it set true).

### DIFF
--- a/api/accounts_api_test.py
+++ b/api/accounts_api_test.py
@@ -31,6 +31,9 @@ from internals import models
 class AccountsAPITest(unittest.TestCase):
 
   def setUp(self):
+    self.app_admin = models.AppUser(email='admin@example.com')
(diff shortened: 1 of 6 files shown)


shivamag00 (commented):
LGTM

shivamag00 (approved):

PR #1363, opened 2021-06-17
Remove a reference to GAE users.

jrobbins:
This follows the pattern that you established for dealing with a page view that requires the user to be signed in.   With the old GAE users library we could redirect them to a sign-in page, but with Google Sign-In we will redirect them to the /features page and ask them to sign in.


shivamag00 (approved):
LGTM

PR #1365, opened 2021-06-18
20210618 gae ndb to cloud ndb

shivamag00:
This PR is the second step in solving  #1075 and #1076. This PR is a follow up to #1352.

This PR will migrate our usage of  GAE NDB to Cloud NDB.


framework/ramcache.py
```diff
@@ -153,11 +156,13 @@ class SharedInvalidateParent(ndb.Model):
...
   PARENT_ENTITY_ID = 1234
-  PARENT_KEY = ndb.Key('SharedInvalidateParent', PARENT_ENTITY_ID)
   SINGLETON_ENTITY_ID = 5678
-  SINGLETON_KEY = ndb.Key(
+  with client.context():
+    PARENT_KEY = ndb.Key('SharedInvalidateParent', PARENT_ENTITY_ID)
+    SINGLETON_KEY = ndb.Key(
       'SharedInvalidateParent', PARENT_ENTITY_ID,
       'SharedInvalidate', SINGLETON_ENTITY_ID)
+
```
shivamag00, line 165: See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1351#discussion_r645886479) link

shivamag00 (commented):
@jrobbins, please review

shivamag00:
@jrobbins. Also, the method `_from_pb()` is no longer used by the Cloud NDB and the alternative method `entity_from_protobuf()` is an independent method in the `google.cloud.ndb.model` module instead of being a method of the `Model` class. Hence, we will now have to use something else for comparing old values with new values as we cannot set the variables for  old values. I will try to come up with a solution over the weekend.

internals/models.py
```diff
@@ -359,6 +360,8 @@ def to_dict(self):
         output[key] = to_dict(value)
       elif isinstance(value, gae_users.User):
         output[key] = value.email()
+      elif isinstance(value, ndb.model.User):
+        output[key] = value.email()
       else:
```
shivamag00, line 365: We need to add this case as unlike the not-so-old NDB Client Library, the Cloud NDB does not use GAE Users Library for creating `User` Objects

framework/ramcache.py
jrobbins: That makes sense.

internals/models.py
jrobbins: OK.

jrobbins:
> @jrobbins. Also, the method `_from_pb()` is no longer used by the Cloud NDB and the alternative method `entity_from_protobuf()` is an independent method in the `google.cloud.ndb.model` module instead of being a method of the `Model` class. Hence, we will now have to use something else for comparing old values with new values as we cannot set the variables for old values. I will try to come up with a solution over the weekend.

We might not need a way to hook into the normal entity loading mechanism.  Instead of having it work in implicitly in call cases, we could make an explicit call to save old data in just the locations in our code where we load a Feature entity for editing.  Right now that should be just in guide.py line 283.

internals/models.py
```diff
@@ -946,18 +949,14 @@ def new_crbug_url(self):
...
-    entity = super(Feature, cls)._from_pb(pb, set_key=set_key, ent=ent, key=key)
     
-    for prop_name, prop in entity._properties.iteritems():
-      old_val = getattr(entity, prop_name, None)
-      setattr(entity, '_old_' + prop_name, old_val)
-
-    return entity
+    for prop_name, prop in self._properties.iteritems():
+      old_val = getattr(self, prop_name, None)
+      setattr(self, '_old_' + prop_name, old_val)
```
shivamag00, line 960: We can use this method for stashing the old values. It seems to work fine. 

FIY, The diffs still show three extra fields that are not useful to the user. These fields belong to None vs False case. Hence, merging main branch in this PR, will solve this problem completely.

pages/guide.py
```diff
@@ -283,6 +283,8 @@ def process_post_data(self, feature_id, stage_id=0):
       feature = models.Feature.get_by_id(feature_id)
       if feature is None:
         self.abort(404, msg='Feature not found')
+      else:
+        feature.stash_values()
```
shivamag00, line 288: Calling the stash_values() here as mentioned [here](https://github.com/GoogleChrome/chromium-dashboard/pull/1365#issuecomment-864134466). Thank you for letting me know there was only one place where this method call was required.

shivamag00 (commented):
@jrobbins, this should solve the problem of stashing old values. Please review these changes.

shivamag00:
@jrobbins, as [this](https://stackoverflow.com/questions/60351198/how-do-you-unit-test-google-cloud-ndb-code) stackoverflow answer and [this](https://groups.google.com/g/google-appengine/c/yuAofPuxYtE/m/GdxzxzZnCQAJ) google group discussion mentions, apparently CLoud NDB does not support testbed. Only solutions are to use Cloud Datastore Emulator or mock the calls to datastore. Please let me know your views on this. 

FIY, the only task left to make this PR ready for merging is migrating the tests to use Cloud NDB and solve any problem that arises on running these tests.

shivamag00:
Also, while trying to run tests today, I was getting the following error

> ImportError: Failed to import test module: api.accounts_api_test
  File "/usr/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/api/accounts_api_test.py", line 26, in <module>
    from api import accounts_api
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/api/accounts_api.py", line 24, in <module>
    from framework import basehandlers
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/framework/basehandlers.py", line 51, in <module>
    django.setup()
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/django/__init__.py", line 22, in setup
    configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/django/conf/__init__.py", line 56, in __getattr__
    self._setup(name)
  File "/mnt/c/Users/Shivam/ubuntu/chromium-dashboard/lib/django/conf/__init__.py", line 39, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
ImproperlyConfigured: Requested setting LOGGING_CONFIG, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.

This error implies that the environment variable `DJANGO_SETTINGS_MODULE` is not set. In `testing_config.py`, we do not set this environment variable. I could find out that after the changes in the commit [45406a9](https://github.com/GoogleChrome/chromium-dashboard/pull/1365/commits/45406a904298a86e82950e737cd76ed18dc7748e)  of this branch, we begin to get this error. 

The only changes that are made is in `ramcache.py` in the commit. I am unable to figure out why we were not getting this error before. Also, this as the stack trace in the error suggests, this error is raised while executing `django.setup()` when `basehandlers.py` is being imported. 

@jrobbins, Do you know the reason for not getting this error before ?

internals/models.py
jrobbins: Sounds good.

jrobbins:
> @jrobbins, as [this](https://stackoverflow.com/questions/60351198/how-do-you-unit-test-google-cloud-ndb-code) stackoverflow answer and [this](https://groups.google.com/g/google-appengine/c/yuAofPuxYtE/m/GdxzxzZnCQAJ) google group discussion mentions, apparently CLoud NDB does not support testbed. Only solutions are to use Cloud Datastore Emulator or mock the calls to datastore. Please let me know your views on this.

Thanks for finding that info.  My suggestion would be to try the emulator first.   The slowdown would be unfortunate, but not terrible, so I think that might be better than needing to re-write a lot of unit tests.    

The email thread that you linked has a way to completely reset the datastore emulator.   I like to keep local test data between runs of the local server for manual testing, however it is probably better to have more reliable unit tests.  So, resetting the data stored in the emulator seems like the right thing to do.


> FIY, the only task left to make this PR ready for merging is migrating the tests to use Cloud NDB and solve any problem that arises on running these tests.

That's great progress so far.  I know that lots of times half the work comes in the final stages of doing something, so keep at it.

jrobbins:
> The only changes that are made is in `ramcache.py` in the commit. I am unable to figure out why we were not getting this error before. Also, this as the stack trace in the error suggests, this error is raised while executing `django.setup()` when `basehandlers.py` is being imported.
> 
> @jrobbins, Do you know the reason for not getting this error before ?

That's a good question.  I never thought about why it worked before.  It would make sense that `django.setup()` would need that environment variable set, and the only place that I see that sets it is env_vars.yaml which seems like it would only apply when running the app locally rather than running unit tests.

However, I see that in the google/appengine/ext/ndb directory the __init__.py file imports some other files, and some of them import google_imports.py, and that imports dev_appserver.   So, probably any import of the non-cloud ndb library indirectly caused `app.yaml` (and therefore `env_vars.yaml`) to be processed.   When you removed the last import of ndb, it probably removed that implicit step.  You could test that theory by adding back that import line.

It's probably for the best.  Unit tests should not need to run in an environment that is implicitly configured.  It would be better to set `DJANGO_SETTINGS_MODULE` explicitly in `testing_config.py`.

shivamag00:
> > The only changes that are made is in `ramcache.py` in the commit. I am unable to figure out why we were not getting this error before. Also, this as the stack trace in the error suggests, this error is raised while executing `django.setup()` when `basehandlers.py` is being imported.
> > @jrobbins, Do you know the reason for not getting this error before ?
> 
> That's a good question. I never thought about why it worked before. It would make sense that `django.setup()` would need that environment variable set, and the only place that I see that sets it is env_vars.yaml which seems like it would only apply when running the app locally rather than running unit tests.
> 
> However, I see that in the google/appengine/ext/ndb directory the **init**.py file imports some other files, and some of them import google_imports.py, and that imports dev_appserver. So, probably any import of the non-cloud ndb library indirectly caused `app.yaml` (and therefore `env_vars.yaml`) to be processed. When you removed the last import of ndb, it probably removed that implicit step. You could test that theory by adding back that import line.
> 
> It's probably for the best. Unit tests should not need to run in an environment that is implicitly configured. It would be better to set `DJANGO_SETTINGS_MODULE` explicitly in `testing_config.py`.

Thank You @jrobbins. This theory seems to be true. I tried running the tests again after importing the not-so-old GAE NDB Library. This error was no longer raised.

shivamag00:
> Thanks for finding that info. My suggestion would be to try the emulator first. The slowdown would be unfortunate, but not terrible, so I think that might be better than needing to re-write a lot of unit tests.
> 
> The email thread that you linked has a way to completely reset the datastore emulator. I like to keep local test data between runs of the local server for manual testing, however it is probably better to have more reliable unit tests. So, resetting the data stored in the emulator seems like the right thing to do.

@jrobbins, Okay. I will try to use Datastore Emulator for running the tests. Thank you for letting me know about the way to reset the datastore entries. 

Also, datastore supports exporting and importing of data just by sending a POST request  as mentioned [here](https://cloud.google.com/datastore/docs/tools/emulator-export-import). We could export the data before running a test. Once all the tests have been executed, we could reset the datastore and import back all the data. This way we could keep local test data  stored in the emulator between runs of the local server for manual testing.

package.json
```diff
@@ -11,6 +11,7 @@
     "travis-deps": "pip install -t lib -r requirements.travis.txt --upgrade",
     "dev-deps": "python -m pip install --no-deps -r requirements.dev.txt",
     "test": "python -m unittest discover -p '*_test.py'  -b",
+    "emulator" : "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk",
     "coverage": "python -m pip install --no-deps -r requirements.dev.txt && python -m coverage erase && python -m coverage run -m unittest discover -p '*_test.py' -b && python -m coverage html",
```
shivamag00, line 15: We can use this script to start the emulator. `--no-store-on-disk` will not store the entries on the disk.

testing_config.py
```diff
@@ -39,13 +39,16 @@
...
+from google.cloud import ndb
 from google.appengine.ext import testbed
 
 os.environ['DJANGO_SECRET'] = 'test secret'
 os.environ['SERVER_SOFTWARE'] = 'test ' + os.environ.get('SERVER_SOFTWARE', '')
 os.environ['CURRENT_VERSION_ID'] = 'test.123'
-
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'
+os.environ['DATASTORE_EMULATOR_HOST'] = 'localhost:15606'
+
```
shivamag00, line 51: These two environment variables are needed to be added while running tests.

shivamag00 (commented):
@jrobbins, Tests are able to use the Datastore Emulator. There is not any visible difference in the speed of their execution. Some tests are still left to be migrated.

shivamag00:
@jrobbins, FIY we will have to add the line `with client.context():` before every method call that interacts with the datastore in each test.

jrobbins:
> @jrobbins, FIY we will have to add the line `with client.context():` before every method call that interacts with the datastore in each test.

That sounds fine.  There might be a way to do it by introducing a base class that we use for all our tests.  Either way seems like it would add about the same amount of complexity.

testing_config.py
jrobbins: Makes sense.

package.json
jrobbins: So, would the user need to run that command before running tests?  Can we put it into the `test` target?

shivamag00: > So, would the user need to run that command before running tests?
Yes the developer will have to run the emulator before running any test.

> Can we put it into the test target?
On executing the npm run emulator, the terminal will get busy with the emulator process. To run the tests, we will have to use another terminal.

jrobbins: OK, that sounds fine for now.   Later we can try running it in the background in the same npm target.

shivamag00:
FIY, all the tests have been migrated, However, sometimes, on running the tests, a RANDOM subset of the tests fail. I tried running one of these test individually. It ran successfully like 9/10 times. I have not yet tracked the cause of this. However, I have a theory that random tests fail because of the queries not being consistent. E.g. when running the website locally, I can recall it took sometime for the datastore to list the feature on the all features page after it was created. Also, this explains the reason for these tests not failing before as we were not using the datastore emulator for testing. To test the theory, I will run the tests from the current main branch using datastore emulator and see if they fail or not. 

@jrobbins, Please let me know your views on this.

jrobbins:
> FIY, all the tests have been migrated, However, sometimes, on running the tests, a RANDOM subset of the tests fail. I tried running one of these test individually. It ran successfully like 9/10 times. I have not yet tracked the cause of this. However, I have a theory that random tests fail because of the queries not being consistent. E.g. when running the website locally, I can recall it took sometime for the datastore to list the feature on the all features page after it was created. Also, this explains the reason for these tests not failing before as we were not using the datastore emulator for testing. To test the theory, I will run the tests from the current main branch using datastore emulator and see if they fail or not.
> 
> @jrobbins, Please let me know your views on this.

Flakey unit tests often indicate real defects in our product source code.   Maybe our app would be improved by specifying strongly consistent reads in models.py.   We could add read_consistency=ndb.STRONG in various fetch() calls and see if that makes the tests not be flakey.   Obviously there could be a performance slowdown for doing that, but I don't think that will be too bad for this app.

package.json
```diff
@@ -11,6 +11,7 @@
     "travis-deps": "pip install -t lib -r requirements.travis.txt --upgrade",
     "dev-deps": "python -m pip install --no-deps -r requirements.dev.txt",
     "test": "python -m unittest discover -p '*_test.py'  -b",
+    "emulator" : "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --consistency=1.0",
```
shivamag00, line 14: Some tests were failing because the default consistency of the emulator is 0.9. As mentioned [here](https://cloud.google.com/sdk/gcloud/reference/beta/emulators/datastore/start), 

> --consistency=CONSISTENCY; default=0.9
Fraction of eventually consistent operations that should succeed immediately. Setting to 1.0 can be useful for unit tests, but may mask incorrect assumptions about non-ancestor queries which are eventually consistent.

Setting the consistency to 1.0 makes the test cases run successfully.

testing_config.py
```diff
@@ -58,14 +61,12 @@ def setUpOurTestbed():
...
   ourTestbed.init_blobstore_stub()
   ourTestbed.init_capability_stub()
-  ourTestbed.init_datastore_v3_stub()
   ourTestbed.init_files_stub()
   ourTestbed.init_logservice_stub()
   ourTestbed.init_mail_stub()
   ourTestbed.init_search_stub()
   ourTestbed.init_urlfetch_stub()
   ourTestbed.init_user_stub()
-  ourTestbed.init_memcache_stub()
```
shivamag00, line 70: We do not require these two stubs. Hence, removed.

shivamag00 (commented):
@jrobbins, please review these changes. One of the comments also includes the reason for the failure of random tests at random runs of testing.

shivamag00:
@jrobbins, FIY, I have resolved the merge conflicts.

package.json
```diff
@@ -11,7 +11,8 @@
     "travis-deps": "pip install -t lib -r requirements.travis.txt --upgrade",
     "dev-deps": "python -m pip install --no-deps -r requirements.dev.txt",
     "test": "curl -X POST 'http://localhost:15606/reset' && python -m unittest discover -p '*_test.py'  -b",
-    "emulator" : "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --consistency=1.0",
+    "start-emulator" : "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --consistency=1.0",
+    "stop-emulator" : "curl -X POST 'http://localhost:15606/shutdown'",
```
shivamag00: These targets are for starting and stopping the emulator. The steps to be executed for running the tests is

1. `npm run start-emulator` 
2. `npm test` - reset the emulator and start runnning the tests. 
3. `npm run stop-emulator`

shivamag00 (commented):
@jrobbins, these will be the new steps to run the tests. Please review

shivamag00:
@jrobbins, FIY,  I have updated README and Developer Docs. See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1365/commits/0aebfa3f26db5687fab41a63463209e8f18b7339) commit

shivamag00:
@jrobbins, FIY, this PR will be ready to merge once all the conversations have been resolved.

jrobbins:
Thanks for doing this.  I thought that I would get a chance to carefully review all of this today but I did not.  So, I will review it tomorrow.  The GAE NDB version is live on the site, so I also want to allow a little time for any problems there to be detected.

api/accounts_api_test.py
```diff
@@ -26,36 +26,44 @@
...
+client = ndb.Client()
 
 
 class AccountsAPITest(unittest.TestCase):
 
   def setUp(self):
     self.app_admin = models.AppUser(email='admin@example.com')
     self.app_admin.is_admin = True
-    self.app_admin.put()
+    with client.context():
```
jrobbins, line 39: These with-statements look correct, but they do get kind of repetitive.  Would it work to override the run() method in each test class like this?

```
  def run(self, result=None):
    with client.context():
      super(AccountsAPITest, self).run(result=result)
```

framework/ramcache.py
```diff
@@ -49,7 +49,10 @@
 
 import logging
 import time as time_module
-from google.appengine.ext import ndb
+# from google.appengine.ext import ndb
```
jrobbins, line 52: This commented-out line can also be removed.

framework/secrets.py
```diff
@@ -23,7 +23,8 @@
 import string
 import time
 
-from google.appengine.ext import ndb
+# from google.appengine.ext import ndb
```
jrobbins, line 26: Same.

internals/models.py
```diff
@@ -23,7 +23,8 @@
 import re
 import time
 
-from google.appengine.ext import ndb
+# from google.appengine.ext import ndb
```
jrobbins, line 26: Same.

internals/notifier.py
```diff
@@ -26,7 +26,8 @@
 import re
 
 from framework import ramcache
-from google.appengine.ext import ndb
+# from google.appengine.ext import ndb
```
jrobbins, line 29: Same.

testing_config.py
jrobbins: Great!

jrobbins (changes requested):
This all looks pretty good.   I just suggest trying the run() override to keep the code cleaner.  I wish I had thought of that when you asked me about this topic earlier, but it would still be worth doing now.

api/accounts_api.py
```diff
@@ -18,7 +18,8 @@
 
 import logging
 
-from google.appengine.ext import ndb
+# from google.appengine.ext import ndb
```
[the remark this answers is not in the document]
shivamag00: Oops...I forgot to remove the commented out lines. Thank you for pointing that out. Done!

framework/ramcache.py
shivamag00: Done!

framework/secrets.py
shivamag00: Done!

internals/models.py
shivamag00: Done!

internals/notifier.py
shivamag00: Done!

api/accounts_api_test.py
shivamag00: Yes, it seems to work fine. Also, as [suggested](https://github.com/GoogleChrome/chromium-dashboard/pull/1365#issuecomment-866246577) by you earlier, I have made a base class for the tests (See [these](https://github.com/GoogleChrome/chromium-dashboard/blob/718bff6e47091cb73159c37c2e87c16d3094ea45/testing_config.py#L128) lines). I have changed the baseclass of all the tests in [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1365/commits/718bff6e47091cb73159c37c2e87c16d3094ea45) commit

The tests work fine. Thank you for pointing out this method. I was unaware about this method as I have never used unittest before in any of my projects.

shivamag00:
@jrobbins, FIY, I have made the requested changes. The code is now DRY with respect to calls to `client.context()` in tests.

jrobbins (approved):
This all looks good.  I tested it in staging and did not find any problems.

Issue #1366, opened 2021-06-18
No dropdown for standardization status


TimothyGu:
**Describe the bug**
When editing a feature, one cannot change the standardization status of the spec.

**To Reproduce**
1. Go to https://chromestatus.com/feature/5658622220566528. Under "Specification" is "Editor's draft".
2. Now go to https://chromestatus.com/guide/editall/5658622220566528 as a Chromium contributor. There is no way to change the status of the specification to be "Established standard" or anything else.

**Expected behavior**
One should be able to change the standardization status.

jyasskin:
I think this is fixed. At least, I can't find "Editor's draft" on https://chromestatus.com/feature/5658622220566528 anymore.

past:
Seems fixed by #1428 and #1514.

PR #1367, opened 2021-06-18
Show spec_link itself rather than old specification field enum.

jrobbins:
This should resolve issue #1366.

IIRC, the standardization field was deprecated because the status of a standardization effort is too nuanced to accurately represent with a simple enum.  Instead, users really need to click through to the spec link and just for themselves what the status of that effort is.

I think the reason that this change was half done before was because I didn't understand all the ways that this enum was used in the UI until digging through it today.

In this CL:
+ Delete the STANDARDIZATION enum from models.py.
+ Mark the DB field as deprecated (we don't delete it because we don't want accidentally reuse the same name).
+ Remove the standardization enum name and int from the JSON representation of features
+ Remove the standardization section from the search help dialog
+ Change the little standardization color block on the expanded feature card from displaying the standardization status enum to just saying "Spec" if there is a spec_link.
+ Change a section heading on the feature detail page from "Consensus &amp; Standardization" to just "Consensus" because the spec link is already displayed under a separate "Specification" heading.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -258,15 +258,6 @@ EDITORS_DRAFT = 4
 PUBLIC_DISCUSSION = 5
 NO_STD_OR_DISCUSSION = 6
 
-STANDARDIZATION = {
-  DEFACTO_STD: 'De-facto standard',
-  ESTABLISHED_STD: 'Established standard',
-  WORKING_DRAFT: 'Working draft or equivalent',
-  EDITORS_DRAFT: "Editor's draft",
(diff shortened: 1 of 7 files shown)


jrobbins:
We decided to keep this field and try an updated set of values.  New PR to come soon.

jpmedley:
>IIRC, the standardization field was deprecated because the status of a standardization effort is too nuanced to accurately represent with a simple enum.

For what it's worth, the folks at MDN reached the same conclusion and have been moving away from reflecting it.

PR #1369, opened 2021-06-22
Update loading primary_blink_components to NDB.

jrobbins:
This looks like one call to load a datastore entity that was missed when upgrading from the old DB API to the GAE NDB API.  It caused an error on the rarely used /admin/subscribers page.


shivamag00 (approved):
LGTM

Issue #1370, opened 2021-06-23
editing UI doesn't show up correctly right after signing in


dbaron:
**Describe the bug**
When I signed in to chromestatus this morning, for the first time, using my chromium.org account, I didn't see any UI for how to edit a page.  I emailed a colleague asking if I was supposed to see such UI, said I was, and described what it looked like.  When I reloaded the page I was on, I was then able to see the editing UI and edit it.  But it's not clear to me why this UI didn't show up the first time after logging in.

I actually have screenshots of both states, which may be interesting.  Here's what things looked like after I logged in initially:
![Screen Shot 2021-06-23 at 09 17 34](https://user-images.githubusercontent.com/248721/123179779-ee89c800-d43e-11eb-8064-54c5f58ef0ff.png)

and then once I reloaded (an hour or so later, maybe a bit more), things looked like this:
![Screen Shot 2021-06-23 at 09 18 48](https://user-images.githubusercontent.com/248721/123179831-03fef200-d43f-11eb-8e74-3ec90afa6087.png)

Note there are two differences:
* the signed-in UI in the upper right corner looks different, and
* the pencil icon shows up at the upper right of the feature

**To Reproduce**
1. Go to a feature page on https://chromestatus.com/ while signed out, such as [the one I was editing](https://www.chromestatus.com/feature/5647948924846080).  It might even require that you've never been signed in before.
2. Click on the "Sign in" button in the upper right corner, and log in to an account that has editing permission (e.g., a chromium.org account).

**Expected behavior**
Editing UI (pencil icon) shows up.

jrobbins:
I think the root cause is that signing in can sometimes fail after Google Sign-In does its work including updating the UI to say "Signed in", but before we reload the page to display the buttons that the user has access to.
shivamag00's PR should resolve this because it eliminates the "Signed in" state of that button, so if something fails, the button will still say "Sign in", so the user will know to try again.

PR #1376, opened 2021-06-29
20210628 UI enhancements

shivamag00:
Review Requested:- @jrobbins 

In this PR, 

- UI Improvements for Narrow View Ports
- Implemented Custom Google Sign In Button - 

Currently, the flow for signing in, as seen by the user, is -
1. There is a button with the text `Sign In`. The User clicks on it. A pop-up appears and the user selects the gmail id to be used.
2. The pop-up disappears and button now reads `Signed In`. Let's say this is `intermediate stage`. 
3. The page is refreshed and in place of button, the gmail id used for logging in is displayed. It means the logging in process is completed.

The intermediate stage may be source of confusion for the user (as mentioned in this issue #1370 . Since we were using the in-built Google Sign In button, we could not change the text of the button. Also, the button was not getting wrapped properly in narrow viewports because of the default CSS. Hence, we needed to use a custom google sign in button.

UI Changes:-
Picture1


jrobbins:
Thanks for this PR.
I looked through this some, but I want to take some time tomorrow to test it before approving it.

jrobbins (approved):

Pravenn18 (commented):
Great Work,the code change is done in a proper manner

PR #1380, opened 2021-06-30
Tell dev_appserver to start datastore emulator

jrobbins:
Based on some previous discussions and the instructions in the README.md file, I thought that 'npm start' was supposed to cause dev_appserver to start the emulator automatically.  However, that was not working for me until I added this option.

### DIFF
--- a/scripts/start_server.sh
+++ b/scripts/start_server.sh
@@ -7,5 +7,6 @@
 # The directory in which this script resides.
 readonly BASEDIR=$(dirname $BASH_SOURCE)
 
-dev_appserver.py -A cr-status --enable_console=1 --datastore_emulator_port=15606 \
+dev_appserver.py -A cr-status --enable_console=1 \
+  --support_datastore_emulator=1 --datastore_emulator_port=15606 \
   $BASEDIR/../app.yaml $BASEDIR/../notifier.yaml


jrobbins:
If 'npm start' was working for you without this option, I imagine two possible causes:
(1) You had the datastore emulator running for the unit tests, or
(2) You might have a different version of dev_appserver installed than I have.

I tried to see if my dev_appserver was old, but it does not seem to support a --version argument.   However, `gcloud components list` says 
```
Your current Cloud SDK version is: 347.0.0
The latest available version is: 347.0.0
```

shivamag00:
That's interesting. When I run npm start, I get the following log which says that Cloud DataStore is started by default. 

> (chromestatus) shivamag00:/mnt/chromium-dashboard$ ./scripts/start_server.sh
> INFO     2021-07-01 03:58:55,720 devappserver2.py:232] Using Cloud Datastore Emulator.
> We are gradually rolling out the emulator as the default datastore implementation of dev_appserver.
> If broken, you can temporarily disable it by --support_datastore_emulator=False
> Read the documentation: https://cloud.google.com/appengine/docs/standard/python/tools/migrate-cloud-datastore-emulator
> Help us validate that the feature is ready by taking this survey: https://goo.gl/forms/UArIcs8K9CUSCm733
> Report issues at: https://issuetracker.google.com/issues/new?component=187272
> 
> INFO     2021-07-01 03:58:55,774 devappserver2.py:289] Skipping SDK update check.
> ....

However, my gcloud SDK is a bit older as `gcloud components list` returns the following:-
```
Your current Cloud SDK version is: 343.0.0
The latest available version is: 347.0.0
```

I will update my SDK and see if Cloud Datastore is no longer started by default.

shivamag00 (approved):
LGTM

PR #1381, opened 2021-06-30
Update instructions to create admin using Cloud NDB.

jrobbins:
The old instructions only worked for the old NDB.


shivamag00 (approved):
LGTM. Thank you for this PR.

In the history section, we have mentioned that it used the NDB Client Library. However, now it uses the Cloud NDB Library. Since I cannot make commits to this PR, I have raised a separate for it. See #1383

PR #1382, opened 2021-06-30
Fix FeatureOwner order() property and improve performance of /admin/blink

jrobbins:
This PR makes two improvements to blink_handlers.py:

(1) Correct the sort order field name for the list of possible subscribers.
  This had originally been specified as 'name', and then your recent upgrade to GAE NDB changed it to models.BlinkComponent.name, which works, but models.FeatureOwner.name is more correct.

(2) Eliminate queries done in loops that made the page very slow to load on the prod site.  
  - There was a loop in python code that where c.owner did a query, and that is done over 700 times.
  - In the HTML template there was a reference to c.subscribers that also did a query and that was done over 700 times
  - The new code uses the list of possible_subscribers that is already fetched and loops over those entities in python code with no additional queries.

### DIFF
--- a/pages/blink_handler.py
+++ b/pages/blink_handler.py
@@ -90,21 +90,29 @@ class BlinkHandler(basehandlers.FlaskHandler):
 
   @permissions.require_admin_site
   def get_template_data(self):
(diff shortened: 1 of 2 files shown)


shivamag00 (approved):
LGTM

PR #1400, opened 2021-07-06
Channels API

shivamag00:
Review Requested: @jrobbins 

In this PR,

- added API for channels.

`schedule.html` requires feattures list and channels with release schedule. This PR creates an API for channels. In a future PR, we will try to use this API for fetching the channels.

### DIFF
--- a/api/channels_api_test.py
+++ b/api/channels_api_test.py
@@ -1,0 +1,90 @@
+from __future__ import division
+from __future__ import print_function
+
+# Copyright 2020 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


api/channels_api_test.py
```diff
@@ -0,0 +1,90 @@
...
+import mock
+import unittest
+
+from api import channels_api
+
+
+class ChannelsAPITest(unittest.TestCase):
+
+  @mock.patch('api.channels_api.fetch_chrome_release_info')
+  @mock.patch('api.channels_api.fetchchannels.get_omaha_data')
```
shivamag00, line 28: I have patched objects in the namespace where they are used.

shivamag00 (commented):
@jrobbins, I have never used `@mock.patch()` before. Please let me know if I have used it incorrectly.

api/channels_api_test.py
jrobbins: Hmm, I would expect it to be
@mock.patch('internals.fetchchannels.get_omaha_data')
If you see tests passing, then I suppose it works either way, but I think specifying the location where a function is defined is more conventional.

jrobbins (changes requested):

api/channels_api_test.py
shivamag00: Yes the test was a success. Anyway, I have changed it to what you suggested thanks.

jrobbins (approved):

Issue #1342, opened 2021-05-31
Use datalist instead


jimmywarting:
On https://chromestatus.com/metrics/feature/timeline/popularity you have a very long list of select options
It's not so search friendly...

![Skärmavbild 2021-06-01 kl  01 09 46](https://user-images.githubusercontent.com/1148376/120248207-7fbdb280-c276-11eb-94e6-c240f1631f5c.png)

Change it to:
```html
<input type="search" list="features">
<datalist id="features">
  <option>
  ...
```
and you will get a input with both a drop down and a search field combo
![Skärmavbild 2021-06-01 kl  01 11 32](https://user-images.githubusercontent.com/1148376/120248262-b85d8c00-c276-11eb-910d-cb26e518df16.png)

(There is no need for a fancy/complicated select2 dropdown with lots of javascript complexity - native is the king!)

jrobbins:
I think to make the change we would need to change the <option value="..."> attribute to be the name of the UseCounter.  That would avoid displaying the UseCounter (bucket) numbers.   And, the JS and/or python code would need to be changed to accommodate that.   We would still need to support existing URLs that contain the bucket ID number.

jimmywarting:
dose number in the screenshots just poped up cuz i just simply switched the `<select>` to be a `<datalist>` in devtool
a unnecessary byproduct if you will. don't remember where they came from or what they stands for

Subhradeep10:
Can I work on it?

Subhradeep10:
@jimmywarting Can I work on it?

jimmywarting:
@Subhradeep10 ofc! Go ahead.

I just submitted a idea, I have no knowledge/experiences in this repo

GouravSingal-code:
can i work on it

rushikeshadhav:
Is this issue open. I am beginner to this project. Can you assign me ?

ankit-dot:
can i work on it \

kevinshen56714:
Hi, I have created PR #1824 for this.

PR #1401, opened 2021-07-08
Use datalist instead of select

shubhammittal588:
This should resolve issue #1342 .

In this PR:
1. Changed select tag with input tag and added attributes type="search" list=“features" .
2. Used datalist tag with attribute id=“features”.
3. Changed the attribute to be the name of the UseCounter in option tag.

### DIFF
--- a/static/elements/chromedash-timeline.js
+++ b/static/elements/chromedash-timeline.js
@@ -253,12 +253,13 @@ ORDER BY yyyymmdd DESC, client`;
 
   render() {
     return html`
-      <select .value="${this.selectedBucketId}" @change="${this.updateSelectedBucketId}">
-        <option disabled value="1">Select a property</option>
-        ${this.props.map((prop) => html`
-          <option value="${prop[0]}">${prop[1]}</option>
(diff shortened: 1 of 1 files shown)


jrobbins:
I think you might be missing a commit.  The diffs for this PR do not match what you wrote in the PR description.

shubhammittal588:
I have committed the diffs, this is the screenshot of the commit I made.
Screenshot 2021-07-10 at 1 08 48 PM

jrobbins:
I think this could be made to work, but it is not working right now there is a UX problem also.

The current code is causing the bucket name to be placed in the URL of the API request.  Instead, it needs to be the bucket ID number.  Or, the API handler needs to be modified to accept either names or IDs.

When the user clicks on the widget, nothing happens initially.  So, it is not so clear what the purpose of this widget is or what options are available.  If the user types "a" they will see options that contain the letter "a", but there does not seem to be a good way to see the entire list.  The prompt "Select a property" is not visible, so there is no indication that the user should do anything with this widget.

jrobbins:
I'm going to close this to clear up our list of pending PRs.  But, I will keep this code in mind when I eventually get around to redesigning this page.

PR #1403, opened 2021-07-09
Don't set DATASTORE_EMULATOR_HOST on GAE

jrobbins:
We only want this variable set when running locally.  If it is set when running on GAE, we get repeated failures with 503 responses from the backend because we are trying to hit an emulator in a situation where there is no such emulator.


shivamag00 (approved):
LGTM

PR #1405, opened 2021-07-13
Use bigger batches with Cloud NDB, add debugging.

jrobbins:
Some pages are timing out when the user requests them.  

For the metrics data, the existing batch query seems to run very fast, but the queries in the loop end up being slow.   So, I am doing a bigger batch request.

I had tried debugging performance of the other pages using LOGPOINTs in the GAE debugger, but apparently Cloud NDB entities in RAM use getter functions to access fields and the cloud debugger does not allow me to use them in a LOGPOINT.

### DIFF
--- a/api/metricsdata.py
+++ b/api/metricsdata.py
@@ -31,7 +31,7 @@ from framework import ramcache
 import settings
 
 CACHE_AGE = 86400 # 24hrs
-
+BATCH_SIZE = 15000
 
 
 def _truncate_day_percentage(datapoint):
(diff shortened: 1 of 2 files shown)


jstenback (approved):
LGTM

jrobbins:
After looking into this more, I don't think that this is going to help enough.  We probably need to precompute the set of data needed to generate the stack rank pages in a cron task.

jrobbins:
I think that #1410 probably improved performance.  I'll keep check later today to see if there are fewer timeouts.

Issue #1372, opened 2021-06-25
The Debuggability section is missing from generated intent message text


jstenback:
**Describe the bug**
When generating an Intent to Experiment message using chromestatus the Debuggability information is not included.

**To Reproduce**
1. Load https://chromestatus.com/admin/features/launch/5740655424831488/3?intent=1
2. Notice how there's no Debuggability section in the generated page.

**Expected behavior**
Include the provided debuggability information from the feature entry.

hashseed:
Ideally this section should also include https://goo.gle/devtools-checklist, which offers a lot of detail on how to think about debuggability from the perspective of the intent author.

mathiasbynens:
> Ideally this section should also include https://goo.gle/devtools-checklist, which offers a lot of detail on how to think about debuggability from the perspective of the intent author.

https://github.com/GoogleChrome/chromium-dashboard/pull/1375 addresses this aspect (but it still keeps the field optional).

mathiasbynens:
Next steps could be:

- [ ] Unconditionally include the debuggability section in generated emails. Patch: https://github.com/GoogleChrome/chromium-dashboard/pull/1409
- [ ] Make the debuggability input field required. Patch: https://github.com/GoogleChrome/chromium-dashboard/pull/1408

mathiasbynens:
With #1375, #1408, and #1409 merged, I believe this issue can now be closed. Thanks everyone! 🥳

PR #1408, opened 2021-07-14
Make “debuggability” field required

mathiasbynens:
Now that #1375 landed, there is clear guidance for feature owners on how to assess debuggability. In cases where debuggability is already covered through existing DevTools functionality, it’s valuable for feature owners to explicitly state that rather than leaving the field empty.

Issue: #1372

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -506,7 +506,7 @@ ALL_FIELDS = {
                    'the Request for Deprecation Trial email thread.')),
 
     'debuggability': forms.CharField(
-        label='Debuggability', required=False,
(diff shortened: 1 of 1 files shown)


jrobbins:
Let's go ahead with this change for now.  But, overall I don't think that making all fields required would work because it would make users feel like they are fighting with the tool whenever they work ahead, and having a few fields required will seem inconsistent.  When I get a chance, I will experiment with replace setting the fields to be always required with some JS that enforces required fields only when the user wants to advance the process stage.

jrobbins (approved):

Issue #1374, opened 2021-06-26
Border-radius use counter stat shows >100% of page loads


juwain:
There is probably an issue with data of the border-radius property [rank usage metrics](https://chromestatus.com/metrics/css/popularity). In the middle of the 2018 stats shows 109% of page loads. 

![55005FCF-F6C3-40DC-89C2-DDEB6A04FBC9](https://user-images.githubusercontent.com/928081/123513252-cdbab000-d694-11eb-98d3-db2890889cdf.jpeg)

1. Go to https://chromestatus.com/metrics/css/timeline/popularity/53.
2. See error in chart. 

![E5967E96-A4B4-4052-AA9B-8C1FC644774D](https://user-images.githubusercontent.com/928081/123513646-d01e0980-d696-11eb-9009-a548e8730376.png)

Expected behavior: percent of page loads should not to be more than 100%.

jrobbins:
The fix was deployed.

The underlying cause was:
+ There was a datapoint for border-radius that had a value of 109% back in 2018 or so.  I don't know the story behind that.
+ We were failing to sort datapoint by date when generating this page, so old datapoints were sometimes shown.
+ We recently upgraded our datastore library, so default sorting may have changed, resulting in a more noticeable incorrect value being surfaced.

PR #1410, opened 2021-07-14
Fix application of query order.

jrobbins:
The DB and NDB query objects are immutable and adding a new filter or setting the query order returns a new query object, which we need to keep in a variable.

This looks like a long-standing defect that has been in our code for at least 9 months.  However, the recent upgrade from DB to NDB to Cloud NDB may have made the defect more visible.  Other wrong results may have been show often, but they happened to not be very different in value from the correct results.


jpchase (approved):
LGTM.

jrobbins:
This resolves issue #1374

PR #1411, opened 2021-07-14
Use async queries to speed up request handlers.

jrobbins:
Use NDB methods fetch_async() and get_async() to do more work in parallel rather than sequentially.

This seems to speed up the cache-miss case for the feature list page and metrics stack rank page by about 30%, which should prevent timeouts on those pages.

I considered a bunch of different ways to speed up these pages and this seems like the biggest win with the least added complexity.  Eventually, these pages will be API-driven and request smaller batches of data incrementally, so I don't want to add complexity (like precomputing results or sharding or serving stale data in certain conditions) that would raise the threshold for replacing these handlers.

### DIFF
--- a/api/metricsdata.py
+++ b/api/metricsdata.py
@@ -141,6 +141,8 @@ class FeatureHandler(basehandlers.FlaskHandler):
   def __query_metrics_for_properties(self):
     datapoints = []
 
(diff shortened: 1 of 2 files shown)


jstenback (approved):
Neat, LGTM

PR #1424, opened 2021-07-16
Query-String Parameter for milestone in features api

### LINKED ISSUE #1414: Add milestone query-string parameter to filter feature list response

shivamag00:
Review Requested: @jrobbins 

Merging this should close #1414 

In this PR,

- Added a Query-String Parameter for milestone in features API to fetch features in a specific milestone

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -29,6 +29,7 @@ class FeaturesAPI(basehandlers.APIHandler):
 
   def do_get(self):
     user = users.get_current_user()
+    show_unlisted_features = permissions.can_edit_feature(user, None)
     feature_list = None
 
     # Query-string parameter 'milestone' is provided
@@ -37,7 +38,7 @@ class FeaturesAPI(basehandlers.APIHandler):
         milestone = int(self.request.args.get('milestone'))
         feature_list = models.Feature.get_in_milestone(
           version=2,
-          show_unlisted=permissions.can_edit_feature(user, None), 
+          show_unlisted=show_unlisted_features, 
           milestone=milestone)
       except ValueError:
         self.abort(400, msg='Invalid  Milestone')
@@ -46,7 +47,7 @@ class FeaturesAPI(basehandlers.APIHandler):
     if feature_list is None:
       feature_list = models.Feature.get_chronological(
           version=2,
-          show_unlisted=permissions.can_edit_feature(user, None))
+          show_unlisted=show_unlisted_features)
 
     return feature_list
 
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -29,9 +29,25 @@ class FeaturesAPI(basehandlers.APIHandler):
 
   def do_get(self):
     user = users.get_current_user()
-    feature_list = models.Feature.get_chronological(
-        version=2,
-        show_unlisted=permissions.can_edit_feature(user, None))
+    feature_list = None
+
+    # Query-string parameter 'milestone' is provided
+    if self.request.args.get('milestone') is not None:
+      try:
+        milestone = int(self.request.args.get('milestone'))
+        feature_list = models.Feature.get_in_milestone(
+          version=2,
+          show_unlisted=permissions.can_edit_feature(user, None), 
+          milestone=milestone)
+      except ValueError:
+        self.abort(400, msg='Invalid  Milestone')
+
+    # No Query-string parameter is provided
+    if feature_list is None:
+      feature_list = models.Feature.get_chronological(
+          version=2,
+          show_unlisted=permissions.can_edit_feature(user, None))
+
     return feature_list
 
   # TODO(jrobbins): do_post
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -37,7 +37,6 @@ class FeaturesAPI(basehandlers.APIHandler):
       try:
         milestone = int(self.request.args.get('milestone'))
         feature_list = models.Feature.get_in_milestone(
-          version=2,
           show_unlisted=show_unlisted_features, 
           milestone=milestone)
       except ValueError:
--- a/internals/models.py
+++ b/internals/models.py
@@ -557,7 +557,7 @@ class Feature(DictModel):
   @classmethod
   def _annotate_first_of_impl_status_in_milestones(self, feature_list, version=2):
     try:
-      versions = [
+      statuses = [
         IMPLEMENTATION_STATUS[BEHIND_A_FLAG],
         IMPLEMENTATION_STATUS[ENABLED_BY_DEFAULT],
         IMPLEMENTATION_STATUS[REMOVED],
@@ -570,8 +570,8 @@ class Feature(DictModel):
         first_of_milestone_func = Feature._first_of_milestone_v2
 
       last_good_idx = 0
-      for i, ver in enumerate(versions):
-        idx = first_of_milestone_func(feature_list, ver, start=last_good_idx)
+      for i, status in enumerate(statuses):
+        idx = first_of_milestone_func(feature_list, status, start=last_good_idx)
         if idx != -1:
           feature_list[idx]['first_of_milestone'] = True
           last_good_idx = idx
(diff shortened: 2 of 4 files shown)


internals/models.py
```diff
@@ -885,6 +885,46 @@ def getSortingMilestone(feature):
...
+    q = q.filter(Feature.shipped_milestone == milestone)
+    desktop_shipping_features = q.fetch(None)
+
+    # Features with an android shipping milestone but no desktop milestone.
+    q = Feature.query()
+    q = q.order(Feature.name)
+    q = q.filter(Feature.shipped_android_milestone == milestone)
+    q = q.filter(Feature.shipped_milestone == None)
+    android_only_shipping_features = q.fetch(None)
+
```
shivamag00, line 906: @jrobbins, Features with desktop or android milestones that are same as the queried milestone, will be returned by the API.

api/features_api.py
```diff
@@ -29,9 +29,25 @@ class FeaturesAPI(basehandlers.APIHandler):
...
+    if self.request.args.get('milestone') is not None:
+      try:
+        milestone = int(self.request.args.get('milestone'))
+        feature_list = models.Feature.get_in_milestone(
+          version=2,
+          show_unlisted=permissions.can_edit_feature(user, None), 
+          milestone=milestone)
+      except ValueError:
+        self.abort(400, msg='Invalid  Milestone')
+
```
shivamag00, line 44: If the query string is anything other than a number, the request will be aborted with error code 400.

shivamag00 (commented):
@jrobbins, please review the following

api/features_api.py
```diff
@@ -29,9 +29,25 @@ class FeaturesAPI(basehandlers.APIHandler):
...
-        version=2,
-        show_unlisted=permissions.can_edit_feature(user, None))
+    feature_list = None
+
+    # Query-string parameter 'milestone' is provided
+    if self.request.args.get('milestone') is not None:
+      try:
+        milestone = int(self.request.args.get('milestone'))
+        feature_list = models.Feature.get_in_milestone(
+          version=2,
```
jrobbins, line 39: You could drop the version=2 argument and the parameter in the function because they are not used.  I think that was a left-over detail of some old upgrade to the chronological feature list feed.  We don't need that for new code, especially since we have a version number in the API URL path.

api/features_api.py
jrobbins: That sounds right.

api/features_api.py
```diff
@@ -29,9 +29,25 @@ class FeaturesAPI(basehandlers.APIHandler):
...
+          show_unlisted=permissions.can_edit_feature(user, None),
```
jrobbins, line 40: The call to permissions.can_edit_feature(user, None) could be factored out to the top of the method.

internals/models.py
```diff
@@ -885,6 +885,46 @@ def getSortingMilestone(feature):
...
+    # Constructor the proper ordering.
+    all_features = []
+    all_features.extend(desktop_shipping_features)
+    all_features.extend(android_only_shipping_features)
+
+    # First sort by name, then sort by feature milestone (latest first).
```
jrobbins, line 912: No longer sorting by milestone.

internals/models.py
```diff
@@ -885,6 +885,46 @@ def getSortingMilestone(feature):
...
+    all_features.sort(key=lambda f: f.name, reverse=False)
+
+    all_features = [f for f in all_features if not f.deleted]
+
+    feature_list = [f.format_for_template(version) for f in all_features]
+
+    self._annotate_first_of_milestones(feature_list, version=version)
```
jrobbins, line 919: I think that this will be needed.

jrobbins (changes requested):

api/features_api.py
shivamag00: If I do not specify `version = 2`, then the API response will be different than what it is for `features.json` and `features_v2.json`. (Both `features.json` and `features_v2.json` use `version = 2` and hence, their response is same.) 

**_How is the response different ?_**
If we do not specify `version = 2`, then 
- the keys in the JSON response will be different. 
- much more information will be present in the response.

shivamag00: Should we make the `format_for_template()` method use `version = 2` by default ?

api/features_api.py
shivamag00: Done. Thank you for pointing this out.

internals/models.py
shivamag00: Oops... I forgot to make changes to this comment. 

Done. Thanks

internals/models.py
shivamag00: The `get_chronological()` method returns the list of all features such that features having similar implementation statuses are listed together. Note that the features that have shipped milestone field or shipped android milestone, are grouped together according to the chromium version number instead of the implementation status.

Hence, features with implementation statuses as `'Behind a flag'`, `'Enabled by default'`, `'Deprecated'`, `'Removed'`, `'Origin trial'` etc., which always have shipped milestone field or shipped android milestone, are always grouped according to the chromium version numbers.

It seems that `_annotate_first_of_milestones()` annotates the first feature of each implementation statuse (which IS NOT associated with version numbers) and each chromium version. It fails to annotate the first feature which is `behind the flag`, `removed`, `deprecated` etc. 

Since, all the features that are returned by the `get_in_milestone()` method belong to a certain version number and hence has one of the implementation statuses from `'Behind a flag'`, `'Enabled by default'`, `'Deprecated'`, `'Removed'`, `'Origin trial'` etc., the method `_annotate_first_of_milestones()` will not be of any use to us.

shivamag00: I did not notice it earlier. To sort the features in the given chromium version according to the implementation status, we will have to use a new method.

shivamag00: See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1424/commits/8f84d1147cd56a3adb440b35663c7b4f635c64bc) commit

api/features_api.py
jrobbins: Yeah, let's make version=2 the default, then it can be omitted from this code.

internals/models.py
jrobbins: Good point.  That looks good now.

internals/models.py
```diff
@@ -554,6 +554,30 @@ def _annotate_first_of_milestones(self, feature_list, version=None):
     except Exception as e:
       logging.error(e)
 
+  @classmethod
+  def _annotate_first_of_impl_status_in_milestones(self, feature_list, version=None):
+    try:
+      versions = [
```
jrobbins, line 560: "versions" sounds like it would be milestones.  "statuses" would be a better name.

internals/models.py
jrobbins: Looks good.

internals/models.py
```diff
@@ -885,6 +909,48 @@ def getSortingMilestone(feature):
...
+    android_only_shipping_features = q.fetch(None)
+
+    # Constructor the proper ordering.
+    all_features = []
+    all_features.extend(desktop_shipping_features)
+    all_features.extend(android_only_shipping_features)
+
+    # Feature list must be first sorted by implementation status and then by name. 
+    # The implementation may seem to be counter-intuitive using sort() method.
+    all_features.sort(key=lambda f: f.name, reverse=False)
```
jrobbins, line 938: You could remove reverse=False here and on the next line.

jrobbins (changes requested):

api/features_api.py
shivamag00: Okay.
Done
See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1424/commits/8ce4987c4f193c24781d866a2a9449a44a40548a) commit

internals/models.py
shivamag00: Okay. Marking this conversation as resolved.

internals/models.py
shivamag00: True. Thanks for pointing this out. Done.

internals/models.py
shivamag00: Done.

jrobbins (approved):

PR #1425, opened 2021-07-16
Add fields for DevTrial milestones.

jrobbins:
In this CL:
 + In models.py: Define fields for the start milestone of a devtrial on each platform.
 + In guideforms.py: Define django form fields for those milestones, and add those fields to the devtrials stage form, the flat form, and the feature detail page.
 + In guide.py: Process the new milestone values when edited.
 + In intent_to_implement.html: Add a new section with a table for all milestones.

Note, this PR does not include logic to display the devtrails in the feature list or releases page, or the features_v2.json feed.

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -642,6 +642,33 @@ ALL_FIELDS = {
             'https://github.com/WICG/idle-detection/blob/main/HOWTO.md'
             '">Example 2</a>.')),
 
+    'dt_milestone_desktop_start': forms.IntegerField(
+        required=False, label='DevTrail on desktop',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on desktop platforms by setting a flag.')),
+
+    'dt_milestone_android_start': forms.IntegerField(
+        required=False, label='DevTrail on Android',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
+
+    'dt_milestone_ios_start': forms.IntegerField(
+        required=False, label='DevTrial on iOS (RARE)',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
+
+    'dt_milestone_webview_start': forms.IntegerField(
+        required=False, label='DevTrial on Webview',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on Webview by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
+
     'flag_name': forms.CharField(
         label='Flag name', required=False,
         help_text='Name of the flag that enables this feature.'),
@@ -734,8 +761,9 @@ Any_DevTrial = define_form_class_using_shared_fields(
 
 ImplStatus_DevTrial = define_form_class_using_shared_fields(
     'ImplStatus_InDevTrial',
-    ('shipped_milestone', 'shipped_android_milestone',
-     'shipped_ios_milestone', 'flag_name'))
+    ('dt_milestone_desktop_start', 'dt_milestone_android_start',
+     'dt_milestone_ios_start', 'dt_milestone_webview_start',
+     'flag_name'))
 
 
 NewFeature_EvalReadinessToShip = define_form_class_using_shared_fields(
@@ -903,6 +931,8 @@ Flat_DevTrial = define_form_class_using_shared_fields(
      # TODO(jrobbins): UA support signals section
 
      # Implementation
+     'dt_milestone_desktop_start', 'dt_milestone_android_start',
+     'dt_milestone_ios_start', 'dt_milestone_webview_start',
      'flag_name'))
   # TODO(jrobbins): api overview link
 
@@ -1010,6 +1040,8 @@ DISPLAY_FIELDS_IN_STAGES = {
         'debuggability',
         'all_platforms', 'all_platforms_descr', 'wpt', 'wpt_descr',
         'sample_links', 'devrel', 'ready_for_trial_url',
+        'dt_milestone_desktop_start', 'dt_milestone_android_start',
+        'dt_milestone_ios_start', 'dt_milestone_webview_start',
         'flag_name'),
     models.INTENT_IMPLEMENT_SHIP: make_display_specs(
         'launch_bug_url',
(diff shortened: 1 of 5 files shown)


pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
             'https://github.com/WICG/idle-detection/blob/main/HOWTO.md'
             '">Example 2</a>.')),
 
+    'dt_milestone_desktop_start': forms.IntegerField(
+        required=False, label='DevTrail on desktop',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
```
jpmedley, line 648:
```suggestion
        help_text=('First milestone that allows developers to try '
```

jpmedley: Follows the best practice of omitting needless words, making the sentence small-screen friendly.

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+                   'this feature on desktop platforms by setting a flag.')),
+
+    'dt_milestone_android_start': forms.IntegerField(
+        required=False, label='DevTrail on Android',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
```
jpmedley, line 654:
```suggestion
        help_text=('First milestone that allows developers to try '
```

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
+
+    'dt_milestone_ios_start': forms.IntegerField(
+        required=False, label='DevTrial on iOS (RARE)',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
```
jpmedley, line 661:
```suggestion
        help_text=('(Rare) First milestone that allows developers to try '
```

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
+
+    'dt_milestone_webview_start': forms.IntegerField(
+        required=False, label='DevTrial on Webview',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
```
jpmedley, line 668:
```suggestion
        help_text=('First milestone that allows developers to try '
```

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+                   'this feature on Webview by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
```
jpmedley, line 670: Same question here.

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on desktop platforms by setting a flag.')),
+
+    'dt_milestone_android_start': forms.IntegerField(
+        required=False, label='DevTrail on Android',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
```
jpmedley, line 656: Are you autofilling based on desktop? If my feature is desktop only, how is this distinguished from desktop-applies elsewhere? Setting this should be explicit, even if it duplicates desktop.

pages/guideforms.py
```diff
@@ -642,6 +642,33 @@ def validate(self, value):
...
+
+    'dt_milestone_ios_start': forms.IntegerField(
+        required=False, label='DevTrial on iOS (RARE)',
+        widget=forms.NumberInput(attrs={'placeholder': 'Milestone #'}),
+        help_text=('First milestone that will allow developers to try out '
+                   'this feature on iOS by setting a flag.  '
+                   'Set this only if different from desktop milestone.')),
```
jpmedley, line 663:
```suggestion
                   'this feature on iOS by setting a flag.  ')),
```

jpmedley: This wording implies that iOS is as common to features. I think I can count on one hand the number of times a Chrome Status feature applied to iOS.

templates/blink/intent_to_implement.html
```diff
@@ -204,6 +204,64 @@ <h4>Contact emails</h4>
...
+    <tr><td>DevTrial on Webview</td>
+    <td>{{feature.dt_milestone_webview_start}}</td></tr>
+  {% endif %}
+
+  {% if feature.ot_milestone_desktop_start %}
+    <tr><td>OriginTrial first</td>
+    <td>{{feature.ot_milestone_desktop_start}}</td></tr>
+  {% endif %}
+
+  {% if feature.ot_milestone_desktop_end %}
```
jpmedley, line 236: Why no origin trial fields for Android and WebView?

jpmedley (changes requested):

jstenback (approved):
LGTM

cwilso (approved):
LGTM

pages/guideforms.py
jrobbins: Done.

pages/guideforms.py
jrobbins: I was not planning on autofilling, but I see what you mean.  If I autofill the desktop value into other platform fields, then users could delete the value to indicate that the feature will not be available on that platform.  I'll try that.

pages/guideforms.py
jrobbins: Removed.

pages/guideforms.py
jrobbins: OK, I added JS for this and added some wording to explain what is happening.   Please take a look.

templates/blink/intent_to_implement.html
jrobbins: I thought we might be removing ot_milestone_android_start and _end.  However, now it seems like we just need to explain exactly what they mean.  I have added them here.

pages/guideforms.py
jpmedley: >Set this only if different from desktop milestone.

I don't think this line should have gone away. This is what tells you that you can override. That may not be obvious.

jrobbins: I thought that your original comment on this line of code called out two possible meanings for a missing platform-specific milestone:
desktop-applies: Missing milestone means that it is the same as the desktop milestone, or
desktop-only: Missing milestone means that the feature will not be available on that platform.
So, I implemented the JS code to auto-fill values so that the desktop-applies use case happens automatically, and the user can clear out one or more platform milestone field to indicate the desktop-only situation (or at least not available on a given platform).

jrobbins: I'm going to go ahead and merge what I have, but I am certainly open to revising it further in a new PR.

PR #1428, opened 2021-07-20
Add standard_maturity field.

jrobbins:
This should partially resolve issue #1366.

In this PR:
+ models.py define a new field standard_maturity and associated enum values and add it to the format_for_template dict.
+ guideforms.py define a django form field for standard_maturity and add the field to various forms and the feature detail page.
+ guide.py process the submitted value for standard_maturity.
+ chromedash-feature-detail.js add logic to access field that are not at the top level of the format_for_template dict.

In a follow-up PR, I will change the logic that generates the feature cards and the top of the feature detail page to use this instead of the old standardization field.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -182,6 +182,7 @@ REVIEW_STATUS_CHOICES = {
     REVIEW_NA: 'Not applicable',
     }
 
+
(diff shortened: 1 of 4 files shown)


cwilso (approved):
LGTM!

PR #1429, opened 2021-07-20
Use version=1 for /features.xml.

jrobbins:
In a recent change, we started using version=2 as the default.  However, the RSS feed served from /features.xml needs version=1, so this specifies that.


shivamag00 (approved):
LGTM

PR #1430, opened 2021-07-21
Use a single npm target to run unit tests

jrobbins:
This simplifies the commands needed to run unit tests.  With this change, the user can simply run `npm test`.

I believe that the overall command still gives an exit code of 0 when tests pass and 1 when tests fail.  So, that should work with a continuous integration tool that relies on exit status to indicate unit test success.  However I have only tested it by using `echo $?`.


package.json
```diff
@@ -10,9 +10,10 @@
     "deps": "python -m pip install -t lib -r requirements.txt --upgrade",
     "travis-deps": "pip install -t lib -r requirements.travis.txt --upgrade",
     "dev-deps": "python -m pip install --no-deps -r requirements.dev.txt",
-    "test": "curl -X POST 'http://localhost:15606/reset' && python -m unittest discover -p '*_test.py'  -b",
+    "do-tests": "curl -X POST 'http://localhost:15606/reset' && python -m unittest discover -p '*_test.py'  -b",
     "start-emulator" : "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --consistency=1.0",
     "stop-emulator" : "curl -X POST 'http://localhost:15606/shutdown'",
+    "test": "(npm run start-emulator > /dev/null 2>&1 &); sleep 3; curl --retry 4 http://localhost:15606/ --retry-connrefused; npm run do-tests; status=$?; npm run stop-emulator; exit $status",
```
KyleJu, line 16: I am not too good with script but would this run if either `start-emulator` or `do-tests` exit?

jrobbins: Yes, they run.   If a command encounters an error, it just gives a non-zero exit status.   The semi-colon means that the next command in the sequence runs regardless of exit status of the first command.   To make one conditional on success of the preceding command, I would use &&.

KyleJu: yea that makes sense. I assume we are disregarding the exit code for `(npm run start-emulator > /dev/null 2>&1 &)` because if the emulator fails to start, the tests would fail right?

KyleJu (approved):

package.json
jrobbins: That's right.  Rather than try to add shell logic to detect if the emulator fails, I just try to give it some time to start, and then go for it.  No need to check its exit status.  If it fails, then the tests will fail so the developer can look into it.  

The reason for having `curl` at all is so that the initial `sleep` can be kind of short rather than making the developer wait for a `sleep 8` every time.

shivamag00 (commented):
LGTM. 
I will try to add export and import data into emulator to this script so that dummy data does not get deleted while running the tests.

Issue #1419, opened 2021-07-15
Implement JS logic whereby toggling a star for any one instance of a feature should toggle stars for all the other occurrences of the feature.


shivamag00:

shivamag00:
After merging #1457, we will be able to test the logic implemented by #1432 for solving this.

shivamag00:
@jrobbins, please close this as #1457 was merged

shivamag00:
@jrobbins, please close this issue.

PR #1432, opened 2021-07-22
Using Webcomponent for each milestone card

### LINKED ISSUE #1416: Create the card Web Component using the existing code and use it to list the milestones on the releases page.

shivamag00:
Review Requested:- @jrobbins 

Merging this should close #1416 and #1419 

This PR is an humble attempt at 

- using web-components for each milestone card in the upcoming page
- a parent web component (`chromedash-upcoming`) manages the state for all the child web component (`chromedash-upcoming-milestone-card`)

The basic principle that I have used is "Properties Down, Events Up" i.e. parent element share data with children by binding properties to them, and child elements share data with parents by raising an event with relevant data in the detail of the event.


shivamag00:
@jrobbins, please let me know if this implementation seems to be good.

shivamag00:
@jrobbins, Besides using web-components in the new upcoming page, following changes to the UI have also been done:-

This is how OLD  RELEASES PAGE looked:-
Releases

This is how the NEW UPCOMING PAGE LOOKS:-
(without Shipping Type Annotations)
Upcoming

(with Shipping Type Annotations)
Upcoming Shipping

shivamag00:
@jrobbins, this PR is ready to be merged please review

static/js-src/upcoming-page.js
```diff
@@ -33,35 +33,32 @@ async function init() {
   document.body.classList.remove('loading');
 
   channelsArray.forEach((channel) => {
-    channels[channel].components = mapFeaturesToComponents(features[channel].filter(f =>
-      f.browsers.chrome.status.milestone_str === channels[channel].version));
+    channels[channel].components = mapFeaturesToShippingType(features[channel]);
```
shivamag00, line 36: We no longer need this filter as we are already fetching features according to the milestone

templates/upcoming.html
```diff
@@ -14,17 +14,17 @@ <h3>Release timeline</h3>
     </div>
     <!-- <paper-toggle-button> doesn't working here. Related links:
       https://github.com/PolymerElements/paper-toggle-button/pull/132 -->
-    <label><input type="checkbox" class="show-blink-checkbox">Show Blink components</label>
+    <label><input type="checkbox" class="show-blink-checkbox">Show Shipping Type</label>
```
shivamag00, line 17: Show Shipping Type replaces the Show Blink Components button. If you approve the UI, we can make the change to show the shipping types by default in a later PR.

shivamag00 (commented):
@jrobbins, please have a look at this

static/elements/chromedash-upcoming-milestone-card.js
```diff
@@ -0,0 +1,170 @@
...
+    // Discard time and time-zone information.
+    const utc1 = Date.UTC(a.getFullYear(), a.getMonth(), a.getDate());
+    const utc2 = Date.UTC(b.getFullYear(), b.getMonth(), b.getDate());
+
+    const MS_PER_DAY = 1000 * 60 * 60 * 24;
+    const daysDiff = Math.floor((utc2 - utc1) / MS_PER_DAY);
+    return {days: Math.abs(daysDiff), future: daysDiff < 1};
+  }
+
+  _computeDate(dateStr) {
```
jrobbins, line 37: Is this computing a date or just formatting it differently?

static/elements/chromedash-upcoming-milestone-card.js
```diff
@@ -0,0 +1,170 @@
...
+      feature: featureId,
+      doStar: newStarred,
+    });
+  }
+
+  _computeDaysUntil(dateStr) {
+    const today = new Date();
+    const diff = this._dateDiffInDays(new Date(dateStr), today);
+    const prefix = diff.future ? 'in' : '';
+    const days = diff.days > 1 ? 's' : '';
```
jrobbins, line 70: The variable name `days` sounds like it would be a number.  A more conventional choice for this would be to use a variable named `plural`.

Internationalization is not a priority for this project, but one tip for internationalization is to never have code that composes letters into words.  Also, zero days should have an 's'.  So, it could be like:
`const dayWord = diff.days == 1 ? 'day' : 'days';`

Likewise, you have a couple of variables here based on diff.future.  It might be more clear and a better practice to have one big conditional on that rather than compose words into a sentence:
```
if (diff.future) {
  return `in ${diff.days} ${dayWord}`;
} else {
  return `${diff.days} ${dayWord} ago`;
}

File: static/elements/chromedash-upcoming-milestone-card.js [hunk 5]
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  _objKeys(obj) {
+    if (!obj) {
+      return [];
+    }
+    return Object.keys(obj).sort();
+  }
+
+  render() {
```
jrobbins, line 82: This would probably be more clear if render() used one helper function to render the info at the top of the card, and a second helper function to render the feature list.  And, that feature list helped function might rely on a third helper function that renders one feature item along with the icons.

static/elements/chromedash-upcoming.js
```diff
@@ -0,0 +1,95 @@
...
+  static styles = style;
+
+  static get properties() {
+    return {
+      // Assigned in upcoming-page.js,
+      channels: {attribute: false},
+      showBlink: {attribute: false},
+      signedIn: {type: Boolean},
+      loginUrl: {type: String},
+      starredFeatures: {type: Object},
```
jrobbins, line 45: It would be good to indicate that this is going to hold a Set.  Maybe you could just do that in a comment.

static/elements/chromedash-upcoming.js
```diff
@@ -0,0 +1,95 @@
...
+
+  render() {
+    if (!this.channels) {
+      return html``;
+    }
+
+    return html`
+      ${['stable', 'beta', 'dev'].map((type) => html`
+        <chromedash-upcoming-milestone-card
+          channel=${JSON.stringify(this.channels[type])}
```
jrobbins, line 79: Rather than convert objects to a string set attributes, I think you can set properties directly using a leading period.  For example, I think you could have
  `.channel=${this.channels[type]}`

Of course, what you have will work and it might be easier to debug, but it seems like it could create an opportunity for future defects if there is ever any value that cannot be serialized.

See: https://lit-html.polymer-project.org/guide/writing-templates#bind-to-properties

static/js-src/upcoming-page.js
jrobbins: looks good.

static/elements/chromedash-upcoming.js
```diff
@@ -0,0 +1,95 @@
+import {LitElement, html} from 'lit-element';
+import '@polymer/iron-icon';
+import style from '../css/elements/chromedash-upcoming.css';
```
jrobbins, line 3: I'd like to eventually phase out SASS and use lit-css instead.  Do you think that the CSS for each of these components could be specified in its .js file?  It's OK to still have a SASS file for the overall page for now.

templates/upcoming.html
jrobbins: OK.

jrobbins (changes requested):

static/elements/chromedash-upcoming-milestone-card.js
shivamag00: This function was present in the OLD schedule page. I have not added anything to it. It seems to convert date string obtained from API to a date object so that it is easier to perform date operations like subtracting two dates.

static/elements/chromedash-upcoming-milestone-card.js
shivamag00: FIY, This function was being used as it is by OLD schedule page. 

Okay, I have noted these points. Will make the suggested changes, Thank you for letting me know.

shivamag00: Done

shivamag00: See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1432/commits/2fbca619fdf1f47b5e606241fd1da1b906a2cee9) commit

static/elements/chromedash-upcoming.js
shivamag00: Okay. Done. See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1432/commits/6b0ee35dea11411c0a7eb5db5c9420a0bc7cae8d) commit.

static/elements/chromedash-upcoming.js
shivamag00: The CSS file for these components import certain variables and styles from other css file. It seems possible to phase out our use of SASS and use lit-css instead. However, It would be better to have separate PRs for this KR.

static/elements/chromedash-upcoming.js
shivamag00: Okay. Thanks for letting me know. 

Done. See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1432/commits/8f2bfe4d221537bac662172628161ddf06c5d14c) commit

static/elements/chromedash-upcoming-milestone-card.js
jrobbins: OK.  I think the name might be more clear with some other verb, but I don't feel strongly about it.

static/elements/chromedash-upcoming.js
jrobbins: ok

static/elements/chromedash-upcoming.js
jrobbins: Why do some values still have single quotes around the expression?  Is that needed?

static/elements/chromedash-upcoming-milestone-card.js
shivamag00: Okay. This seems to be the better approach. 

Done. See [this](https://github.com/GoogleChrome/chromium-dashboard/pull/1432/commits/b77f2604c3870df414b27bfc7cb4e7026e96b161) commit.

static/elements/chromedash-upcoming.js
shivamag00: Oops... I had removed them. But later, when I undo-ed some of the changes, I forgot to remove them again. Thanks for pointing it out. 

Done

static/elements/chromedash-upcoming.js
```diff
@@ -0,0 +1,95 @@
...
+      ${['stable', 'beta', 'dev'].map((type) => html`
+        <chromedash-upcoming-milestone-card
+          .channel=${this.channels[type]}
+          .templateContent=${TEMPLATE_CONTENT[type]}
+          ?showdates=${SHOW_DATES}
+          .removedStatus=${REMOVED_STATUS}
+          .deprecatedStatus=${DEPRECATED_STATUS}
+          .starredFeatures=${[...this.starredFeatures]}
+          ?signedin=${this.signedIn}
+          ?showShippingType=${this.showShippingType}
```
shivamag00, line 86: I have renamed the `showBlink` variable to `showShippingType`

static/elements/chromedash-upcoming-milestone-card.js
```diff
@@ -0,0 +1,192 @@
...
+            `)}
+          </ul>
+        `)}
+      </div>
+    `;
+  }
+
+  render() {
+    return html`
+      <section class="release ${this.showShippingType ? '' : 'no-components'}">
```
shivamag00, line 184: Earlier, the true condition would return the object `nothing` . This would show up in the html as `[Object object]`. Hence, I have replaced it with empty string.

shivamag00 (commented):
@jrobbins, please review these

static/elements/chromedash-upcoming.js
```diff
@@ -0,0 +1,95 @@
...
+    channelLabel: 'Dev',
+    h1Class: 'chrome_version--dev',
+    downloadUrl: 'https://www.google.com/chrome/browser/canary.html',
+    downloadTitle: 'Download Chrome Canary',
+    dateText: 'coming',
+    featureHeader: 'Features planned in this release',
+  },
+};
+const REMOVED_STATUS = ['Removed'];
+const DEPRECATED_STATUS = ['Deprecated', 'No longer pursuing'];
```
jpmedley, line 32: Granting that I don't know how this constant is being used, 'deprecated' and 'no longer pursuing' aren't the same thing.

jpmedley (changes requested):

static/elements/chromedash-upcoming.js
shivamag00: Hmm... This constant is being used to display the tooltip and icon for 'deprecated' and 'no longer pursuing' features on the CURRENT (OLD) releases page. The developers, who developed the OLD releases page, used the same icon and tooltip for both the deprecated and no longer pursuing features. Hence, it  may be the reason for them to define the constant this way. 

Anyway, Is it correct and helpful for the users that the same icon be used for both the types of feature on the releases page?

jpmedley: I think this is a problem. If I'm scrolling through the list with everything, this icon is the only thing that tells me the status of a feature. It provides incorrect information.

@jrobbins should I open a ticket on this?

jrobbins: Would a feature that is no longer pursed belong on the roadmap view at all?

I'm thinking that this might be code that was copied-and-pasted a while back from the feature list page, where such a feature might appear in the section of the feature list for those features.  However, this code is only for the "upcoming" roadmap view.

If this logic case is needed for this page, then let's track a separate issue for it.

jpmedley: @cwilso Please comment. I believe there is a reason to display this. The one that occurs to me is to have a record of things considered and rejected so that later engineers don't re-propose the same or similar ideas. I wonder if we should go so far as to require an explanation for why something is no longer being pursued. It might save someone from starting down a dead-end road.

static/elements/chromedash-upcoming-milestone-card.js
jrobbins: Looks good.

static/elements/chromedash-upcoming.js
jrobbins: Looks good.

static/elements/chromedash-upcoming.js
jrobbins: Features that are no-longer-pursued would still be in the feature list.  I just don't see a place for them on the "Upcoming" page because they are not part of any upcoming release.

jpmedley (approved):

PR #1433, opened 2021-07-22
Access samples at the version=2 pathname.

jrobbins:
This is a follow-up to a recent CL that set version=2 as the default for the format_for_template() method.   This web component was expecting to find sample links at feature.sample_links, but in version 2 those are at feature.resources.samples.


shivamag00 (approved):
LGTM

PR #1434, opened 2021-07-22
Trying single threading to avoid CPU contention

jrobbins:
I have a theory that our current timeouts and general inconsistent latency on datastore queries is because of CPU contention while decoding (somewhat) large result sets.  This change should precent contention by basically spinning up more GAE instances, and using each GAE instance for only one HTTP request at a time.

This change should have no user visible effect, other than more consistent latency on cache misses.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,7 +1,11 @@
 runtime: python27
-threadsafe: true
+threadsafe: false
 api_version: 1
 instance_class: F4
+
+automatic_scaling:
+  min_idle_instances: 1
+  max_pending_latency: 0.2s
 
 # default_expiration: "30s"
 


jrobbins:
I already tested it briefly on prod and saw that the app seems to work as normal.  GAE seems to create about 7 instances rather than the 2 instances that we normally have, but that should not be a problem.  After this change is merged, I will deploy it and watch the GAE Error Reporting page closely to see if the timeouts decrease.

jstenback (approved):
LGTM

app.yaml
```diff
@@ -1,8 +1,12 @@
 runtime: python27
-threadsafe: true
+threadsafe: false
```
KyleJu, line 2: https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/config-files#updating_appyaml. This field is depreciated in python 3 and `All applications are presumed to be threadsafe`.

app.yaml
```diff
@@ -1,8 +1,12 @@
...
 api_version: 1
 instance_class: F4
 
+automatic_scaling:
+  min_idle_instances: 1
```
KyleJu, line 7: It means we will have a minimum one instance running at all time, rather than 0. Unsure that billing is a concern for this project?

app.yaml
jrobbins: I think of this as an experiment to resolve a current operational problem even if it cannot be the long-term solution.
Hopefully, ndb works faster in py3.  If not, we can look into a new way to access datastore.

KyleJu: Ack

KyleJu (approved):

app.yaml
jrobbins: It is not a big concern.  Our GAE bill has been about $11/day.  It recently went up to $17/day right about the time that we started using Cloud NDB.  I'll keep an eye on new daily charges, but I expect that it won't be too much higher than what we are already paying.

PR #1437, opened 2021-07-27
Fix the XML feed to match version=2

jrobbins:
This is another instance where existing code assumed the version=1 format, so I updated it to match the version=2 format.


shivamag00 (approved):
LGTM

PR #1439, opened 2021-07-27
Fix another XML feed defect and add tests.

jrobbins:
Update more of the RSS feed generation logic to work with version=2.

I should have caught this defect at the same time that I made my other recent fix, but I didn't realize we were lacking test coverage there, and my local data didn't have any matching features.  So, I have added a test.

I believe that we can phase out version=1 now because it is not used anywhere, however I will do that in a separate PR to keep each PR focused.


shivamag00 (approved):
LGTM

PR #1443, opened 2021-07-29
Update feature detail page to match format_for_template(vesion=2).

jrobbins:
This is more follow-up to the change that makes version=2 the default.


shivamag00 (approved):
LGTM

PR #1447, opened 2021-08-03
Try logging to stdout

jrobbins:
I couldn't figure out what was happening with Cloud Logging.

On reading the AppEngine py2 to py3 upgrade guide again, I notice that there is an alternative available in both py2 and py3, which is simply to write to stdout.    That has some disadvantages in that it loses the source file and line number, and everything seems to get logged at the warning level.  However, it is really simple, so maybe we could just try it for now and come back to it once we are on py3.  I tested it on staging and it seems to nest the log lines under the requests.

### DIFF
--- a/api/accounts_api.py
+++ b/api/accounts_api.py
@@ -16,11 +16,10 @@
 from __future__ import division
 from __future__ import print_function
 
-import logging
-
 from google.cloud import ndb
 
 from framework import basehandlers
+from framework import logging
 from framework import permissions
 from framework import ramcache
 from internals import models
--- a/api/approvals_api.py
+++ b/api/approvals_api.py
@@ -16,9 +16,8 @@
 from __future__ import division
 from __future__ import print_function
 
-import logging
-
 from framework import basehandlers
+from framework import logging
 from framework import permissions
 from internals import approval_defs
 from internals import models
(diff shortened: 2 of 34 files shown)


jrobbins:
I'm not sure that this would capture any lines written by libraries that use the standard python logging.  
And, I am not sure that it will log exceptions in py3.  We could test that by making a tiny py3 GAE app that causes an exception.

jrobbins:
Actually, I don't even think that we need this.  The standard logger will write to stdout itself, so we don't need to use print().  We just need to configure the standard logger to print INFO level entries because it defaults to only printing WARN and higher.

So, I think this could be reduced to just one line in settings.py or basehandlers.py:
logging.basicConfig(level=logging.INFO)

I tested it out with a py3 hello-world application:
https://github.com/jrobbins/gae-py3/blob/main/main.py

It logs to the terminal when running locally.  And, when running in GAE for py3, it produces log entries that are readable but lack nesting or severity indicators.  When I hit the /crash page, the exception shows up.  So, that seems good enough.

image

I'll send a separate PR for that one-line change.

jrobbins:
This is superseded by https://github.com/GoogleChrome/chromium-dashboard/pull/1449

PR #1449, opened 2021-08-04
Adjust logging level for py3.

jrobbins:
Try a super-simple way to get logging to work well enough for GAE py3.

The standard logging library will write to stdout, and GAE captures any lines written to stdout as log entries.  They are not nested under the request and they do not have severity info parsed, but it is good enough for debugging.

This line is needed because the standard python logging library defaults to only outputting log entries with severity WARN or higher, but we want INFO to match how it worked under py2.  It is not set when running unit tests to avoid excessive output.


KyleJu (approved):
Yea that could work! If we realize the need for nested entries/structured logging, we can still set up separately

PR #1450, opened 2021-08-04
Display the standards maturity value, or backfill based on standardization

jrobbins:
In this PR:
+ Define a mapping from old standardization to standard_maturity
+ Update templates and components

Also, eliminate a color computation function that used to set the overall color of a feature.  It has not been used in months.


cwilso (approved):
LGTM

PR #1453, opened 2021-08-05
Avoid "gapi not defined" JS error.

jrobbins:
While debugging another problem, I noticed that the JS console includes lines saying "gapi not defined".   I believe that this is because this line runs startApp() before gapi is loaded.  I don't think that this <script> is needed at all since startApp() is run as part of the gapi loading.


shivamag00 (approved):
LGTM

PR #1464, opened 2021-08-13
Map feature to Shipping Milestone on backend instead of frontend

shivamag00:
Review Requested:- @jrobbins 

In this PR,

- Map feature to Shipping Milestone will be done on backend instead of frontend
- If a milestone release has no features, a string stating the same will be displayed on the frontend
- Fixed failing tests

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -955,17 +955,34 @@ class Feature(DictModel):
     if milestone == None:
       return None
 
+    all_features = {}
+    all_features[IMPLEMENTATION_STATUS[ENABLED_BY_DEFAULT]] = []
+    all_features[IMPLEMENTATION_STATUS[DEPRECATED]] = []
+    all_features[IMPLEMENTATION_STATUS[REMOVED]] = []
+    all_features[IMPLEMENTATION_STATUS[INTERVENTION]] = []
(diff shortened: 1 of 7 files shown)


internals/models.py
```diff
@@ -955,38 +955,62 @@ def get_in_milestone(
...
+    
+    # Push feature to list corresponding to the implementation status of feature in queried milestone
     for feature in desktop_shipping_features:
-      if feature.impl_status_chrome not in {ENABLED_BY_DEFAULT, DEPRECATED, REMOVED}:
-        if feature.feature_type == FEATURE_TYPE_DEPRECATION_ID and Feature.dt_milestone_desktop_start != None:
-          feature.impl_status_chrome = DEPRECATED
-        if feature.feature_type == FEATURE_TYPE_INCUBATE_ID:
-          feature.impl_status_chrome = ENABLED_BY_DEFAULT
+      if feature.impl_status_chrome == ENABLED_BY_DEFAULT:
+        all_features[IMPLEMENTATION_STATUS[ENABLED_BY_DEFAULT]].append(feature)
```
jrobbins, line 975: I was thinking that the features would be grouped by the milestone that matched the requested milestone, not by the current status of the feature.  However, let's go ahead with what you have here.  We can always come back to it.

jrobbins (approved):

Issue #1461, opened 2021-08-13
Not all crbug components are listed.


mfalken:
**Describe the bug**
The crbug components are not all listed.

**To Reproduce**

1. Go to https://chromestatus.com/guide/edit/5355965538893824
2. Set Blink Component to Internals>Preload>Prerender to match the bug https://crbug.com/1126305

**Expected behavior**
Component is changed.

**Additional context**
Error message: Select a valid choice. Internals>Preload>Prerender is not one of the available choices.

jpchase:
Likely duplicate of #496 or #565. Short version is that the list of components available for selection is not the full list of components in crbug.

Issue #565, opened 2018-07-24
Missing component "Blink>ReportingObserver"


jpchase:
The component "Blink>ReportingObserver" is missing from the list when editing a entry.

As in #496, this component is missing from https://blinkcomponents-b48b5.firebaseapp.com/blinkcomponents, and it's underlying source: https://storage.googleapis.com/chromium-owners/component_map.json.

Upon investigation, I see that component_map.json is populated by scraping OWNERS files in the Chromium code base. See https://cs.chromium.org/chromium/src/tools/checkteamtags/extract_components.py.

This is where we run into a problem for "Blink>ReportingObserver". The OWNERS file only supports one assigned Blink component per directory. The ReportingObserver code lives in a shared directory. We can't reasonably change the existing component mapping. In the longer term, it probably makes sense to move the ReportingObserver to a separate directory.

Regardless, it seems like a problem that the list of available Blink components in chromestatus relies on source code organization in Chromium. I would expect that if the components are defined in crbug.com, then they should be made available for selection when editing an entry.

aglaforge:
Do we have any ways to query into Monorail to get a list of components?  The indirection seems to be problematic here.

jpchase:
Yes, it's possible to use the Monorail API to get a list of components. I've been able to get that working in another project.
@jrobbins I can point you at the internal code (although it's in Go, I think there are Python examples around).

Issue #496, opened 2017-08-25
"Blink>Media>Video" is missing


beaufortfrancois:
https://www.chromestatus.com/feature/5682169347309568 tells me "Blink>Media>Video" is not a valid choice while it was clearly one at some point.

Any idea what happened?

For now, I've updated it to "Blink>Media" but it would be nice to have it fixed.

ebidel:
Some used it on the https://bugs.chromium.org/p/chromium/issues/detail?id=310450. We pull the first tag from the bug if you save the feature with the generic "Blink" component when it's first created.

Blink>Media>Video might be new or non-standard. It's not listed in blinkcomponents-b48b5.firebaseapp.com/blinkcomponents, which is where that list comes from.

LMK if you run into others like this. I'm making a list for the TPMs :)

beaufortfrancois:
Thanks @ebidel.
Where is the source code for blinkcomponents-b48b5.firebaseapp.com/blinkcomponents and where I can contribute?

beaufortfrancois:
I can see "Blink>Media>Video" in component drop list at https://bugs.chromium.org/p/chromium/issues/list

beaufortfrancois:
@ebidel I'm still running in the same issue with `Blink>Media` and now `Blink>Media>PictureInPicture`.
Let me know where I can update https://blinkcomponents-b48b5.firebaseapp.com/blinkcomponents if possible.

Thank you!

ebidel:
You should ask the owners to add them to https://storage.googleapis.com/chromium-owners/component_map.json.

jpmedley:
@jrobbins These still seem to be issues.

jpmedley:
Not sure when this was fixed (#1466?).

PR #1466, opened 2021-08-13
Updated hardcoded list of crbug components.

jrobbins:
This should resolve issue #1461, issue #496, and issue #565.

In this PR:
+ Update the hard-coded list of blink components based on info that I got from a monorail API call.
+ Eliminate wf_components because they were not used anywhere.
+ A few whitespace changes crept in


jrobbins:
The diff of hack_components.py is hard to view on GitHub because it looks like every line changed even though it did not.  I think the problem may have been a conversion of tabs to spaces.   You can view a better diff by using the gear menu just above the diffs and choosing to ignore whitespace.

jpmedley:
@jrobbins If the source of this information in code search were listed in a file comment, it would be easy for anyone on the team to open a PR to update this.

jpchase (approved):
LGTM.

jpchase:
> @jrobbins If the source of this information in code search were listed in a file comment, it would be easy for anyone on the team to open a PR to update this.

As I understand, the source of this data is an API call to monorail. That is, not easily found in code search.  Unless the API call is easy for anyone to do, perhaps it best to leave out the comment.

jrobbins:
This time was still a manual hack.  I literally viewed the monorail API response for the "project config" in chrome developer tools and then edited it down to just the list of the names of components.  And, I made decisions about which top-level components to include or exclude.  It's not a process that we should repeat.  I'll track a new issue to implement a new automated way to do it, which will involve getting allow-listed for the monorail API.  Eric's old node.js server seems to be completely down at this point, always returning status 410 (Gone).

PR #1467, opened 2021-08-14
Added Aria-Label attributes

### LINKED ISSUE #1421: Add ARIA attributes to side-scrolling buttons and ensure that the flow of the page makes sense when using a screen reader.

shivamag00:
Review Requested:- @jrobbins 

Merging this should close #1421.

In this PR,

- Added some aria-labels for the newly added elements.

### DIFF
--- a/templates/upcoming.html
+++ b/templates/upcoming.html
@@ -21,14 +21,15 @@
 {% block content %}
   <section id="releases-section">
     <div class="timeline-controls">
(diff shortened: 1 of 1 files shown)


shivamag00:
Also, the order in which the screen reader reads makes sense i.e. it would first read a milestone and then the features within that milestone.

jrobbins (approved):

PR #1469, opened 2021-08-14
Implement milestone visualization

jrobbins:
This implements a simple gantt-like set of bars for the milestones of an individual feature on the feature detail page.

Details:
* If the feature has no milestones, a message is shown instead.
* Rows are shown for each platform that has milestones.
* When an OT is planned on a platform, the background bar has two color to visually indicate the progression
* Diamond shapes are used for the milestones, with milestone numbers centered in them
* Any milestone with the same number as the current stable milestone will have a dark border to indicate "you are here"
* Missing milestones are simply now shown, allowing the background image to show through, indicating undetermined future or past
* Within a given column, distinct milestone numbers are offset from each other to avoid misleading alignment
* Simple markup allows screen reader to read milestone numbers, but there is probably room for improvement


static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,198 @@
...
+      .offset_3 { padding-left: 24px; }
+    `];
+  }
+
+  renderDevTrial(milestone) {
+    if (!milestone) return nothing;
+    return html`
+      <div class="diamond dev_trial
+                  ${milestone == this.stableMilestone ? 'stable' : ''}"
+       title="First milestone with this feature available to developers"
```
jpmedley, line 101:
```suggestion
       title="First milestone with this feature available to developers behind a flag."
```

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,198 @@
...
+      ><span>${milestone}</span></div>`;
+  }
+
+  renderOriginTrial(milestone) {
+    if (!milestone) return nothing;
+    return html`
+      <div class="diamond origin_trial
+                  ${milestone == this.stableMilestone ? 'stable' : ''}"
+       title="First milestone with this feature enabled on specific origins"
```
jpmedley, line 110:
```suggestion
       title="First milestone with this feature enabled on participating origins."
```

jpmedley: @jpchase His phrasing avoids the trap of a circular definition. I think 'specific origins' is a bit vague. What do you think?

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,198 @@
...
+      ><span>${milestone}</span></div>`;
+  }
+
+  renderShipping(milestone) {
+    if (!milestone) return nothing;
+    return html`
+      <div class="diamond shipping
+                  ${milestone == this.stableMilestone ? 'stable' : ''}"
+       title="Milestone with this feature enabled by default"
```
jpmedley, line 119:
```suggestion
       title="Milestone with this feature enabled by default."
```

static/elements/chromedash-gantt.js
jpmedley: Would it be possible to put a conditional in here such that if the milestone has not been reached, the line would read, "First expected milestone..."? The same would apply to the origin trial and shipping.

jpmedley (changes requested):

jpmedley:
🎉

static/elements/chromedash-gantt.js
jrobbins: Is it correct to use a period?  These phrases are not complete sentences.

jpmedley: Not necessarily. This is hover text, right? What does your other hover text do? It should be consistent.

jrobbins: The other hover text does not include periods.  E.g., title="Adds a new feature to the site"
I'll adopt all your wording changes.

static/elements/chromedash-gantt.js
jrobbins: Done.

static/elements/chromedash-gantt.js
jrobbins: I went with "participating".

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,206 @@
...
+      }
+      .offset_0 { padding-left: 0px; }
+      .offset_1 { padding-left: 8px; }
+      .offset_2 { padding-left: 16px; }
+      .offset_3 { padding-left: 24px; }
+    `];
+  }
+
+  firstPhrase(milestone) {
+    if (milestone <= this.stableMilestone) {
```
KyleJu, line 97: Haven't seen this `<=` expression before. What does this mean?

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,206 @@
...
+  render() {
+    let f = this.feature;
+    let dtDesktop = f.dt_milestone_desktop_start;
+    let otDesktop = f.ot_milestone_desktop_start;
+    let shipDesktop = f.browsers.chrome.desktop;
+    let dtAndroid = f.dt_milestone_android_start;
+    let otAndroid = f.ot_milestone_android_start;
+    let shipAndroid = f.browsers.chrome.android;
+    let dtIos = f.dt_milestone_ios_start;
+    let otIos = null;
```
KyleJu, line 167: `otIos` and `otWebview` are null but never assigned to a real value in the code

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,206 @@
...
+         </li>
+         <li class="${'offset_' + shipOffset}">
+           ${this.renderShipping(shippingMilestone)}
+         </li>
+       </ul>
+    `;
+  }
+
+  render() {
+    let f = this.feature;
```
KyleJu, line 159: All of these declarations are never reassigned. Maybe use `const` rather than `let`

static/elements/chromedash-gantt.js
jpmedley: The listed milestone of the feature is less than the currently available version of Chrome. If the proposed milestone hasn't come, we want to communicate that the listed milestone is the target, but not guaranteed. The use of the returned string is on lines 109, 118, and 127.

jrobbins: The `<=` operator in JS is just a plain old `less than` comparison.  It is not related to the arrow operator or anything like that, even though it looks like a backwards one.

static/elements/chromedash-gantt.js
jrobbins: I added a comment to explain that those two platforms do not support origin trials.  I defined them as variables rather than using null in the templates below so that the fact that they were always null would be more obvious and the set of variables more consistent.  Let me know if you think using in-line null values would be better.

static/elements/chromedash-gantt.js
jrobbins: Good point.  Done.

static/elements/chromedash-gantt.js
KyleJu: Yea I was totally confused it with the arrow operator...

KyleJu: Also another note that when `this.stableMilestone` is `null`, null will evaluate to 0, which I guess work out in this use case.

static/elements/chromedash-gantt.js
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      return 'First expected milestone with this feature';
+    }
+  }
+
+  renderDevTrial(milestone) {
+    if (!milestone) return nothing;
+    return html`
+      <div class="diamond dev_trial
+                  ${milestone == this.stableMilestone ? 'stable' : ''}"
```
KyleJu, line 108:
```suggestion
                  ${milestone === this.stableMilestone ? 'stable' : ''}"
```
nit

KyleJu (approved):
LGTM

static/elements/chromedash-gantt.js
jrobbins: Fixed.

jpmedley:
🎉

PR #1474, opened 2021-08-18
Update privacy link to skip a redirect.

jrobbins:
While working through our system test spreadsheet, I was doing the row about footer links and I noticed that the URL we were using takes the user through an interstitial page.   So, I just used the destination URL.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -14,7 +14,6 @@ indexes:
 - kind: "Feature"
   properties:
   - name: "shipped_milestone"
-  - name: "shipped_android_milestone"
     direction: desc
   - name: "name"
 - kind: "Feature"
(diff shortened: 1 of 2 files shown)


index.yaml
```diff
@@ -14,12 +14,12 @@ indexes:
 - kind: "Feature"
   properties:
   - name: "shipped_milestone"
+  - name: "shipped_android_milestone"
```
KyleJu, line 17: Should this be updated?

jrobbins: Sigh.  No, that is something that should be part of a different PR.  I have reverted the change via the GitHub UI.  GitHub still shows a diff, but it is now removing the line that I had added.  So, I think it is fine.  I tried pulling this PR branch to my workstation and it seems right.

KyleJu (approved):

jrobbins:
FYI, I reverted and relanded this PR to avoid that index.yaml change.

PR #1477, opened 2021-08-18
GitHub Actions CI

### LINKED ISSUE #1360: Get our continuous integration (CI) system working again
The old Travis-CI setup that we inherited has stopped working.

Research the options available today (e.g., GitHub Actions, Google Cloud Build, Travis, Circle, etc.)  choose one and get it working.

Right now we only use CI to run unit tests for github PRs.    In the future, we will also want to run javascript unit tests for PRs.   And, of course, we need Py2 for now and Py3 later, ideally with the possibility of running tests in both during the final stages of our upgrade to py3..

KyleJu:
Fix #1360.

Implements a hacky workaround due to https://stackoverflow.com/questions/42697026/install-google-cloud-components-error-from-gcloud-command. The `google-cloud-sdk` installed through any sort of package managers, e.g. apt-get, cannot run `cloud-datastore-emulator`. This PR instead manually downloads `google-cloud-sdk` and substitutes the existing `google-cloud-sdk` package on Ubuntu.


KyleJu:
@jrobbins I cannot get `npm run test` to work locally either. Any specific instructions to set up emulator?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          npm install -g gulp
+          npm install
+          npm run ci-dep
+
+      - name: lint
+        run: npm run lint
+
+      - name: test
+        run: npm run test
```
KyleJu, line 41:
```suggestion
        run: npm test
```

jrobbins:
> @jrobbins I cannot get `npm test` to work locally either. Any specific instructions to set up emulator? When I run locally, I got
> 
> `Failed to connect to localhost port 15606`

I think you need to add the emulator as a gcloud component.    You can check if you already have it by doing `gcloud components list`.   I believe that I installed it with the command:
`gcloud components install app-engine-python app-engine-python-extras cloud-datastore-emulator`

I think that dependency is undocumented.  It should be added to README.md.  It was not added to the old .travis.yml because travis has not been working lately and I didn't think through the fact that it was needed there.

KyleJu:
> > @jrobbins I cannot get `npm test` to work locally either. Any specific instructions to set up emulator? When I run locally, I got
> > `Failed to connect to localhost port 15606`
> 
> I think you need to add the emulator as a gcloud component. You can check if you already have it by doing `gcloud components list`. I believe that I installed it with the command:
> `gcloud components install app-engine-python app-engine-python-extras cloud-datastore-emulator`
> 
> You can test it by running `npm run start-emulator`. I believe that the emulator requires Java, but I didn't do anything to install java.
> 
> I think that dependency is undocumented. It should be added to README.md. It was not added to the old .travis.yml because travis has not been working lately and I didn't think through the fact that it was needed there.

Turns out it also requires installing `gcloud Beta Commands`. I will explicitly add that to the CI

KyleJu:
@jrobbins Ready for review

ci/install_google_cloud_sdk.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-# Note: We now install and update the Google Cloud SDK in .travis.yml.
+# Note: We now install and update the Google Cloud SDK in ci.yml.
```
jrobbins, line 4: Do we need this file or this whole directory now?  I don't see how it is used.  I think it might not have even been used under travis.

jrobbins (approved):
Thanks for doing this.  It will be so nice to have tests run for each PR again.

LGTM after considering one change for this or a follow-up PR.

ci/install_google_cloud_sdk.sh
KyleJu: This file was used for automatic staging deployment for Travis. I thought we would keep it just in case we want to deploy staging instances again. wdyt

jrobbins: OK, it would be good to eventually have automatic deployment to staging, especially if we had automated system tests.  I assume that we could limit such deployments to only PRs that are sent by users on a short allow-list.   I am not sure that we would need this particular file, but we can certainly keep it around.  Or, if we delete it, it would not be hard to undelete.

KyleJu: Yea I will delete this for now

Issue #1496, opened 2021-08-27
Chrome 93 is gone from the drop down menu


tomayac:
**Describe the bug**

![2454E03D-B47B-4AB1-B5F8-FDCFEAEC601D](https://user-images.githubusercontent.com/145676/131085157-a8b1db20-46be-4feb-bdf7-eba482540c5f.png)

PR #1505, opened 2021-08-30
Show dates on schedule "gap" column.

jrobbins:
This PR finishes resolving issue #1496.

In this PR:
* When we need to display a "gap" release, fetch the milestone dates for that milestone number and insert that info into he CHANNELS object that is used by chrome dash-schedule.  
* Display a blank space above the Chrome milestone number so that it lines up with other columns.
* Don't use English prepositions "was" or "between" because they don't add much value and choosing the correct one would require logic that is not worth adding.

### DIFF
--- a/static/js-src/schedule-page.js
+++ b/static/js-src/schedule-page.js
@@ -19,14 +19,23 @@ async function init() {
   const features = await featuresPromise;
   let columns = ['stable', 'beta', 'dev'];
 
+  let gapMilestone = null;
+
   // If there is a gap between stable, beta, or dev then display "gap" column.
   if (CHANNELS['beta'].version > CHANNELS['stable'].version + 1) {
-    CHANNELS['gap'] = {version: CHANNELS['stable'].version + 1};
+    gapMilestone = CHANNELS['stable'].version + 1;
     columns = ['stable', 'gap', 'beta'];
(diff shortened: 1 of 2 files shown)


jrobbins:
Here's how this looks visually:
image

static/js-src/schedule-page.js
```diff
@@ -19,14 +19,23 @@ async function init() {
...
-    CHANNELS['gap'] = {version: CHANNELS['stable'].version + 1};
+    gapMilestone = CHANNELS['stable'].version + 1;
     columns = ['stable', 'gap', 'beta'];
   } else if (CHANNELS['dev'].version > CHANNELS['beta'].version + 1) {
-    CHANNELS['gap'] = {version: CHANNELS['beta'].version + 1};
+    gapMilestone = CHANNELS['beta'].version + 1;
     columns = ['stable', 'beta', 'gap'];
   }
+  if (gapMilestone) {
+    const gapInfo = await window.csClient.getSpecifiedChannels(
```
KyleJu, line 33: The similar occurence of `getSpecifiedChannels` is used in a [try and catch](https://github.com/GoogleChrome/chromium-dashboard/blob/866dc3c43ff3d9432b6115a6125260cb8ad30df5/static/elements/chromedash-upcoming.js#L191)

KyleJu (approved):

Issue #1507, opened 2021-08-31
"Estimated milestone" appears when there are no milestones


jpmedley:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
Update notification emails show an "Estimated milestone" heading when there are no estimated milestones.

![image](https://user-images.githubusercontent.com/8204171/131552176-f2f7d5a6-3e1c-4b7d-a066-e173ed05c6a3.png)

**Expected behavior**
Heading should show "None yet entered".

**Additional context**
I considered suggesting removal of the heading, but I think showing "None yet entered" will serve as a reminder of future work.

PR #1508, opened 2021-08-31
Show a message when a feature has no milestones

jrobbins:
This should resolve issue #1507.

This change impacts that feature change notification emails and also the intent email text that feature owners are supposed to copy and paste.  When a feature has no milestones specified, there will now be a message that says "No milestones specified".


jrobbins:
BTW, I went with "No milestones specified" rather than the suggested "None yet entered" because I already use "No milestones specified" in the chromedash-gantt element when there are no milestones.   Also, just in case this line of text is ever separated from the surrounding context, it should still be clear.

Sorry about the super-long condition line.  Django templates do not allow line breaks inside `{% ... %}` tags.

jpmedley:
I was more interested in the heading being kept than I was the actual message that we show. So no problem.

jpmedley (approved):

Issue #1494, opened 2021-08-26
Gantt charts should list end versions for origin trials


jpmedley:
**Is your feature request related to a problem? Please describe.**
Sometimes there is a gap between the end of an origin trial and shipping of a feature. The current gantt chart implementation gives the impression of no gap.

Consider this chart for JS [Self-Profiling API](https://www.chromestatus.com/feature/5170190448852992). This it appears that the origin trial ran from 78 to 93.

![image](https://user-images.githubusercontent.com/8204171/130890907-6eeafe6b-4dfd-431f-85d3-1b2572f1fe18.png)

A look at the underlying data shows that it ran from 78 to 80, a gap of 13 releases.

![image](https://user-images.githubusercontent.com/8204171/130890943-69c51ca3-4a41-46db-b8fc-80bd13526338.png)

**Describe the solution you'd like**
The gantt chart needs to show the precise range of an origin trial.

**Additional context**
The origin trial dashboard already does this. If the long-term plan is to integrate the dashboard's functionality into Chrome Status, we will eventually need to show this too.

![image](https://user-images.githubusercontent.com/8204171/132883614-a726ac3c-da4d-45f5-a8b0-081990b8ab83.png)

jpmedley:
#1509 doesn't really solve the problem. 

* Showing a gap doesn't tell you if a feature was missing completely for a version or more.
* The gab also doesn't tell you when a current origin trials is scheduled end. (I've added additional context to the but description to show how the OT dashboard does this. See above.)

jpmedley:
Duplicate of #1480

PR #1509, opened 2021-08-31
Add OT gap to milestone visualization.

jrobbins:
This should resolve issue #1494.

I have already committed a new background image that shows the dark-blue area for origin trails being shorter than the overall gantt bar.

In this PR:
+ Add a new JS function chooseBackground() that chooses among the three backgrounds that we have:
 - without-ot is all light blue.  It is used when there is no OT, or if the OT start and ends in the same milestone.
 - with-ot is half light blue and then half dark blue.  This is now used only for gapless OTs that extend all the way up to one milestone before shipping.
 - with-ot-gap is mostly light blue and then only about 1/4 dark blue.  This is used for the remaining cases where the OT explicitly ends well before the shipping milestone, or if there is simply no end specified.

### DIFF
--- a/static/elements/chromedash-gantt.js
+++ b/static/elements/chromedash-gantt.js
@@ -52,6 +52,9 @@ class ChromedashGantt extends LitElement {
       .with-ot {
         background-image: url(/static/img/gantt-bar-with-ot.png);
       }
+      .with-ot-gap {
+        background-image: url(/static/img/gantt-bar-with-ot-gap.png);
+      }
       .platform {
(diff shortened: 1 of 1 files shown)


jrobbins:
Here is what it looks like on some local test data that has both a gapless OT, a regular (with gap) OT, and no OT.
image

jpmedley:
I approve of this being done, especially so quickly. Thank you, Jason. I shouldn't approve this because it's code. @KyleJu

KyleJu (approved):

PR #1517, opened 2021-09-07
Fix docs & samples links on /feature/ pages

jeffposnick:
It looks like #1443 introduced some changes to account for a new data schema, but there are a couple of places that did not get updated to account for the fact that `feature.doc_links` and `feature.sample_links` is now `feature.resources.docs` and `feature.resources.samples`.

This PR remediates that, ensuring that the docs and samples links show up on the standalone feature pages (and the `intent_to_implement.html` template).

Please note that I haven't run a local copy of this project in a while, so I haven't actually tested these changes out—please confirm that the change works before merging!

### DIFF
--- a/templates/blink/intent_to_implement.html
+++ b/templates/blink/intent_to_implement.html
@@ -34,7 +34,7 @@
   <br><br><h4>Specification</h4>
   {{feature.standards.spec|urlize}}
 
-  {% if feature.doc_links %}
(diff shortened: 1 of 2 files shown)


jrobbins (approved):

jrobbins:
Thanks for catching this mistake.

jrobbins:
This is live now.

PR #1518, opened 2021-09-07
Fix logical-and syntax.

jrobbins:
This is a follow-up to 
https://github.com/GoogleChrome/chromium-dashboard/pull/1517


jrobbins:
I'm going to TBR this because the change is very straight forward and landing it will unblock the deployment of the original fix.

jeffposnick:
👍

PR #1519, opened 2021-09-08
First attemp at a unit test for templates.

jrobbins:
We had a couple of defects recently that were in .html files.   They slipped through because we don't have any unit tests for templates.  Here's one way we could start implementing some of those tests.

The main thing that this kind of test proves is simply that the template can be rendered without hitting an error in the django template syntax.  E.g., JS-like '&&' is wrong because django uses python-like 'and' for the logical-AND operator.

The main idea is to isolate the template itself rather than running the whole request processing code path.  Ideally, each test would just create a fake (but fully formed) template_data dictionary, and then we can render it to a string and check that the that rendering step worked.   However, writing a constant for a fully formed template_data value seems like it might be kind of tedious, so here I have called the page handler to create an initial template_data, which unit tests can then modify.

Verifying that rendering worked is done in a very simple way by looking for substrings.  In the future, we could get fancy by parsing the generated HTML to ensure that it is valid, that href attributes occur within an <a> tag, etc.

### DIFF
--- a/pages/featuredetail_test.py
+++ b/pages/featuredetail_test.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 
 import testing_config  # Must be imported first
 
+import os
 import flask
 import werkzeug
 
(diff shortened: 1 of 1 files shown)


jrobbins:
Let me know what you think.  The alternative would be to test the overall `get()` method, which does the template rendering as part of the whole request processing code path.  That would probably make the unit tests shorter, but it could also make test failures harder to debug because each such unit test would be executing a lot more code.  That would be like a simple integration test, which would also be useful: it comes down to choosing a strategy that we think would be productive for writing tests and effective at catching defects.

KyleJu:
> Let me know what you think. The alternative would be to test the overall `get()` method, which does the template rendering as part of the whole request processing code path. That would probably make the unit tests shorter, but it could also make test failures harder to debug because each such unit test would be executing a lot more code. That would be like a simple integration test, which would also be useful: it comes down to choosing a strategy that we think would be productive for writing tests and effective at catching defects.

I am a little bit confused about the `get()` method. It is almost the same as testing the `FeatureDetailHandler` itself? From my experience with testing templates in wpt.fyi, there are three things we try to cover:

- Testing if whole rendering succeeds or not, by checking the respond code. Basically `FeatureDetailHandlerTest`
- Template data itself, testing `get_template_data` method
- Template html, by checking if the template_text contains certain substrings.

It looks like `featuredetail_test.py` pretty much covered all of them and I think it is good enough for a first attempt. 

IIUC, to catch " template can be rendered without hitting an error in the django template syntax. E.g., JS-like '&&' is wrong because django uses python-like 'and' for the logical-AND operator", an integration test like `FeatureDetailHandlerTest` is good enough?

KyleJu (approved):

jrobbins:
> > Let me know what you think. The alternative would be to test the overall `get()` method, which does the template rendering as part of the whole request processing code path. That would probably make the unit tests shorter, but it could also make test failures harder to debug because each such unit test would be executing a lot more code. That would be like a simple integration test, which would also be useful: it comes down to choosing a strategy that we think would be productive for writing tests and effective at catching defects.
> 
> I am a little bit confused about the `get()` method. It is almost the same as testing the `FeatureDetailHandler` itself? From my experience with testing templates in wpt.fyi, there are three things we try to cover:
> 
> * Testing if whole rendering succeeds or not, by checking the respond code. Basically `FeatureDetailHandlerTest`
> * Template data itself, testing `get_template_data` method
> * Template html, by checking if the template_text contains certain substrings. (or through parsing HTML)
> 
> It looks like `featuredetail_test.py` pretty much covered all of them and I think it is good enough for a first attempt.
> 
> IIUC, to catch " template can be rendered without hitting an error in the django template syntax. E.g., JS-like '&&' is wrong because django uses python-like 'and' for the logical-AND operator", an integration test like `FeatureDetailHandlerTest` is good enough?

Actually, `FeatureDetailHandlerTest` does not test rendering at all.  Those tests only run the `get_template_data()` which returns a python dict of template data but does not do anything with the template.  Only with the addition of `FeatureDetailTemplateTest` do the tests start dealing with templates at all.

I'll go with this and see how it works out as I add a few more tests.

PR #1533, opened 2021-09-16
Implement myfeatures page boxes for starred-by:me and owner:me

jrobbins:
This is progress on implementing the myfeatures page.

In this PR:
* Make the chromedash-feature-table element fetch a feature list from the server, and display "Loading..." or a message when there are no matching features.
* Add logic to the features API to accept those queries
* Add `internals/search.py` to process those queries
* Add associated unit tests

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -854,7 +854,6 @@ class Feature(DictModel):
         if feature_id in result_dict]
     return result_list
 
-
   @classmethod
   def get_chronological(
       self, limit=None, update_cache=False, version=None, show_unlisted=False):
--- a/internals/models.py
+++ b/internals/models.py
@@ -776,7 +776,10 @@ class Feature(DictModel):
 
       # TODO(ericbidelman): Support more than one filter.
       if filterby:
-        query = query.filter(Feature.category == filterby[1])
+        if filterby[0] == 'category':
+          query = query.filter(Feature.category == filterby[1])
+        elif filterby[0] == 'owner':
+          query = query.filter(Feature.owner == filterby[1])
 
       features = query.fetch(limit)
 
@@ -824,6 +827,33 @@ class Feature(DictModel):
         ramcache.set(KEY, feature)
 
     return feature
+
+  @classmethod
+  def get_by_ids(self, feature_ids, update_cache=False):
+    result_dict = {}
+    futures = []
+
+    for feature_id in feature_ids:
+      KEY = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, feature_id)
+      feature = ramcache.get(KEY)
+
+      if feature is None or update_cache:
+        futures.append(Feature.get_by_id_async(feature_id))
+
+    for future in futures:
+      unformatted_feature = future.get_result()
+      if unformatted_feature and not unformatted_feature.deleted:
+        feature = unformatted_feature.format_for_template()
+        feature['updated_display'] = unformatted_feature.updated.strftime("%Y-%m-%d")
+        feature['new_crbug_url'] = unformatted_feature.new_crbug_url()
+        ramcache.set(KEY, feature)
+        result_dict[unformatted_feature.key.integer_id()] = feature
+
+    result_list = [
+        result_dict.get(feature_id) for feature_id in feature_ids
+        if feature_id in result_dict]
+    return result_list
+
 
   @classmethod
   def get_chronological(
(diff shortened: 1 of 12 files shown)


jrobbins:
This is what it looks like with some quick test data on my local server:
image

internals/models.py
```diff
@@ -825,6 +828,33 @@ def get_feature(self, feature_id, update_cache=False):
...
+        feature['new_crbug_url'] = unformatted_feature.new_crbug_url()
+        ramcache.set(KEY, feature)
+        result_dict[unformatted_feature.key.integer_id()] = feature
+
+    result_list = [
+        result_dict.get(feature_id) for feature_id in feature_ids
+        if feature_id in result_dict]
+    return result_list
+
+
```
KyleJu, line 857:
```suggestion
```

internals/search.py
```diff
@@ -0,0 +1,63 @@
...
+from framework import users
+from internals import models
+from internals import notifier
+
+
+def process_pending_approval_me_query():
+  """Return a list of features needing approval by current user."""
+  # TODO(jrobbins): write this
+  return []
+
```
KyleJu, line 30: Should be single space between functions

KyleJu (commented):
I need a bit more to understand changes. Will try to get it done by Friday

internals/search.py
jrobbins: I think that two blank lines is correct in this case because these are top-level functions.  If they were methods, then it would be one line.
https://google.github.io/styleguide/pyguide.html#s3.5-blank-lines
```Two blank lines between top-level definitions, be they function or class definitions. One blank line between method definitions and between the class line and the first method.```

File: internals/search.py [hunk 3]
```diff
@@ -0,0 +1,63 @@
...
+
+def process_starred_me_query():
+  """Return a list of features starred by the current user."""
+  feature_ids = []
+  user = users.get_current_user()
+  if user:
```
KyleJu, line 36: Optional nits:
```suggestion
  if not user:
   return []
  feature_ids = notifier.FeatureStar.get_user_stars(user.email())
  return models.Feature.get_by_ids(feature_ids)
```

internals/models.py
```diff
@@ -825,6 +828,32 @@ def get_feature(self, feature_id, update_cache=False):
...
+  def get_by_ids(self, feature_ids, update_cache=False):
+    result_dict = {}
+    futures = []
+
+    for feature_id in feature_ids:
+      KEY = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, feature_id)
+      feature = ramcache.get(KEY)
+
+      if feature is None or update_cache:
+        futures.append(Feature.get_by_id_async(feature_id))
```
KyleJu, line 841: Does `Feature.get_by_id_async` update `ramcache` automatically? (nm, it is cached below)

KyleJu: No action if `feature is not None`?

static/elements/chromedash-feature-table.js
```diff
@@ -38,11 +48,25 @@ class ChromedashFeatureTable extends LitElement {
     `];
   }
 
+  renderMessages() {
```
KyleJu, line 51: How is this function being notified if `this.loading` became `False`? In ploymer you can achieve it by creating a simple observers, eg. `loading` to the parameter

internals/models.py
jrobbins: Right.  If it is not none, then that means that we got a ramcache hit on line 838.

internals/search.py
jrobbins: Change made.

static/elements/chromedash-feature-table.js
jrobbins: Good catch.  This was a minor defect.

In lit-element, the render method is called after anything modified one of the element properties as defined that are the properties() method.  I had this.features listed as a property, but not this.loading.  The code worked because both this.features and this.loading were set at the same time.  However, you are correct that this.loading should also be a property.  That would make the code do the right thing if we somehow set this.loading at some other time, or if the render() method was called between those two assignments (but I don't think that that is possible in JS).

KyleJu: I see. That seems like a much simpler design than creating many observable changes. Thanks for the explanation :)

KyleJu (approved):

PR #1537, opened 2021-09-16
On the upcoming page, always group features by launch type.

jrobbins:
The old schedule page had an option to group by blink component.  That was not super-useful, so it was unchecked by default.  That choice was copied over onto the new upcoming page, but it should be removed.

On the upcoming page, we want to clearly show what kind of launch is happening for each feature in each milestone.  So, we should always group feature by launch type.  We don't need the option to have features shown in a list without the grouping.

Also, tweak the wording for the message when a milestone has nothing in it yet.


KyleJu (approved):

shivamag00 (commented):
LGTM

PR #1538, opened 2021-09-16
Use background highlights on Upcoming page.

jrobbins:
In this PR:
* Use a background color for highlighting rather than changing font size and color
* Highlight the feature `<li>` element rather than just the `<a>` element so that the highlighted regions will have a more consistent size
* Use a little padding on all four sides of the `<li>` element so that the text is not right up against the edge of the highlighted area


shivamag00 (approved):
LGTM

PR #1540, opened 2021-09-17
Rename "Upcoming" page to "Roadmap".

jrobbins:
Rename the yet-to-be-launched "Upcoming" page to be "Roadmap".

The name "Upcoming" was selected to emphasize the default of showing current milestones, but the problem with that name is that the page can also show past milestones.

While the change could have been made only in the URL, it will be less confusing for future developers if we align the names used in the UI and all the code.


shivamag00 (approved):
LGTM

jpmedley:
I thought this would be at `https://cr-status-staging.appspot.com/features/roadmap`, but it's not. Is it somewhere else or is it not yet staged.

jrobbins:
It is just /roadmap.
So the staging URL is https://cr-status-staging.appspot.com/roadmap
And you can see it running in prod here: https://chromestatus.com/roadmap

PR #1541, opened 2021-09-18
Speed up the roadmap page by querying in parallel and caching

jrobbins:
This PR reduces the latency on requests to `/api/v0/features?milestone=<number>`.

On staging for a cache hit, the time is reduced from 400-500ms to 140-200ms.  A roughly 60% speedup.
On prod for a cache miss, the time is reduced from 400-700ms to 250-450ms. A roughly 40% speedup.
On prod for a cache miss, the time is reduced from 400-700ms to 120-400ms. A roughly 50% speedup.
The absolute number of ms saved on the page load is 200-300ms and does feel faster.
Scrolling to past or future releases is also a little faster, which fells more responsive when the user clicks rapidly.

In this PR:
+ In `models.Feature.get_in_milestone()`, move all the queries closer to the top of the function, and make them all use fetch_async().  After all the queries have been started, wait for them all to finish.
+ Add a check for cached results at the top of that function and set the cached results at the bottom.
+ Rename some of the result variables to clarify the structure of those values.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -924,6 +924,12 @@ class Feature(DictModel):
     if milestone == None:
       return None
 
+    cache_key = '%s|%s|%s|%s' % (
+        Feature.DEFAULT_CACHE_KEY, 'milestone', show_unlisted, milestone)
+    cached_allowed_features_by_type = ramcache.get(cache_key)
+    if cached_allowed_features_by_type:
+      return cached_allowed_features_by_type
+
     all_features = {}
     all_features[IMPLEMENTATION_STATUS[ENABLED_BY_DEFAULT]] = []
     all_features[IMPLEMENTATION_STATUS[DEPRECATED]] = []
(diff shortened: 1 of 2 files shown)


shivamag00 (approved):
LGTM

KyleJu (approved):
LGTM % a question - whats the memory impact of this change? Whats the capacity on `ramcache`? e.g. The free tier on redis is 1GB

jrobbins:
> LGTM % a question - whats the memory impact of this change? Whats the capacity on `ramcache`? e.g. The free tier on redis is 1GB

ramcache doesn't use redis.  I had looked into using that as a way to stop using the old GAE py2 memcache API, however setting up redis with py2 seemed too complicated.  We could come back to try making that upgrade once we are completely on py3 and it should be simpler.    The ramcache implementation just uses python RAM in each running instance.  It does distributed invalidation by having each request handler check a singleton datastore entity.  For details, see:
https://github.com/GoogleChrome/chromium-dashboard/blob/main/framework/ramcache.py

As for the RAM usage, we seem to have a good amount of spare capacity.  Our app uses F4 instances which have a max of 1024MB RAM usage.  Our instances are currently using 37% to 68% of that.  That leaves at least 300MB free in each instance.

I don't know exactly how much RAM the new cache entries will consumer, but they likely total less than 6x the size of the response size for features_v2.json (1MB).  6x because it will be one copy for signed in users who can view unlisted features and one copy for other users, multiplied by a max of 3 because features can have a few different kinds of launches (dev trial, origin trial, and shipping).  That is 6 * 1MB = 6MB.  Which seems pretty small compared to the amount of unused RAM in each instance.

PR #1542, opened 2021-09-19
Add two icons to copy to clipboard.

jrobbins:
This addresses two requests from users:
1. It should be easer to copy the URL of a feature detail page.
2. Sometimes when copying the intent text, the formatting gets messed up.  This seems to happen if the user selects a small area outside the parent element of the content that they wish to copy.  So, a single click to copy it should avoid that.

In this PR:
* Define link and content_copy icons
* Add the link icon to the feature detail page and make it call navigator.clipboard.writeText().
* Add the content_copy icon to the intent preview page and make it do an old-school copy command.
* Simplify the toast element auto-hiding.

navigator.clipboard is the preferred approach, but I could not find a working example of copying formatted text using that API.


static/elements/chromedash-toast.js
```diff
@@ -86,39 +84,20 @@ class ChromedashToast extends LitElement {
       }
     }, {once: true});
 
-    // Override duration just for this toast.
-    const originalDuration = this.duration;
-    if (typeof optDuration !== 'undefined') {
-      this.duration = optDuration;
+    let duration = optDuration || DEFAULT_DURATION;
```
liyangguang, line 87: nit: `const`

static/js-src/feature-page.js
```diff
@@ -32,6 +35,14 @@ const shareFeature = () => {
   }
 };
 
+function copyURLToClipboard() {
+  event.preventDefault();
+  const url = copyLinkEl.href;
+  navigator.clipboard.writeText(url).then(() => {
```
liyangguang, line 41: Do we need to handle non-supporting browser? (like when this function is not available, or failed, show a failed copy toast, or fallback to the selection + execute('copy') method)

liyangguang (approved):

static/elements/chromedash-toast.js
jrobbins: Changed.

static/js-src/feature-page.js
jrobbins: No, I think our core users are all on modern browsers, which have supported this for many releases already.  For non-supporting browsers, the user has the work-around of simply copying the current browser URL the old fashioned way.

PR #1546, opened 2021-09-21
Delete unneeded code in scripts/.

jrobbins:
I don't believe that any of this code was needed or actually used.  It looks like it was previously useful for running locally while hitting the production datastore, but I think that is not a good thing to do.

If we somehow do need any of this, we would probably need an updated version of it anyway.


KyleJu (approved):
One less test error on the CI!!

PR #1549, opened 2021-09-21
Fix a test failure in api/.

jrobbins:
I think this was a long standing defect in the code that has nothing to do with python 2 or 3.
I guess the python2 version of the flask library had a wildcard for these keywords, while it is checked more strictly in py3.


KyleJu (approved):
Nice!

PR #1555, opened 2021-09-23
Get test coverage working for py3.

jrobbins:
In this PR:
+ Update the test coverage target to follow the same pattern of using the datastore emulator that the test target uses.
+ Update the coverage tool command line to the way that it is supposed to be run under py3.
+ Add a new view-coverage target that starts a web server for the generated HTML.


KyleJu (approved):
Not sure why the py2 target is failing

PR #1560, opened 2021-09-24
Add targets for pylint and test2

jrobbins:
Adding these targets should make it easier for developers to run them.


KyleJu (approved):
Good work!

PR #1564, opened 2021-09-28
Make the myfeatures page be one big webcomponent.

jrobbins:
This is progress toward launching the MyFeatures page.  My goals in this step are:
* Improve the appearance of the page a little
* Conditionally offer the pending-approvals box only to users who have permission to approve intents
* Keep the star icons consistently on or off even if the same feature is shown multiple times on the page

In this PR:
* Reimplement the myfeatures.html template to use a single web component instead of using three parts in the django template.
* Implement a new web component chromedash-myfeatures to do that rendering, and to keep track of starred features.
* Define django variable user.canApprove and pass it through to JS conditionals to control what is offered
* Show icons for approvals, editing, and starring as appropriate
* Tweak the visual appearance of the page
* Define an approvals stamp icon in icons.js

There is a user-visible "TODO" that can pop up if the user clicks the approvals icon.  However, this page is not linked from anything in the site navigation yet, and that icon is only offered to a small list of users who have permission to approve.  I've leaving it in for now as I continue to work on this page.


jrobbins:
This is what this looks like with some local test data:
image

static/elements/chromedash-feature-table.js
```diff
@@ -46,9 +50,49 @@ class ChromedashFeatureTable extends LitElement {
         padding: var(--content-padding-half);
         border-bottom: var(--table-divider);
       }
+      td.name_col {
+        width: 100%;
+      }
+      td.icon_col {
+        white-space: nowrap;
+      }
+
```
KyleJu, line 59:
```suggestion
```

static/elements/chromedash-feature-table.js
```diff
@@ -46,9 +50,49 @@ class ChromedashFeatureTable extends LitElement {
...
+  toggleStar(e) {
+    console.log('toggleStar');
+    e.preventDefault();
+    e.stopPropagation();
+
+    const iconEl = e.target;
+    const featureId = Number(iconEl.dataset.featureId);
+    const newStarred = !this.starredFeatures.has(featureId);
+
+    // handled in myfeatures-page.js
```
KyleJu, line 89: you mean chromedash-myfeatures.js?

static/elements/chromedash-myfeatures.js
```diff
@@ -0,0 +1,84 @@
...
+        title="${title}"
+        opened>
+
+        <chromedash-feature-table
+          query="${query}"
+          ?signedin=${this.signedIn}
+          ?canedit=${this.canEdit}
+          ?canApprove=${this.canApprove}
+          .starredFeatures=${this.starredFeatures}
+          @star-toggle-event=${this.handleStarToggle}
```
KyleJu, line 53: Is this how you listen to the `'star-toggle-event` event?

static/elements/chromedash-feature-table.js
```diff
@@ -22,6 +24,8 @@ class ChromedashFeatureTable extends LitElement {
     this.starredFeatures = new Set();
```
KyleJu, line 24: `starredFeatures` is being passed from the parent DOM. Does it still need to be intialized?

KyleJu (approved):
LGTM with some comments

static/elements/chromedash-feature-table.js
jrobbins: IIRc, this is the recommended way to initialize properties as per the lit-element docs.  In this case, the value is immediately set again based on an attribute set in the DOM, however web components are meant to be reused in a variety of contexts so it might not always be set that way.

static/elements/chromedash-feature-table.js
jrobbins: Good catch.  Fixed.

static/elements/chromedash-myfeatures.js
jrobbins: Yes, the @ syntax adds an event listener.  
https://lit-element.polymer-project.org/guide/events#add-declarative-event-listeners

PR #1570, opened 2021-09-30
Add My Features tab and indicate active tab.

jrobbins:
This adds a top-tab navigation link to the myfeatures page, essentially launching the page.
The tab is only shown when the user is signed in.

In this PR:
+ Add the top tab link to /myfeatures
+ Change the CSS to indicate the active tab, in the spirit of the MD spec:
  https://material.io/components/tabs#anatomy
  "To differentiate an active tab from an inactive tab, apply an underline and color change to the active tab’s text and icon."
+ Add a little bit of JS to add the "active" CSS class to the active tab, if any.
+ Also, add a little space between the sections of the myfeatures page

### DIFF
--- a/templates/header.html
+++ b/templates/header.html
@@ -54,8 +54,8 @@
   </nav>
 </header>
 
-{% if user %}
-  <script nonce="{{nonce}}">
+<script nonce="{{nonce}}">
+  {% if user %}
     const signOutEl = document.querySelector('#sign-out-link');
     if (signOutEl) {
         signOutEl.addEventListener('click', (e) => {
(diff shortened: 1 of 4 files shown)


jrobbins:
This is how it looks
image

templates/header.html
```diff
@@ -61,5 +63,13 @@
             signOut();
         });
     }
+
+    const maintabsEl = document.querySelector('#maintabs');
```
KyleJu, line 67: what is this for?

KyleJu, line 67: This is run every time users sign in and sign out right?

KyleJu (approved):

templates/header.html
```diff
@@ -61,5 +63,13 @@
...
+    for (let tabEl of maintabs.children) {
```
KyleJu, line 68: Since it is going through all children, does it mean             
(<li><a href="/metrics/css/popularity">CSS</a></li> or  <li><a href="/metrics/feature/popularity">JS/HTML</a></li>)
 and 
<div class="nav-dropdown-container flex-item" href="/metrics">

will all be `active`? when the page is on

templates/header.html
jrobbins: Yes, the whole page is reloaded, so all page content that is specific to the user is recomputed.

jrobbins: Ah, but I just caught a mistake in that the whole <script> was only included if the user is signed in.  Will fix in a sec.

templates/header.html
jrobbins: maintabs.children is a list of only the direct children, so only the <div> for the menu header gets the underline.

PR #1577, opened 2021-10-07
Add an OIDC Authorization header to uma-export request.

jrobbins:
This PR adds an Authorization header as described in this p2->p3 upgrade doc:
https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-app-identity#token

Our requests to uma-export were being rejected because that app looks for a X-Appengine-Inbound-Appid header that was set by URLFetch in py2 but that there is no way to pass in py3.   Now we with identify our app via the Authorization header instead.


KyleJu:
@jrobbins looks like it is trying to make real requests in the tests

jrobbins:
OK, it passed now.  It's odd that I didn't need to mock the token generation when I tested it locally.

PR #1588, opened 2021-10-21
Make footer fixed on very long pages.

jrobbins:
This should resolve issue #1096.

Footer links are infrequently used, so I don't want them to always be taking up screen real estate on pages like the Guide UI.  However, having the footer be position:relative on a few pages is a pain because the pages are very long.  I think the right long-term fix is to make those pages not so long, however in the near term we can make the footer fixed (always displayed) on just those pages.

In this PR:
* Add CSS to make the footer fixed position to the CSS files for the feature list and metrics list pages.  These rules are inside a media condition for desktop devices because we don't want to dedicate screen space to the footer on mobile devices.
* Combine some existing CSS rules
* Delete an ::after rule that does not seem to be needed.

### DIFF
--- a/static/sass/features/features.scss
+++ b/static/sass/features/features.scss
@@ -53,10 +53,12 @@ chromedash-featurelist {
 }
 
 @media only screen and (min-width: 701px) {
-  #content {
-    &::after {
-      width: $max-content-width;
-    }
+
+  footer {
+    position: fixed;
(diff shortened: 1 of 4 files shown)


jrobbins:
Demo link:
https://99e508f-dot-app-py3-dot-cr-status-staging.appspot.com/

PR #1596, opened 2021-11-02
Outline the implementation of the approvals dialog

jrobbins:
This replaces a previous alert("TODO") with a real dialog box that still says "TODO".

In this CL:
* Defined a custom element chromedash-approvals-dialog that wraps a chromedash-dialog element.
* For now, that dialog just says "TODO".
* Generate an event when the approvals-stamp icon is clicked, and handle that event by opening the dialog.
* Pass only the feature ID in that event because we will reuse this dialog on pages where only the ID is available.
* Change from using a signedIn boolean to a signedInUser email address string.  The full implementation of the dialog will need the string.
* Fix a couple of mismatched HTML tags that were caught by JS lint.

### DIFF
--- a/static/components.js
+++ b/static/components.js
@@ -14,6 +14,7 @@ import '@polymer/paper-styles/color.js';
 // chromedash components
 import './elements/icons';
 import './elements/chromedash-accordion';
+import './elements/chromedash-approvals-dialog';
 import './elements/chromedash-banner';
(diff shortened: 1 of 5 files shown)


jrobbins:
This is what it looks like when the user clicks a stamp icon:
image

liyangguang (approved):

PR #1597, opened 2021-11-04
When an intent review is complete, make it no longer pending.

jrobbins:
Each feature can have a list of associated approval values which are grouped by approval field (e.g., prototype, experiment, ship).  When an intent email thread is detected, a NEEDS_REVIEW value is added to indicate the start of that review process.  When viewing the My Features page, the list of pending reviews is generated by searching for NEEDS_REVIEW approval values.

This change implements the end of the review process.  When users have added enough APPROVED or NOT_APPROVED values, then we detect that fact and delete the original NEEDS_REVIEW value.  This makes that item no longer appear on the My Features page (until it starts the review of a later intent with a different field id).

In this PR:
* Implement boolean methods `is_approved()` and `is_resolved()`.
* Implement a models.py method `clear_request()` to delete the relevant NEEDS_REVIEW entity.
* Add logic to approvals_api.py to `check is_resolved()` and call `clear_request()`.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -20,6 +20,7 @@ import requests
 
 from framework import permissions
 from framework import ramcache
+from internals import models
 import settings
 
 CACHE_EXPIRATION = 60 * 60  # One hour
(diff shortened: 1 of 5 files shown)


internals/approval_defs.py
```diff
@@ -116,3 +117,35 @@ def fields_approvable_by(user):
...
+    return count >= 1
+  elif afd.rule == THREE_LGTM:
+    return count >= 3
+  else:
+    logging.error('Unexpected approval rule')
+    return False
+
+
+def is_resolved(approval_values, field_id):
+  """Return true if the review is done (approved or not approved)."""
```
KyleJu, line 142: very optional nit:
```suggestion
  # Any NOT_APPROVED value means that the review is no longer pending.
  for av in approval_values:
    if av.state == models.Approval.NOT_APPROVED:
      return True
  return is_approved(approval_values, field_id)
```

KyleJu: and perhaps remove line 128 and 129 if above is accepted

KyleJu (approved):

internals/approval_defs.py
jrobbins: Thanks for the comment.  I thought about it and I think that the original doc string is more accurate because an APPROVED value is also no longer pending.

PR #1598, opened 2021-11-04
Make py2 inbound email handler call py3 via appspot.com.

jrobbins:
This is a second try at fixing a problem that we are still seeing in production.  The python 2 code for handling an inbound email message is failing to make a request to the python 3 code for further processing.  The python three handler rejects the request because it is missing the X-AppEngine-Inbound-Appid header that proves that the request came from our app.   This exact same code works on staging.

After my previous fix was deployed and it continued to fail, I looked at the docs again:
https://cloud.google.com/appengine/docs/standard/python/appidentity
"Only calls made to your app's appspot.com domain will contain the X-Appengine-Inbound-Appid header. Calls to custom domains do not contain the header."

This PR makes the py2 code send a request to the fully spelled out appspot.com URL of the py3 code rather than using the "chromestatus.com" custom domain name.

### DIFF
--- a/internals/sendemail.py
+++ b/internals/sendemail.py
@@ -134,7 +134,7 @@ def call_py3_task_handler(handler_path, task_dict):
   """Request that our py3 code handle the rest of the work."""
   handler_host = 'http://localhost:8080'
   if settings.APP_ID == 'cr-status':
-    handler_host = 'https://chromestatus.com'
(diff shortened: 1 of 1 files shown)


KyleJu (approved):
Worth a try! I wonder why custom domains worked before py3

PR #1599, opened 2021-11-04
Fix defect in read-only display of approval values.

jrobbins:
The existing code used a state integer as an index into a list of [id, name] pairs.  However, that was wrong because the whole point of having that list was that the ids did not line up with the indexes.

In this PR:
* Add a lookup function findStateName() to get the correct state name for read-only display.

I didn't notice this sooner because it only shows up when testing with two different users.  Each user sees the other user's approval values as read-only.

### DIFF
--- a/static/elements/chromedash-approvals-dialog.js
+++ b/static/elements/chromedash-approvals-dialog.js
@@ -4,14 +4,15 @@ import './chromedash-dialog';
 import SHARED_STYLES from '../css/shared.css';
 
 const STATE_NAMES = [
-  [0, 'Needs review'],
+  // Not used: [0, 'Needs review'],
   [1, 'N/a or Ack'],
+  [2, 'Review requested'],
   [3, 'Review started'],
   [4, 'Need info'],
   [5, 'Approved'],
   [6, 'Not approved'],
 ];
-const ACTIVE_STATES = [0, 3, 4];
+const PENDING_STATES = [2, 3, 4];
 
 const APPROVAL_DEFS = [
   {name: 'Intent to Prototype',
@@ -147,7 +148,7 @@ class ChromedashApprovalsDialog extends LitElement {
       (res) => {
         this.approvals = res.approvals;
         const numPending = this.approvals.filter((av) =>
-          ACTIVE_STATES.includes(av.state)).length;
+          PENDING_STATES.includes(av.state)).length;
         this.subsetPending = (numPending > 0 &&
                               numPending < APPROVAL_DEFS.length);
         this.showAllIntents = numPending == 0;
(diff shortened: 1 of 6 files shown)


static/elements/chromedash-approvals-dialog.js
```diff
@@ -171,6 +171,15 @@ class ChromedashApprovalsDialog extends LitElement {
     return dateStr.split('.')[0]; // Ignore microseconds.
   }
 
+  findStateName(state) {
+    for (let item of STATE_NAMES) {
+      if (item[0] == state) {
+        return item[1];
+      }
+    }
+    throw new Error(`Unexpected approval state ${state}`);
```
liyangguang, line 180: Do we need to do anything here, like an alert?

(if the render part of the code throws error, will it just result in a blank/noninteractive component?)

liyangguang (commented):
I don't quite follow the error vs code here.

I see the code changes is because value of [2] is left out in `STATE_NAMES`. Where does 2 users and readonly come into play?

(not about this commit - but is the number value coming from backend? why do we need to leave out `2`? If have to, maybe add a comment between 1 & 3, so easier to notice the gap.)

jrobbins:
> I don't quite follow the error vs code here.
> 
> I see the code changes is because value of [2] is left out in `STATE_NAMES`. Where does 2 users and readonly come into play?
> 
> (not about this commit - but is the number value coming from backend? why do we need to leave out `2`? If have to, maybe add a comment between 1 & 3, so easier to notice the gap.)

Yes, the state numbers come from the backend.  Those values follow the enums defined for Monorail's feature launch tracking (FLT) functionality.
https://chromium.googlesource.com/infra/infra/+/refs/heads/main/appengine/monorail/schema/tracker.sql#526

Thanks to your comment, I went back and reread the FLT design doc.  There are enum values for NEEDS_REVIEW and REVIEW_REQUESTED.  NEEDS_REVIEW is used when the system initiates the review, whereas REVIEW_REQUESTED is used when a user initiates the review.  In chromestatus, the feature owners always initiate a review by posting to the blink-dev mailing list, so we should be using REVIEW_REQUESTED.  I have added that change to this PR.

As you suggested, I also now have a comment line explaining the value that is not used so that there are no gaps in the numbers.

static/elements/chromedash-approvals-dialog.js
jrobbins: I made this fail in a less dramatic way by just displaying the state number to the user.  In practice that should never happen, but in the case of a defect or some kind of data corruption, the dialog will still load and we will be able to see visible evidence of the problem.

liyangguang (approved):

PR #1600, opened 2021-11-05
Implement a "Recently reviewed features" box on the myfeatures page

jrobbins:
As mentioned at yesterday's team meeting, I realized that it seems like a problem that once a feature is reviewed it just disappears from the pending approvals box.  People like to see their recent progress in a visible way.

In this PR:
* Add a new "Recently reviewed features" box on the myfeatures page.  The accordion for that box is initially closed.
* Implement a new query to get the 10 most recently reviewed features to populate that box.
* Add new unit tests and improve some existing ones.

### DIFF
--- a/internals/models_test.py
+++ b/internals/models_test.py
@@ -339,10 +339,17 @@ class ApprovalTest(testing_config.CustomTestCase):
     self.feature_1.put()
     self.feature_1_id = self.feature_1.key.integer_id()
     self.appr_1 = models.Approval(
-        feature_id=self.feature_1_id, field_id=1, state=2,
-        set_on=datetime.datetime.now(),
+        feature_id=self.feature_1_id, field_id=1,
+        state=models.Approval.REVIEW_REQUESTED,
+        set_on=datetime.datetime.now() - datetime.timedelta(1),
         set_by='one@example.com')
     self.appr_1.put()
+    self.appr_2 = models.Approval(
+        feature_id=self.feature_1_id, field_id=1,
+        state=models.Approval.APPROVED,
+        set_on=datetime.datetime.now(),
+        set_by='two@example.com')
+    self.appr_2.put()
 
   def tearDown(self):
     self.feature_1.key.delete()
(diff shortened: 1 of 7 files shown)


internals/models_test.py
```diff
@@ -339,10 +339,17 @@ def setUp(self):
     self.feature_1.put()
     self.feature_1_id = self.feature_1.key.integer_id()
     self.appr_1 = models.Approval(
-        feature_id=self.feature_1_id, field_id=1, state=2,
-        set_on=datetime.datetime.now(),
+        feature_id=self.feature_1_id, field_id=1,
+        state=models.Approval.REVIEW_REQUESTED,
+        set_on=datetime.datetime.now() - datetime.timedelta(1),
```
KyleJu, line 344: Not sure what kinds of clocks it uses for testing, but hope `datetime.datetime.now() - datetime.timedelta(1)` doesn't go below 0?

internals/search.py
```diff
@@ -80,6 +99,8 @@ def process_query(user_query):
     return process_starred_me_query()
   if user_query == 'owner:me':
     return process_owner_me_query()
+  if user_query == 'is:recently-reviewed':
```
KyleJu, line 102: Are these queries documented anywhere or it is set by the client side?

static/elements/chromedash-myfeatures.js
```diff
@@ -78,9 +78,12 @@ class ChromedashMyFeatures extends LitElement {
...
   }
 
-  renderPendingApprovals() {
-    return this.renderBox(
+  renderPendingAndRecentApprovals() {
+    const pendingBox = this.renderBox(
       'Features pending my approval', 'pending-approval-by:me', 'approvals');
+    const recentBox = this.renderBox(
+      'Recently reviewed features', 'is:recently-reviewed', 'approvals', false);
+    return [pendingBox, recentBox];
```
KyleJu, line 86: What does it mean to return a list of html here? Does it get rendered side by side?

internals/models_test.py
jrobbins: I believe that it uses the regular clock.  It seems to work fine.

internals/search.py
jrobbins: For now, each query is hard-coded into the client to populate specific boxes in the UI.  

We might generalize this to a user query language at some point in the future.  The fact that these queries are not visible to users means that they do not need end-user documentation and we could rework the query language without users needing to relearn anything.

static/elements/chromedash-myfeatures.js
jrobbins: Not side-by-side, just one string of HTML after the other.   The resulting page looks like this:

image

KyleJu (approved):

PR #1601, opened 2021-11-09
Add stamp icons to feature list and feature detail pages.

jrobbins:
This adds the feature approval icon to more places where it might be useful.  It is only shown to users who have permission to approve intents.

In this PR:
* Add the approval stamp icon to the feature detail page next to the edit icon.  And, add event handlers to make it actually open the dialog.
* Add the approval stamp icon to the feature cards on the feature list page.  And, event handlers.  And, pass some permission info down so that the cards know when to show the icon.

### DIFF
--- a/static/js-src/feature-page.js
+++ b/static/js-src/feature-page.js
@@ -1,6 +1,7 @@
 (function(exports) {
 const toastEl = document.querySelector('chromedash-toast');
 const copyLinkEl = document.querySelector('#copy-link');
+const approvalsIconEl = document.querySelector('#approvals-icon');
 
 // Event handler. Used in feature.html template.
 const subscribeToFeature = (featureId) => {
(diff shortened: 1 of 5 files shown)


static/js-src/feature-page.js
```diff
@@ -36,13 +37,17 @@ const shareFeature = () => {
 };
 
 function copyURLToClipboard() {
-  event.preventDefault();
```
KyleJu, line 39: should this be removed?

KyleJu (approved):

static/js-src/feature-page.js
jrobbins: Good point.  It is needed.  The way that that line was written looked like a defect to me because `event` is not defined anywhere.  I rewrote calls to `preventDefault()` for each of the listeners and tested them.

PR #1611, opened 2021-11-19
Implement posting review comments to blink-dev mailing list.

jrobbins:
This implements the ability for a API Owner to make a review comment in the approvals dialog box and opt into having that comment also sent to the discussion thread for that intent on the blink-dev mailing list.

On staging we use a separate test mailing list rather than blink-dev.

In this PR:

 * Change the UI from a checkbox to a `<select>` to allow the user to specify which intent thread the comment is intended for.  Usually each feature should only have one intent discussion thread at a time, but that expectation might not be true given that we have incomplete data for existing features.  Also, we only offer the option when we are sure to be able to parse the thread ID out of the discussion URL.

* Pass the approval field ID from the chromedash-approvals-dialog element via cs-client to the comments API handler.

 * In the comment handler, call a new function in notifier.py.

 * In notifier.py, implement `post_comment_to_mailing_list()` to create a task with all the info needed for the email.  The email body is formatted using a template so that all of our sent emails are HTML rather than plain text, which simplifies sendemail.py.

 * In sendemail.py, add support for an optional parameter for the sender so that posted messages are attributed to the user who posted the comment.  We cannot actually send as that user, so we just use their user name in the "friendly" part of the address.  The sending email address must also be different for each sender to avoid confusing gmail, so we use the "plus" part of the address.

* Add a settings.py value for the email address of the mailing list to post to.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -273,7 +273,7 @@ TEST_ARCHIVE_URL_PREFIX = (
     'https://groups.google.com/d/msgid/jrobbins-test/')
 
 def get_thread_id(feature, approval_field):
-  """If we have the URL of the Google Groups thread, we can get it's ID."""
+  """If we have the URL of the Google Groups thread, we can get its ID."""
   if approval_field == approval_defs.PrototypeApproval:
     thread_url = feature.intent_to_implement_url
   # TODO(jrobbins): Ready-for-trial threads
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -1,6 +1,3 @@
-
-
-
 # -*- coding: utf-8 -*-
 # Copyright 2017 Google Inc.
 #
@@ -27,8 +24,6 @@ import os
 from framework import ramcache
 from google.cloud import ndb
 import requests
-# from google.appengine.api import users
-from framework import users
 
 from django.template.loader import render_to_string
 from django.utils.html import conditional_escape as escape
@@ -36,6 +31,7 @@ from django.utils.html import conditional_escape as escape
 from framework import basehandlers
 from framework import cloud_tasks_helpers
 import settings
+from internals import approval_defs
 from internals import models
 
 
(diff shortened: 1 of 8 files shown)


jrobbins:
Thanks for the other reviews.  I just wanted to make sure that you saw this one too.

internals/notifier.py
```diff
@@ -269,3 +265,70 @@ def process_post_data(self):
...
     return {'message': 'Done'}
+
+
+BLINK_DEV_ARCHIVE_URL_PREFIX = (
+    'https://groups.google.com/a/chromium.org/d/msgid/blink-dev/')
+TEST_ARCHIVE_URL_PREFIX = (
+    'https://groups.google.com/d/msgid/jrobbins-test/')
+
+def get_thread_id(feature, approval_field):
+  """If we have the URL of the Google Groups thread, we can get it's ID."""
```
KyleJu, line 276:
```suggestion
  """If we have the URL of the Google Groups thread, we can get its ID."""
```

internals/sendemail.py
```diff
@@ -55,19 +55,26 @@ def handle_outbound_mail_task():
   require_task_header()
 
   to = get_param(flask.request, 'to')
+  from_user = get_param(flask.request, 'from_user', required=False)
   subject = get_param(flask.request, 'subject')
   email_html = get_param(flask.request, 'html')
+  references = get_param(flask.request, 'references', required=False)
 
-  if settings.SEND_ALL_EMAIL_TO:
+  if settings.SEND_ALL_EMAIL_TO and to != settings.REVIEW_COMMENT_MAILING_LIST:
```
KyleJu, line 63: Is `settings.REVIEW_COMMENT_MAILING_LIST` a list?

KyleJu: Actually it is a string

KyleJu (approved):

PR #1617, opened 2021-11-24
Add a field for next_review_date.

jrobbins:
This addresses some of the feedback I got from the demo to API Owners about 15 days ago.

In this PR:
* Add a new field date next_review_action to the Feature entity type.
* Define a django form field for it and configure it to use `<input type="date">`.
* Add the new form field to each of the forms in which it might be relevant, and display it once on the issue detail page.
* Add a case to the guide UX form POST handler logic to store this date field
* Make the preparation for the diff email pass the date as a string because that is needed before it gets used in a JSON string

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -1,6 +1,3 @@
-
-
-
 # -*- coding: utf-8 -*-
 # Copyright 2020 Google Inc.
 #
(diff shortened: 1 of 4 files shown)


pages/guide.py
```diff
@@ -354,6 +351,11 @@ def process_post_data(self, feature_id, stage_id=0):
     if self.touched('i2s_lgtms'):
       feature.i2s_lgtms = self.split_emails('i2s_lgtms')
 
+    if self.touched('next_review_date'):
+      date_str = self.form.get('next_review_date')
+      date_val = datetime.date.fromisoformat(date_str)
```
KyleJu, line 356: Maybe have a test for it so we know the format of `date_str`?

KyleJu (approved):
LGTM with a comment

pages/guide.py
jrobbins: Added.

jrobbins:
I am reverting because I found a bug and I have a better way to do this.
The bug is that the ISO string parsing fails when the field is not set in the UI, causing an empty string to be passed.
The better way to do it is to keep track of next action on a per-approval basis rather than having just one value for the entire feature.

PR #1625, opened 2021-11-30
Implement a new approval config datastore entity type and API

jrobbins:
This replaces my previous attempt at representing the next-action date for reviews (#1617).

In this PR:
* Define new datastore entity for ApprovalConfig which will be stored per-feature and per-approval-field.
* Implement get and set methods in models.py.
* Implement a new API class for ApprovalConfigs with the ability to get and set.
* Unit tests 
* cs-client.js methods to call the new API.

Nothing uses this new API yet, but I will work on the UI and JS for these features in my next PR.

### DIFF
--- a/api/approvals_api.py
+++ b/api/approvals_api.py
@@ -13,9 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-
-
+import datetime
 import logging
 
 from framework import basehandlers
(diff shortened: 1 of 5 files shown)


api/approvals_api.py
```diff
@@ -61,3 +59,45 @@ def do_post(self, feature_id=None):
...
+    if not permissions.can_approve_feature(user, feature, approvers):
+      self.abort(403, msg='User is not an approver')
+
+    owners = None
+    if owners_str:
+      owners = self.split_emails('owners')
+
+    next_action = None
+    if next_action_str:
+      next_action = datetime.date.fromisoformat(next_action_str)
```
KyleJu, line 97: Check `ValueError` here? what if `next_action_str` is malformatted

KyleJu (approved):
LGTM with a comment

api/approvals_api.py
jrobbins: Good point.  I added a try-except block for that and unit tests for it.

PR #1630, opened 2021-12-02
Treatempty uma-export result as not ready.

jrobbins:
This copes with a change in uma-export that happened back on 2021-11-21 or so.

The old behavior of uma-export in the case where metrics data was not yet ready was something like:
`{e: 'data is not ready'}`.  But, as a side-effect of that change, uma-export started returning `{r: {}}`.  The downside of that is that when our cron task gets today's metrics data and it is not ready yet, then we store that empty result, set a capstone value, and then don't ask again the next time the cron job runs.

In this PR:
* Treat an empty 'r' value as the data not being ready.  So, nothing is stored and the cron job will later try again.
* Add more logging.
* Add unit tests.

I have exchanged some emails with chrome-metrics-team and they seem likely to change the behavior or uma-export back to the way that it was, but our code should handle these cases more robustly anyway.

### DIFF
--- a/internals/fetchmetrics_test.py
+++ b/internals/fetchmetrics_test.py
@@ -13,10 +13,9 @@
 # limitations under the License.
 
 
-
-
 import base64
 import datetime
+import json
 import testing_config  # Must be imported before the module under test.
 import urllib.request, urllib.parse, urllib.error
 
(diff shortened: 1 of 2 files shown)


KyleJu:
Uh installation errors for py2 CI...

internals/fetchmetrics_test.py
```diff
@@ -98,6 +97,68 @@ def testHasCapstone__found(self):
 
     self.assertTrue(actual)
 
+  @mock.patch('internals.fetchmetrics._FetchMetrics')
+  def test_FetchData__ready(self, mock_fetch_metrics):
+    """When the uma-export data is ready, we parse and return it."""
+    r = {'123': {'rate': 0.001},
+         '234': {'rate': 0.002}}
+    XSSI_PROTECTION = ')]}\' // XSSI prefix (go/xssi).'
```
KyleJu, line 105: Had a brief read about how to avoid XSSI. Interesting!

KyleJu (approved):

KyleJu:
> Uh installation errors for py2 CI...

Looks like a `npm` issue. Maybe `npm` is outdated?

PR #1633, opened 2021-12-03
Implement a UI for editing approval configs

jrobbins:
This is progress on the requests that API owners made a couple of weeks ago to be able to set a next-action date, an owner, and specify that certain origin trials require 3 LGTMs.

In this PR:
* In chromedash-approvals-dialog: Add new fetching, rendering, and input handling logic for the new fields.  The fields are initially not shown but are shown if they have a value or if the user clicks an expansion icon.
* Change the API and tests to treat all three fields (owner, nextAction, and additionalReview) as required.  That is so that the user can clear each field by providing an empty value.  Also, add a list of possible owners to choose from.
* Tweak cs-client to provide empty values of the right type.
* In approval_devs.py: Add a missing reference to the approval for extending an experiment.

### DIFF
--- a/static/elements/chromedash-approvals-dialog.js
+++ b/static/elements/chromedash-approvals-dialog.js
@@ -45,9 +45,11 @@ class ChromedashApprovalsDialog extends LitElement {
       approvals: {type: Array},
       comments: {type: Array},
       configs: {type: Array},
+      possibleOwners: {type: Object},
       showConfigs: {type: Object},
       showAllIntents: {type: Boolean},
       changedApprovalsByField: {attribute: false},
+      changedConfigsByField: {attribute: false},
       needsSave: {type: Boolean},
       loading: {attribute: false},
     };
@@ -63,9 +65,11 @@ class ChromedashApprovalsDialog extends LitElement {
     this.comments = [];
     this.configs = [];
     this.subsetPending = false;
+    this.possibleOwners = {};
     this.showConfigs = new Set();
     this.showAllIntents = false;
     this.changedApprovalsByField = new Map();
+    this.changedConfigsByField = new Map();
     this.needsSave = false;
     this.loading = true;
   }
@@ -173,6 +177,8 @@ class ChromedashApprovalsDialog extends LitElement {
       (res) => {
         this.configs = res.configs;
         this.showConfigs = new Set(this.configs.map(c => c.field_id));
+        this.changedConfigsByField = new Map();
+        this.possibleOwners = res.possible_owners;
       });
     Promise.all([p1, p2, p3, p4]).then(() => {
       this.loading = false;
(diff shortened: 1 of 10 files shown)


jrobbins:
Here's a screenshot:
image

static/elements/chromedash-approvals-dialog.js
```diff
@@ -405,6 +523,16 @@ class ChromedashApprovalsDialog extends LitElement {
   handleCancel() {
     this.shadowRoot.querySelector('chromedash-dialog').close();
   }
+
+  toggleConfig(approvalDef) {
+    let newConfigs = new Set([...this.showConfigs]); // Make a copy.
```
KyleJu, line 528: Why are we making a copy? Is it to trigger certain functions?

static/elements/chromedash-approvals-dialog.js
```diff
@@ -230,6 +251,63 @@ class ChromedashApprovalsDialog extends LitElement {
...
+    const additionalReviewWidget = html`
+      <input id="additional-review-${fieldId}" data-field="${fieldId}"
+             @change=${this.handleConfigChanged}
+             type=checkbox ?checked=${additionalReview}>
+    `;
+
+    return html`
+     <table class="config-area">
+       <tr>
+         <td>Owner:</td>
```
KyleJu, line 294: Owners?

static/elements/chromedash-approvals-dialog.js
```diff
@@ -159,7 +173,14 @@ class ChromedashApprovalsDialog extends LitElement {
...
       });
-    Promise.all([p1, p2, p3]).then(() => {
+    const p4 = window.csClient.getApprovalConfigs(this.featureId).then(
+      (res) => {
+        this.configs = res.configs;
+        this.showConfigs = new Set(this.configs.map(c => c.field_id));
+        this.changedConfigsByField = new Map();
+        this.possibleOwners = res.possible_owners;
+      });
+    Promise.all([p1, p2, p3, p4]).then(() => {
```
KyleJu, line 183: IIUC, when `p1,p2,p3,p4` promises have resolved, this is resolved. What if one of them get rejected, is `this.loading` still true?

static/elements/chromedash-approvals-dialog.js
jrobbins: Yes, it is to trigger a new call to render(). 

My understanding is that modifying a collection is not detected as a change to a lit-element property.  Instead, we need to set the property itself to point to a new collection.  This page is not explicit about it, but it says that the first step in the process is that `the property setter is called`.  That is not going to happen if we simply mutate an existing property value.
https://lit.dev/docs/v1/components/properties/#what-happens-when-properties-change

static/elements/chromedash-approvals-dialog.js
jrobbins: Changed.

static/elements/chromedash-approvals-dialog.js
jrobbins: Yes, that is true.  If any one of the individual promises is rejected, then the Promise.all() also rejects.  There is no reject case, so the user would just sit there seeing a spinner.

To improve that UX, I have added a catch() clause that shows an error message and closes the dialog.

static/elements/chromedash-approvals-dialog.js
```diff
@@ -230,6 +255,63 @@ class ChromedashApprovalsDialog extends LitElement {
...
 
+  renderConfigWidgets(approvalDef) {
+    const fieldId = approvalDef.id;
+    const isOpen = this.showConfigs.has(fieldId);
+    if (!isOpen) {
+      return nothing;
+    }
+
+    const config = this.configs.find(c => c.field_id == fieldId);
+    const owners = (config && config.owners || []).join(', ');
```
KyleJu, line 266: This line is not very intuitive. I understand it used as non-Boolean operands but according to the API `[a5 = 'Cat' && 'Dog'      // t && t returns "Dog"]`. It is not clear what precedence is here.

static/elements/chromedash-approvals-dialog.js
KyleJu: Yup thanks for addressing it

static/elements/chromedash-approvals-dialog.js
```diff
@@ -230,6 +255,63 @@ class ChromedashApprovalsDialog extends LitElement {
...
+    const nextActionWidget = html`
+      <input id="next-action-${fieldId}" data-field="${fieldId}"
+             @change=${this.handleConfigChanged}
+             type=date name="next_action_${fieldId}"
+             value="${nextAction}">
+    `;
+    const additionalReviewWidget = html`
+      <input id="additional-review-${fieldId}" data-field="${fieldId}"
+             @change=${this.handleConfigChanged}
+             type=checkbox ?checked=${additionalReview}>
```
KyleJu, line 292:
```suggestion
             type=checkbox ? checked=${additionalReview}>
```

KyleJu (approved):
LGMT

static/elements/chromedash-approvals-dialog.js
KyleJu: `That is not going to happen if we simply mutate an existing property value.` 
> Yup that is the case for Polymer as well. Thanks for the explanation

static/elements/chromedash-approvals-dialog.js
jrobbins: The leading `?` controls whether the attribute is present or not.  It is normally written without a space.
https://lit.dev/docs/v1/components/templates/#bind-properties-to-templated-elements

KyleJu: Aha I see.

static/elements/chromedash-approvals-dialog.js
jrobbins: I rewrote these as conditional expressions (like `c ? x : y`) and I think that that is an improvement.  Thanks.

Issue #1629, opened 2021-12-02
Signed into chromestatus.com vs. www.chromestatus.com


jrobbins:
**Describe the bug**
Our app keeps separate session cookies for chromestatus.com vs. www.chromestatus.com, which can be very confusing.

**To Reproduce**
1. Click a link to https://www.chromestatus.com
2. Sign in
3. Click a link to https://chromestatus.com
4. See that you are not signed in.  You could even sign in to a different account.

**Expected behavior**
One session for the site, regardless of domain.  We should choose to make one version of the domain conical and redirect to that.

**Additional context**
I think this has not been noticed before because when a user types the URL, the browser autocompletes it to the form that that user had used before.  It is more evident when following links, e.g., from intent emails on blink-dev.

jrobbins:
Sure, to see the fix, take a look at pull request #1635.

This change is live on the site now.

Issue #632, opened 2019-03-31
Resolve www. vs naked domain


samthor:
Follow-on from #420.

We could redirect all users from www. to the naked domain, and serve a pure redirect there, so that folks don't have both SWs installed.

robdodson:
That seems like a good thing to do. If someone already has a sw installed for the www version do you know if that would interfere with things at all?

samthor:
We'd have to serve a blank SW from "www." so we'd force an update, so folks can get the real HTTP response, which'd redirect them.

It sounds like you prefer the naked domain @robdodson?

robdodson:
I personally prefer the naked domain but am very cautious about zombie service workers :D

PR #1635, opened 2021-12-04
Redirect from www.DOMAIN to bare domain.

jrobbins:
This should resolve issue #1629.

In this PR:
* In FlaskHandler.get(), check for `www.` in the request host name and if so, redirect to the full URL without the `www.`/
* For URLs that we generate, use the non-www form.

We only redirect GETs.  We allow POSTs and API GETs to be processes regardless of the domain name.  That will avoid a problem for anyone who is signed in on the www.chromestatus.com domain at the time that this change is deployed.  If they submit a form after the new version is live, it should work, but then redirect them after the form is processed.  They will see themselves logged out because of the domain change, but at least the form data will have been processed.


KyleJu (approved):

KyleJu:
I didn't know `www.chromestatus.com` and `chromestatus.com` are not canonical? I would assume if you sign in one, you sign in to both

jpmedley:
Addresses #632.

PR #1638, opened 2021-12-08
Show review activity in list

jrobbins:
This addresses some feedback from our demo to API Owners last week.  Basically, the tool should be more like their old spreadsheet and not require opening the approvals dialog quite so much.

In this PR:
* Fetch approval-related info in the list view it is showing the approvals columns.
* Render that info in some `<div>s` under the feature row.

This PR lays out the code and implements about half the info that I want to ultimately show.  The rest of it is left as TODOs for now.


static/elements/chromedash-feature-table.js
```diff
@@ -234,29 +264,87 @@ class ChromedashFeatureTable extends LitElement {
     return nothing;
   }
 
+  getEarliestReviewDate(feature) {
+    const configs = this.configs[feature.id];
+    const allDates = configs.map(c => c.next_action).filter(d => d);
+    if (allDates.length > 0) {
+      allDates.sort();
```
KyleJu, line 271: Is this a native `sort()` function? I couldn't find the API for it

KyleJu (approved):
LGTM with a comment

static/elements/chromedash-feature-table.js
jrobbins: Yeah, it is just the native sort() function:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

One thing to keep in mind about JS array sort() is that it sorts everything as strings, unless a conversion function is supplied.  That can give unexpected results when trying to sort numbers.  However, in this case, the data are dates in format YYYY-MM-DD, so sorting them as strings seems fine.

PR #1639, opened 2021-12-08
Hide non-working UI elements.

jrobbins:
In a previous PR I laid out some UI elements for quick-action buttons, but the buttons were not functional.
Let's hide those buttons and the thread links, until they actually work.


KyleJu (commented):
LGTM

PR #1640, opened 2021-12-08
Add navigational context for feature detail page back arrow.

jrobbins:
This addresses a request from last week's demo to API Owners.  When they click through a feature name link on the myfeatures pages, they see the feature detail page.  The request was to make the back arrow icon on the feature detail page navigate back to the myfeatures page in that case, rather than always going to the all features page.

In this PR:
* Add `?context=myfeatures` to the links generated on the myfeatures page
* Calculate the context link for the back arrow in feature.py (without blindly trusting any query string parameter)
* Use that context link for the back arrow icon href in the feature.html template.


KyleJu (commented):
LGTM

Issue #1647, opened 2021-12-09
Features sometimes show up twice in a box on the MyFeatures page


jrobbins:
**Describe the bug**
Entries in the myfeatures boxes are sometimes duplicated.  This is an intermittent error.      

**To Reproduce**
1. Repeatedly reload the myfeatures page.
2. Within a few tries you will notice that one of the features toward the end of the list shows up twice.
3. Reload again and the list is (probably) back to normal. 

The length of the list is unchanged, meaning that the duplicate row is shown instead of some other row.

**Expected behavior**
Each feature should only be listed once.  And, reloading the page without anyone editing the features should produce the same feature list each time.

**Additional context**
The intermittent nature indicates some kind of race condition.  However, the code for populating those boxes is pretty simple.

jrobbins:
The bug was that the local variable KEY had the wrong value.  So, cached features were being written to the wrong location in the ramcache.  Looking at the source code with that in mind, it becomes easy to spot the problem:
* KEY is being set in the first for-loop, and has a final value that is appropriate for the last feature ID in the list on the last iteration of that first for-loop.
* KEY is then used in the second for-loop, but it is never recalculated.  So, every retrieved feature is stored at the KEY location that was intended for the final feature in the list.

Our existing unit tests don't catch this defect because the code actually produces the correct result so long as it is run a single thread and the list feature ID in the list was a cache miss.   The promises may overlap in time, but we resolve them in order.  That means that if there are N iterations of the second for-loop, N-1 of them write incorrect features to the same entry in ramcache and then the final iteration actually overwrites those with the correct feature.

However, in the case where there are two requests being served concurrently:
* Processing of request A can be on one of the non-final iterations, meaning that the cache has the wrong feature written to it.
* Processing of request B can try to read that feature from the cache and actually get the wrong feature.

KyleJu:
> Our existing unit tests don't catch this defect because the code actually produces the correct result so long as it is run a single thread and the list feature ID in the list was a cache miss. The promises may overlap in time, but we resolve them in order. That means that if there are N iterations of the second for-loop, N-1 of them write incorrect features to the same entry in ramcache and then the final iteration actually overwrites those with the correct feature.

Great analysis!

jrobbins:
This fix was deployed on Thursday (4 days ago).

PR #1648, opened 2021-12-09
Fix incorrect caching logic that caused transient duplicates on myfeatures.

jrobbins:
This resolves issue #1647.

Basically, `Feature.get_by_ids()` was caching features using the wrong cache key.  But, the problem only shows up when there are concurrent requests.  See the issue comments for a full explanation.

In this PR:
* Fix the call to `memcache.set()` to use the correct cache key for the feature being cached.
* Rename both of the variables involved so that a similar error would be more obvious in the future.
* Add a unit test.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -839,8 +839,8 @@ class Feature(DictModel):
     futures = []
 
     for feature_id in feature_ids:
-      KEY = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, feature_id)
-      feature = ramcache.get(KEY)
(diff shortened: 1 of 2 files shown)


jrobbins:
I deployed this change to staging but did not make it live:
* You can repro the defect by signing in as an admin on staging and repeatedly reloading /myfeatures
* The defect does not repro on https://rc-dot-cr-status-staging.appspot.com/myfeatures

KyleJu (approved):

Issue #1644, opened 2021-12-08
Unable to comment on an approved item in "My Features"


dbratell:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**

In the "My Features" section I approved an item, and after that I was unable to save a log update. This may be related to Item #1641 where I was unable to change the status. Maybe both happened for the same reason. Or maybe not.

**To Reproduce**
I've only one item set to approved and I do not want to "approve" anything just to test so I will go from memory here.

I performed the following steps. many of them are probably unnecessary.

1. Opened an item by clicking the stamp icon.
2. Set the dropdown to approved.
3. Clicked the save button, though nothing seemed to happen.
4. Wrote a few words in case the save button needed that to work. 
5. Clicked the save button. Nothing seemed to happen.
6. Closed the item.
7. Opened the item by clicking the stamp icon.
8. Noticed that the item was set to approved but the comment was not in the log.

**Expected behavior**
The text I wrote should have been in the log.

jrobbins:
In the logs I see a bunch of 403 responses at around this time from a user using Opera 81.0.4196.27.   So, I think the underlying cause of this problem is probably that your Google Sign-In session expired.   That is our probably mistake rather than anything that you did.  

Our Google Sign-In sessions seem to last one hour, which is too short to do real work.  We need to figure out how to configure it to be longer and/or refresh the session as needed.  This might also be the right time for us to upgrade to Google Identity Services since Google Sign-In is deprecated.

Also, we don't show user-visible error messages when XHR requests fail.  I need to add that.

jrobbins:
A (poor) work-around for now is to sign out and sign back in if you notice things not working.

dbratell:
Closing this since it was just a symptom of issue 1649.

Issue #1502, opened 2021-08-30
Sign-In breaks silently and quickly


npm1:
**Describe the bug**
When trying to create a new feature, I signed in and the sign in timed out by the time I tried to submit this new feature description.

**To Reproduce**
1. Go to https://www.chromestatus.com/features
2. Click on sign in.
3. Click to add new feature
4. Forget about it and work on other things for a couple of hours.
5. Go back to the new feature page. No errors show up.
6. Write the feature description and click on submit.
7. See error page which says I'm not logged in. I lost what I wrote on the summary so have to do this again.

**Expected behavior**
The page should prompt me to sign in again if it expires. I would not expect it to expire in a few hours though, that seems inadequate. And if it does expire it should store the current state so that you can copy-paste the fields and don't lose your work.

PR #1649, opened 2021-12-14
Refresh Google Sign-In session.

jrobbins:
This should address issues #1502, issue #1644, and probably some others.

One underlying cause of problems for users is that the Google Sign-In session expires after 1 hour, however there is no UI indication that things are going wrong.  Google Sign-In does have a way to refresh the session to start the 1-hour limit over again, as documented in this reference page, however the reason why we need to use these calls is not spelled out anywhere in the Google Sign-In docs, it is only discussed on Stack Overflow.
https://developers.google.com/identity/sign-in/web/reference#googleusergetid

In this PR: 
* Add a new JS function refreshSession() that calls reloadAuthResponse() to get a new Google Sign-In session token that has a new expiration date 1 hour in the future.  And, call our login API to encode that into a flask session cookie and set that cookie so that the next form submission will work.
* Call the new function immediately after each page load and repeat it again every 10 minutes that the user sits on a page.  The reason to use setInterval() is so that it will be called if the user reactivates an idle tab, and to keep retrying if they have intermittent connectivity.
* Tell the user about any failed API call so that they will know to try again.

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -118,21 +118,44 @@ limitations under the License.
             window.location.reload();
         });
     }
-    
+
+    function refreshSession() {
+        console.log('refreshing google sigin-in session');
+        let cu = auth2.currentUser.get();
+        cu.reloadAuthResponse().then(function(unused_auth_response) {
+            // Calling this sets the flask session cookie.
+            csClient.signIn(auth2.currentUser.get());
+        }).catch(() => {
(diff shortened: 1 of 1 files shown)


jrobbins:
FYI, this is currently live on staging.

KyleJu (approved):

PR #1205, opened 2021-03-16
20210316 urlfetch to native python

shivamag00:
Merging this will close #1074 (Raised by @jrobbins )

- The repository has to be migrated to Python 3 from Python 2. For this, it was necessary to replace urlfetch to native python libraries. 
- I have used `requests` library of Python as suggested by [Google Cloud Docs ](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-outbound-requests#making_simple_https_requests)

PR #1651, opened 2021-12-21
Clean up requirements.txt

KyleJu:
Clean up requirements.txt. 

- Rearrange the dependencies in terms of Py2 and Py3
- Remove pins in #1205, #1335, #1141 
- Keeps the google-auth pin in #1365, where google-cloud-core 1.7.0 requires google-auth<2.0dev,>=1.24.0,

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,13 +9,11 @@ google-cloud-tasks
 google-cloud-ndb
 google-cloud-logging
 google-auth==1.24.0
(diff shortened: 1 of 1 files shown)


KyleJu:
@jrobbins Would be easier to review commit by commit

jrobbins (approved):
Thanks for cleaning up the cruft that was here.

Issue #1655, opened 2021-12-24
CSS ISSUE : larger Property names not breaking up and sliding into Percentage column.


nirajbhutada:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
CSS ISSUE : larger Property names not breaking up and sliding into Percentage column.
URL : https://chromestatus.com/metrics/feature/popularity

**To Reproduce**
1. Go to 'https://chromestatus.com/metrics/feature/popularity'
2. See error for larger Property names  like 'MixedContentInSecureFrameThatDoesNotRestrictMixedContent', 'HTMLMetaElementReferrerPolicyMultipleTokensAffectingRequest', etc.
3. PFA screenshot for the same.
![image](https://user-images.githubusercontent.com/4961480/147316861-c4cdb56a-9384-47b1-85ad-ce999b72ab1a.png)

**Expected behavior**
Large property names should wrap to 2nd line.
The word-break property specifies how words should break when reaching the end of a line.
![image](https://user-images.githubusercontent.com/4961480/147317483-bb12ed62-cfeb-406d-9380-d131efeb45fd.png)
Solution : Need to apply below CSS to property names.
{
    word-break: break-all;
}

**Additional context**

jrobbins:
This has been live for a while.  Thanks again.

PR #1656, opened 2021-12-24
#1655 fix-CSS ISSUE : larger Property names not breaking up and sliding into Percentage column. 

nirajbhutada:
CSS ISSUE : larger Property names not breaking up and sliding into Percentage column. GoogleChrome#1655 fix

Large property names should wrap to 2nd line.
The word-break property specifies how words should break when reaching the end of a line.
added { word-break: break-all; } to all inside

PFA screenshot for the same.
image
Expected behavior
Large property names should wrap to 2nd line.
The word-break property specifies how words should break when reaching the end of a line.
image
Solution : Need to apply below CSS to property names.
{
word-break: break-all;
}
Before Fix :
image
After Fix :
image

### DIFF
--- a/static/elements/chromedash-metrics.js
+++ b/static/elements/chromedash-metrics.js
@@ -58,6 +58,7 @@ class ChromedashMetrics extends LitElement {
       li > :first-child {
         flex: 1;
         margin-right: 10px;
+        word-break: break-all;
       }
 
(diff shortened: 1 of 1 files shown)


nirajbhutada:
closing this pull request as its blocking because of CLA issue

PR #1653, opened 2021-12-21
Upgrade Django to 2.2.25

KyleJu:
Upgrade Django to 2.2.25 by going through [release notes from 2.0 to 2.2.25](https://docs.djangoproject.com/en/3.2/releases/). The only relevant note that I am unsure of is

- [Silencing of exceptions raised while rendering the {% include %} template tag will be removed](https://docs.djangoproject.com/en/3.2/releases/2.1/#features-removed-in-2-1)

PR #1657, opened 2021-12-24
Further separate py2 code.

jrobbins:
This will help up make updates to the py2 and py3 requirements.txt files independently.

In this PR:
* Move sendemail.py and its tests into a new directory `py2/`.
* Move app-py2.yaml into `py2/` as well.  Change handler paths because `py2/` is the root directory of this service now.  And, add a `/py2` handler just to show that that service is responding.
* Update dispatch.yaml to route `/py2` health check handler to the default app (app-py2).
* Move testing_config_py2.py into `py2/`.
* Add a symbolic link from `py2/settings.py` up to the main `settings.py` file.
* Create a new `py2/requirements.txt` file with just the dependencies for our py2 code.  All of which are pinned.
* Update the py3 version of Django because that is what we were trying to do in the first place.
* Update `package.json` targets and scripts to use files in the new locations.

There are also some changes here to fix the way that cloud tasks are created, but those should disappear after my other PR is merged.


jrobbins:
This is currently live on staging and I was able to successfully make an edit and generate a notification email (via py2).

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # Also see requirements.dev.txt
```
KyleJu, line 3: Also remove the py2 libraries below

package.json
```diff
@@ -7,13 +7,13 @@
     "node": ">=6.0.0"
   },
   "scripts": {
-    "deps": "python3 -m pip install -t lib -r requirements.txt --upgrade",
+    "deps": "python3 -m pip install -t lib -r requirements.txt --upgrade; cd py2; python3 -m pip install -t lib -r requirements.txt --upgrade",
```
KyleJu, line 10:
```suggestion
    "deps": "python3 -m pip install -t lib -r requirements.txt --upgrade; cd py2; python2 -m pip install -t lib -r requirements.txt --upgrade",
```
Should this be python2 or I guess python3 has the backward compatibility to install py2 deps. I think the reason we can use python3 hers is because they are no overlapping dependencies between py2 and py3 services?

KyleJu: Inspired by this comment, I have a workaround for https://github.com/GoogleChrome/chromium-dashboard/pull/1653. I will install all dependencies in py3 but run in py2. WDYT? (Sorry for going back and forth on this discussion)

jrobbins: I was thinking of keeping it as python3 so that we are able to run `npm run deps` even on a workstation that has no python2 installed.  It seems to work.

requirements.txt
jrobbins: I added TODOs to explain why we still need them for a while.  

I think I could do an immediate follow-up PR to remove the need for the mock dep.   

We will still need flask for py3, but we could upgrade to a more recent version.  Upgrading flask might take some research, code changes, and testing.

KyleJu (approved):

PR #1662, opened 2022-01-06
Bring Django to the latest release

KyleJu:
Upgrade Django to 3.2.11 by going through release notes. The only relevant bit is 

- django.utils.html.escape() now uses html.escape() to escape HTML. This converts ' to `&#x27` instead of the previous equivalent decimal code `&#39` in [the 3.0 release](https://docs.djangoproject.com/en/3.2/releases/3.0/)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 # Also see requirements.dev.txt
 
 # Py3 Libraries
-Django==2.2.25
+Django==3.2.11
 funcsigs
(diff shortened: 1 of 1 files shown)


jrobbins:
As discussed today: This looks good, but let's hold it until after the upcoming deployment so that we get more operational experience with the django version that we have now before upgrading in the next step.

jrobbins (approved):

Issue #1659, opened 2022-01-03
Process change: question regarding non-open-source-code or APIs


chrishtr:
Please add the following bolded text to the intent-to-ship required questions list:

**Does the feature depend on any code or APIs outside the Chromium open source repository and its open-source dependencies to function? (e.g. server-side APIs, operating system APIs or closed-source code bundles)
Yes or no. If yes, explain why this is necessary.**

Proposal/approval evidence here:

https://groups.google.com/a/chromium.org/g/blink-api-owners-discuss/c/fbAyrB6fT-g/m/ZMLZ_JveEAAJ

past:
I wonder how likely it is that engineers encountering this question will take "operating system APIs" literally. I suspect the API owners don't care if the feature is using malloc, mmap or the like, but rather the more exotic operating system APIs that don't have equivalents elsewhere?

chrishtr:
Maybe we could rephrase as "operating system APIs tailored to this feature"?

past:
Yes, that seems clearer.

jrobbins:
This change was displayed back in January.

PR #1679, opened 2022-01-14
Add non_oss_deps field to guide and intent email template.

jrobbins:
This should resolve issue #1659.

In this PR:
* Add a new field `non_oss_deps` in models.py.
* Add it in guideforms.py  so that is shows up in the prepare-to-ship stage.
* Add it to guide.py so that it is saved when the form is submitted.
* Add it to intent_to_implement.html so that it is included in the intent email template.

### DIFF
--- a/templates/blink/intent_to_implement.html
+++ b/templates/blink/intent_to_implement.html
@@ -203,6 +203,16 @@
   {{feature.measurement|urlize}}
 {% endif %}
 
+{% if feature.non_oss_deps %}
+  <br><br><h4>Non-OSS dependencies</h4>
+  <div>Does the feature depend on any code or APIs outside the Chromium
+  open source repository and its open-source dependencies to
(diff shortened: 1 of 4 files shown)


templates/blink/intent_to_implement.html
```diff
@@ -203,6 +203,16 @@ <h4>Contact emails</h4>
...
+{% if feature.non_oss_deps %}
+  <br><br><h4>Non-OSS dependencies</h4>
+  <div>Does the feature depend on any code or APIs outside the Chromium
+  open source repository and its open-source dependencies to
+  function?</div>
+
+  {{feature.non_oss_deps|urlize}}
+{% endif %}
+
+
```
KyleJu, line 215:
```suggestion
```suggestion
```
```

KyleJu (approved):

PR #1690, opened 2022-01-20
Create codeql-analysis.yml

KyleJu:
Per https://github.blog/2022-01-19-reducing-security-risk-oss-actions-opensff-scorecards-v4/

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'javascript', 'python', 'ruby' ]
+        language: [ 'javascript', 'python' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
         # Learn more about CodeQL language support at https://git.io/codeql-language-support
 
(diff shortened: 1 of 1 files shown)


jrobbins (approved):

KyleJu:
The annoying part about this is it will make runs look "red" although `Code scanning results` is a non-blocking check

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,70 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'python', 'ruby' ]
```
KyleJu, line 35:
```suggestion
        language: [ 'javascript', 'python' ]
```

KyleJu:
> The CodeQL results look pretty good. Maybe we should consider them to be failed tests. I'll try to send you PRs to clear up some of the flagged security warnings later today. If we can resolve them all then your current configuration of this tool might be OK.

sounds good. I did find a way to make it [`optional`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error), but resolving them is even better

PR #1699, opened 2022-01-25
Addressed some code scanning warnings.

jrobbins:
This should resolve about 4 of the 10 warnings that are currently listed on
https://github.com/GoogleChrome/chromium-dashboard/security/code-scanning


KyleJu (approved):

jrobbins:
Cool, it looks like the analysis shows green when the changes in the PR don't have problems, even if there are other problems in the code elsewhere.  So, we should be able to keep this analysis as it is currently configured, even if we don't get around to resolving all the problems immediately.

PR #1700, opened 2022-01-25
Resolve some more code scanning issues.

jrobbins:
This should clear up the four items currently listed on 
https://github.com/GoogleChrome/chromium-dashboard/security/code-scanning

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -170,10 +170,6 @@ limitations under the License.
       startApp();
     }
 
-    function getQueryStringValue (key) {
-      return decodeURIComponent(window.location.search.replace(new RegExp("^(?:.*[&\\?]" + encodeURIComponent(key).replace(/[\.\+\*]/g, "\\$&") + "(?:\\=([^&]*))?)?.*$", "i"), "$1"));
-    }
-
     function promptSignIn(e) {
       e.preventDefault();
       e.stopPropagation();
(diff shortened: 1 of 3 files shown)


KyleJu (approved):

templates/_base.html
```diff
@@ -185,7 +181,8 @@
     // This happens when an anon visitor tries to view a page that
     // requires being signed in.  We cannot promptSignIn(), because
     // lacking a user-initiated event, the popup will be blocked.
-    if (getQueryStringValue("loginStatus") == 'False') {
+    const urlParams = new URLSearchParams(window.location.search);
+    if (urlParams.get("loginStatus") == 'False') {
```
KyleJu, line 185:
```suggestion
    if (urlParams.get("loginStatus") == null) {
```
The boolean in javascript should be true and false but here specifically we should use `null`, per. https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get

KyleJu (changes requested):
Left one comment

templates/_base.html
jrobbins: It's actually set as a string spelled out as "False" in the URL rather than being a boolean.  

It's defined here
https://github.com/GoogleChrome/chromium-dashboard/blob/main/settings.py#L67
and used here and a few other places
https://github.com/GoogleChrome/chromium-dashboard/blob/main/framework/permissions.py#L97

KyleJu: aha I see

KyleJu (approved):

Issue #1677, opened 2022-01-12
Implement unit tests for django templates


jrobbins:
Right now we have one unit test test for our templates.

We should have unit tests that touch every template, even if not every conditional within each template is covered yet.

jrobbins:
The one existing unit test that we have now is:
https://github.com/GoogleChrome/chromium-dashboard/blob/main/pages/featuredetail_test.py#L89

The basic format of such tests is:
* It is located with the tests for the handler class that uses that template
* In the setUp method: Create the handler, use the handler to generate template_data for some common usage case.
* In the test methods: Modify the template_data to exercise various aspects of the template, call render() with the template_data to run the template, then check that the resulting HTML text contains some expected substrings.

It would be good to have that basic level of testing for some of the other 35 HTML and email template files in the the templates/ directory.

Also, it would be good to have some basic HTML validation on the resulting HTML text.  Individual templates do not usually generate a full HTML response, they are just fragments.  But, each fragment should have more or less matching HTML elements, attributes should have matching quote marks, etc.  Maybe this could be done using an HTML parsing library.
https://docs.python.org/3.9/library/html.parser.html
https://github.com/html5lib/html5lib-python

dlaliberte:
OK, I'll work on this.

I don't see how to assign myself.  Maybe you need to assign me?  Or do I need to add myself as a contributor somehow?  I'm still pretty new to GitHub, as you can probably tell.

jrobbins:
I assigned this issue to you.  And, I also added you as a collaborator on this project with write access.

dlaliberte:
First part done: https://github.com/GoogleChrome/chromium-dashboard/pull/1715
I forgot to link to this issue, but we don't want to close it yet anyway.

dlaliberte:
All django templates have html validation tests now, as far as I can tell, except for the email notifier templates (update-feature-email.html, new-feature-email.html, and review-comment-email.html) since it  not clear to me how these are used, and testing is likely to be very different given the email context compared to normal web page usage.

We could leave this issue open until everything is done.

jrobbins:
We'll follow the test pattern that you established when we change from django to jinja2 templates.

PR #1715, opened 2022-02-05
Add test that template renders valid html.

dlaliberte:


dlaliberte:
If there is a parser error, the message is not very helpful for finding where the error is.

dlaliberte:
Do I need to add html5lib to some configuration file?

jrobbins:
> If there is a parser error, the message is not very helpful for finding where the error is.

Well, that is better than what we have now which is nothing.

jrobbins (approved):

dlaliberte:
This change was for: https://github.com/GoogleChrome/chromium-dashboard/issues/1677

PR #1729, opened 2022-02-10
Migrate sign-in Google Identity Services library

KyleJu:
Migrate Google sign-in to [Google Identity Services library](https://developers.google.com/identity/gsi/web). As a result of this change, user sign-in status to their Google Account and your app are independent of each other. Chromestatus manages user sign-in solely through session cookies.

This is a follow-up PR for #1773

- The new sign-in uses [popup mode](https://developers.google.com/identity/gsi/web/guides/migration#the_new_way), instead of redirect mode. 
- Remove the use of refresh and access tokens. The new sign-in uses a simplified [JWT token](https://developers.google.com/identity/gsi/web/reference/js-reference#CredentialResponse).
- Remove the old [session refresh logic](https://developers.google.com/identity/gsi/web/guides/migration#session_state_and_listeners). User sign-in status is independent of their Google Account.


templates/header.html
```diff
@@ -38,12 +38,10 @@
           </ul>
         </div>
       {% else %}
-        <!-- <div class="g-signin2" data-onsuccess="onSignIn"></div> -->
-        <div id="gSignInWrapper">
-          <div id="gSignInCustomBtn" class="customGPlusSignIn">
-              <span class="gSignInIcon"></span>
-          </div>
+        <div id="g_id_onload" data-client_id="{{google_sign_in_client_id}}" data-ux_mode="redirect"
+          data-login_uri="${window.location.hostname + '/login'}">
```
KyleJu, line 42: Replace this with [data-callback](https://developers.google.com/identity/gsi/web/reference/html-reference#data-callback)

KyleJu:
Seeing following errors:
```
Refused to load the script 'https://accounts.google.com/gsi/client' because it violates the following Content Security Policy directive: "script-src 'unsafe-inline' https://www.gstatic.com 'nonce-1QXK1E0KUdYb4mslVryclWqTzoGzd1'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
```

jrobbins:
> Seeing following errors:
> 
> ```
> Refused to load the script 'https://accounts.google.com/gsi/client' because it violates the following 
> Content Security Policy directive: "script-src 'unsafe-inline' https://www.gstatic.com 'nonce-1QXK1E0KUdYb4mslVryclWqTzoGzd1'". 
> Note that  'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
> ```

Maybe you could add 'https://accounts.google.com/gsi/client' to framework/csp.py line 34.

KyleJu:
> > Seeing following errors:
> > ```
> > Refused to load the script 'https://accounts.google.com/gsi/client' because it violates the following 
> > Content Security Policy directive: "script-src 'unsafe-inline' https://www.gstatic.com 'nonce-1QXK1E0KUdYb4mslVryclWqTzoGzd1'". 
> > Note that  'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.
> > ```
> 
> Maybe you could add 'https://accounts.google.com/gsi/client' to framework/csp.py line 34.

Yea I figured that part out and it worked. Although I wonder why `apis.google.com/` wasn't added there

KyleJu:
After configure csp, sign-in shows up on the page
Screen Shot 2022-02-15 at 3 30 49 PM (2)

jrobbins:
Nice.  Is this still a draft PR?

KyleJu:
> Nice. Is this still a draft PR?

Yea it is still a draft. I have to clean up the session refresh and figure out the callback

KyleJu:
Encountering following errors:

`[GSI_LOGGER]: The given origin is not allowed for the given client ID.`. Looks like I need to configure the localhost to the OAuth client ID credentials

KyleJu:
> Encountering following errors:
> 
> `[GSI_LOGGER]: The given origin is not allowed for the given client ID.`. Looks like I need to configure the localhost to the OAuth client ID credentials

`http://127.0.0.1:8080` already exists in the credentials. Adding `http://127.0.0.1` as well per https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid

KyleJu:
> > Encountering following errors:
> > `[GSI_LOGGER]: The given origin is not allowed for the given client ID.`. Looks like I need to configure the localhost to the OAuth client ID credentials
> 
> `http://127.0.0.1:8080` already exists in the credentials. Adding `http://127.0.0.1` as well per https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid

Also adding `http://localhost` on top of `http://localhost:8080`

KyleJu:
Able to get it to work with the latest change
Screen Shot 2022-02-17 at 5 51 52 PM (2)

templates/_base.html
```diff
@@ -111,70 +76,17 @@
...
-          }, function(error) {
-              console.log(error);
-          });
-    }
-
-    function onLoad() {
-      startApp();
-    }
-
     function promptSignIn(e) {
```
KyleJu, line 86: @jrobbins I am unclear about this function - How is this function being used? Now that we have a one-tap signin in a popup mode, do we still need this function?

jrobbins (changes requested):
Once the nonce={{nonce}} part is done, this should be good.

Also, I want to test to see if there is still a one-hour session expiration.  I'll try that now.

jrobbins:
> I think that we still need refreshSession(). 

Actually, maybe we are approaching this incorrectly.  Maybe we should not be trying to refresh the user's session on "Sign in with google".  This docs page says:
https://developers.google.com/identity/gsi/web/guides/overview

     User sign-in status to their Google Account and your app are independent of each other, 
     except during the sign-in moment itself when you know that the user has successfully 
     authenticated and is signed into their Google Account. Users may remain signed-in, 
     sign-out, or switch to a different Google Account while maintaining an active, signed-in 
     session on your website.

So, that would seem to mean that once our JS code calls /api/v0/login, that we should then just set the Flask session cookie and let it define the user session.  But, that means we need to decide what goes in that session cookie instead of an id_token, and we would need to make sure that its encryption protects against someone being able to forge a cookie.

KyleJu:
> But, that means we need to decide what goes in that session cookie instead of an id_token, and we would need to make sure that its encryption protects against someone being able to forge a cookie.

That is correct. Essentially, we have to figure out our "user identifier" and decide what goes into cookies and what is its TTL. The "user identifier" can be as simple as user's email and needs to be encrypted on the server side before set in the response.

KyleJu:
> I think that we still need refreshSession(). With this PR, I tried signing in, opening the form to create a new feature entry, and then leaving it there for an hour. The result was a 403 because I was signed out when hour elapsed.
> 
> I don't see an equivalent to the old `auth2.currentUser.get().reloadAuthResponse()`. However, I am thinking that we can probably find our how that method is implemented and just copy it into our code. I looked around but didn't find it yet. However, by using that old method temporarily, I did prove that the old id_token and the new credentials string are the same (or at least interchangeable).

I believe that anything related to the use of `auth2` has been [removed on the client side](https://developers.google.com/identity/gsi/web/guides/migration#object_migration_reference_for_user_sign-in). Per your response below, `User sign-in status to their Google Account and your app are independent of each other` after the migration

KyleJu:
> > But, that means we need to decide what goes in that session cookie instead of an id_token, and we would need to make sure that its encryption protects against someone being able to forge a cookie.
> 
> That is correct. Essentially, we have to figure out our "user identifier" and decide what goes into cookies and what is its TTL. The "user identifier" can be as simple as user's email and needs to be encrypted on the server side before set in the response.

If we use email address as the identifier, we can simply have 24-hour TTL and all we need to check if the access to our app has been revoked by user through [google.accounts.id.revoke](https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.revoke)`



[google.accounts.id.revoke](https://developers.google.com/identity/gsi/web/reference/js-reference#google.accounts.id.revoke)

jrobbins:
I think we still need to do some kind of periodic refresh rather than just choosing a long TTL.   If we choose a short TTL (even 24 hours) then users who leave a chromestatus tab open on Friday will find themselves signed out when they try to continue the following Monday.  If we choose a long TTL (like 7 or 30 days), it is not so secure because an attacker has more time to forge a valid token or steal one from the victim, and it is not a great user experience because they will feel that the randomly get signed out.   If we have an infinite TTL (basically don't check any expiration) I think that gives an attacker too much opportunity to forge or steal a cookie.   We do have a separate XSRF token that has a 2 hour TTL with working refresh logic, but those are only used for writes and they depend on knowing the current user's email address via the session.

In terms of how to encrypt the user's email address in the session, maybe we can still keep that in the same format as the id_token with the same level of security.   I am not so clear on how this all works, but it looks like the id_token or credentials string is a JWT (JSON Web Token) and there is a library for dealing with them that includes a refresh() method.
https://google-auth.readthedocs.io/en/stable/_modules/google/auth/jwt.html#Credentials.refresh

My first choice would be to keep the JWT issuer as accounts.google.com and somehow get that refresh() method to reach out to accounts.google.com to make the refreshed JWT.   I think that that *might* be what happens if things are set up correctly, but I am not sure.

My second choice would be for our app to make the refreshed JWT itself using some secret key stored in GAE (not part of our source code).

jrobbins:
After reading about this more, I don't think that accounts.google.com will generate a new token with a expiration date farther in the future.  So, my first choice seems like unlikely to be possible.

Minting our own JWT token seems possible, but it is a hassle.  We would need to have a secret key in PEM format stored in datastore.  This approach would be better than just relying on the Flask session, but I am not sure that it is enough of a difference to justify the complexity.

So, the most practical option would seem to be simply putting the user's email address into the Flask session.  I just looked at the flask.sessions code and it does sign the session with a secret key that we already have set up.   I think that is good enough.  I would still want it to have an expiration timestamp and refresh logic, but that expiration timestamp could just be another field in the flask session rather than dealing with JWTs.  Our login_api handler (or a separate refresh handler) could simply set a new expiration timestamp in the flask session.

What are your thoughts?  Would you like me to take a hack at implementing the change in the session management in a prequel PR?

KyleJu:
> After reading about this more, I don't think that accounts.google.com will generate a new token with a expiration date farther in the future. So, my first choice seems like unlikely to be possible.
> 
> Minting our own JWT token seems possible, but it is a hassle. We would need to have a secret key in PEM format stored in datastore. This approach would be better than just relying on the Flask session, but I am not sure that it is enough of a difference to justify the complexity.

I am unsure how to "mint" a JWT token. However I am still suspecting that there is a better way than deal with JWT Token. Looking at the [related documentation](https://developers.google.com/identity/protocols/oauth2#serviceaccount). There supposed to be a `refresh token` that never expires, but with limited scope, e.g. can only access to email address and name.
The concept of refresh token is similar to other OAuth flow, such as GitHub and it allows us to manage user sign-in status without any "refresh" logic, because it never expires. All we need to do is to check if user has revoked the permission of sign-in in our app. I am trying to figure out how we can exchange a `JWT` token into a refresh token.

> 
> So, the most practical option would seem to be simply putting the user's email address into the Flask session. I just looked at the flask.sessions code and it does sign the session with a secret key that we already have set up. I think that is good enough. I would still want it to have an expiration timestamp and refresh logic, but that expiration timestamp could just be another field in the flask session rather than dealing with JWTs. Our login_api handler (or a separate refresh handler) could simply set a new expiration timestamp in the flask session.

Yea I think storing email address into the flask session is a standard approach, similar to how we deal with GitHub sign-in in wpt.fyi. I am tempted to say that we can get rid of any sort of refresh logic for the JWT tokens, but implement a refresh logic for cookies.

> 
> What are your thoughts? Would you like me to take a hack at implementing the change in the session management in a prequel PR?

Let clarify a few things in a meeting before the prequel PR? I

templates/_base.html
KyleJu: Let's do this in a follow-up PR

jrobbins: I'll send a separate CL to change the signed-out star behavior.

jrobbins (approved):

PR #1773, opened 2022-03-04
Use new session contents

jrobbins:
This is a prequel PR for the switch from Google Sign-In to Sign in with Google.

In this PR:
* `login_api.py`: store a dictionary of user info that we need, plus a signature.
* `users.py`: use either the old JWT or our new user info from the session.  Separate the three code paths for ways to be signed in.
* `basehandlers.py`: remove an unused import.


jrobbins:
This is work-in-progress because I need to add unit tests.  But, you can understand the approach from this first commit.

jrobbins:
This is ready to review.

framework/users.py
```diff
@@ -185,28 +189,41 @@ def is_current_user_admin(self):
...
-      return current_user
-
-    token = session.get('id_token')
-    current_user = None
-    if token:
+        user_via_env = User(email=os.environ['USER_EMAIL'])
+      return user_via_env
+
+    # TODO(jrobbins): Remove this code path after 30 days.
+    jwt = session.get('id_token')
```
KyleJu, line 198: Is this for the legacy code `session["id_token"] = token`?

api/login_api.py
```diff
@@ -34,12 +32,14 @@ def do_post(self):
...
+      idinfo = id_token.verify_oauth2_token(
           token, requests.Request(),
           settings.GOOGLE_SIGN_IN_CLIENT_ID)
-      # userid = idinfo['sub']
-      # email = idinfo['email']
-      session["id_token"] = token
+      user_info = {
+          'email': idinfo['email'],
+          }
+      signature = xsrf.generate_token(str(user_info))
```
KyleJu, line 41: We should also encode `user_info`. User shouldn't able to view what the value for `signed_user_info` looks like

api/login_api.py
```diff
@@ -34,12 +32,14 @@ def do_post(self):
...
+      session['signed_user_info'] = user_info, signature
```
KyleJu, line 42: How often does this signature expire?

framework/users.py
jrobbins: Yes.  Because, when we deploy this, we don't want existing users to be suddenly signed out.  Instead, we keep accepting their current sessions.  All those current sessions should be long gone within 30 days.

api/login_api.py
jrobbins: Flask encrypts the entire session using app.secret_key (set in basehandlers.py).

That raises the question, why do we need to sign the user_info if it is already encrypted using some kind of secret?  Encryption + signing seems wasteful.  I think it is wise to sign the user_info separately so that we can implement key rotation on the signing key.  I don't see a way to implement key rotation for Flask's app.secret_key.

api/login_api.py
jrobbins: 10 days.  It is determined by the timeout specified when we verify the signature in `framework/users.py`.

(And, an implication of having it limited rather than infinite is that we should some kind of refresh logic, otherwise a user could be using our site for 10 days and then be suddenly signed out.   I think that we can add that into the existing code in `api/token_refresh_api.py` rather than trying to get a new id_token like we did with "sign-in with google".  That can be part of a future PR.)

api/login_api.py
KyleJu: I see. Well I can speak for the [securecookie library](https://github.com/gorilla/securecookie#securecookie) we use in wpt.fyi. This library `hashKey`, `blockKey` where `hashkey` uses for cookie singing and `blockkey` encrypts the value itself. I cannot find the documentation for the flask cookie encryption, so my question is does it do both encryption and signature? If so, I am open not to have a key rotation

jrobbins: I only found it by looking at the flask source code.  I don't see anything there about key rotation.  So, I think we will need to do it for `user_info` but let's see what they suggest in the security review rather than adding complexity now.

api/login_api.py
jrobbins: Whoops.  I was trying to address this comment by adding a 10 day expiration to the session cookie and I push that change to main rather than to this branch.  Please consider it to be part of this PR.  The commit is here:
https://github.com/GoogleChrome/chromium-dashboard/commit/cc182d544e75f0fd9e9e0494790c1f7cfbffae0a

api/login_api.py
KyleJu: Yea lets do that

KyleJu (approved):

PR #1783, opened 2022-03-10
Relabel owners from "Contact emails" to "Feature owners"

jrobbins:
Today, while answering a question for a user, I noticed that the existing UI the field to edit the set of feature owners is labeled "Contact emails" where it is edited, but it is displayed as "owners" on the feature detail page.  I always refer to it as owners.  We should make it consistent so that users will recognize the same field name in the list of available search fields.


cwilso (approved):
LGTM

PR #1784, opened 2022-03-10
Handle more prefixes and "+" in intent detection.

jrobbins:
This week there were a couple of intent email threads that were not correctly detected:
* One had the mailing list prefix for another mailing list.  I suppose that the feature owner discussed it on that list first.
* One used "+" instead of the word "and".

In this PR:
* Strip prefixes using regexes.  Also, strip "Fwd:" for forwarded messages.
* Replace "+" or "&" with "and", which simplifies the cases in the code below it.
* Normalize whitespace
* Add several more unit tests


jstenback (approved):
LGTM

KyleJu (approved):

PR #1785, opened 2022-03-12
Accept either old or new login requests during transition.

jrobbins:
This avoids errors caused by existing pages that users have open in their browser.  The old code tries to refresh the login session every 10 minutes by calling this API with the old parameter name.    

Also, if a user had an old page open and they sign in using the Google Sign-In button that is on that existing page, it will make a call to the login API using the old parameter.


KyleJu (approved):
Thanks for catching it!

api/login_api.py
```diff
@@ -28,7 +28,9 @@ class LoginAPI(basehandlers.APIHandler):
   """Create a session using the credential generated by Google Sign-In."""
 
   def do_post(self):
-    token = self.get_param('credential')
+    # TODO(jrobbins): Remove id_token after next deployment.
+    token = (self.get_param('id_token', required=False) or
+             self.get_param('credential'))
     message = "Unable to Authenticate"
```
KyleJu, line 34: Let make it more actionable. "Unable to Authenticate and please log in again"?

jrobbins: Done.

api/login_api.py
```diff
@@ -28,8 +28,10 @@ class LoginAPI(basehandlers.APIHandler):
   """Create a session using the credential generated by Google Sign-In."""
 
   def do_post(self):
-    token = self.get_param('credential')
-    message = "Unable to Authenticate"
+    # TODO(jrobbins): Remove id_token after next deployment.
+    token = (self.get_param('id_token', required=False) or
+             self.get_param('credential'))
+    message = "Unable to Authenticate.  Please sign in again."
```
KyleJu, line 34:
```suggestion
    message = "Unable to Authenticate. Please sign in again."
```

PR #1787, opened 2022-03-12
Remove obsolete pages: /schedule and /samples.

jrobbins:
Remove python, html, JS, and CSS for two pages that were removed from our site information architecture last year.
The samples page was removed from the site because of low usage and unclear usefulness.
The schedule page was replaced by the roadmap page.

There was one call from other code to a function defined in schedule.py, however that function (and its tests) were already copied into channels_api.py (and channels_api_test.py) in anticipation of deleting schedule.py.


jrobbins:
FYI, GAE logs show no usage of the HTML pages or the XML feed.  The JSON feed gets some hits, but some of the ones that I saw were from some kind of feed aggregator service and the client string mentioned "4 subscribers".  So, I think that no many people are using these feeds.  They might even have been long forgotten.

KyleJu (approved):

PR #1795, opened 2022-03-15
Updating Readme.md - fixing the broken lighthouse badge image

NinjaZMY:
In the Readme.md file:
lighthouse badge image isn't rendered so it was fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Chrome Platform Status
 ==================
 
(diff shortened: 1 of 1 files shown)


jrobbins:
Thanks for the reminder that this is broken.

However, it was not supposed to be a static image of a lighthouse.  It was supposed to be displaying the result of a compatibility test.  That test is not working and is probably not relevant now because we simplified the design of the app.  I'll just remove those lines from README.md.

PR #1801, opened 2022-03-22
Define namedtuples for progress items and actions.

jrobbins:
In this PR:
* Create named tuples for ProgressItem and Action and change the process definitions to use them.
* For most ProgressItems, specify the field that must be edited to complete that item.
* For each Action, list the prerequisite ProgressItems that must be completed to avoid the interstitial dialog (and therefore which fields must be edited) even if some of those progress items were from previous process stages.
* Update chromedash-process-overview.js to handle the new format of progress items and actions.
* Now that progress item fields are specified, offer the user an "Edit" link when they hover over a progress item.
* Remove title="Blink Process" because it shows up as a tooltip whenever the user hovers anywhere in the process.

### DIFF
--- a/internals/processes.py
+++ b/internals/processes.py
@@ -1,6 +1,3 @@
-
-
-
 # -*- coding: utf-8 -*-
 # Copyright 2020 Google Inc.
 #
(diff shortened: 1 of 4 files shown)


jrobbins:
This is still a work in progress.  I have made the changes to specify the progress item fields, but I still need to specify the action prerequisites.  And, I need to update chromedash-process-overview to handle the new format of the process JSON.  I'm sharing this now so that you can get a feel for what I have in mind.

dlaliberte:
Looks good so far. I was going to say, it seems you will need to update chromedash-process-overview, but you figured that out, of course.  I'm not sure how it could work without that update, however.

dlaliberte:
Looks good. I'm looking forward to integrating with my changes for the dialog.

dlaliberte (approved):

jrobbins:
OK, this is ready for review now.

I did have to change chromedash-process-overview to handle the new format since each progress item is another layer of JSON objects rather than just a string.  I also took advantage of the new "field" part of the progress item to offer a direct link to the field that needs to be edited.  

In this PR, I assume that each progress item refers to fields that are in the same stage as the progress item.  That's pretty easy.  To make your dialog box have "Edit" links that jump straight to the field that needs to be edited, you will need to add logic that loops over the stages looking for the one that has the needed progress item.  Then use that stage ID number in the link URL path.

PR #1803, opened 2022-03-23
Add `npm stop`.

dlaliberte:


jrobbins:
I think we still need dev-deps and test2 for those rare times in which we need to make edits to our remaining python 2 source code.   We would not be able to run those on our corporate workstation, but they are still run as part of CI and I could run them using py2 on my laptop.

`npm stop` looks useful.

dlaliberte:
OK, added the py2 scripts back in.  Useful exercise to see if I can do these simple git tasks via vscode.

jrobbins (approved):

PR #1804, opened 2022-03-23
Refresh user session when refreshing XSRF token.

jrobbins:
In this PR:
* Refactor code that creates the signed user_info in the session

*Call it from both login_api and refresh_token_api

### DIFF
--- a/api/token_refresh_api.py
+++ b/api/token_refresh_api.py
@@ -13,14 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-
-
 import logging
 
 from framework import basehandlers
 from framework import xsrf
-from internals import models
+from framework import users
 
 
 class TokenRefreshAPI(basehandlers.APIHandler):
@@ -41,8 +38,9 @@ class TokenRefreshAPI(basehandlers.APIHandler):
 
   # Note: we use only POST instead of GET to avoid attacks that use GETs.
   def do_post(self):
-    """Return a new XSRF token for the current user."""
+    """Refresh the session and return a new XSRF token for the current user."""
     user = self.get_current_user()
+    users.add_signed_user_info_to_session(user.email())
     result = {
         'token': xsrf.generate_token(user.email()),
         'token_expires_sec': xsrf.token_expires_sec(),
--- a/api/token_refresh_api.py
+++ b/api/token_refresh_api.py
@@ -40,7 +40,7 @@ class TokenRefreshAPI(basehandlers.APIHandler):
   def do_post(self):
     """Refresh the session and return a new XSRF token for the current user."""
     user = self.get_current_user()
-    users.add_signed_user_info_to_session(user.email())
+    users.refresh_user_session()
     result = {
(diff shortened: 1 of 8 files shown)


api/token_refresh_api.py
```diff
@@ -41,8 +38,9 @@ def validate_token(self, token, email):
 
   # Note: we use only POST instead of GET to avoid attacks that use GETs.
   def do_post(self):
```
KyleJu, line 40: Is this being called when a session cookie expires, every 10 days?

api/token_refresh_api.py
```diff
@@ -41,8 +38,9 @@ def validate_token(self, token, email):
...
-    """Return a new XSRF token for the current user."""
+    """Refresh the session and return a new XSRF token for the current user."""
     user = self.get_current_user()
+    users.add_signed_user_info_to_session(user.email())
```
KyleJu, line 43: Is user always signed up when this function is called? Can user be None?

jrobbins: Yes, the user will always be signed in at this point.  It will not be None.

If the user is not signed in, the code in basehandlers.py will generate a 403 response before this do_post() method is called.

KyleJu (approved):

api/token_refresh_api.py
jrobbins: Now that you mention it, I think I probably need to make sure this happens more often.

I added this session refresh to the existing XSRF token refresh requests.  Those happen whenever the user submits a form or causes the UI to do a POST request to the API, unless the XSRF token is less than two hours old.  So, if a user is actually working with the site they are likely to call this API method, but that is not always the case.  For example, they could simply navigate around for 10 days and then they would be suddenly signed out.

I'll add a call to setInterval() to refresh the session daily even if the user never makes any changes.  And, also refresh the session on every navigation.

jrobbins: I was only able to do half of what I wanted to do.

I was able to make it refresh the session on each page navigation.  That should be pretty good.  As long as the user does something to cause a new page load with the 10-day window, their session should be refreshed and they should stay signed in.

However, I was not able to make the setInterval() part work.  We use the fetch() method and the session cookie is httpOnly.  Apparently that combination makes it impossible for an API response from the server to set a cookie.  So, if a user opens a form, and sits there 10 days, they won't be able to submit the form.  That's not ideal, but is probably not going to be problem very often, especially since loading a chromestatus page in another tab would set the cookie.

KyleJu: I think you might need [`credentials`](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) header for fetch(), `credentials: "same-origin"`.  It is not set be default https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters. Regardless, the likelihood that someone leaves a form open for 10 days is small. So LGTM

jrobbins: We already set `credentials: "same-origin"`.  I am thinking that the httpOnly aspect of the session cookie is the reason, but I don't want to change that.   We'll see if this is good enough.

Issue #1774, opened 2022-03-04
Lists on /roadmap view should use case-insensitive sort


jpmedley:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
Sometimes a feature entry has a title that begins with a language keyword (JS, CSS, etc.). Such words cannot be capitalized as required by the English rules for titles. The current title sorting on the [/roadmap view](https://chromestatus.com/roadmap) does not take this into account. Any feature with an uncapitalized title appears after features with capitalized titles. (Sorting appears to be by unicode value, rather than English letter value.)

**To Reproduce**
There is an example of this on the [/roadmap view](https://chromestatus.com/roadmap) at the time of writing. If an example is no longer visible, see the image below.

![image](https://user-images.githubusercontent.com/8204171/156825001-49370d49-63b5-40ca-b1ae-e42829574814.png)

**Expected behavior**
All features will be sorted by English letter regardless of whether that letter is upper case or lower case.

Spectre-ak:
Hi, I've created a PR for this issue https://github.com/GoogleChrome/chromium-dashboard/pull/1809

jrobbins:
Thanks again.  This has been live for a while.

PR #1809, opened 2022-03-28
Added case insensitive sort for features list 

Spectre-ak:
Fixes Lists on /roadmap view should use case-insensitive sort #1774

Added case insentive sort for features list using custom comparator

`(a, b) => a.name.localeCompare(b.name, 'fr', {ignorePunctuation: true})`

Samples after sorting

image
image

### DIFF
--- a/static/elements/chromedash-roadmap.js
+++ b/static/elements/chromedash-roadmap.js
@@ -161,7 +161,7 @@ class ChromedashRoadmap extends LitElement {
       newMilestonesInfo[milestoneNum].version = milestoneNum;
       newMilestonesInfo[milestoneNum].features = milestoneFeatures[milestoneNum];
       Object.keys(newMilestonesInfo[milestoneNum].features).forEach(status => {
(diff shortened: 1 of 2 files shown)


static/elements/chromedash-roadmap.js
```diff
@@ -159,6 +160,9 @@ class ChromedashRoadmap extends LitElement {
     milestoneNumsArray.forEach((milestoneNum) => {
       newMilestonesInfo[milestoneNum].version = milestoneNum;
       newMilestonesInfo[milestoneNum].features = milestoneFeatures[milestoneNum];
+      Object.keys(newMilestonesInfo[milestoneNum].features).forEach(status => {
+        newMilestonesInfo[milestoneNum].features[status].sort( (a, b) => compareFeatures(a, b));
```
jrobbins, line 164: I think the last part of this line could just be `sort(compareFeatures)`.
Same thing for three other occurrences below.

jrobbins (changes requested):

jrobbins (approved):

Issue #1678, opened 2022-01-12
Set up javascript unit tests


jrobbins:
**Describe the solution you'd like**
* Implement some javascript unit tests.
* Choose a test runner and testing library.
* Configure an npm target to run the javascript unit tests.
* Add these JS tests to our GitHub CI.

jrobbins:
Thanks to Dan's work, we can now run JS unit tests locally.  But, we need to set it up in GitHub Actions CI.

dlaliberte:
Thanks also to Jason's work fixing the problems with my earlier attempts to add a web test runner test.  

This items in the description of this issue are now satisfield, so this issue can be closed.

PR #1816, opened 2022-04-06
Replace scss files used by chromedash elements with -css.js files.

dlaliberte:
In order to set up javascript unit tests of the lit elements, for #1678, we need to dynamically load css, which is apparently problematic with the available tools.  So we can avoid loading css by converting the scss code to use lit css`...` wrappers around the scss code.  

For this PR, no significant changes were required.   The nested import of _vars requires that we instead export SHARED_STYLES as an array of styles, which we then "spread" into the styles array in each lit element that uses it.


dlaliberte:
I still have a messed up process.  More previous commits are accumulating.  And now there is a conflict that I thought I just resolved, but still appears unresolved.

static/elements/chromedash-process-overview_test.js
```diff
@@ -0,0 +1,39 @@
+import {html, fixture, expect} from '@open-wc/testing';
+
+import './chromedash-process-overview.js';
+
+describe('MyElement', () => {
+  it('has a default title "Hey there" and counter 5', async () => {
+    const el = await fixture(html`
+      <chromedash-process-overview></chromedash-process-overview>
+    `);
+
```
dlaliberte, line 10: Oops, I didn't mean to include this.  I'll figure out how to remove it.

static/sass/_layout-css.js
```diff
@@ -0,0 +1,18 @@
+import { css } from "lit";
+
+export const LAYOUT_CSS = css`.layout {
```
jrobbins, line 3: This would be more clear if you added a newline right at the start of the tagged string.

static/sass/elements/chromedash-roadmap-css.js
```diff
@@ -0,0 +1,15 @@
...
+import { SHARED_STYLES } from "../shared-css.js";
+import { LAYOUT_CSS } from "../_layout-css.js";
+
+export const ROADMAP_CSS = [
+  SHARED_STYLES,
+  LAYOUT_CSS,
+  css`
+    :host {
+      display: inline-flex;
+      padding: 0 0em $content-padding * 5;
```
jrobbins, line 11: These are SCCS variables that would need to be replaced with JS constants or CSS variables.   There is no multiplication in CSS expressions, so you would need a constant that is 5 times the $content-padding value.  Or, you could just do the multiplication yourself and use 80px here.

static/sass/shared-css.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+  a:hover {
+    text-decoration: underline;
+    color: var(--link-hover-color);
+    cursor: pointer;
+  }
+
+  input:not([type="submit"]),
+  textarea {
+    border: 1px solid $bar-border-color;
```
jrobbins, line 58: This is an example of a SCSS variable that could be replaced with a CSS variable.  We define most of our CSS variables in theme.css.

static/sass/shared-css.js
```diff
@@ -0,0 +1,127 @@
...
+
+  .button.secondary {
+    margin: var(--content-padding);
+    border: var(--default-border);
+    border-color: var(--primary-border-background);
+    border-radius: var(--border-radius);
+  }
+
+  button:not(:disabled):hover {
+    border-color: $gray-4;
```
jrobbins, line 97: Another SCSS variable that needs to be replaced with a CSS variable.

shared-css.js
```diff
@@ -0,0 +1,116 @@
+import { css } from "lit";
```
jrobbins, line 1: Are you adding this file twice?  I see it both at the top level and under static/sass/

jrobbins (changes requested):

static/sass/elements/chromedash-roadmap-css.js
dlaliberte: Good catch!  I noticed in other places.  But the math would need some kind of expression evaluation.  I wonder if something like this might be possible css`... ${cssvar`--content-padding` * 5} ...`, assuming cssvar could be defined to do the right thing.

shared-css.js
dlaliberte: I assumed we need it in every file that uses the css literal.  The compiler (or import interpreter) ought to avoid importing something again if it has already been imported, but I don't know that it works that way.  The editor would not know about a higher-level import of css in whatever imports shared-css.js, so the editor ought to complain about css being undefined.   

I'm getting an obscure error now, after I started making more changes to add a first test, so I can't test what happens if I comment out this import.

static/sass/elements/chromedash-roadmap-css.js
dlaliberte: However, the next line with margin-right: $content-padding * -1  is more likely to be intended as relative to content-padding.  Maybe we should define more things similar to --content-padding-half, like --content-padding-negative, and --content-padding-huge.

shared-css.js
dlaliberte: Removed all my extra changes, got a new package-lock.json, etc, and that fixed the obscure error.  

Now I see that commenting out the import of lit for css doesn't complain.  But nor does it complain with the import.  It seems more correct to do the import, following the principle that every file should be correct on its own.

jrobbins: I was making a file-level comment on line 1 of this file rather than referring to the actual content of line 1.  Sorry for being unclear about thati.

It looks like you have mistakenly added two copies of shared-css.js to this PR: one at the top level and one under static/sass.

jrobbins (changes requested):

dlaliberte:
Ah, yes I was confused.  No idea how that top-level copy got created.  But it is curious that the import to define css is not needed, in the correct shared-css.js file.

jrobbins (approved):

jrobbins:
If you merge changes with `git merge main` you will need to update the new file that I just added `static/elements/chromedash-new-feature-list.js` to make it import SHARED_STYLES the same way.   Or, I could send you a small PR for it.

dlaliberte:
Got it.   Since the old shared.scss is still there, your code will work that way too, I expect.  But I can make the same change as for the other files.

PR #1818, opened 2022-04-07
Fix secure cookie header

jrobbins:
This is a follow-up to #1808 to correct a typo.  Also, only set these cookie details when not running locally because localhost is not using SSL.


jrobbins:
Merging because this is blocking local development and the fix is very straightforward.

PR #1819, opened 2022-04-07
Handle quoted search values.

jrobbins:
Instead of just breaking the user query into terms by splitting on spaces, we now look for all substrings that match one of these two patterns:
* Full-text search term: A single word or a quoted string
* A structured search term: A field name, operator, and then a single word or quoted string value.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -121,11 +121,21 @@ def process_queriable_field(field_name, operator, val_str):
   return promise
 
 
+# A full-text query term consisting of a single word or quoted string.
+# The single word case cannot contain an operator.
+# We do not support any kind of escaped quotes in quoted strings.
 TEXT_PATTERN = r'[^":=><! ]+|"[^S]+"'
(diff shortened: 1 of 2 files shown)


internals/search.py
```diff
@@ -121,19 +121,21 @@ def process_queriable_field(field_name, operator, val_str):
   return promise
 
 
-FIELD_NAME_PATTERN = r'[.a-z_0-9]+'
+TEXT_PATTERN = r'[^":=><! ]+|"[^S]+"'
```
dlaliberte, line 124: Since regex notation is often very difficult to read, it would be good to describe what it does in a less precise but easier to understand english comment.

dlaliberte (approved):

internals/search.py
jrobbins: Good suggestion.  I added the comments.

dlaliberte (approved):
Great comments.  Thanks.

PR #1821, opened 2022-04-07
Set up the new feature list page initially as a feature table.

jrobbins:
This is the starting point for the new feature list page.   The new search widgets are not added yet, those will come in a separate PR.

In this PR:
* main.py: Register a new page for server-side routing at url `/newfeatures`.
* pages/newfeatures.py: A basic handler for the new page that provides no page data.
* templates/new-feature-list-page.html: a simple template that just contains one Lit element.
* static/componoents.js: Add the new page-level custom element to the list of files to compile.
* static/elements/chromedash-new-feature-list.js:  New page-level element that just contains a plain old feature table so far.
* static/js-src/new-feature-list-page.js: remove the spinner and fetch the list of features that the user has starred.  Someday this will be in a top-level SPA element, but for now it is the same small bits of JS for most pages.

### DIFF
--- a/static/elements/chromedash-new-feature-list.js
+++ b/static/elements/chromedash-new-feature-list.js
@@ -1,0 +1,86 @@
+import {LitElement, css, html} from 'lit';
+import './chromedash-approvals-dialog';
+import './chromedash-feature-table';
+import SHARED_STYLES from '../css/shared.css';
+
+
+class ChromedashNewFeatureList extends LitElement {
+  static get properties() {
+    return {
+      signedInUser: {type: String},
+      canEdit: {type: Boolean},
+      canApprove: {type: Boolean},
+      loginUrl: {type: String},
+      starredFeatures: {type: Object}, // will contain a set of starred features
+    };
+  }
+
+  constructor() {
+    super();
+    this.signedInUser = ''; // email address
+    this.starredFeatures = new Set();
+    this.canEdit = false;
+    this.canApprove = false;
+  }
+
+  static get styles() {
+    return [
+      SHARED_STYLES,
+      css`
+      `];
+  }
+
+  // Handles the Star-Toggle event fired by any one of the child components
+  handleStarToggle(e) {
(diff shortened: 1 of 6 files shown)


static/elements/chromedash-new-feature-list.js
```diff
@@ -0,0 +1,86 @@
...
+    this.signedInUser = ''; // email address
+    this.starredFeatures = new Set();
+    this.canEdit = false;
+    this.canApprove = false;
+  }
+
+  static get styles() {
+    return [
+      SHARED_STYLES,
+      css`
```
dlaliberte, line 29: Is an empty css needed here, or is this just a placeholder?

static/elements/chromedash-new-feature-list.js
```diff
@@ -0,0 +1,86 @@
...
+      .then(() => {
+        if (e.detail.doStar) {
+          newStarredFeatures.add(e.detail.featureId);
+        } else {
+          newStarredFeatures.delete(e.detail.featureId);
+        }
+        this.starredFeatures = newStarredFeatures;
+      })
+      .catch(() => {
+        alert('Unable to star the Feature. Please Try Again.');
```
dlaliberte, line 46: Do we want a browser alert here, or is this something that normally never gets used?

dlaliberte (approved):
LGTM

static/elements/chromedash-new-feature-list.js
jrobbins: It is just a placeholder.  I suppose that I could delete it, but I kind of like having consistency among all the custom elements.

static/elements/chromedash-new-feature-list.js
jrobbins: It should almost never happen so I went with the simplest handler.  We also have a "toast" element, so I suppose that I could have used that.  I'll keep that in mind if eventually improve our error handing.

PR #1823, opened 2022-04-08
Fix icons not visible in some views

### LINKED ISSUE #1802: Icons not visible in some views

**Describe the bug**
In views smaller than desktop, I cannot see the action icons. 

image

**To Reproduce**
1. Open the site.
2. Shrink the window. Eventually, the icons will cease to be visible.

image

**Expected behavior**
Icons should be visible at all times.

kevinshen56714:
This is to fix #1802 where the action icons disappear in small views.

In this PR:
- feature.html: Added a new css class `tooltips` for the div that wraps the action icons.
- main.scss: Excluded `tooltips` class from `display:none` in small view.

Before:

After:

### DIFF
--- a/static/sass/main.scss
+++ b/static/sass/main.scss
@@ -420,7 +420,7 @@ footer {
 
   #subheader {
     div {
-      &:not(.search):not(.actionlinks) {
+      &:not(.search):not(.actionlinks):not(.tooltips) {
         display: none;
       }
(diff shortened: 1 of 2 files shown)


jrobbins (approved):

jrobbins:
Thanks for contributing this!

PR #1824, opened 2022-04-09
Use datalist instead of select for timeline pages

kevinshen56714:
This is to address #1342 and provide a search friendly dropdown list for the timeline pages.

In this PR:
- chromedash-timeline.js: Changed `<select>` to `<input>` + `<datalist>` and updated the event and API handler logics (so that API requests still use bucket ID number not the displayed property name).
- shared-css.js: Excluded the datalist input from `border: 1px solid var(--bar-border-color);`

Results:
Screen Shot 2022-04-09 at 5 04 03 PM
Screen Shot 2022-04-09 at 5 01 39 PM

(I manually set the input box width to 560px, which is about the same width of current select element, please let me know if it's fine)

### DIFF
--- a/static/elements/chromedash-timeline.js
+++ b/static/elements/chromedash-timeline.js
@@ -86,7 +86,7 @@ class ChromedashTimeline extends LitElement {
       }
 
       #datalistinput {
-        width: 560px;
+        width: 20em;
       }
     `];
   }
@@ -104,7 +104,7 @@ class ChromedashTimeline extends LitElement {
   }
 
   updateSelectedBucketId(e) {
-    const feature = this.props.find((el) => el[0] === e.currentTarget.value);
+    const feature = this.props.find((el) => el[1] === e.currentTarget.value);
     if (feature) {
       this.selectedBucketId = feature[0];
     }
@@ -266,7 +266,7 @@ ORDER BY yyyymmdd DESC, client`;
       <input id="datalistinput" type="search" list="features" placeholder="Select or search a property" @change="${this.updateSelectedBucketId}" />
       <datalist id="features">
         ${this.props.map((prop) => html`
-          <option value="${prop[1]}" />
+          <option value="${prop[1]}" dataset-debug-bucket-id="${prop[0]}" />
         `)}
       </datalist>
       <label>Show all historical data: <input type="checkbox" ?checked="${this.showAllHistoricalData}" @change="${this.toggleShowAllHistoricalData}"></label>
--- a/static/elements/chromedash-timeline.js
+++ b/static/elements/chromedash-timeline.js
@@ -84,6 +84,10 @@ class ChromedashTimeline extends LitElement {
         background: var(--table-alternate-background);
         display: inline-block;
       }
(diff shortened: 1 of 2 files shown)


kevinshen56714:
@jrobbins please review and let me know if this is good. Thank you.

KyleJu (approved):
LGTM on the code. Leave it to Jason to review the feature itself

static/elements/chromedash-timeline.js
```diff
@@ -84,6 +84,10 @@ class ChromedashTimeline extends LitElement {
         background: var(--table-alternate-background);
         display: inline-block;
       }
+
+      #datalistinput {
+        width: 560px;
```
jrobbins, line 89: Let's go with 20em here.

static/elements/chromedash-timeline.js
```diff
@@ -253,12 +263,12 @@ ORDER BY yyyymmdd DESC, client`;
...
     return html`
-      <select .value="${this.selectedBucketId}" @change="${this.updateSelectedBucketId}">
-        <option disabled value="1">Select a property</option>
+      <input id="datalistinput" type="search" list="features" placeholder="Select or search a property" @change="${this.updateSelectedBucketId}" />
+      <datalist id="features">
         ${this.props.map((prop) => html`
-          <option value="${prop[0]}">${prop[1]}</option>
-          `)}
-      </select>
+          <option value="${prop[1]}" />
```
jrobbins, line 269: This is good.  Not specifying a body to the option avoids the two-line display that we hit on a previous attempt to use datalist.

For debugging, I do like to be able to see the bucket IDs spelled out when I inspect this element.  So, please add a datalist-debug-bucket-id="${prop[0]}" here.

jrobbins (changes requested):
Sorry for the delay in reviewing this.
Overall this looks good and I think that we could merge it if you make the suggested changes.

static/elements/chromedash-timeline.js
```diff
@@ -100,7 +104,10 @@ class ChromedashTimeline extends LitElement {
   }
 
   updateSelectedBucketId(e) {
-    this.selectedBucketId = e.currentTarget.value;
+    const feature = this.props.find((el) => el[0] === e.currentTarget.value);
```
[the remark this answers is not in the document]
kevinshen56714: My mistake, have changed it to check el[1].

static/elements/chromedash-timeline.js
kevinshen56714: Changed it to 20em

static/elements/chromedash-timeline.js
kevinshen56714: Added the ```dataset-debug-bucket-id```

kevinshen56714:
@jrobbins Thanks for reviewing. I have made the requested changes.

jrobbins:
Thanks again.

jimmywarting:
tested it today, it got so much more user friendlier, thanks 🙂 👍

Issue #1817, opened 2022-04-06
[Process improvement] Add intent-to-ship question about open interop/compat-impacting issues


chrishtr:
Reference: https://groups.google.com/a/chromium.org/g/blink-api-owners-discuss/c/Jzij3fKN5kA/m/I6nqUxUZBwAJ

Please add this required question to the intent-to-ship template:

_Please list known spec issues which may cause web compat risk if resolved (e.g. resulting in change to naming or structure of the API: <answer here>_

chrishtr:
We're also making the following changes to OTs:

_1. Remove the requirement to ask for a gapless Origin Trial
2. Any Origin Trial may request up to N milestones with the current policy (i.e. ~ no spec required, no TAG review required; only need an explainer and description of how the API works).
3. Extensions of the Origin Trial beyond N can only request up to M at a time. Approval of the first extension is blocked on evidence of the following:
  * A draft spec (early draft is ok, but must be spec-like and associated with the appropriate standardization venue, or WICG)
  * A TAG review in progress
  * [bit.ly/blink-signals](http://bit.ly/blink-signals) requests
  * Substantial evidence of outreach for feedback from the community
  * Progress on WPT tests_

@jrobbins does anywhere in chromestatus have to be changed to reflect the above? I'm planning to make the relevant changes to the Chromium site now.

jpmedley:
What do you mean by 'can only request up to M at a time'?

chrishtr:
N=6 and M=3 in the published version of the text.

It means "you can start an OT for up to 6 releases with the usual rule; each extension can be for at most 3 releases and needs substantial progress towards an I2S each time.

jpmedley:
Jason is currently implementing a way to force entry of fields based on other states. For example, this would enforce the fields need to be done before an I2S can be sent. What was not considered was that requirements might be different between the first and second OT. @jrobbins would know better, but I don't think that can be enforced without more coding.   There's nothing to prevent starting another origin trial. But the Chrome owners will need to police the requirements.

On a side note, Jason is about to implement a new way of showing milestones on the rendered card which will show end milestones and gaps.

jrobbins:
I have a PR pending.  I tweaked the wording to make it more clear to me:
"Please list known spec issues whose resolution may introduce web compat risk (e.g., change to naming or structure of the API)."

Let me know if I got that wrong.

chrishtr:
I added a comment on the landed PR.

[...]

PR #1832, opened 2022-04-12
Add field for anticipated spec changes before i2s.

jrobbins:
This should resolve issue #1817.   There's a lot of FYI context in that issue, but the actual requested change is to add a field to the database which then shows up in the intent-to-ship emails.

Changes in this PR:
* models.py: Add the anticipated_spec_changes field and define a query term for it
* guideforms.py: Define the django form field for anticipated_spec_changes and locate it on the "prepare to ship" form, flat form, and on the feature detail page.
* processes.py: indicate that anticipanted_spec_changes should be shown on the intent-to-ship email, but not necessarily on other emails.
* intent_to_implement.html: This template is used for all intent emails.  Add a conditional to show anticipated_spec_changes if it has a defined value or if the user is preparing an intent-to-ship email.

Also in this PR: Correct the display name of another field to use sentence-case capitalization.

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -386,7 +386,7 @@ ALL_FIELDS = {
          'when designing this feature.')),
 
     'webview_risks': forms.CharField(
-        label='WebView Application Risks', required=False,
+        label='WebView application risks', required=False,
         widget=forms.Textarea(attrs={'cols': 50, 'maxlength': 1480}),
         help_text=
         ('Does this intent deprecate or change behavior of existing APIs, '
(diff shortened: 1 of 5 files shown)


pages/guideforms.py
```diff
@@ -500,6 +500,16 @@ def validate(self, value):
         ('If your feature was available as an origin trial, link to a summary '
          'of usage and developer feedback. If not, leave this empty.')),
 
+    'anticipated_spec_changes': forms.CharField(
+        required=False, label='Anticipated spec changes',
+        widget=forms.Textarea(
+            attrs={'rows': 4, 'cols': 50, 'maxlength': 500,
+                   'placeholder': 'https://\nhttps://'}),
+        help_text=
+        ('Please list known spec issues whose resolution may introduce '
```
chrishtr, line 509: @glenrob gave some feedback on Monday that the text I proposed was unclear to non-insiders, e.g. my wording of "spec issues" was vague, and he wasn't sure what the point of the question was. How about:

"Open questions about a feature may be a source of future web compat or interop issues. Please list open issues (e.g. links to known github issues in the project for the feature specification) whose resolution may introduce web compat/interop risk (e.g., changing to naming or structure of the API in a non-backward-compatible way)."

The purpose of the first sentence is to make clear the *why* of this question, and the secnod is the what, but with a bit more examples.

PR #1833, opened 2022-04-13
Add index for new ot_milestone_webview_start field.

jrobbins:
This is a follow-up to my recent PR #1830.

Because a query is done using a combination of conditions on ot_milestone_webview_start, ot_milestone_desktop_start, and name, datastore requires that exact index to be specified in index.yaml and the index built on the server.

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -111,6 +111,11 @@ indexes:
   - name: "name"
 - kind: "Feature"
   properties:
+  - name: "ot_milestone_webview_start"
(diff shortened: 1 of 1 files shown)


jrobbins:
I have already applied this change to the staging and production datastore indexes and it seems to work.

Here's a page on prod that relies on the new index:
https://1ce851f-tainted-dot-app-py3-dot-cr-status.appspot.com/roadmap
Since I already built the index, that page is working

dlaliberte (approved):

PR #1837, opened 2022-04-15
Replace gantt diamond with boxy strips to clarify OT gaps

jrobbins:
Replace the feature milestone gantt chats that we currently show with boxy strips that are labeled with the milestone type and numbers.  Also, on small screens, show it as a bullet list rather than a chart.

In this PR:
* Rewrite the chart generation to use `<ul>` and `<li>` tags.
* Add grid-row and grid-column properties to position the elements.  The labeled element comes last in the DOM so that it is above the other elements so that long labels are visible.
* Use CSS media query to style the elements as a chart when the screen is large enough.
* Compute the set of grid column headers (which are not actually shown) based on the milestones that have values.  Try three approaches to scale to different situations.
* Delete the images used for the old gantt chart background bars.


jrobbins:
Some screenshoots:
image

image

image

dlaliberte (approved):

PR #1840, opened 2022-04-16
Remove highlight on mouse leave from roadmap

Spectre-ak:
Added mouse leave event to remove highlight from features on roadmap section.

This PR is for a new feature and not related to any issue.


jrobbins (approved):

jrobbins:
Thanks for doing this.  Small details like this can improve users' feeling about the site.

PR #1842, opened 2022-04-18
Add shoelace and gulp-concat. Replace process-overview dialog with shoelace dialog

dlaliberte:
This PR adds the dependency on @shoelace-style/shoelace, and we initially use its [dialog component](https://shoelace.style/components/dialog) and [button component](https://shoelace.style/components/button).

To use the shoelace light theme, via its light.css file, this PR adds a "css" task to the gulp config, and concat's the .css file into base.css, which is then imported via _base.html.


static/components.js
```diff
@@ -11,6 +11,15 @@ import '@polymer/paper-listbox';
 import '@polymer/paper-ripple';
 import '@polymer/paper-styles/color.js';
 
+// Shoelace components
+// css is imported via _base.html in base.css, built by gulpfile.babel.js.
+import '@shoelace-style/shoelace/dist/components/button/button.js';
+import '@shoelace-style/shoelace/dist/components/dialog/dialog.js';
+import { setBasePath } from '@shoelace-style/shoelace/dist/utilities/base-path.js';
+
+// Set the base path to the folder you copied Shoelace's assets to
```
dlaliberte, line 20: This comment about the shoelace assets and the path used with setBasePath should be changed to whatever is appropriate for our installation.  I'm not sure yet whether the assets (images) are available this way.

templates/_base.html
```diff
@@ -23,6 +23,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
 
   <meta name="theme-color" content="#366597">
+
+  <link rel="stylesheet" href="/static/css/base.css" />
```
jrobbins, line 27: Since you have this, can you remove line 45? (the one that calls inline_file on main.css)

jrobbins (changes requested):

dlaliberte:
Here is what it looks like when I click "Draft Ready for Trial email":

image

We should probably add text to the dialog that mentions what this list of missing prerequisites is for.  In this example, it is for drafting the "Ready for Trial" email, but you can't tell that from the dialog.

templates/_base.html
dlaliberte: But base.css doesn't include main.css.  We could probably also append main.css in base.css.  How about I do that in another PR?

static/components.js
dlaliberte: I'll fix the assets as you suggest in an upcoming PR, when I actually need the assets for something.

templates/_base.html
jrobbins: Ah, my mistake.  I thought that your new "css" rule was concatenating two things together, but it is really just copying one source to one destination.

jrobbins (approved):

dlaliberte:
A couple more TODOs for this dialog:

 * The dialog position is horizontally and vertically centered in the container, but It should instead be positioned near the mouse.  I don't see any options for positioning the dialog.
 * The dialog is not draggable.  It always annoys me when a dialog that covers content I might need to see cannot be repositioned.
 * The container (the whole overview table) is grayed out while the dialog is open.   Not sure it makes sense because you can close the dialog by clicking anywhere outside of it. This might be OK, but if we want to keep it, then the entire page should be grayed out.

PR #1845, opened 2022-04-18
Replace chromedash-dialog with sl-dialog.

dlaliberte:
This PR replaces the remaining uses of chromedash-dialog with sl-dialog, and deletes chromedash-dialog.


dlaliberte:
The chromedash-approvals-dialog uses sl-dialog now and looks the same:

image

And the features legend looks like this:

image

static/elements/chromedash-approvals-dialog.js
```diff
@@ -77,37 +76,37 @@ class ChromedashApprovalsDialog extends LitElement {
     return [
       ...SHARED_STYLES,
       css`
-        .loading {
+        :host .loading {
```
jrobbins, line 79: Is the :host needed?  I thought everything in the shadow DOM is under :host.

dlaliberte: You're right.  I was confused,thinking my styles were starting to work once I added the :host context.  But it must have been for some other reason.

jrobbins (approved):

PR #1846, opened 2022-04-19
Replace chromedash-accordion with sl-details

dlaliberte:
This PR replaces the chromedash-accordion component with sl-details.  The style is almost identical except for the toggle icon being on the right side, which might be better anyway since the other control-icons on feature lines are also on the right.


dlaliberte:
The features list looks like this:

image

and with one row closed: 

image

The "Additional fields.." section looks like this:

image

dlaliberte:
The default animation when hiding is a combo fade and shrink.  

I removed the style change for the sl-icon since the default aligns better, even though it is considerably larger than current iron-icon style.

PR #1847, opened 2022-04-20
Extract URLs from text fields before rendering

### LINKED ISSUE #1484: URL fields should handle extraneous information

**Describe the bug**
The form fields that require URLs allow other text that breaks links. The image below shows this for a feature that's about to ship. The feature owner was allowed to enter '(merged)', which the rendered card makes part of the URL. Following this "link" returns a 404.

image

**To Reproduce**
1. In the staged test version of the site, open any feature.
2. Enter text like the following in any field intended to hold a URL: "`https://github.com/whatwg/html/pull/6562 (merged)`".
3. Save the entry.
4. On the rendered card attempt to follow the link.

**Expected behavior**
Here are a couple of options. 
* disallow non-URL text in URL fields.
* Extract URLs from text fields and render them before concatenating additional text.

kevinshen56714:
This is to fix #1484. Currently, the form fields that require URLs allow incomplete urls or other text that breaks links (even though all URL fields are validated on entry).

In this PR, I enhanced the ```parse_link``` function in ```basehandler.py``` to extract URLs from text fields and render them before concatenating additional text.

As an example, if you enter these urls:

Before (broken links)

After (either removed or trimmed)

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -413,11 +413,11 @@ class FlaskHandler(BaseHandler):
   def parse_link(self, param_name):
     link = flask.request.form.get(param_name) or None
     if link:
-      if not link.startswith('http'):
-        link = str('http://' + link)
-      else:
-        link = str(link)
-    return link
+      # see https://www.regextester.com/93901 for url regex
+      url_pattern = (r'\b([\w_-]+(?:(?:\.[\w_-]+)+))'
+                     r'([\w.,@?^=%&:/~+#-]*[\w@?^=%&/~+#-])?\b')
+      match_obj = re.search(url_pattern, str(link))
+    return 'http://' + match_obj.group() if match_obj else None
 
   def parse_int(self, param_name):
     param = flask.request.form.get(param_name) or None
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -417,9 +417,12 @@ class FlaskHandler(BaseHandler):
       url_pattern = (r'\b(http:\/\/|https:\/\/)?([\w_-]+(?:(?:\.[\w_-]+)+))'
                      r'([\w.,@?^=%&:/~+#-]*[\w@?^=%&/~+#-])?\b')
       match_obj = re.search(url_pattern, str(link))
-      link = match_obj.group() or None
-      if not link.startswith(('http://', 'https://')):
-        link = 'http://' + link
+      if match_obj:
+        link = match_obj.group()
+        if not link.startswith(('http://', 'https://')):
+          link = 'http://' + link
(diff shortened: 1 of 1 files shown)


kevinshen56714:
@jrobbins please review this. Thank you.

jpmedley:
Awesome! Thanks you.

framework/basehandlers.py
```diff
@@ -413,11 +413,11 @@ def split_emails(self, param_name):
...
-      if not link.startswith('http'):
-        link = str('http://' + link)
-      else:
-        link = str(link)
-    return link
+      # see https://www.regextester.com/93901 for url regex
+      url_pattern = (r'\b([\w_-]+(?:(?:\.[\w_-]+)+))'
+                     r'([\w.,@?^=%&:/~+#-]*[\w@?^=%&/~+#-])?\b')
+      match_obj = re.search(url_pattern, str(link))
+    return 'http://' + match_obj.group() if match_obj else None
```
jrobbins, line 420: Here, you are making every URL use the `http://` scheme.   It would be better to keep the `https://` scheme if the user entered it.    And, in the case where the user entered no scheme or something that we do not support (like `ftp`, `file`, or `about`), then use `http://`.

kevinshen56714: I modified the regex slightly so that it reserves the ```http://``` or ```https://``` of the URL. And now we concatenate the link with ```http://``` only when it does not start with ```http://``` or ```https://```.

framework/basehandlers.py
```diff
@@ -413,10 +413,13 @@ def split_emails(self, param_name):
...
-      if not link.startswith('http'):
-        link = str('http://' + link)
-      else:
-        link = str(link)
+      # see https://www.regextester.com/93901 for url regex
+      url_pattern = (r'\b(http:\/\/|https:\/\/)?([\w_-]+(?:(?:\.[\w_-]+)+))'
+                     r'([\w.,@?^=%&:/~+#-]*[\w@?^=%&/~+#-])?\b')
+      match_obj = re.search(url_pattern, str(link))
+      link = match_obj.group() or None
+      if not link.startswith(('http://', 'https://')):
```
jrobbins, line 421: This will cause an error if link is None.

jrobbins (changes requested):

framework/basehandlers.py
kevinshen56714: I made a slight change, now the script would check if there's actually a ```match_obj``` before moving on to process it.

jrobbins (approved):

jrobbins:
Thanks for doing this and making the requested changes.

PR #1849, opened 2022-04-21
Fix placement of dialog on process-overview. Misc cleanup.

dlaliberte:
The whole problem with the placement of the dialog on the process overview, which was being centered in the table and the table was grayed out, was caused by the presence of the "contains: content" style, which creates a containing block for position: fixed, as used y the dialog.   See: https://developer.mozilla.org/en-US/docs/Web/CSS/contain

### DIFF
--- a/static/elements/chromedash-process-overview.js
+++ b/static/elements/chromedash-process-overview.js
@@ -1,5 +1,4 @@
 import {LitElement, css, html, nothing} from 'lit';
-import '@polymer/iron-icon';
 import './chromedash-callout';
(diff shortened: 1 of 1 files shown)


dlaliberte:
This PR also moves the dialog into the renderAction method, which avoids some complexity in how data was passed to the dialog.  Also the "pending" terminology was wrong - it should have been changed to "prerequisits".

I also added some text to the dialog, to give it sufficient context to be clear on its own.

dlaliberte:
Here is how it looks with this PR:

image

jrobbins (approved):

PR #1850, opened 2022-04-21
Added href to chromedash-x-meter

### LINKED ISSUE #1543: chromedash-x-meter doesn't produce valid links
**Describe the bug**
From [metrics](https://www.chromestatus.com/metrics/feature/popularity) you can access a page that shows more detail on each metric.  You can only do that by a simple click because the chart, which uses a custom `chromedash-x-meter` element, does not render the link in a way that is visible to browsers.

**To Reproduce**
1. Go to https://www.chromestatus.com/metrics/feature/popularity
2. Right click on any of the graph parts.
3. [BUG] The browser does not recognize these as links.  There is no option to copy a link or open it in a new tab/window.
4. Left click on the graph parts, outside of the box where the text percentage is rendered.
5. [BUG] The browser does not navigate, though it shows a cursor of style 'pointer'.

**Expected behavior**
The intent is that these act as links to pages with more detail.  A cursor style of 'pointer' is set and a left click will result in an onclick handler being invoked with the browser navigating to the page identified in the href attribute of the element, if the percentage is clicked on.

If these were truly links then browser functions like "open in new tab" would be able to function.

**Additional context**
This is a very heavy page (Firefox reports this as taking almost 5 times more memory than an active GMail tab).  Being able to open detail pages in a new tab is a useful part of managing the costs associated with navigation to and from this page, even with a bfcache.

Spectre-ak:
Fixes #1543 chromedash-x-meter doesn't produce valid links

Wrapped chromedash-x-meter html inside `<a>` tag and added text-decoration: none on hover.
I also removed showTimeline function, because the href will be able to redirect on click 

Before 
Screenshot (22)

After
Screenshot (23)

### DIFF
--- a/static/elements/chromedash-x-meter.js
+++ b/static/elements/chromedash-x-meter.js
@@ -38,19 +38,29 @@ class ChromedashXMeter extends LitElement {
         padding: 3px 0;
         text-indent: 5px;
       }
-    `];
-  }
+      
+      a {
+        text-decoration: none;
+        color: var(--barchart-color);
+      }
 
-  showTimeline() {
-    window.location.href = this.href;
+      a:hover {
+        text-decoration: none;
+        color: var(--barchart-color);
+        cursor: pointer;
+      }
+
+    `];
   }
 
   render() {
     return html`
-      <div @click="${this.showTimeline}" style="width: ${(this.value / this.max * 100)}%">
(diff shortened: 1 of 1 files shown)


Spectre-ak:
@jrobbins please review this. Thanks

jrobbins (approved):
Nice.  It's always better to use native `<a>` elements so that middle-click, tab order, keyboard navigation, and the "copy link" context menu item work.

PR #1851, opened 2022-04-22
Initial test of using Web Test Runner

dlaliberte:
Got the basics working with minimal test.  Note there is a significant difference between using the 'fixture' method with the type parameter or without.


jrobbins (approved):
Thanks for doing this.
I'll try to learn more about web-test-runner today.

Issue #1716, opened 2022-02-07
"Intent to Ship LGTMs by" field shows validation error for multiple emails


scheib:
When entering Intent to Ship LGTM details there is a field which requests "Comma separated list of full email addresses of API owners who LGTM'd the Intent to Ship email thread."  However, when multiple emails are entered an error is shown "Enter a valid email address."

Screenshot:

> ![image](https://user-images.githubusercontent.com/386971/152840827-ac9f6136-12a0-4ec9-b3fe-fc31d7e54e98.png)

**To Reproduce**
1. Enter a single email address into the "Intent to Ship LGTMs by" field.
2. Save, and go back to edit, notice that the email is accepted.
3. Complete entering the multiple LGTM emails, comma separated.
4. Save, and go back to edit, notice the error message, "Enter a valid email address."

**Expected behavior**
Comma separated emails are accepted without error.

past:
Fixed by #1870 and also this field is deprecated now.

PR #1870, opened 2022-05-04
Add styling for invalid form field values

dlaliberte:
In response to #1716, this PR adds styles to highlight invalid form field values.  The form fields already have the appropriate type="email" or type="url" attributes, and also "multiple" is generally true.  Consequently, the values are automatically validated, once entered.  Empty strings are not invalid, unless a form field is required.

The added styles include the ":invalid" pseudo-class.  We also include :not(:focus) so the style change does not show up until you leave the form field, since it is rather distracting to always see errors while still entering a value. 

Finally, we append " (currently invalid)" after the help text.

### DIFF
--- a/static/sass/_vars-css.js
+++ b/static/sass/_vars-css.js
@@ -16,6 +16,7 @@ export const VARS = css`
 
   --bar-shadow-color: rgba(0, 0, 0, .065);
   --bar-border-color: #D4D4D4;
+  --error-border-color: #FF0000;
 
   --chromium-color-dark: #366597;
   --chromium-color-medium: #85b4df;
--- a/static/sass/_vars.scss
+++ b/static/sass/_vars.scss
@@ -12,6 +12,7 @@ $nav-link-color: #444;
 
 $bar-shadow-color: rgba(0, 0, 0, .065);
 $bar-border-color: #D4D4D4;
+$error-border-color: #FF0000;
 
 $chromium-color-dark: #366597;
 $chromium-color-medium: #85b4df;
--- a/static/sass/shared-css.js
+++ b/static/sass/shared-css.js
@@ -59,6 +59,15 @@ export const SHARED_STYLES = [
   textarea {
     border: 1px solid var(--bar-border-color);
   }
+  input:not([type="submit"]):not([type="search"]):invalid:not(:focus) {
+    outline: 1px dotted var(--error-border-color);
+    background-color: #FFEDF5;
+  }
+  input:not([type="submit"]):not([type="search"]):invalid:not(:focus) ~ span::after {
+    color: $error-border-color;
+    content: ' (Currently invalid)';
+  }
+
   input:not([type="submit"])[disabled],
   textarea[disabled],
   button[disabled] {
--- a/static/sass/shared.scss
+++ b/static/sass/shared.scss
@@ -45,6 +45,14 @@ a:hover {
 input:not([type="submit"]), textarea {
   border: 1px solid $bar-border-color;
 }
+input:not([type="submit"]):not([type="search"]):invalid:not(:focus) {
+  outline: 1px dotted $error-border-color;
+  background-color: #FFEDF5;
+}
+input:not([type="submit"]):not([type="search"]):invalid:not(:focus) ~ span::after {
+  color: $error-border-color;
+  content: ' (Currently invalid)';
+}
 input:not([type="submit"])[disabled], textarea[disabled], button[disabled] {
   opacity: 0.5;
 }


dlaliberte:
Here is what it looks like for an email field:

image

Similar for url fields, but the allowed format is probably too generous and we should probably add a pattern.

jrobbins:
I like that you were able to do this with just CSS.   And, I kind of like the additional background color to make it more obvious that a field has failed to validate.  However, I think it would be better to stick with the browser default style for invalid input.  The browser default is kind of subtle, but it is standard.   So, let's close this PR and go with the other one that you are working on.

dlaliberte:
> However, I think it would be better to stick with the browser default style for invalid input.

I don't see any browser default style for invalid input.  I do see a tooltip-like error message pointing at the form field when I forget to include a ',' between two email addresses and then try to submit the form, saying "A part following '@' should not contain the symbol ' '." but that rather indirect as well as confusing.  Is this the default style you mean?   If we can get that error message to show up as soon as the user leaves the field, that would be better.

The style change I made not only changes the background color but also adds a dotted outline.  This is more subtle, but at least it is a visible change.   I also added the ::after " (Currently invalid)" text, partly as a challenge, but it is not a terribly robust way of indicating an invalid value.

So if the only default style for invalid input is that error message in the tooltip when submitting the form, I don't think that is good enough.  Some kind of visible style change and/or error display should show up immediately, at least when leaving the form field, though probably not while still editing the field.

dlaliberte:
> If we can get that error message to show up as soon as the user leaves the field, that would be better.

I notice now that, when the input is invalid according to the browser, the text of same error message is displayed in a tooltip when you hover over the input field, but the style is just a default tooltip.  There are a few ways we can deal with this better.  See 

* https://stackoverflow.com/questions/38930221/styling-default-validation-tooltip
* https://www.telerik.com/blogs/building-html5-form-validation-bubble-replacements

jrobbins:
I was wrong about the default browser styling for invalid input values.  You are right that there is no such default.   When I make a small edit in the form field for intent-to-ship LGTMer emails, I see a thin green or red outline, like this:
image
image
However, those are not from the browser, I see that they are set in our own forms.scss file:
https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/sass/forms.scss#L51

So, let's go with your color indicators.

Regarding displaying the message with an ::after pseudo element.  I think that placing it after the help text makes it seem to be related to the help text.  It would be better if it could be placed directly to the right or below the input field, even if it pushes the help text down a little.  The phrase "(Currently invalid)" would not need the parenthesis if it is more clearly attached to the input element rather than the other text.

In that phrase, the word "currently" seems like it would be clear from context because the phrase appears when the user has interacted with the field.   Two words is probably the right length, but I think a better word would be "format" to clarify that it is the syntax of the value that is invalid rather than the semantics (e.g., the user might wonder if the person identified by that email address is not eligible to LGTM an intent).

No need to customize the email validator.  Just using `MultiEmailField` should be close enough.  
https://github.com/GoogleChrome/chromium-dashboard/blob/main/pages/guideforms.py#L26

dlaliberte:
I forgot to post this yesterday: I noticed we have border coloring for :valid and :invalid states in static/sass/forms.scss

```
  .interacted:valid {
    border: 1px solid green;
  }

  .interacted:invalid {
    border: 1px solid $invalid-color;
  }
```

The 'interacted' class is added by feature_form.js. I'm not sure why I wasn't seeing this border change previously.

dlaliberte:
> No need to customize the email validator. Just using `MultiEmailField` should be close enough.

The MultiEmailField function will generate the display on the server side regarding the invalid format of the email field, but it does so only after the form has been previously submitted and when the page is redisplayed.  Furthermore, the message doesn't disappear immediately after the user corrects the value, but only after the form is submitted again and the page is redisplayed.  So relying only on MultiEmailField lacks the immediate feedback about the error, and also the immediate feedback about the correction.

So to avoid this problem, we do need to customize the email validation done in the browser.  (Alternatively, if there was a way to immediately submit each form field change and get it validated on the server side, and then immediately update the display regarding the validation without redisplaying the whole page, that would work, but it is much more complex.)

jpmedley:
It seems like this discussion is beyond email addresses. Email validation may get you something for free. But don't you want the response to invalid input to be consistent across the app? Wouldn't that mean that you need to design for situation s where the browser provides nothing and apply those idioms to all fields whether the browser provides something or not?

dlaliberte:
> Doesn't using `<input type="email" mutliple>` set up the validation that we need?

It does set up validation, but it is insufficient validation.  It is too generous about the format, more so than the validation done by MultiEmailField, since it allows "f@f", for example.  That's why I said above that we would need to do the same validation as MultiEmailField but in the browser, possibly by using the input pattern attribute.

> Oh, we could make MultiEmailField do that so that we don't need to specify the widget on each field.

Not sure what you mean.  The django email.html template is used now, via the forms.EmailInput call, which simply includes the input.html tag, which generates the input tag for the email field as well as other types of fields. I think the MultiEmailField() wrapper call also does the server-side validation and adds the error message if that is appropriate, but I am not quite sure how it catches the error and adds the message.

jrobbins:
OK, try using the pattern attribute.  However, I don't think that the client-side and server-side validation regexes need to be perfectly aligned.  For the most part, the oddball input cases that they might treat differently would occur rarely in practice.

pages/guideforms.py
```diff
@@ -55,6 +55,17 @@ def validate(self, value):
   'Write it from a web developer\'s point of view.\n\n'
   'Follow the example link below for more guidance.')
 
+# This email address pattern is more restrictive than actually allowed, but it covers most reasonable uses.
+EMAIL_ADDRESS_PATTERN = '[ ]*([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9\-\.]+)\.([a-zA-Z]{2,5}){1,25}[ ]*'
```
jrobbins, line 59: In the last part for the top-level domain: `([a-zA-Z]{2,5}){1,25}`, why have both `2 to 5` and `1 to 25` rather than a single `2 to 125` constraint or `2 or more` written as `{2,}`?

pages/guideforms.py
```diff
@@ -188,10 +198,9 @@ def validate(self, value):
         help_text=('The spec document has details in a specification language '
                    'such as Web IDL, or there is an exsting MDN page.')),
 
-    'spec_mentors': forms.EmailField(
+    'spec_mentors': MultiEmailField(
         required=False, label='Spec mentor',
-        widget=forms.EmailInput(
-            attrs={'multiple': True, 'placeholder': 'email'}),
+        widget=forms.EmailInput(attrs=MULTI_EMAIL_FIELD_ATTRS),
```
jrobbins, line 203: Can this widget constructor argument be specified once in a new MultiEmailField constructor with a call to super() rather than each time?

pages/guideforms.py
```diff
@@ -55,6 +55,17 @@ def validate(self, value):
...
+# This email address pattern is more restrictive than actually allowed, but it covers most reasonable uses.
+EMAIL_ADDRESS_PATTERN = '[ ]*([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9\-\.]+)\.([a-zA-Z]{2,5}){1,25}[ ]*'
+
+EMAIL_ADDRESSES_PATTERN = '^' + EMAIL_ADDRESS_PATTERN + '(,' + EMAIL_ADDRESS_PATTERN + ')*$'
+
+MULTI_EMAIL_FIELD_ATTRS = {
+    'title':"Enter one or more comma-separated complete email addresses.",
+    'multiple': True, 
+    'placeholder': 'address1@domain.com, address2@chromium.org',
+    'pattern': EMAIL_ADDRESSES_PATTERN
```
jrobbins, line 67: If we use `pattern`, should remove `multiple`?

pages/guideforms.py
```diff
@@ -55,6 +55,17 @@ def validate(self, value):
...
 
+# This email address pattern is more restrictive than actually allowed, but it covers most reasonable uses.
+EMAIL_ADDRESS_PATTERN = '[ ]*([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9\-\.]+)\.([a-zA-Z]{2,5}){1,25}[ ]*'
+
+EMAIL_ADDRESSES_PATTERN = '^' + EMAIL_ADDRESS_PATTERN + '(,' + EMAIL_ADDRESS_PATTERN + ')*$'
+
+MULTI_EMAIL_FIELD_ATTRS = {
+    'title':"Enter one or more comma-separated complete email addresses.",
+    'multiple': True, 
+    'placeholder': 'address1@domain.com, address2@chromium.org',
```
jrobbins, line 66: I think the placeholder should use the word `user` rather than `address`, because the whole thing is the address rather than just that one part.

pages/guideforms.py
```diff
@@ -55,6 +55,17 @@ def validate(self, value):
   'Write it from a web developer\'s point of view.\n\n'
   'Follow the example link below for more guidance.')
 
+# This email address pattern is more restrictive than actually allowed, but it covers most reasonable uses.
+EMAIL_ADDRESS_PATTERN = '[ ]*([a-zA-Z0-9_\-\.]+)@([a-zA-Z0-9\-\.]+)\.([a-zA-Z]{2,5}){1,25}[ ]*'
+
+EMAIL_ADDRESSES_PATTERN = '^' + EMAIL_ADDRESS_PATTERN + '(,' + EMAIL_ADDRESS_PATTERN + ')*$'
```
jrobbins, line 61: I don't think that you need the `^` and the `$` because those are implicit.
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#constraint_validation

pages/guideforms.py
jrobbins: Also, a email address username part can contain `+` and `=`.    It can actually contain more kinds of characters than that, but I think those would be good enough because we are getting into very uncommon cases if we complicate it more than adding those.

jrobbins (changes requested):

pages/guideforms.py
dlaliberte: I missed that. Thanks.

pages/guideforms.py
dlaliberte: Adding + and =.  

I copied this email address pattern from stackoverflow.  I think it is not so great, now that you mention these issues.  I'll look at the default django pattern again, since it is mentioned as being comprehensive and well tested.

pages/guideforms.py
dlaliberte: I believe the multiple attribute is still required because the default pattern is still used by the input element, and it needs to know that it should use the default pattern in a multi-email list wrapper pattern, which, by the way, assumes the delimiter is ','.  The pattern attribute, however, needs to do its own multi-email list wrapper.  Not sure about ignoring whitespace.

BTW, I only found this out by experimenting.  I don't see it in the MDN documentation.  The "pattern" page doesn't mention "multiple", and vice versa, the "multiple" page doesn't mention "pattern".  It does seem inconsistent, however, and perhaps this reflects a bug in Chrome.

pages/guideforms.py
dlaliberte: I'm not sure, but it sounds reasonable.   There is more in the python implementation of the super() that I didn't want to try to figure out just for this PR.

pages/guideforms.py
jrobbins: If we keep `multiple` then we should probably specify `type="email"` too, otherwise there is nothing to answer the question "multiple what?".

You might also test it out with and without `type` and `multiple`.   caniuse.com says that `pattern` is widely supported, and I _assume_ that `pattern` supersedes those, but it would be nice to know for sure.  If you like, we can leave that for a future PR.

pages/guideforms.py
dlaliberte: I can't figure out the django pattern.  But here is a pretty good site for several variations of email patterns:   https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s01.html

jrobbins: That's a good resource.  I would pick their "Simple, with all valid local part characters"
```
[A-Z0-9_!#$%&'*+/=?`{|}~^.-]+@[A-Z0-9.-]+
```

It just needs `A-Za-z` in two places to make it case-insensitive.

And... I suggest taking out the `!` and `%`, because those have special meaning for some older email routing systems. They could be more of a risk of being exploited than is justified by their very rare legitimate usage. https://tldp.org/LDP/nag/node189.html
If I'm wrong about that and users complain, we can always add them back, but I haven't seen those characters used on our site so far.

Likewise, allowing the single quote and backtick seem like asking for trouble.

So that would lead to this pattern for an individual address:
```
[A-Za-z0-9_#$&*+/=?{|}~^.-]+@[A-Za-z0-9.-]+
```

dlaliberte: OK, also removing '|' from the user part since that can also be dangerous, according to the o'reilly doc.

And I made the domain part require at least two components, the last of which must be only 2-6 letters.

PR #1875, opened 2022-05-07
Remove CORS all permissive header

KyleJu:
Fixed one of the security suggestions from the security review.  Remove CORS all permissive "*" header

### DIFF
--- a/framework/basehandlers_test.py
+++ b/framework/basehandlers_test.py
@@ -279,7 +279,6 @@ class ConstHandlerTests(testing_config.CustomTestCase):
     actual_text, actual_status, actual_headers = actual_tuple
     self.assertIn('Hi Guest,', actual_text)
     self.assertEqual(200, actual_status)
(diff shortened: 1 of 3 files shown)


framework/basehandlers_test.py
```diff
@@ -522,7 +519,6 @@ def test_get__html_page(self):
 
     self.assertIn('Hi testing', actual_html)
     self.assertEqual(200, actual_status)
-    self.assertIn('Access-Control-Allow-Origin', actual_headers)
```
jrobbins, line 525: It would be better to use assertNotIn().  We don't want to "not care" if the header is there, we want to affirm that it is absent.

jrobbins (changes requested):

framework/basehandlers_test.py
KyleJu: Yea that's right. Thanks!

jrobbins (approved):

PR #1885, opened 2022-05-15
Add run webtest to ci.yml

dlaliberte:
This may be all that is required to run the webtest for each PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,3 +46,6 @@ jobs:
 
       - name: build
         run: npm run build
+
+      - name: webtest
+        run: npm run webtest


dlaliberte:
This much appears to be working.  I started setting up more playwright test code, but it looks like we don't need that yet.

.github/workflows/ci.yml
```diff
@@ -47,5 +47,7 @@ jobs:
       - name: build
         run: npm run build
 
+      - run: npx playwright install-deps
```
jrobbins: Should this use the `run: |` syntax like some of the other rules use?

jrobbins (approved):
LGTM after you consider one question.

.github/workflows/ci.yml
dlaliberte: On  [Workflow syntax for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions), it says

```
Each run keyword represents a new process and shell in the runner environment. When you provide multi-line commands, each line runs in the same shell.
```

The "|" notation is for running multi-line commands in the same shell.  I suspect that since these are installation steps, they have side effects on the global environment so they don't need to run in the same shell.  

This this PR works as is, I'd rather first submit that, and then we can try alternatives.  Maybe all the installation steps can be run in one shell, for example.

PR #1891, opened 2022-05-17
Show "Edit" links for active stage and in missing prereqs dialog

dlaliberte:
As per discussion in WebStatus meeting, the "Edit" links should be visible even when not hovering in the "missing prereqs" dialog and in the active stage.


dlaliberte:
image

jpmedley:
👏

jrobbins (approved):

PR #1892, opened 2022-05-17
Remove app-layout and rduce unneeded polymer deps

dlaliberte:
This PR will redo #1757 but also replace the use of app-layout (app-header-layout and app-header elements) with divs.  Everything seems unchanged.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,14 +6,9 @@
     "": {
       "name": "chromestatus-dashboard",
       "dependencies": {
-        "@polymer/app-layout": "^3.1.0",
         "@polymer/iron-collapse": "^3.0.1",
         "@polymer/iron-icon": "^3.0.1",
         "@polymer/iron-iconset-svg": "^3.0.1",
-        "@polymer/paper-item": "^3.0.1",
-        "@polymer/paper-listbox": "^3.0.1",
-        "@polymer/paper-ripple": "^3.0.2",
-        "@polymer/paper-styles": "^3.0.1",
         "lit": "^2",
         "node-fetch": ">=3.2.4",
         "node-sass": ">=4.13.1",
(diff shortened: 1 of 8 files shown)


jrobbins:
OK, this looks simple enough.  I am going to check it out and test it in a a few browsers to see if I spot anything.

jrobbins:
I am seeing the same visual problems as when I tried this change months ago.    

Specifically, most of the headings have very thin-line fonts that look not-great in any browser but look especially bad in chrome.  

This is how it looks with this change:
image

And, this is the way that it is supposed to look:
image

The thin font is harder to read and some of the characters are look a little mis-shaped because some parts of the letters look darker than the rest of the strokes:  
image
These shape problems do not occur without the changes in this PR.

I had no luck tracking down exactly what is causing these differences in fonts.  The font-* CSS properties seem to be identical between this change and main: `font-family: Roboto, sans-serif; font-size: 25pt; font-weight: 300;`.  And, yet, it looks different.  So, maybe it is some font related CSS property that does not start with `font-`.

dlaliberte:
OK, thanks for your detailed response.  I thought the changes would be more dramatic, but these are more subtle. There is indeed some difference, probably involving substitute fonts that are the right size, but rendered slightly differently.  Should be interesting trying to track this down.

jrobbins:
It is something to do with
`import '@polymer/paper-item';  `
If I start with the code in head, and then comment out that one line, I see the weak fonts.

Also, I tried a super-simple stand-alone HTML page that has 
```
<body style="background: rgb(250, 250, 250)">
    <h2	style="                                                                                                                                                                  
          font-family: Roboto, sans-serif;                                                                                                                                       
          font-size: 25px;                                                                                                                                                       
          font-weight: 300;                                                                                                                                                      
          color: rgba(0, 0, 0, 0.62);"
      >This has the overly weak font</h2>
</body>
```
and it shows the weak font.

So, the weak font is normal on chrome for font-weight:300.  It is something in paper-item that makes the font look OK even though we still use font-weight:300.

jrobbins:
A-ha!  I think that I figured it out.   `paper-item` must indirectly include `@polymer/font-roboto/roboto.js', which has JS to dynamically add a `<link>` element that loads a specific set of versions of the roboto font.

Here is a stand-alone HTML file that does it right:
```
<head>
  <link	rel="stylesheet" type="text/css" cross-origin="anonymous"
        href="https://fonts.googleapis.com/css?family=Roboto+Mono:400,700|Roboto:400,300,300italic,400italic,500,500italic,700,700italic">
</head>
<body style="background: rgb(250, 250, 250)">
    <h2	style="                                                                                                                                                                  
          font-family: Roboto, sans-serif;                                                                                                                                       
          font-size: 25px;                                                                                                                                                       
          font-weight: 300;                                                                                                                                                      
          color: rgba(0, 0, 0, 0.62);                                                                                                                                            
          "
      >This font looks good!!!  Needed &lt;link>.</h2>
</body>
```


So, we just need to add that `<link>` to our _base.html `<head>`.

jrobbins:
This might be a better set of link elements.  It is the way that is currently recommended by fonts.google.com.   I don't know if we use any weights other than 300, but I guess we could use it. If we spot any weak-looking fonts then we could always add more weights on fonts.google.com and regenerate a new URL.
```
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
<link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300&display=swap" rel="stylesheet">
```

Also, in the CSS rule, the font name `Roboto` is supposed to be in single quotes: `font-family: 'Roboto', sans-serif; `

dlaliberte:
Great detective work.  Would have taken me longer to narrow it down.

I'll try adding these link tags to the end of the head of _base.html.  Seems to be working.

dlaliberte:
image

jrobbins (approved):

PR #1894, opened 2022-05-19
Fix bolding in various places

dlaliberte:
Bold headers on Process Overview and Form Field editors got lost due to default font: inherit, and missing bold fonts.


dlaliberte:
After this change:

image

jrobbins (approved):

PR #1901, opened 2022-05-24
Add URL validation pattern and MultiUrlField validator

dlaliberte:


jrobbins:
This PR is missing a description.

pages/guideforms.py
```diff
@@ -71,6 +95,24 @@ def validate(self, value):
     'pattern': EMAIL_ADDRESSES_REGEX
 }
 
+# From https://rodneyrehm.de/t/url-regex.html#imme_emosol+ht-%26f-tp%28s%29
+URL_REGEX = '[ ]*(https?|ftp)://(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?[ ]*'
```
jrobbins, line 99: Let's not accept the ftp scheme.

pages/guideforms.py
```diff
@@ -71,6 +95,24 @@ def validate(self, value):
...
+URL_FIELD_ATTRS = {
+    'title': 'Enter a full URL https://...',
+    'placeholder': 'https://...',
+    'pattern': URL_REGEX
+}
+
+MULTI_URL_FIELD_ATTRS = {
+    'title': 'Enter one or more full URLs, one per line:\nhttps://...\nhttps://...',
+    'placeholder': 'https://...\nhttps://...',
+    'rows': 4, 'cols': 50, 'maxlength': 500
```
jrobbins, line 112: Let's bump up the maxlength to 5000.  I don't know why it was set so small before.

pages/guideforms.py
```diff
@@ -38,6 +41,27 @@ def validate(self, value):
         for email in value:
             validate_email(email.strip())
 
+
+def validate_url(url):
```
jrobbins, line 45: Please add some unit tests for this function and the new class.

jrobbins (changes requested):

pages/guideforms.py
dlaliberte: Sounds good.  I was wondering about that.  Is there any limitation on the database side?

dlaliberte:
> This PR is missing a description.

done.

dlaliberte:
Another problem not addressed by this PR is that if a URL or URLs are invalid, the server-side doesn't update the data that is stored, so the change is essentially thrown away.  We need to avoid that problem, perhaps by storing whatever the user has entered even if it is invalid.  This applies to email fields as well.

pages/guideforms.py
dlaliberte: I added tests of the validate_url.  Curiously, the "with" statement seems to only apply to one "suite" statement, so I had to repeat the "with" clause for each test.

I haven't figured out how to test the MultiFormField yet.   I haven't found an example, and it seems there is a fair amount of context that is required to test properly.

jrobbins:
Now that my pinning change has merged, try rebasing and pushing your changes to this branch again.  Hopefully the python tests will pass this time.

dlaliberte:
The continuous integration tests are now passing.

pages/guideforms.py
```diff
@@ -71,6 +100,24 @@ def validate(self, value):
     'pattern': EMAIL_ADDRESSES_REGEX
 }
 
+# From https://rodneyrehm.de/t/url-regex.html#imme_emosol+ht-%26f-tp%28s%29
+URL_REGEX = '[ ]*(https?)://(-\.)?([^\s/?\.#-]+\.?)+(/[^\s]*)?[ ]*'
```
jrobbins, line 104: I don't see why we need the `((-\.)?` part.

And, I think the domain part could be more clearly separate from the path-query-anchor part.  E.g., there should not be a `#` in the domain part.

Let's try:
```
URL_REGEX = '[ ]*((https?)://)?([\w-]+(\.[\w-]+)+)(/[^\s]*)?[ ]*'
```

jrobbins (changes requested):

jrobbins (approved):

pages/guideforms.py
dlaliberte: Added comment about using the imme_emosol pattern.

I don't understand how that pattern works, but testing e.g. http;//-www.com or http://.www.com, it actually correctly fails to match.  See http://regexr.com for a handy regex tester.

So I think we should leave this as is, for now anyway.

PR #1906, opened 2022-05-26
Pin all py3 library deps to avoid unexpected breakage.

jrobbins:
Another PR seems to be failing python unit tests in CI, maybe because it is getting a new version of a library that has some problem.  This PR pins all library versions to known-good versions.


dlaliberte (approved):
This is likely to be tighter than we need, but we can loosen the constraints later as needed.

PR #1921, opened 2022-06-07
Define and use chromedash-textarea.  Use sl-input.

dlaliberte:
This PR defines a chromedash-textarea component, subclassing sl-textarea.    We do this primarily to add validation for multiple URLs, which applies when the multiple attribute is defined, and the chromedash_split_pattern and chromedash_single_pattern attributes are defined.   If the pattern attribute is defined, this is also used regardless whether multiple is defined.

This is follow-up work after #1901, to do client-side validation mentioned there.

Also, use sl-input for all input form fields.


pages/guideforms.py
```diff
@@ -66,6 +66,21 @@ def validate(self, value):
...
+    template_name = 'django/forms/widgets/text.html'
+
+class ChromedashTextarea(forms.widgets.Textarea):
+    template_name = 'django/forms/widgets/chromedash-textarea.html'
+
+    def __init__(self, attrs=None):
+        # Use slightly better defaults than HTML's 20x2 box
+        default_attrs = {
+            'cols': 50, 'rows': 10,  'maxlength': 500,
+            'placeholder': SUMMARY_PLACEHOLDER_TXT}
```
jrobbins, line 79: It seems oddly specific to have the class definition default to using a placeholder that is meant for the summary field.   How about having no placeholder specified in the class but add back `widget=ChromedashTextarea(attrs={'placeholder': SUMMARY_PLACEHOLDER_TXT})` on line 159.

pages/guideforms.py
```diff
@@ -57,7 +57,7 @@ def to_python(self, value):
         # Return an empty list if no input was given.
         if not value:
             return []
-        return value.split('\n')
+        return value.split(r'\s+')
 
     def validate(self, value):
```
jrobbins, line 62: We discussed removing the sever-side validation today.   However, I don't think that we can do that quite yet.  One reason why this validation is here is to prevent against an attacker inputting a URL with the `javascript:` scheme.  That won't make it past client-side validation, but an attacker could use `wget` to make a POST request to our server rather than using a browser.  So, before we can remove this, we need to add defensive validation at every place where we currently output a URL.

jrobbins (commented):
This is coming together.  Thanks for figuring all this out so far.

static/elements/chromedash-textarea.js
```diff
@@ -0,0 +1,58 @@
...
+    if (!this.originalCheckValidity) {
+      // Set up custom validation.
+      this.originalCheckValidity = this.input.checkValidity;
+      this.input.checkValidity = () => {
+        const value = this.input.value;
+        // First do default validity check.
+        if (this.originalCheckValidity && !this.originalCheckValidity.apply(this.input)) {
+          return false;
+        }
+        // If multiple, then use chrome
```
jrobbins, line 31: This comment is incomplete.

pages/guideforms.py
jpmedley: `SUMMARY_PLACEHOLDER_TXT` has instructions specific to the summary field. We definitely DON'T want that placeholder text in other fields.

static/sass/forms.scss
```diff
@@ -116,3 +116,14 @@ form section.flat_form + h3, .final_buttons {
 #metadata-readonly div + div {
   margin-top: 4px;
 }
+
+sl-input::part(base) {
+  overflow: visible;
+}
+
+sl-input[invalid]::part(input):not(:focus),
+chromedash-textarea[invalid]::part(textarea):not(:focus)
```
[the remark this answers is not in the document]
dlaliberte: done

static/elements/chromedash-textarea.js
dlaliberte: Yup, writing the comment got me to think about generalizing the process a bit... and I forgot to return to finish the thought.

pages/guideforms.py
```diff
@@ -66,6 +66,19 @@ def validate(self, value):
...
 
+class ChromedashTextInput(forms.widgets.Input):
+    template_name = 'django/forms/widgets/text.html'
+
+class ChromedashTextarea(forms.widgets.Textarea):
+    template_name = 'django/forms/widgets/chromedash-textarea.html'
+
+    def __init__(self, attrs=None):
+        # Use slightly better defaults than HTML's 20x2 box
+        default_attrs = {'cols': 50, 'rows': 10,  'maxlength': 500}
```
jrobbins, line 77: maxlength needs to be 5000.  Allowing only 500 is too short.

static/elements/chromedash-textarea.js
```diff
@@ -0,0 +1,60 @@
...
+    super();
+  }
+
+  updated() {
+    if (!this.input) {
+      return;
+    }
+    const value = this.input.value;
+
+    const customCheckValidity = () => {
```
jrobbins, line 25: Does this need to be a nested function?  I think you could define it as a regular function and then call it as `this.customCheckValidity(value)`.

jrobbins (changes requested):

static/elements/chromedash-textarea.js
dlaliberte: Yes, that is possible now.  I was previously using a more complex process that turned out to not work anyway.

pages/guideforms.py
dlaliberte: I think the 10 rows x 50 cols == 500 was the rationale.  But the actual size of the textarea content can have many more rows and columns, so 5000 is probably fine.

static/elements/chromedash-textarea.js
dlaliberte: Moved into a method, since it uses 'this'.

dlaliberte:
It appears there is "1 change requested" but I can't see it.  Maybe things got out of sync.

jrobbins (approved):
Thanks for seeing this through.

jrobbins:
Now that this has merged and I am trying it out, I see two new problems with single-line text fields.  These are release blockers.

* Text in the sl-input fields is a light blue color which is very hard to read.  It is set via `-webkit-text-fill-color: var(--sl-color-primary-500)`.  It should be almost black `--sl-input-color`.
* Checkboxes are now being rendered as `<sl-input type="checkbox"...>`  which shows up as a single-line text input field rather than any kind of checkbox.  So, this needs to go back to using the standard HTML `<input type="checkbox"...>` or `<sl-checkbox>`.

Do you see the same problems when you run locally?

jrobbins:
> * Text in the sl-input fields is a light blue color which is very hard to read. 

I just realized that this only happens when I initially use the browser autocomplete feature.  Once I use that on a given field, it stays light blue from then on.  I think that this is still a problem because users might use autocomplete, e.g., if the need to fill in values that are the same for a few related features.

image

dlaliberte:
> * Text in the sl-input fields is a light blue color which is very hard to read.  It is set via `-webkit-text-fill-color: var(--sl-color-primary-500)`.  It should be almost black `--sl-input-color`.

I didn't notice this problem, but it should be relatively easy to fix, with the appropriate style changes.

> * Checkboxes are now being rendered as `<sl-input type="checkbox"...>`  which shows up as a single-line text input field rather than any kind of checkbox.  So, this needs to go back to using the standard HTML `<input type="checkbox"...>` or `<sl-checkbox>`.

This problem with checkboxes is obviously very wrong, and also possibly easy to fix.  I only changed the input template, but didn't think that I also need to change all other uses of input elements, in particular checkboxes and radio buttons.   Maybe best to rollback this PR.  I haven't that yet, and I am not seeing any obvious way to do it, so if we want to fix this quickly, perhaps you could start the rollback.

PR #1923, opened 2022-06-08
Send and receive emails in py3 using GAE legacy libraries

jrobbins:
In this PR:
* Move sendemail.py and unit tests from py2/ to framework/
* Update dispatch.yaml to route requests to the default service instead of app-py3
* Stop using service app-py3 and make our py3 code the default service
* Delete everything in py2 and the associated scripts in package.json


dispatch.yaml
```diff
@@ -14,11 +14,5 @@ dispatch:
...
 
-- url: "*/tasks/outbound-email"
-  service: default
-
-- url: "*/py2"
-  service: default
-
 - url: "*/*"
-  service: app-py3
+  service: default
```
KyleJu, line 18: Remove the comments from line 3 to line 8

requirements.txt
```diff
@@ -2,14 +2,15 @@
...
 funcsigs==1.0.2
 google-python-cloud-debugger==2.18
 google-cloud-tasks==2.7.0
 google-cloud-ndb==1.11.1
 google-cloud-logging==3.0.0
-google-auth==1.24.0
+google-auth==1.31.0
 requests==2.27.1
 
 # TODO(jrobbins): Update to a 2.x version of flask, which may require source
```
KyleJu, line 16: Maybe we can clean up the Py2 dependencies in a follow-up PR

scripts/start_server.sh
```diff
@@ -28,7 +28,7 @@ gunicorn --bind :8080 --workers 4 main:app
 #  $BASEDIR/../dispatch.yaml \
 #  $BASEDIR/../notifier.yaml \
 #  $BASEDIR/../dev-default.yaml \
```
KyleJu, line 30: is `dev-default.yaml` still needed? We can delete the file at this point.

KyleJu (approved):

dispatch.yaml
jrobbins: I shortened the comments, but I still wanted something in there to explain that inbound email goes to the default service.

requirements.txt
jrobbins: Yep, I'll try doing that after the next deployment.

scripts/start_server.sh
jrobbins: Good point.  Gone.

PR #1933, opened 2022-06-14
Avoid type="email" for multi emails.

jrobbins:
With the current code, we generate HTML that looks like
`<input type=email multiple pattern="...">`
That displays with the invalid pink background whenever multiple email addresses are entered.
Removing the `type="email"` by using the regular input widget class allows validation to rely solely on the pattern.


jrobbins:
Before:
image

After:
image

dlaliberte:
I thought that the default validation for type="email" was very liberal, so it should basically never fail before the pattern match fails.  But maybe it actually conflicts with use of "multiple".

dlaliberte (approved):

PR #1934, opened 2022-06-14
Replace select and option with sl-select and sl-menu-item

dlaliberte:
Replace select and option with sl-select and sl-menu-item.  Also uses sl-icon.

Also fixes the shoelace base path, because we need to use the check-lg and chevron-right icons.

Here is what it looks like with an opened select box and preselected item;

image


jrobbins (approved):
This all looks good.

In our meeting today, I had mentioned that I might deploy this today, but instead I think I will deploy without this change.  The reason is that I'd like to wait for your follow-up PR (or another commit and re-review on this one) that hides the list of choices until after the component JS loads.

dlaliberte:
> In our meeting today, I had mentioned that I might deploy this today, but instead I think I will deploy without this change. The reason is that I'd like to wait for your follow-up PR (or another commit and re-review on this one) that hides the list of choices until after the component JS loads.

That's fine.  I tried the body.loading sl-select selector, but it doesn't seem to be enough. since the shoelace code doesn't convert the web components soon enough.  However, this general problem is addressed in the shoelace docs:
https://shoelace.style/getting-started/usage?id=waiting-for-components-to-load
I'll try a few things.  I think we need to use display:none on the sl-menu-item elements, however, since we don't want them to take up any space.

static/sass/forms.scss
```diff
@@ -131,4 +131,21 @@ chromedash-textarea[invalid]::part(textarea)
...
+}
+
+/* menu items for selects should not be displayed at all, until defined */
+sl-select sl-menu-item:not(:defined) {
+   display: none 
+}
+
+/* Hide extra br following display:block elements */
+chromedash-textarea+br,
+sl-select+br,
```
jrobbins, line 148: This would be more readable with spaces around the `+`.

jrobbins (commented):
LGTM after one small change.

static/sass/forms.scss
jrobbins: Oh, you already merged it.  No big deal, just a style nit.

PR #1937, opened 2022-06-15
Convert part of feature template to a new element

kevinshen56714:
This PR converts a huge part of `feature.html` to the new `chromedash-feature-page` Lit element.

In this PR:
1. `chromedash-feature-page.js` is created
2. `feature.scss` is safely removed (all the required css are now included in the new `chromedash-feature-page` component)
3. `chromedash-feature-page_test.js` is created
4.  `sinon` package is added for stubbing out API calls for the unit test


jrobbins:
This all looks good so far. 

Regarding the unit test output, I think that the lines like `[Errno 2] No such file or directory: 'static/css/blink.css'` can be ignored for now.  They offer when tests are run before the first `npm start` or `npm run build`.  So, fixing those is more an issue of CI configuration.  They are not test failures, it's just noise.

The failing test case `test_links` can probably be reduced to have fewer assertions now, and it will ultimately be removed when you are done with converting this page to be all in components.   Instead, you should add a simple JS unit test to exercise your render() method.  There are currently only two `*_test.js` files in our code for you to refer to as examples.

The feature description, motivation, and comments should be "URLized" or "auto linked".  You can see an example of how to do that in JS code in chrome dash-feature-detail.js via the `autolink()` function.

kevinshen56714:
@jrobbins I attempted to create a js test for the new `chromedash-feature-page`. However, I'm not entirely sure how to deal with the part where we fetch data in `connectedCallback` through `window.csClient.getFeature()`, and the test is failing because of that. 
Do you have any suggestions? Thank you!

jrobbins:
@dlaliberte Dan, what do you recommend for mocks in JS unit tests?

I have not done a lot of JS unit tests, so this is a learning experience for me too.  It looks like `Sinon` is a popular JS mock library.  It was used by some other engineers who worked on the monorail frontend, e.g.,
https://chromium.googlesource.com/infra/infra/+/refs/heads/main/appengine/monorail/static_src/elements/issue-detail/mr-issue-page/mr-issue-page.test.js
In that example prpcClient is analogous to our csClient.  On line 32, a method of that object is stubbed out.  Then on lines 218 there are some assertions made about a method on that object being called with certain arguments.  Once you have a mock method, you can call `returns()` on it to make it return some test data.  See https://sinonjs.org/releases/latest/mocks/.  Also, it looks like we might need to add a couple of lines when using fixtures, see https://open-wc.org/guides/knowledge/testing/stubs/.

dlaliberte:
> Dan, what do you recommend for mocks in JS unit tests?

To be honest, I tend to avoid using mocks, stubs, fakes and anything else that is not testing the real thing.  Otherwise, it seems like extra work that risks not doing adequate tests of real behavior.  Consequently, I haven't looked into the alternatives that might be available.    There is a fair amount of debate about how testing should be done, in general, and regarding this aspect as well.  See https://medium.com/@june.pravin/mocking-is-not-practical-use-fakes-e30cc6eaaf4e

I tend to prefer integration tests that cover typical as well as extreme cases, and then when something breaks, it may be a bit more work to narrow down exactly what broke, but at least we don't have to worry about whether the tests are valid.

jrobbins:
Let's keep integration tests on our list of things to do.  But, I do think that we need JS unit tests to catch small mistakes early and to prevent regressions, and some unit tests need mocks or fakes.   Sinon seems pretty widely used, or we could build fakes, either way is fine with me.

kevinshen56714:
@jrobbins I finished the new js unit test, and I believe this PR is ready for review.

One thing to note however, is that in the unit test, I have to first manually initialize `window.csClient` and then stub out the `getFeature` call. Please let me know your thoughts on that.

Thank you!

static/elements/chromedash-feature-page_test.js
```diff
@@ -0,0 +1,87 @@
...
+      safari: {view: {text: 'fake safari view text'}},
+      webdev: {view: {text: 'fake webdev view text'}},
+    },
+    resources: {
+      samples: ['fake sample link one', 'fake sample link two'],
+      docs: ['fake doc link one', 'fake doc link two'],
+    },
+    standards: {
+      spec: 'fake spec link',
+      maturity: {test: 'Unknown standards status - check spec link for status'},
```
jrobbins, line 29: test -> text.

jrobbins (changes requested):
This is a great start.  I have a few requested changes but they should be easy to resolve.

package.json
```diff
@@ -93,6 +93,7 @@
     "lit": "^2",
     "node-fetch": ">=3.2.6",
     "node-sass": ">=4.13.1",
+    "sinon": "^14.0.0",
```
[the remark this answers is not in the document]
kevinshen56714: Moved to devDependecies!

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+          max-width: var(--max-content-width);
+        }
+        #feature ul {
+          list-style-position:inside;
+          list-style:none;
+        }
+        section {
+          margin-bottom:1em;
+        }
+        section h3{
```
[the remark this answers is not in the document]
kevinshen56714: Fixed.

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+          border: var(--card-border);
+          box-shadow: var(--card-box-shadow);
+
+          box-sizing: border-box;
+          word-wrap: break-word;
+          margin-bottom: var(--content-padding);
+          max-width: var(--max-content-width);
+        }
+        #feature ul {
+          list-style-position:inside;
```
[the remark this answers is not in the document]
kevinshen56714: Fixed.

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+        <br>
+        <p>
+          <label>Implementation status:</label>
+          <b>${this.feature.browsers.chrome.status.text}</b>
+          ${this.feature.browsers.chrome.bug ? html`
+            <a href=${this.feature.browsers.chrome.bug} target="_blank" rel="noopener">
+              tracking bug
+            </a>
+          `: nothing}
+          <chromedash-gantt feature=${JSON.stringify(this.feature)}></chromedash-gantt>
```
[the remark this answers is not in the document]
kevinshen56714: Fixed.

static/elements/chromedash-feature-page_test.js
kevinshen56714: Fixed.

jrobbins:
The JS tests pass, but if you look at the test run output there is a warning:
```
🚧 Browser logs on Chromium:
[23](https://github.com/GoogleChrome/chromium-dashboard/runs/6995446321?check_suite_focus=true#step:11:24)
      An error was thrown in a Promise outside a test. Did you forget to await a function or assertion?
[24](https://github.com/GoogleChrome/chromium-dashboard/runs/6995446321?check_suite_focus=true#step:11:25)
      Error: Got error response from server
[25](https://github.com/GoogleChrome/chromium-dashboard/runs/6995446321?check_suite_focus=true#step:11:26)
        at p.<anonymous> (static/elements/chromedash-feature-page_test.js:10:48)
[26](https://github.com/GoogleChrome/chromium-dashboard/runs/6995446321?check_suite_focus=true#step:11:27)
        at static/elements/chromedash-feature-page_test.js:8:1
```

Ideally, we should have pretty clean test runs so that later when there is a new test failure, we can debug it without needing to find the relevant details among a bunch of stuff that looks like potential errors.

I think a way to resolve this warning would be to:
* move the definition of `invalidFeaturePromise` and the `returns(invalidFeaturePromise)` call into the `'renders with no data'` test case.
* For consistency, move `validFeaturePromise` and `returns(validFeaturePromise)` into the `'renders with fake data'` test case.

If you feel that that makes those test cases too hard to read, you could try moving those lines into a helper function that is called from the test case.

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+              ${this.feature.browsers.ff.view.url ? html`
+                <a href=${this.feature.browsers.ff.view.url}>${this.feature.browsers.ff.view.text}</a>
+              `: html`
+                ${this.feature.browsers.ff.view.text}
+              `}
+            </li>
+          `: nothing}
+          ${this.feature.browsers.safari.view.val ? html`
+            <li>
+              <label>Firefox:</label>
```
jrobbins, line 194: This one should be "Safari".

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+      ${this.feature.motivation ? html`
+        <section id="motivation">
+          <h3>Motivation</h3>
+          <p class="preformatted">${autolink(this.feature.motivation)}</p>
+        </section>
+      `: nothing}
+
+      ${this.feature.resources && this.feature.resources.samples ? html`
+        <section id="demo">
+          <h3>${this.feature.resources.length == 1 ? 'Demo' : 'Demos'}</h3>
```
jrobbins, line 115: `resources` is an object, I think you need `resources.samples` which is either null or an array.

static/elements/chromedash-feature-page.js
```diff
@@ -0,0 +1,264 @@
...
+            `)}
+          </p>
+        `: nothing}
+        <br>
+        <p>
+          <label>Implementation status:</label>
+          <b>${this.feature.browsers.chrome.status.text}</b>
+          ${this.feature.browsers.chrome.bug ? html`
+            <a href=${this.feature.browsers.chrome.bug} target="_blank" rel="noopener">
+              tracking bug
```
jrobbins, line 170: The old version of the page used parenthesis: `(<a href=...>tracking bug</a>)`.  I don't think that it is a great visual treatment, but it does help prevent the user from incorrectly attempting to read the line of text like a sentence.

jrobbins (changes requested):
This is looking very good.  I deployed it as a non-live version on our staging server.  You can check for any visual differences by opening two browser tabs on:
https://db38687-dot-cr-status-staging.appspot.com/feature/5699445609988096
and
https://cr-status-staging.appspot.com/feature/5699445609988096
and then quickly flipping between the two tabs.
That's how I spotted a few minor issues.

static/elements/chromedash-feature-page.js
kevinshen56714: Fixed!

static/elements/chromedash-feature-page.js
kevinshen56714: Fixed!

static/elements/chromedash-feature-page.js
kevinshen56714: Added!

kevinshen56714:
It is a very good way to check! Requested changes are made.

I will definitely be more careful when translating a html template to a component!

jrobbins (approved):

PR #1940, opened 2022-06-15
Specify current value of sl-select widgets.

jrobbins:
This sets the current value of a sl-select widget.

The value of widget.value seems to be an array of strings for some reason, so I am using the django `|first` filter to get the first string.  This seems to also work with the field has no value.


dlaliberte (approved):
I suppose this is to support the multiple option that applies to select elements.

PR #1943, opened 2022-06-16
Make non-django form header fields match django-generated ones.

jrobbins:
I think the django-generated ones used to be bold, so the ones I added in django templates had `<b>` to match, but somewhere along the line the django-generated ones stopped being bold.


dlaliberte (approved):
We keep losing the bold style.  Our css rules are fairly complex, so no idea why the broke again.

PR #1945, opened 2022-06-16
Add CSS rules for chromedash-select.

jrobbins:
You had made some refinements to the CSS for sl-select elements.
When we replaced sl-select with chromedash-select, those tweaks stopped being applied.
This PR adds them back for chrome status-select.

### DIFF
--- a/static/sass/forms.scss
+++ b/static/sass/forms.scss
@@ -128,7 +128,8 @@ chromedash-textarea[invalid]::part(textarea)
   background-color: #FFEDF5;
 }
 
-sl-select[size="small"] sl-menu-item::part(base) {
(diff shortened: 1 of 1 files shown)


jrobbins:
Is your other PR going to make this one unneeded?

dlaliberte:
> Is your other PR going to make this one unneeded?

Actually, we can delete chromedash-select now.  Your last update of shoelace code included the fix.   So I think maybe this PR won't be needed.

My other PR is also removing the hiding of the extra brs, with the subclassing of the form to override as_table.

jrobbins:
Yep, I agree this PR is no longer needed.

PR #1960, opened 2022-06-21
Form cleanups (focus for radio buttons, extra br)

dlaliberte:
This PR just adds a highlight around the radio button when focused.  We need this since some other style is hiding the focus style, and we can't use sl-radio yet since it doesn't support validation.

image

Also defines and uses ChromedashForm to override the as_table method, and remove the extra br tag.


jrobbins:
The PR description seems out-of-date because you are using sl-radio now.   Also, the reference to validation is unclear.

PR #1965, opened 2022-06-23
Change the guide form field layout for all tables

dlaliberte:
This PR moves the feature form fields to be under the corresponding header and the associated help text remains on the right, with width:50% for each.  The header for each field is moved to a separate row, so it stands out and the form field and help text align on their tops.

This change is applied to the new.html form, the metadata.html form (used via the edit.html page), the stage.html form and the editall.html form.

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -253,7 +253,9 @@ ALL_FIELDS = {
                    'feature implements?')),
 
     'unlisted': forms.BooleanField(
-      required=False, initial=False, label="Set to: unlisted",
+      label="Unlisted",
+      widget=forms.CheckboxInput(attrs={'label': "Unlisted"}),
+      required=False, initial=False,
       help_text=('Check this box for draft features that should not appear '
                  'in the feature list. Anyone with the link will be able to '
                  'view the feature on the detail page.')),
@@ -266,7 +268,9 @@ ALL_FIELDS = {
                    'spec link when available.')),
 
     'api_spec': forms.BooleanField(
-        required=False, initial=False, label='API spec',
+        label='API spec',
+        widget=forms.CheckboxInput(attrs={'label': "API spec"}),
+        required=False, initial=False,
         help_text=('The spec document has details in a specification language '
                    'such as Web IDL, or there is an existing MDN page.')),
 
@@ -623,7 +627,9 @@ ALL_FIELDS = {
          'DevTools support checklist</a> for guidance.')),
 
     'all_platforms': forms.BooleanField(
-        required=False, initial=False, label='Supported on all platforms?',
+        label='Supported on all platforms?',
+        widget=forms.CheckboxInput(attrs={'label': "Supported on all platforms"}),
+        required=False, initial=False, 
         help_text=
         ('Will this feature be supported on all six Blink platforms '
          '(Windows, Mac, Linux, Chrome OS, Android, and Android WebView)?')),
@@ -637,7 +643,9 @@ ALL_FIELDS = {
                 'supported on all platforms.')),
 
     'wpt': forms.BooleanField(
-        required=False, initial=False, label='Web Platform Tests',
+        label='Web Platform Tests',
+        widget=forms.CheckboxInput(attrs={'label': "Web Platform Tests"}),
+        required=False, initial=False, 
         help_text='Is this feature fully tested in Web Platform Tests?'),
 
     'wpt_descr': forms.CharField(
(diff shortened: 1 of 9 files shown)


dlaliberte:
Here is the new.html page:

image

jrobbins:
This seems pretty good, but I do have a few suggestions.  Rather than give rationale for each, I'll just list them here.  I'm listing them here rather than in github code review comments because a lot of this is on lines that are not changed in the PR yet.
* On the new.html table, change the table width from 60em to 80em.
* Rather than a 50/50 split for the td cells, use a 60/40 split to give a little more width to the input fields.  That can be accomplished by using css selector td:first-of-kind { width: 60% } instead of the td { width: 50%} that you have.
* Remove the .helptext CSS attribute font-style: italic and change font-size: smaller to small.

On new.html's feature_type field, I see what you are doing with putting the help for each radio button across from it, however it gives the visual impression of just having several paragraphs in the help column rather than being obvious that they are one-to-one.   I think that text is not actually field-level help text, it is option-level help text and it should be closer to the radio buttons that it describes.  So, maybe put that whole field back the way it was and have it in one big `<td span=2>`.

When I look through these forms, I feel that the checkboxes get lost visually because only the heading is near them and then there is a really big horizontal gap between the checkbox and the help text.  Possible solution below.

Mobile is not our primary target platform, but we should try to keep it somewhat usable.  This table overflows horizontally on mobile, which looks bad.    Maybe it could work if we introduce a `@media` rule for narrow screens that changes the display attributive of `tr`, `th`, and `td`.  Or, we could change the markup to use semantic elements `label`, then the input element, then `aside` for the help text.  CSS flex box could make those lay out in one, two, or even three columns based on the container width.  We could use CSS to make a checkbox appear on the same row as its label.

I like the way that the three fields for signals form a natural visual grouping under one label.

dlaliberte:
> This seems pretty good, but I do have a few suggestions. Rather than give rationale for each, I'll just list them here. I'm listing them here rather than in github code review comments because a lot of this is on lines that are not changed in the PR yet.

Glad you like it so far.  This is a rather radical change, and I was wondering about most of the same issues you raised, but I didn't have an obvious or quick solution.  

> * On the new.html table, change the table width from 60em to 80em.

The table and column widths should probably be shared styles.  But we can deal with that later once we have more unified generation of the "table"s in a web component.

> * Rather than a 50/50 split for the td cells, use a 60/40 split to give a little more width to the input fields.  That can be accomplished by using css selector td:first-of-kind { width: 60% } instead of the td { width: 50%} that you have.
> * Remove the .helptext CSS attribute font-style: italic and change font-size: smaller to small.

Sounds good. 

> On new.html's feature_type field, I see what you are doing with putting the help for each radio button across from it, however it gives the visual impression of just having several paragraphs in the help column rather than being obvious that they are one-to-one. I think that text is not actually field-level help text, it is option-level help text and it should be closer to the radio buttons that it describes. So, maybe put that whole field back the way it was and have it in one big `<td span=2>`.

I'll try that.  

> 
> When I look through these forms, I feel that the checkboxes get lost visually because only the heading is near them and then there is a really big horizontal gap between the checkbox and the help text. Possible solution below.

We could repeat a variation of the label, e.g. "Set to: unlisted", similar to the Stage setting.  

Not sure what the label should be for "API spec" since the help text indicates that there are a couple alternatives for what it means to select the checkbox: "The spec document has details in a specification language such as Web IDL, or there is an existing MDN page."

> 
> Mobile is not our primary target platform, but we should try to keep it somewhat usable. This table overflows horizontally on mobile, which looks bad.

I agree we should deal with this, probably by changing the layout based on the screen width as you suggest.  It's no worse than the current layout, right?    Once we unify the layout of all the tables with one web component, and a component for each "row", it will be easier to experiment with the alternatives.

> Maybe it could work if we introduce a `@media` rule for narrow screens that changes the display attributive of `tr`, `th`, and `td`. Or, we could change the markup to use semantic elements `label`, then the input element, then `aside` for the help text. CSS flex box could make those lay out in one, two, or even three columns based on the container width. We could use CSS to make a checkbox appear on the same row as its label.

> I like the way that the three fields for signals form a natural visual grouping under one label.

The grouping is good.  The extra vertical space between the fields bugs me though :)

dlaliberte:
After the updates:

https://screenshot.googleplex.com/APRm8p74V89u5zH

jrobbins:
Thanks for these changes.  It is looking better.

For the checkboxes, I think that the best way to present them would be to have the checkbox on the same row as the header.  E.g., (in ascii art): 
```Unlisted [_]                   Check this box for draft features that should not appear...```

I realize that might be tricky to do in a django table.  So, my second choice suggestion is to repeat the label after the checkbox:
```Unlisted
      [_] Unlisted                Check this box for draft features that should not appear...```
The second one should be clickable.  Try `<sl-checkbox>{{label}}</sl-checkbox>`.

Regarding mobile, I had thought that this PR made it a bit worse, but today I compared more pages and I see that the old design had plenty of the same overflow problems on mobile.  So, I won't block this PR on mobile, but let's try to improve the mobile UX a little in a follow-up PR.

jrobbins (approved):

PR #1969, opened 2022-06-24
Use long Expires with GAE version as buster.

jrobbins:
When I deployed a new version of the site today it caused a brief outage.  I quickly rolled back and worked around the problem to successfully finish the deployment.  This is a better fix.

The problem was that we had no setting in `app.yaml` for `default_expiration`.  It defaults to 10 hours, so we were sending all images, css, and JS responses with headers telling the browser that it could be cached for 10 hours.  When a new version of the site makes changes to HTML content or some JS content that gets inlined into the HTML pages, that can cause exceptions in JS code that was deployed the week prior but is still being used by some users for up to 10 hours.  In my case, I saw it as a JS error on the feature list page that occurred before `chromedash-feature.js` was parsed, so the page looked pretty blank.  I might do a force-reload, but users who didn't know that I did a deployment would just see an unusable page.

One solution could have been to make a shorter expiration window, e.g., 10 minutes, but that would still potentially cause a brief outage every time we do a deployment.  And, it would slow down frequent users because our growing body of JS code would need to be fetched and parsed while those users are in the middle of using the site.

So, the solution I went with is this (similar to the way Monorail did it):
* Set a fairly long expiration window: 10 days
* Make each `<script src=...>` tag and stylesheet reference include a query parameter in the URL that acts as a "cache buster" to cause an HTTP cache miss when that parameter changes.  Our server code does not see this parameter or do anything with it because the JS is a static file.
* Use the GAE version number as that cache buster value because it changes exactly when we deploy a new version of the app.  FYI, we use the git hash of the latest commit as the GAE version.
* We don't do this for images because we don't have very many of them and they very rarely change.  If we ever need to change an image, we should give it a new filename.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -8,6 +8,10 @@ automatic_scaling:
   min_idle_instances: 1
   max_pending_latency: 0.2s
 
+# Use a long HTTP cache expiration time for css and JS, but also
+# include the GAE version number as a query string parameter so that on
+# each deployment, users get new static files to match the new app version.
+default_expiration: "10d"
 
 handlers:
 # Static handlers ---------------------------------------------------------------
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -296,6 +296,8 @@ class FlaskHandler(BaseHandler):
   def get_common_data(self, path=None):
     """Return template data used on all pages, e.g., sign-in info."""
     current_path = path or flask.request.full_path
+    # Used to make browser load new JS and CSS for each GAE version.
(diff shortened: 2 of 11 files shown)


kevinshen56714 (approved):

dlaliberte (approved):
Your solution seems reasonable, but it seems like there ought to be a solution that doesn't require us to remember to add the query parameter.  Some kind of server configuration should be able to do the equivalent of cache busting when a new version is deployed.

PR #1970, opened 2022-06-24
Add `do_get` for `processes_api`, `fielddefs_api`, and `cues_api`

kevinshen56714:
The main goal of this PR is to move the `<sl-detail>` in feature.html into the `chromedash-feature-page` component.

To achieve that,
1. `processes_api.py` and `processes_api_test.py` are added with a do_get method
2. `fielddefs_api.py` and `fielddefs_api_test.py` are added with a do_get method
3. do_get method method is also added to the `cues_api.py` (`cues_api_test.py` is also updated)
4. corresponding `getProcess`, `getFieldDefs`, and `getDismissedCues` functions are added to the csClient and called from the `chromedash-feature-page` component

### DIFF
--- a/main.py
+++ b/main.py
@@ -96,10 +96,9 @@ api_routes = [
      comments_api.CommentsAPI),
     (API_BASE + '/features/<int:feature_id>/approvals/<int:field_id>/comments',
      comments_api.CommentsAPI),
+    (API_BASE + '/features/<int:feature_id>/process', processes_api.ProcessesAPI),
 
     (API_BASE + '/fielddefs', fielddefs_api.FieldDefsAPI),
-
-    (API_BASE + '/processes/<int:feature_id>', processes_api.ProcessesAPI),
 
     (API_BASE + '/login', login_api.LoginAPI),
     (API_BASE + '/logout', logout_api.LogoutAPI),
--- a/main.py
+++ b/main.py
@@ -22,9 +22,11 @@ from api import channels_api
 from api import comments_api
 from api import cues_api
 from api import features_api
+from api import fielddefs_api
 from api import login_api
 from api import logout_api
 from api import metricsdata
+from api import processes_api
 from api import stars_api
 from api import token_refresh_api
 from framework import basehandlers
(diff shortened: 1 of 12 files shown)


main.py
```diff
@@ -95,6 +97,10 @@
     (API_BASE + '/features/<int:feature_id>/approvals/<int:field_id>/comments',
      comments_api.CommentsAPI),
 
+    (API_BASE + '/fielddefs', fielddefs_api.FieldDefsAPI),
+
+    (API_BASE + '/processes/<int:feature_id>', processes_api.ProcessesAPI),
```
jrobbins, line 102: If the URL is `/processes/<int>`, I would expect that that int is the feature_type.   I think it is a fine idea to use the feature_id in the request because that allows you to initiate this request without waiting for the feature itself to load.  However, if the parameter is going to be the feature_id, then the url should probably be `/feature/<int:feature_id>/process`.  That logic does not apply to fielddefs because that request has no ID in the URL.

static/elements/chromedash-feature-detail.js
```diff
@@ -11,18 +11,18 @@ class ChromedashFeatureDetail extends LitElement {
   static get properties() {
     return {
       feature: {type: Object},
-      process: {type: Array},
+      process: {type: Object},
```
jrobbins, line 14: Thanks for noticing and fixing this.

static/js-src/cs-client.js
```diff
@@ -239,6 +243,16 @@ class ChromeStatusClient {
     return this.doGet(`/features?q=${userQuery}`);
   }
 
+  // Fielddefs API
+  getFieldDefs() {
+    return this.doGet(`/fielddefs`);
+  }
+
+  // Processes API
+  getProcess(featureId) {
```
jrobbins, line 252: Like with the URLs, if the name is `getProcess`, I would expect the argument to be a feature_type because that is how we identify processes.  It's fine to use `featureId`, but in that case I think a better name would be `getProcessFor(featureId)` or `getProcesByFeature(featureId)`, or `getFeatureProcess(featureId)`.

jrobbins (changes requested):

static/elements/chromedash-feature-detail.js
kevinshen56714: NP!

static/js-src/cs-client.js
kevinshen56714: Changed it to `getFeatureProcess(featureId)`

main.py
kevinshen56714: Changed it to `/feature/<int:feature_id>/process`. I think for now it's simpler and easier to debug to just use Promise.all to send out all requests at once.

jrobbins (approved):

PR #1976, opened 2022-06-27
Convert My Features page into a single component page

kevinshen56714:
This PR turns the `myfeatures.html` into a single Lit element page. Specifically,

1. The original `chromedash-myfeatures.js` is renamed to `chromedash-myfeatures-page.js` with data fetched from APIs instead of from Django template data.
2. `chromedash-myfeatures-page-test.js` is added to test the rendering of my features page.
3. `static/js-src/myfeatures-page.js` is safely removed.
4. The content of myfeatures.html is now a single Lit element!
5. `chromedash-feature-table.js` is refined with the following changes:
    5.1. Added `<a>` tag for star icons so that hovering over them would change the cursor.
    5.2. Added a `updated()` method so that the "Features I starred" section would be rerendered every time a feature is unstarred or starred for a more dynamic UI. See the videos below:

## Before
https://user-images.githubusercontent.com/11501902/176015140-7079a196-3e12-4686-baa2-8cb17a98747f.mov

## After
https://user-images.githubusercontent.com/11501902/176015170-2d4f8720-9a02-4ee7-aacb-aa77f85fe324.mov

### DIFF
--- a/static/elements/chromedash-feature-table.js
+++ b/static/elements/chromedash-feature-table.js
@@ -148,7 +148,7 @@ class ChromedashFeatureTable extends LitElement {
     const featureId = Number(iconEl.dataset.featureId);
     const newStarred = !this.starredFeatures.has(featureId);
 
-    // handled in chromedash-myfeatures.js
+    // handled in chromedash-myfeatures-page.js
     this._fireEvent('star-toggle-event', {
       featureId: featureId,
       doStar: newStarred,
@@ -170,7 +170,7 @@ class ChromedashFeatureTable extends LitElement {
   }
 
   openApprovalsDialog(featureId) {
-    // handled in chromedash-myfeatures.js
+    // handled in chromedash-myfeatures-page.js
     this._fireEvent('open-approvals-event', {
       featureId: featureId,
     });
--- a/static/elements/chromedash-feature-table.js
+++ b/static/elements/chromedash-feature-table.js
@@ -73,6 +73,15 @@ class ChromedashFeatureTable extends LitElement {
     }
   }
 
+  // For rerendering of the "Features I starred" section when a feature is starred
+  // only re-fetch if !this.loading to avoid double fetching on first update.
+  updated(changedProperties) {
+    if (this.query == 'starred-by:me' && !this.loading &&
+      changedProperties.has('starredFeatures')) {
+      this.fetchFeatures();
+    }
+  }
+
   handleSearch(event) {
     this.loading = true;
     this.query = event.detail.query;
@@ -140,7 +149,6 @@ class ChromedashFeatureTable extends LitElement {
   }
 
   toggleStar(e) {
-    console.log('toggleStar');
     e.preventDefault();
     e.stopPropagation();
 
@@ -207,17 +215,16 @@ class ChromedashFeatureTable extends LitElement {
 
   renderStarIcon(feature) {
     return html`
-      <span class="tooltip"
+      <a href="#" class="tooltip" data-tooltip @click=${this.toggleStar}
         title="Receive an email notification when there are updates">
         <iron-icon
           icon="${this.starredFeatures.has(Number(feature.id)) ?
           'chromestatus:star' :
           'chromestatus:star-border'}"
           class="pushicon"
-          data-feature-id="${feature.id}"
-          @click="${this.toggleStar}">
+          data-feature-id="${feature.id}">
         </iron-icon>
-      </span>
+      </a>
     `;
   }
 
@@ -381,7 +388,7 @@ class ChromedashFeatureTable extends LitElement {
       ${this.renderSearch()}
        <table>
         ${this.renderMessages() ||
-          this.features.map(this.renderFeature.bind(this))}
+          this.features.map((feature) => this.renderFeature(feature))}
        </table>
      `;
   }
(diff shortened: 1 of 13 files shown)


static/elements/chromedash-myfeatures-page.js
```diff
@@ -64,9 +81,9 @@ class ChromedashMyFeatures extends LitElement {
 
         <chromedash-feature-table
           query="${query}"
-          ?signedIn=${Boolean(this.signedInUser)}
-          ?canEdit=${this.canEdit}
-          ?canApprove=${this.canApprove}
+          ?signedIn=${Boolean(this.user)}
+          ?canEdit=${this.user && this.user.can_edit}
```
jrobbins, line 85: I like the way these expressions came out.

static/elements/chromedash-myfeatures-page.js
```diff
@@ -1,36 +1,53 @@
...
   static get properties() {
     return {
-      signedInUser: {type: String},
-      canEdit: {type: Boolean},
-      canApprove: {type: Boolean},
-      loginUrl: {type: String},
-      starredFeatures: {type: Object}, // will contain a set of starred features
+      user: {type: Object},
+      starredFeatures: {type: Set}, // will contain a set of starred features
+      toastEl: {type: Element},
```
jrobbins, line 21: Does this need to be a property, or even an instance variable?  The element could be looked up whenever we need to display a toast message.   I think the main idea of a property is that it could change value and trigger the element to re-render, which does not seem to apply here.

static/elements/chromedash-myfeatures-page.js
```diff
@@ -1,36 +1,53 @@
 import {LitElement, css, html, nothing} from 'lit';
-import './chromedash-approvals-dialog';
```
jrobbins, line 2: Why remove these two imports?  Because they are imported into components.js and thus available everywhere?  I've gone back and forth on whether each component should follow the principle of "include what you use" or whether components.js should just include all the components that anything uses so that we have one comprehensive list.  What do you think?

static/elements/chromedash-feature-table.js
```diff
@@ -73,6 +73,15 @@ class ChromedashFeatureTable extends LitElement {
     }
   }
 
+  // For rerendering of the "Features I starred" section when a feature is starred
+  // only re-fetch if !this.loading to avoid double fetching on first update.
+  updated(changedProperties) {
```
jrobbins, line 78: Wow, that came out really simple.

jrobbins (commented):
Good work on this.  I have a few comments for discussion.

static/elements/chromedash-myfeatures-page.js
kevinshen56714: Yes, it turns out to be very readable!

static/elements/chromedash-myfeatures-page.js
kevinshen56714: What you says makes sense, and I removed the toastEl property.

But I'm trying to reduce the amount of `document.querySelector` (especially the toast is used quite often). 

In the new commit, I made a util method `showToastMessage` and import it to the elements. This way we only query once for the toast and at the same time avoid code repetition. Let me know what you think!

static/elements/chromedash-myfeatures-page.js
kevinshen56714: I added them back in the new commit. I think it makes sense to import the elements to where it's used. 

My thought is that once we have the SPA set up, we can then import only the `*-page.js` components in the component.js, while the other components will only be imported to whichever components that use it!

jrobbins: OK, that certainly works.  

It might make sense to have the shoelace-style components imported in components.js because they are widely used and there is a little initialization needed.

static/elements/chromedash-myfeatures-page.js
jrobbins: OK that sounds good.

That reminds me, we should consider replacing chromedash-toast with https://shoelace.style/components/alert?id=toast-notifications.   That might allow us to eliminate one source file.  I am not sure if their component is really better than ours in a meaningful way.

kevinshen56714:
@jrobbins I finally fixed the unit tests. The approval dialogs are moved back to Django templates temporarily as we discussed. I also added minimal new API stubs for the imported child components to not error out in the unit tests. Please review the PR again, thank you!

jrobbins (approved):
The new utility function to open the approvals dialog looks good.  

Our discussion today about what it would take to make a highly reusable footer element makes me think that maybe the utility functions for a given element should be in the source file for that element rather all together in utils.js.   Such a utility function could go a step further than what we have now by eliminating the requirement to put the element in the django template at all.  If it is not found on the page, the utility function could create the element and append it to document.   Just an idea to think about.

kevinshen56714:
> The new utility function to open the approvals dialog looks good.
> 
> Our discussion today about what it would take to make a highly reusable footer element makes me think that maybe the utility functions for a given element should be in the source file for that element rather all together in utils.js. Such a utility function could go a step further than what we have now by eliminating the requirement to put the element in the django template at all. If it is not found on the page, the utility function could create the element and append it to document. Just an idea to think about.

I love this idea! This can
1. avoid creating a toast and dialog that may not be used at all
2. avoid querying them all the time
3. make unit tests easier

I may try to implement it later this week!

PR #1981, opened 2022-06-27
Add chromedash-form-table and -form-field and use for all feature field forms

dlaliberte:
This PR replaces all feature form tables with chromedash-form-table and all rows with chromedash-form-field.  

Since tables and trs depend on each other, initial dom parsing will remove all content of tables that is not trs, so it is important to replace all the tables that contain any chromedash-form-field, and thus all trs also must be replaced by chromedash-form-field.  This applies to guide/{new.html, metadata.html, stage.html, editall.html}.

The resulting layout is the same as before, except for a few cases where we were explicitly using colspan="2" in manually generated table rows.  e.g. the new.html radio button group for feature type.

image

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -832,12 +832,17 @@ class ChromedashForm(forms.Form):
 
     def as_table(self):
         "Return this form rendered as HTML <tr>s -- excluding the <table></table>."
-        header = '<tr%(html_class_attr)s><th colspan="2"><b>%(label)s</b></th></tr>'
-        html = header + '<tr%(html_class_attr)s><td>%(errors)s%(field)s</td><td>%(help_text)s</td></tr>'
+        # header = '<tr%(html_class_attr)s><th colspan="2"><b>%(label)s</b></th></tr>'
+        # html = header + '<tr%(html_class_attr)s><td>%(errors)s%(field)s</td><td>%(help_text)s</td></tr>'
+        label = '<span slot="label">%(label)s</span>'
+        field = '<span slot="field">%(field)s</span>'
+        error = '<span slot="error">%(errors)s</span>'
+        help =  '<span slot="help">%(help_text)s</span>'
+        html = '<chromedash-form-field %(html_class_attr)s>' + label + field + error + help + '%(label)s' + '</chromedash-form-field>'
         return self._html_output(
             normal_row=html,
-            error_row='<tr><td colspan="2">%s</td></tr>',
-            row_ender='</td></tr>',
+            error_row='<chromedash-form-field error_text="%s"></chromedash-form-field>',
+            row_ender='</chromedash-form-field>',
             help_text_html='<span class="helptext">%s</span>',
             errors_on_separate_row=False,
         )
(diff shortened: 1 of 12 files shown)


dlaliberte:
image

jrobbins:
I see what you are doing, but I'm not 100% clear on why we need these custom elements instead of using standard HTML `<table>` elements directly.    This PR description needs a sentence or two about the problem that you are trying to solve.

dlaliberte:
> I see what you are doing, but I'm not 100% clear on why we need these custom elements instead of using standard HTML `<table>` elements directly. This PR description needs a sentence or two about the problem that you are trying to solve.

The second paragraph of the description explains why we need components for both the row groups and the containing table:  `Since tables and trs depend on each other, initial dom parsing will remove all content of tables that is not trs, ...`

I initially tried only defining the chromedash-form-field, but then found that the dom parsing moved them out of the table container to be after (or before?) the table, so those rows were displayed out of order, and styling was inconsistent.  Maybe if I add style="display:table-row-group" to each chromedash-form-field, they will behave like rows and stay where they were defined.  

But furthermore, defining a custom element for the table (chromedash-form-table) and replacing all the table rows with chromedash-form-field allows us to more easily change the layout of all the tables and rows.  For example, we can add more columns across all the rows, or use a grid layout instead of table layout.

pages/guideforms.py
```diff
@@ -832,12 +832,17 @@ class ChromedashForm(forms.Form):
 
     def as_table(self):
         "Return this form rendered as HTML <tr>s -- excluding the <table></table>."
-        header = '<tr%(html_class_attr)s><th colspan="2"><b>%(label)s</b></th></tr>'
-        html = header + '<tr%(html_class_attr)s><td>%(errors)s%(field)s</td><td>%(help_text)s</td></tr>'
+        # header = '<tr%(html_class_attr)s><th colspan="2"><b>%(label)s</b></th></tr>'
```
jrobbins, line 835: Unneeded lines should be deleted.

static/elements/chromedash-form-table.js
```diff
@@ -0,0 +1,32 @@
+import {LitElement, css, html} from 'lit';
+// import {ifDefined} from 'lit/directives/if-defined.js';
```
jrobbins, line 2: Delete if not needed.

jrobbins (approved):
LGTM after addressing some comments.  I think I put in the same comments on an earlier version but didn't send them, so they may be duplicated.

static/elements/chromedash-form-table.js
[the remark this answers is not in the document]
dlaliberte: Deleting SHARED_STYLES too.

pages/guideforms.py
dlaliberte: done.

static/elements/chromedash-form-field.js
```diff
@@ -0,0 +1,87 @@
+import {LitElement, css, html} from 'lit';
+import {ifDefined} from 'lit/directives/if-defined.js';
+// import {live} from 'lit/directives/live.js';
```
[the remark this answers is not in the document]
dlaliberte: done.  also removed SHARED_STYLES.

static/elements/chromedash-form-table.js
dlaliberte: done

PR #1982, opened 2022-06-28
Make the footer sticky on every page

kevinshen56714:
In this PR, the footer css that was originally in the `metrics.scss` and `features.scss` is moved to `main.scss`, meaning the sticky footer is applied to all pages.

Basically, I found that the footer is currently only sticky on the metrics and features pages. However, I think it generally looks better on all the other pages as well if the footer always sticks at the bottom of the page.

For example,

#### Without sticky footer:
Screen Shot 2022-06-27 at 8 15 00 PM
#### With sticky footer:
Screen Shot 2022-06-27 at 8 16 00 PM

#### Without sticky footer:
Screen Shot 2022-06-27 at 8 19 50 PM

#### With sticky footer:
Screen Shot 2022-06-27 at 8 18 18 PM

### DIFF
--- a/static/sass/features/features.scss
+++ b/static/sass/features/features.scss
@@ -38,7 +38,6 @@ chromedash-legend {
 
 chromedash-featurelist {
   max-width: $max-content-width;
-  margin-bottom: 54px + $content-padding * 2; // bottom bar + some.
 }
 
 #subheader {
--- a/static/sass/features/features.scss
+++ b/static/sass/features/features.scss
@@ -52,16 +52,6 @@ chromedash-featurelist {
   }
 }
 
-@media only screen and (min-width: 701px) {
-
-  footer {
-    position: fixed;
-    bottom: 0;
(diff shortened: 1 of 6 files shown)


jrobbins:
One challenge with having a `position:fixed` footer is that it can sometimes overlap content at the bottom of a long page.  E.g., the user might not be able to access a submit button at the bottom of a form.  Also, a mobile user might not want to give up an entire row of screen real estate all the time.

I think the cure for overlapping the last element of a long page is to always have an extra margin at the bottom.  E.g., `chromedash-feature-list` has `margin-bottom: 86px`.  However, we should have that consistently throughout the app.  Maybe put it into `#content-column` instead of any element or individual page.

And, the cure for mobile is probably to make it not be `position:fixed` on a narrow or short screen.  I.e., if either dimension is less than 600px.

jrobbins (approved):

PR #1985, opened 2022-06-30
Try skeletons on feature detail page.

jrobbins:
This is just an experiment, it's not really needed but I think it might be a small UX improvement.

Right now, when loading the feature detail page, the user sees a blank page with a spinner, then the sl-details grey bar appears for a fraction of a second, then the feature information box is rendered which pushes that sl-details bar down to its final position.  The whole process takes about half a second, so it is not too distracting, but it is noticeable.

In this PR, I try replacing the spinner that is inside this virtual page with shimmering sl-skeleton elements.  The shape of the skeletons is generic and will not match the sections shown for any particular feature.  With this PR, the user initially sees a white box that looks like the eventual feature info box, with some skeleton placeholder content.  Then, the data loads and the actual content is shown a half second later.  There is still a visual pop, but I think that it is a little less noticeable.

Here's what it looks like in that half second:
image

### DIFF
--- a/static/components.js
+++ b/static/components.js
@@ -14,6 +14,7 @@ import '@shoelace-style/shoelace/dist/components/details/details.js';
 import '@shoelace-style/shoelace/dist/components/icon/icon.js';
 import '@shoelace-style/shoelace/dist/components/input/input.js';
 import '@shoelace-style/shoelace/dist/components/menu-item/menu-item.js';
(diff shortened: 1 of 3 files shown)


dlaliberte (approved):
I look forward to trying it.

kevinshen56714 (approved):
LGTM! I like they are within the feature div so that the outer frame is the same as once it's loaded.

PR #1991, opened 2022-07-01
Try on-demand `chromedash-approval-dialog` creation

kevinshen56714:
Following our [previous discussion](https://github.com/GoogleChrome/chromium-dashboard/pull/1976#pullrequestreview-1024066817), I moved the `openApprovalDialog` util function to the `chromedash-approval-dialog` file. The new function creates the element and appends it to the document body on demand. Once it's created, the function will use the same instance (like a singleton) that it has created. This allows us to not put the element in Django templates or other element's render function at all! If this looks good, I will create another PR for the toast util function as well.

I also slightly modified the Promises in the openWithFeature function because I think the original code does each fetch twice (one at variable definition and one at Promise.all)

### DIFF
--- a/static/elements/chromedash-approvals-dialog.js
+++ b/static/elements/chromedash-approvals-dialog.js
@@ -1,5 +1,7 @@
 import {LitElement, css, html, nothing} from 'lit';
 import '@polymer/iron-icon';
+import {showToastMessage} from './utils.js';
+
 import {SHARED_STYLES} from '../sass/shared-css.js';
 
 export const STATE_NAMES = [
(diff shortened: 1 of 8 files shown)


jrobbins (approved):
Awesome.  This certainly does clean up some code.

I don't think that the data was fetched twice because Promise.all() just waits for the promises to be resolved without running them again.  Also, in the chrome devtools network tab I can see a non-duplicated set of 4 requests each time I open the dialog.  Having said that, dealing with all 4 responses at once does simplify the code, and since loading was true until all responses were processed, there was no early partial rendering anyway.

jrobbins:
Idea for a possible followup PR: consider passing in an entire feature rather than the feature ID.  That would eliminate one request.

kevinshen56714:
> Awesome. This certainly does clean up some code.
> 
> I don't think that the data was fetched twice because Promise.all() just waits for the promises to be resolved without running them again. Also, in the chrome devtools network tab I can see a non-duplicated set of 4 requests each time I open the dialog. Having said that, dealing with all 4 responses at once does simplify the code, and since loading was true until all responses were processed, there was no early partial rendering anyway.

Cool, I'll use devtools to check requests more often as well! Thanks for checking this.

PR #1993, opened 2022-07-03
Add settings API and settings page component

kevinshen56714:
This PR converts the content of settings.html into a single component. Specifically,

1. `settings_api.py` and `settings_api_test.py` are created (under the /currentuser/settings route) for fetching user settings
2. `chromedash-settings-page.js` and `chromedash-settings-page_test.js` are created.
3. Improved the toast transition. Now, if the toast is open while another open toast instruction is fired, the previous toast would slide away before the new toast slides in. See the following video demo:

https://user-images.githubusercontent.com/11501902/177060325-9a01856c-1cd5-41f7-96ff-cf3e731a7730.mov

### DIFF
--- a/api/settings_api.py
+++ b/api/settings_api.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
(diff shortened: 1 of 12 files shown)


jrobbins (approved):
Good job.  This is one more page done.
I like the pattern of disabling the submit button while submitting.

PR #1997, opened 2022-07-04
Convert footer into a component

kevinshen56714:
This PR converts `footer.html` into `chromedash-footer.js` component (`footer.html` can be safely removed). 

This is a necessary step for getting rid of django templates in the `_base.html` and for setting up a SPA.


jrobbins (approved):
This looks good.

Feel free to inline the CSS as lit-css in this PR or in a follow-up PR.
It should be pretty easy to identify the blocks of SASS that need to be moved and translated:
https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/sass/main.scss#L197
and
https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/sass/main.scss#L331

kevinshen56714:
> This looks good.
> 
> Feel free to inline the CSS as lit-css in this PR or in a follow-up PR. It should be pretty easy to identify the blocks of SASS that need to be moved and translated: https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/sass/main.scss#L197 and https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/sass/main.scss#L331

I moved the footer CSS into the component, please review the changes. Thanks!

static/sass/main.scss
```diff
@@ -328,15 +309,8 @@ footer {
   color: var(--warning-color);
 }
 
-// Enable sticky footer when the screen is wide and long enough
+// Add margin-bottom when sticky footer is enabled
```
jrobbins, line 312: This is fine.  However, I think that you could take it a step further by eliminating this whole block of SASS.   Instead of the margin being added to @content-column, that space would be in the height or margin of a `<div id="footer-space"></div>` element inside `chromedash-footer` as a sibling to the `<footer>` element.  I'm just brainstorming, I haven't tried it myself.

jrobbins (approved):

static/sass/main.scss
kevinshen56714: This is a great idea! As you suggested, I added a `<div>` in the footer component as a spacer and removed the css in main.scss.

jrobbins (approved):

PR #2002, opened 2022-07-07
Label checkboxes via django template.

jrobbins:
This resolves a problem where checkboxes where not being POSTed as part of a form.

The problem with our previous approach was that shoelace form elements cannot be nested inside custom elements that are themselves within a `<form>`.  Shoelace relies on the standard JS method `closest()` to find the `<form>` element, and that function does not cross shadowRoot boundaries.

Instead of using `chromedash-checkbox` to place a label beside the box, we use the django template to pass the label as the body of `sl-checkbox`.  This avoids the level of nesting that caused the problem.

Django does not allow the expression `widget.attrs.get("label")` or `widget.attrs["label"]`, so we need to loop and check.


kevinshen56714:
@jrobbins this will break the settings page, which still uses `chromedash-checkbox`: https://github.com/GoogleChrome/chromium-dashboard/blob/c730fca07a1bc02acf045e102f8aadf350892961/static/elements/chromedash-settings-page.js#L101

I will submit a PR soon to fix this.

PR #2003, opened 2022-07-07
Remove a missed reference to chromedash-checkbox.

jrobbins:
I missed these references earlier.

I think that unit tests passed before because the settings page did not explicitly include chromedash-checkbox, so there was no import failure to raise an error.  In such a situation, the element does exist, even if it is not defined as a custom element.


dlaliberte (approved):
We can delete chromedash-checkbox now, by the way.

jrobbins:
> We can delete chromedash-checkbox now, by the way.

I already did in my previous PR.  That's why I was expecting JS unit tests to fail on that previous PR.

PR #2004, opened 2022-07-07
Convert header into a component

kevinshen56714:
This converts `header.html` into the `chromedash-header` component. Specifically,

1. Added `chromedash-header.js` and `chromedash-header_test.js`
2. Moved all header-related css from `main.scss` into the new component
3. Removed `header.html`
4. Changed to using the JavaScript API of the Google Sign-In service

Note that as I mentioned, the Google Sign-In library cannot find elements in a shadowDOM. So the Google sign-in button is created at the document level and then slotted into the header element.


jrobbins:
This all looks good.
I really like your solution for the sign in button.

I'd like to just hold this PR for a couple of days because I will deploy a new version of the site tomorrow and sometimes there is a need to follow up with fixes or reverting commits.  After that is all settled, I will approve this PR.

jrobbins (approved):
Looks good. Go ahead and merge now.

PR #2006, opened 2022-07-07
Only show inactive feature statuses in the title.

jrobbins:
The old django template only showed certain status values in the page `<h2>` title.   That logic was left out of https://github.com/GoogleChrome/chromium-dashboard/pull/1972.

This PR restores that behavior.


kevinshen56714 (approved):
This is my mistake. Thanks for fixing it!

PR #2007, opened 2022-07-08
Migrate remaining help_text to form-field-specs.js.

dlaliberte:
In addition to migrating all the remaining help_text to the form-field-specs.js, we can simplify chromedash-form-field.js to not use the "help" slot, and simplify ChromedashForm in guideforms.py to not process any help text.

### DIFF
--- a/static/elements/form-field-specs.js
+++ b/static/elements/form-field-specs.js
@@ -1,9 +1,18 @@
 import {html} from 'lit';
 
+
+const SHIPPED_HELP_TXT = html`
+  First milestone to ship with this status. Applies to: Enabled by
+  default, Browser Intervention, Deprecated and Removed.`;
+
+const
+  SHIPPED_WEBVIEW_HELP_TXT = html`
+  First milestone to ship with this status.
+  Applies to Enabled by default, Browser
+  Intervention, Deprecated, and Removed.`;
+
 // Map of specifications for all form fields.
-// TODO:
-//   * Finish migrating remaining fields.
-//   * Migrate other properties.
+// TODO: Migrate other properties.
 export const ALL_FIELDS = {
 
   'name': {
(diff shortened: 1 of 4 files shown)


static/elements/form-field-specs.js
```diff
@@ -81,19 +89,26 @@ export const ALL_FIELDS = {
...
+        default, be sure to set the equivalent Process stage.`,
   },
 
   'bug_url': {
     help_text: html`
         Tracking bug url (https://bugs.chromium.org/...). This bug
         should have "Type=Feature" set and be world readable.
-        Note: This field only accepts one URL.`,
+        Note: This field only accepts one URL.
+        <br/><br>
```
jrobbins, line 102: Any reason why one of these BR's has the slash and the other does not?  It is a self-closing tag regardless.

static/elements/form-field-specs.js
```diff
@@ -357,20 +385,18 @@ export const ALL_FIELDS = {
       designs? Double check that your experiment makes sense given that 
       a large developer (e.g. a Google product or Facebook) likely 
       can't use it in production due to the limits enforced by origin 
-      trials.\n\nIf Intent to Extend Origin Trial, highlight new/different 
-      areas for experimentation. Should not be an exact copy of goals 
+      trials.\n\nIf you send an Intent to Extend Origin Trial, highlight
```
jrobbins, line 388: Do the `\n\n` need to be `<br>`s ?

static/elements/form-field-specs.js
```diff
@@ -357,20 +385,18 @@ export const ALL_FIELDS = {
       designs? Double check that your experiment makes sense given that 
       a large developer (e.g. a Google product or Facebook) likely 
       can't use it in production due to the limits enforced by origin 
-      trials.\n\nIf Intent to Extend Origin Trial, highlight new/different 
-      areas for experimentation. Should not be an exact copy of goals 
+      trials.\n\nIf you send an Intent to Extend Origin Trial, highlight 
+      areas for experimentation. They should not be an exact copy of the goals
```
jrobbins, line 389: Why take out "new/deferent"?  I assume those words were used oringally because extending an experiment should include be something that is different from the original experiment.

static/elements/form-field-specs.js
```diff
@@ -454,20 +480,24 @@ export const ALL_FIELDS = {
...
+      email addresses of API owners who LGTM'd 
+      the Intent to Ship email thread.  `,
   },
 
   'r4dt_lgtms': {
     help_text: html`
       Full email addresses of API owners who LGTM\'d 
-      the Request for Deprecation Trial email thread.`,
+      the Request for Deprecation Trial email thread.
+      This field fills automatically
```
jrobbins, line 499: Same.

static/elements/form-field-specs.js
```diff
@@ -454,20 +480,24 @@ export const ALL_FIELDS = {
   'i2e_lgtms': {
     help_text: html`
       Full email address of API owner who LGTM\'d the 
-      Intent to Experiment email thread.`,
+      Intent to Experiment email thread. 
+      This field fills automatically
```
jrobbins, line 484: Sadly, that part of internals/detect_intent.py does not work.  So, I think it would be best to leave this new sentence off.  Eventually, we'll phase out this field.

static/elements/form-field-specs.js
```diff
@@ -571,31 +602,31 @@ export const ALL_FIELDS = {
...
       this feature on desktop platforms by setting a flag. 
       When flags are enabled by default in preparation for 
-      removal, please use the fields in the ship stage.`,
+      shipping or removal, please use the fields in the ship stage.`,
   },
 
   'flag_name': {
     help_text: html`
-      Name of the flag that enables this feature.'`,
+      Name of the flag on chrome://flags that enables this feature.'`,
```
jrobbins, line 629: I didn't notice it before, but there is an extra apostrophe at the end of this string.

jrobbins (changes requested):
These help content changes generally look very good, but I have a few comments for you to address.

static/elements/form-field-specs.js
dlaliberte: I thought I used `<br/>` for both.  There used to be a good reason to use `<br/>` but maybe it no longer matters.
But it is generally better to avoid br altogether and instead use block tags around neighboring text, so the extra whitespace is inferred.  That's more work :)

static/elements/form-field-specs.js
dlaliberte: Yes.  Missed those.

static/elements/form-field-specs.js
dlaliberte: Joe's change removed the "new/different" text, probably because it is moderately unessential.  But perhaps there is nothing new or different, but they just need another round of experimentation, so maybe it should say "highlight the reason for the extension".

static/elements/form-field-specs.js
dlaliberte: OK.

static/elements/form-field-specs.js
dlaliberte: done

static/elements/form-field-specs.js
dlaliberte: Thanks for noticing that.  The migration was rather messy, with many cases I could not automate consistently.

jrobbins (approved):

PR #2011, opened 2022-07-11
New permissions system implementation

DanielRyanSmith:
This PR adds a new permissions system to ChromeStatus, which allows feature ownership and more strictly regulates who can update information for each feature.

## Changelist
- New user access level: Site Editor.
  - A site editor will have access to edit any feature, but will not be granted admin privileges, such as designating new admins/site editors or deleting features.
  - A user can be designated as a site editor on the admin user permissions page.
- New feature-level permission: Editor.
  - Any number of editors can now be added to any feature. This designation is similar to owners, but editors are not publicly displayed in the list of owners on the feature.
  - Editors can be added at any time during feature creation or editing.
- Features now recognize ownership and edit access is only allowed to certain users associated with the feature.
  - Aside from admins and site editors, only feature owners, editors, and creators can make changes to a given feature.
  - Unlisted features are only visible to users who have edit access to the unlisted feature (including admins and site editors).
  - The dropdown in the "My Features" view has been renamed "Features I can edit" rather than "Features I own". It now displays all features in which the user is designated as an owner, editor, or creator.
  - The ability to create features is still dictated by the old permission structure (any email address from certain domains).

## Implementation information
Some new fields have been added to datastore.
 - The `Feature` entity contains the new fields:
   - `creator`, which represents the original creator of the feature. This has been added to better query for this information, as the default user field `created_by` comes with baggage that makes it difficult to use this way.
   - `editors`, which represents a list of users designated to have edit access to the feature.
 - The `AppUser` entity contains the new field `is_site_editor`, which designated that user as having edit access to every feature.

## Screenshots
Because a user designated as an admin will have elevated privileges compared to a site editor, the site editor checkbox will be checked and disabled when the admin access is added to a new user.
Screen Shot 2022-07-11 at 11 05 36 AM

"Features I own" dropdown has been renamed "Features I can edit", which displays all features in which the user is designated as an owner, editor, or creator.
Screen Shot 2022-07-11 at 11 46 52 AM

Users will have edit access and can view unlisted features in which they are an owner, editor, or creator of (To display for this example, the descriptions in these mock-features describe the permissions of the current user)
Screen Shot 2022-07-11 at 12 20 29 PM

### DIFF
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -84,8 +84,7 @@ def feature_edit_list(user):
     return []
 
   # Query features to find which can be edited.
-  features_editable = models.Feature.get_all(filterby=('owner', user.email()))
-  features_editable += models.Feature.get_all(filterby=('editors', user.email()))
+  features_editable = models.Feature.get_all(filterby=('can_edit', user.email()))
   # Return a list of unique ids of features that can be edited.
   return list(set([f['id'] for f in features_editable]))
 
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -65,9 +65,12 @@ def can_edit_any_feature(user):
   """Return True if the user is allowed to edit all features."""
   if not user:
     return False
-  app_user = models.AppUser.get_app_user(user.email())
+  app_user = models.AppUser.get_app_user(user.email())  
+  if not app_user:
+    return False
+
   # Site editors or admins should be able to edit any feature.
-  return app_user and (app_user.is_admin or app_user.is_site_editor)
+  return app_user.is_admin or app_user.is_site_editor
 
 
 def feature_edit_list(user):
@@ -87,20 +90,24 @@ def feature_edit_list(user):
   return list(set([f['id'] for f in features_editable]))
 
 
-def can_edit_feature(user, feature_id=None):
+def can_edit_feature(user, feature_id):
   """Return True if the user is allowed to edit the given feature."""
   # If the user can edit any feature, they can edit this feature.
   if can_edit_any_feature(user):
     return True
 
-  if not feature_id:
+  if not feature_id or not user:
     return False
 
-  # Check if the user is an owner or editor for this feature. If yes, can edit.
   feature = models.Feature.get_feature(feature_id)
-  return (feature and
-            (user.email() in feature['browsers']['chrome']['owners'] or
-            user.email() in feature['editors']))
+  if not feature:
+    return False
+  
+  # Check if the user is an owner or editor for this feature. If yes, can edit.
+  is_owner = ('owners' in feature['browsers']['chrome'] and
+                user.email() in feature['browsers']['chrome']['owners'])
+  is_editor = 'editors' in feature and user.email() in feature['editors']
+  return is_owner or is_editor
 
 
 def can_approve_feature(user, feature, approvers):
@@ -168,10 +175,20 @@ def require_create_feature(handler):
   return check_login
 
 
-def require_edit_feature(handler):
-  """Handler decorator to require the user can edit the current feature."""
-  def check_login(self, *args, **kwargs):
-    return _reject_or_proceed(
-        self, handler, args, kwargs, can_edit_feature)
+def require_edit_feature_permission(obj, feature_id):
+  """Check if user has permission to edit feature and abort if not."""
+  """Redirect, abort(403), or call handler_method."""
+  user = obj.get_current_user()
+  req = obj.request
 
-  return check_login
+  # Give the user a chance to sign in
+  if not user and req.method == 'GET':
+    return obj.redirect(settings.LOGIN_PAGE_URL)
+
+  # Redirect to 404 if feature is not found.
+  if models.Feature.get_by_id(int(feature_id)) is None:
+    obj.abort(404, msg='Feature not found')
+  
+  # Redirect to 403 if user does not have edit permission for feature.
+  if not can_edit_feature(user, feature_id):
+    obj.abort(403)
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -84,7 +84,8 @@ def feature_edit_list(user):
     return []
 
   # Query features to find which can be edited.
-  features_editable = models.Feature.get_all(filterby=('can_edit', user.email()))
+  features_editable = models.Feature.get_all(
+    filterby=('can_edit', user.email()))
   # Return a list of unique ids of features that can be edited.
   return list(set([f['id'] for f in features_editable]))
 
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -125,7 +125,6 @@ def can_approve_feature(user, feature, approvers):
 def _reject_or_proceed(
     handler_obj, handler_method, handler_args, handler_kwargs,
     perm_function):
-
   """Redirect, abort(403), or call handler_method."""
   user = handler_obj.get_current_user()
   req = handler_obj.request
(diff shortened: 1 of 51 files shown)


internals/search.py
```diff
@@ -211,7 +214,13 @@ def process_query(
       logging.info('combining result so far with %r', feature_ids)
       result_id_set.intersection_update(feature_ids)
   result_id_list = list(result_id_set or [])
-  total_count = len(result_id_list)
+
+  # TODO(danielrsmith): Update this to more efficiently count the
```
jrobbins, line 218: I think that you have hit upon a defect in my existing code for paginating with unlisted features: search results do not show unlisted features that the signed in user could edit.

However, I don't think that this is the right fix because retrieving the entire list of features before pagination means that we are retrieving a lot more features.  One of the main goals of this API is to be faster than our old approach by avoiding doing that.

So, I think that you could back out your changes to this file and just leave a TODO for now.   E.g., # TODO(jrobbins): results should include unlisted features that the signed in user can edit.

internals/models.py
```diff
@@ -1236,6 +1272,7 @@ def put(self, notify=True, **kwargs):
 
   # General info.
   category = ndb.IntegerProperty(required=True)
+  creator = ndb.StringProperty()
```
jrobbins, line 1275: We don't need a new field for creator because we already have created_by.

pages/guide.py
```diff
@@ -139,6 +140,8 @@ def process_post_data(self):
         intent_stage=models.INTENT_NONE,
         summary=self.form.get('summary'),
         owner=owners,
+        editors=editors,
+        creator=signed_in_user.email(),
```
jrobbins, line 144: We don't need to set creator because we already set created_by.

static/elements/chromedash-feature-page.js
```diff
@@ -225,7 +225,7 @@ export class ChromedashFeaturePage extends LitElement {
               </a>
             </span>
           `: nothing}
-          ${this.user && this.user.can_edit ? html`
+          ${this.user && this.user.can_edit_all ? html`
```
jrobbins, line 228: Don't you also need to check if this feature's ID is in the  list of IDs that the signed in user can edit?

static/elements/chromedash-roadmap-milestone-card.js
```diff
@@ -1,6 +1,6 @@
 import {LitElement, html, nothing} from 'lit';
 import {ROADMAP_MILESTONE_CARD_CSS} from
-  '../sass/elements/chromedash-roadmap-milestone-card-css.js';
+'../sass/elements/chromedash-roadmap-milestone-card-css.js';
```
jrobbins, line 3: Usually we use indentation to show that the line is being continued from the previous line.  So, let's keep this indent.

static/elements/chromedash-userlist.js
```diff
@@ -55,6 +55,32 @@ class ChromedashUserlist extends LitElement {
...
+        return -1;
+      }
+      if ((b.is_admin && !a.is_admin) || (b.is_site_editor && (!a.is_site_editor && !a.is_admin))) {
+        return 1;
+      }
+      return a.email.localeCompare(b.email);
+    });
+  }
+
+  _onAdminToggle() {
```
jrobbins, line 70: Both of your enhancement on this page are good.

framework/permissions.py
```diff
@@ -130,11 +175,20 @@ def check_login(self, *args, **kwargs):
   return check_login
 
 
-def require_edit_feature(handler):
-  """Handler decorator to require the user can edit the current feature."""
-  # TODO(jrobbins): make this per-feature
-  def check_login(self, *args, **kwargs):
-    return _reject_or_proceed(
-        self, handler, args, kwargs, can_edit_any_feature)
+def validate_feature_edit_permission(obj, feature_id):
```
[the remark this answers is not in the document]
DanielRyanSmith: Yes, it might be better to retain the name `handler_obj` as it does in `_reject_or_proceed()`.

As for keeping this in decorator form, I couldn't find a consistent approach that was possible in that regard. From what I could tell, it is not possible to pass a dynamic parameter to a decorator function, and we now need to know the feature ID to check if the user has permission to view that feature (we have to check if they're an owner, editor, or creator).
I originally had an implementation that changed `_reject_or proceed()` to check `handler_kwargs` to see if the original function was passed a `feature_id` argument, and pass it into the `perm_function` invocation. This ended up feeling like a bad practice and odd corner cases kept popping up when running, and in testing. The current implementation was what I settled on for simplicity's sake.

internals/models.py
DanielRyanSmith: I thought as much, but then I saw this comment:
https://github.com/GoogleChrome/chromium-dashboard/blob/1b142660f5abbeae171b6146913022e7e926339a/internals/models.py#L1367-L1370

I did try and query `created_by` anyway, but I was not able to get it to work properly. After reading some common problems, it seemed like other devs recommended keeping a separate `created` field if it needs to be referenced, as `created_by` comes with additional baggage being a `UserProperty`.
However, there might be a way to use it - this is just the solution I came up with.

internals/search.py
DanielRyanSmith: I agree that this is not the right fix here (hence the TODO 😅). Maybe I can take a deeper look and implement a fix here that satisfies the efficiency needs and displays the correct number.

static/elements/chromedash-roadmap-milestone-card.js
DanielRyanSmith: I think this was automatic formatting - I'll remove this.

static/elements/chromedash-userlist.js
DanielRyanSmith: Thanks! 😊

static/elements/chromedash-feature-page.js
DanielRyanSmith: Yes that's true! I this this is a WIP page created by Kevin, so the tests don't run against this page. I'll add the functionality regardless so that it works with the new changes.

framework/permissions.py
```diff
@@ -62,16 +62,52 @@ def can_create_feature(user):
...
+  if not user:
+    return False
+  
+  # If the user can edit any feature, we don't need the full list.
+  # We can just assume they will have edit access to all features.
+  if can_edit_any_feature(user):
+    return []
+
+  # Query features to find which can be edited.
+  features_editable = models.Feature.get_all(
```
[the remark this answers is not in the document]
DanielRyanSmith: `can_edit` will run a query that checks if the user's email is listed in the `owner` field, listed in the `editor` field, or is equal to `creator` field, so it should catch all of these scenarios.

jrobbins (changes requested):

internals/models.py
```diff
@@ -1000,20 +1024,18 @@ def getSortingMilestone(feature):
...
 
   @classmethod
   def get_in_milestone(
       self, show_unlisted=False, milestone=None):
     if milestone == None:
       return None
 
-    cache_key = '%s|%s|%s|%s' % (
-        Feature.DEFAULT_CACHE_KEY, 'milestone', show_unlisted, milestone)
+    cache_key = '%s|%s|%s' % (
```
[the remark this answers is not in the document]
DanielRyanSmith: I see what you mean. I updated the code to cache the result before filtering unlisted.

internals/search.py
```diff
@@ -184,6 +194,13 @@ def process_query(
   terms = TERM_RE.findall(user_query + ' ')[:MAX_TERMS] or []
   if not show_deleted:
     terms.append(('deleted', '=', 'false', None))
+
+  # If part of the query represents the user's feature ownership
```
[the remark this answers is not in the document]
DanielRyanSmith: Sure! Added.

pages/intentpreview.py
```diff
@@ -36,6 +36,7 @@ class IntentEmailPreviewHandler(basehandlers.FlaskHandler):
 
   @permissions.require_edit_feature
   def get_template_data(self, feature_id=None, stage_id=None):
+
```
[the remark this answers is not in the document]
DanielRyanSmith: Removed!

PR #2016, opened 2022-07-12
Workaround conflict with Shoelace event listeners.

dlaliberte:
This PR simply adds type="module" to the script tag that attaches event listeners to the form for each of the form fields.  This is required to correctly validate form fields before submission, and block submission if there are any invalid fields.

### DIFF
--- a/templates/_base.html
+++ b/templates/_base.html
@@ -129,7 +129,12 @@ limitations under the License.
   <script src="https://www.googletagmanager.com/gtag/js?id=UA-179341418-1"
           async nonce="{{nonce}}"></script>
 
-  <script nonce="{{nonce}}">
+  {% comment %}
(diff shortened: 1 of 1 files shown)


dlaliberte:
See https://github.com/shoelace-style/shoelace/issues/718 for the suggested workaround involving type="module".

jrobbins (approved):
I tested this in FF and it worked fine.

I think I understand the root cause now.  This line in our cs-client.js file causes the form to be submitted.  If that logic runs before the shoelace validation logic, then we will submit regardless of validation.  Changing the order of event listener registration by adding type="module" also changes the order of event listener execution, which give the shoelace validation code a chance to cancel the event before it reaches our code.
https://github.com/GoogleChrome/chromium-dashboard/blob/main/static/js-src/cs-client.js#L67

PR #2017, opened 2022-07-12
Enable new permissions system

DanielRyanSmith:
This PR fully enables the permissions system described in #2011. This PR will enact the new changes, and the application will behave as intended in the original design.


jrobbins:
The number of editors values filled in automatically seems to have leveled off.  So, we can go ahead with rebasing this PR, updating any newly added code, re-reviewing it, and landing it.

PR #2018, opened 2022-07-13
Add chromedash-metrics-page

kevinshen56714:
This PR moves all the metrics-related html content into the `chromedash-metrics-page` component. The changes include:

1. Added `chromedash-metrics-page.js` in which the /data/ (for metrics data) or the /data/blink/ (for feature buckets data; this exists but was never called before) API endpoint is called.
2. Safely removed`static/js-src/metrics-page.js`, `static/js-src/metrics-timeline-page.js`, `templates/metrics/css/_css_metric_nav.html`, and `templates/metrics/feature/_feature_metric_nav.html`.
3. Modified all of the 6 metrics html files and they all now contain only one component.
4. Enhanced chromedash-timeline UI by adding a new error message if the entered feature name does not exist and a new progress bar to indicate the timeline plot is loading (see the demo below).

Overall, this PR only transferred content into the new component. The next step is to implement the UI changes that we discussed as well as to add a unit test.

https://user-images.githubusercontent.com/11501902/178640229-3970b239-39c3-4fbd-89cc-626458f45a23.mov

https://user-images.githubusercontent.com/11501902/178640213-0a375ce2-5cac-43cf-8ad8-f060bfb6844c.mov


static/elements/chromedash-timeline.js
```diff
@@ -234,42 +237,44 @@ class ChromedashTimeline extends LitElement {
...
       hadEl.src = dsEmbedUrl;
 
-      const bigqueryEl = this.shadowRoot.getElementById('bigquery');
+      const bigqueryEl = this.shadowRoot.querySelector('#bigquery');
       bigqueryEl.textContent = `#standardSQL
-SELECT yyyymmdd, client, pct_urls, sample_urls
-FROM \`httparchive.blink_features.usage\`
-WHERE feature = '${featureName}'
-ORDER BY yyyymmdd DESC, client`;
+        SELECT yyyymmdd, client, pct_urls, sample_urls
```
jrobbins, line 258: Does this add visible space that would be copied onto the clipboard if the user tries to use this code?

static/elements/chromedash-metrics-page.js
```diff
@@ -0,0 +1,263 @@
...
+          }
+        }
+      `];
+  }
+
+  static get properties() {
+    return {
+      type: {type: String}, // "css" or "feature"
+      view: {type: String}, // "popularity" or "animated"
+      mode: {type: String}, // "metrics" or "timeline"
```
jrobbins, line 98: Why have a mode rather than two separate page elements?

kevinshen56714: Updated!

static/elements/chromedash-timeline.js
kevinshen56714: It actually does, so I have changed it back.

PR #2021, opened 2022-07-13
Create scheduled DataStore backups.

DanielRyanSmith:
Addresses #917

This PR implements a new scheduled job that runs daily to export the project's DataStore data to a backup bucket. This data is set to persist for 60 days before being deleted.

Created following the [GCP schedule export guide](https://cloud.google.com/datastore/docs/schedule-export).

### DIFF
--- a/internals/data_backup.py
+++ b/internals/data_backup.py
@@ -15,6 +15,7 @@ import os
 
 from googleapiclient.discovery import build
 from googleapiclient.discovery_cache.base import Cache
+import settings
 
 from framework import basehandlers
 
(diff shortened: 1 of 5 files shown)


internals/data_backup.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def set(self, url, content):
+        MemoryCache._CACHE[url] = content
+
+
+class BackupExportHandler(basehandlers.FlaskHandler):
+  """Triggers a new Datastore export."""
+
+  def get_template_data(self):
+    bucket = os.environ.get("BACKUP_BUCKET")
```
jrobbins, line 35: What sets this environment variable?
If you want to use an environment variable, it could be set in app.yaml, but that file is the same for staging and prod.
However, it might be more clear to set it as a constant in settings.py.  There are conditional sections there for staging and prod.

internals/data_backup.py
```diff
@@ -0,0 +1,48 @@
...
+    # The default cache (file_cache) is unavailable when using oauth2client >= 4.0.0 or google-auth,
+    # and it will log worrisome messages unless given another interface to use.
+    datastore = build("datastore", "v1", cache=MemoryCache())
+    project_id = os.environ.get("GOOGLE_CLOUD_PROJECT")
```
jrobbins, line 39: This value is available as settings.APP_ID.   Using that variable makes it slightly easier to set a mock value during testing.

jrobbins:
FYI, this PR will be significant progress toward resolving issue #917.

internals/data_backup.py
DanielRyanSmith: Yes, this is the right approach, and I've changed it now to use it.

jrobbins (approved):

PR #2022, opened 2022-07-14
Move the new feature button to the heading.

jrobbins:
This addresses feedback from Joe that we should make the "Add feature" button easier to find before making the roadmap page the site home page.

In this PR:
* Remove the "Add new feature" button from the all features page, along with associated CSS.
* Add a "Add feature" sl-button in the header near the account menu, iff the user is signed in, has permission to create a feature, and is not already on the feature creation page.

image

### DIFF
--- a/static/elements/chromedash-header.js
+++ b/static/elements/chromedash-header.js
@@ -261,6 +261,11 @@ export class ChromedashHeader extends LitElement {
     return html`
       <div class="flex-container flex-container-inner-second">
       ${this.user ? html`
+        ${this.user.can_create_feature && !this.isCurrentPage('/guide/new') ? html`
+          <sl-button href="/guide/new" variant="primary" size="small">
+            Add feature
+          </sl-button>
+        `: nothing }
         <div class="nav-dropdown-container">
           <a class="nav-dropdown-trigger">
             ${this.user.email}
--- a/static/sass/features/features.scss
+++ b/static/sass/features/features.scss
@@ -64,11 +64,6 @@ chromedash-featurelist {
   .feature-count {
     min-width: 170px;
   }
-  .actionlinks {
-    .blue-button {
-      margin-left: 4px;
-    }
-  }
 }
 
 @media only screen and (max-width: 1100px) {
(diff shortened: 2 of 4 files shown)


jrobbins:
This is what it looks like on a small device.  I did not keep the circle-plus icon and the @media query to reduce the button to only that icon, because I think this looks reasonable as-is.  In the future, we should redo our site to have a hamburger menu on small screens.  In such a case, this button can go inside that menu.

image

dlaliberte:
This seems fine, but somewhat tangentially, I still think it would be better to show the "Add feature" button *only* on the "My features" page, since you can't add a feature unless you are logged in, and users don't need to see it otherwise.  
Adding a feature is rare enough that it is not like we have to advertise the ability to add a feature even when the user is not logged in.  Taking up space for the button on the heading, along with the list of tabs and log-in/settings dropdown, seems out of place.

Furthermore, I'd probably recommend merging "My features" and "All features" into one "Features" tab, and on that page, have a filter for "My features", and also allow easy filtering by other criteria, such as by some other user, or group, or topic area.  Then it would make sense, if you are logged in, to show the "Add feature" button.   But again, we shouldn't show "My features" unless the user is logged in.

dlaliberte (approved):

jrobbins:
The button is only shown when signed in.

I understand the point about adding a feature being an infrequent operation.  However, it is often the first thing that a new logged in user is looking for: a feature owner just knows that they want to launch a feature, they are told that they need to follow the launching-features doc, and that doc tells them to create a feature in chromestatus.  They may have never used chromestatus before, and are already starting off the interaction on a bad foot because they see the whole process as bureaucratic overhead.  So, given that situation, think we help those users out by having the button available on the first page they see, which will often be the roadmap page, but it could be any page.  

I considered having the my-features functionality as just a query term on the regular features page.  In fact, that is basically the way that it is implemented under the covers.   However, we went with a separate page to (a) increase discoverability, and (b) put multiple boxes on that page whereas the all features page is one large box.

PR #2023, opened 2022-07-14
Stack rank page UI refactor

kevinshen56714:
This PR is the first step towards the new metrics page. The changes include:

1. Added a timeline button for each property/feature
2. Added a new SORT BY drop down menu and slightly modified the sorting method
3. Percentage numbers are shown on the side of the bar (so that the color won't be affected by the background color of the bar)
4. Hovering property name will show a link symbol on the side and clicking it allows users to copy the hash link
5. Added loading skeleton animations

## Video demo

https://user-images.githubusercontent.com/11501902/179089892-f3bb66c1-82dc-44a7-8359-b2aa2fe9ee4f.mov

## Mobile view
Mobile view


static/elements/chromedash-stack-rank.js
```diff
@@ -93,94 +157,141 @@ class ChromedashStackRank extends LitElement {
...
+    `;
   }
 
-  _updateSortIcons() {
-    this.propertyNameSortIcon = this._getOrderIcon('property_name');
-    this.percentSortIcon = this._getOrderIcon('percentage');
+  renderStackRank() {
+    return html`
+      ${this.viewList.map((item) => html`
+        <li class="stack-rank-item" id="${item.property_name}" tabindex="0">
```
jrobbins, line 226: Do we need `tabindex="0"`?  I think that makes the row selectable by using the tab key, but it is not a link or anything, so I don't see the purpose of selecting the row.  If someone was trying to navigate using tab and enter keys, having all these extra tab-able elements would mean that they need to press tab 1/3 more often to get to what they want.

static/elements/chromedash-x-meter.js
```diff
@@ -14,53 +14,46 @@ class ChromedashXMeter extends LitElement {
...
-        color: var(--barchart-color);
-        cursor: pointer;
-      }
-
     `];
   }
 
   render() {
     return html`
-      <a href="${this.href}">
```
jrobbins, line 58: Could we keep the href?  That would allow users to click on the bar or percentage number if that is the way that they remember to do it.

static/elements/icons.js
```diff
@@ -91,10 +91,14 @@ const template = html`
 
 <g id="share"><path d="M18 16.08c-.76 0-1.44.3-1.96.77L8.91 12.7c.05-.23.09-.46.09-.7s-.04-.47-.09-.7l7.05-4.11c.54.5 1.25.81 2.04.81 1.66 0 3-1.34 3-3s-1.34-3-3-3-3 1.34-3 3c0 .24.04.47.09.7L8.04 9.81C7.5 9.31 6.79 9 6 9c-1.66 0-3 1.34-3 3s1.34 3 3 3c.79 0 1.5-.31 2.04-.81l7.12 4.16c-.05.21-.08.43-.08.65 0 1.61 1.31 2.92 2.92 2.92 1.61 0 2.92-1.31 2.92-2.92s-1.31-2.92-2.92-2.92z"/></g>
 
+<g id="sort"><path d="M3 18h6v-2H3v2zM3 6v2h18V6H3zm0 7h12v-2H3v2z"></path></g>
```
jrobbins, line 94: Good job adding icons.  I always thought that was one of the more obscure aspects of this code.

jrobbins (changes requested):
Overall this looks great.  It's a big improvement to the UX and the code quality.

static/elements/chromedash-stack-rank.js
kevinshen56714: I see. It is removed!

static/elements/chromedash-x-meter.js
kevinshen56714: Added back on the progress bar (not on the percentage number).

static/elements/icons.js
kevinshen56714: Thanks!

PR #2024, opened 2022-07-14
Start migrating form field labels to form-field-specs.js

dlaliberte:
* All the fields for new features are migrated.
* Was necessary to leave label='' in guideforms.py for migrated labels, otherwise the field name is used.

Note that we can't migrate the placeholder and title text so easily since these are passed only into the django form field as attributes and not yet exposed to the chromedash-form-field element.  We should first add the form field type and then we can construct the form fields with Shoelace elements directly, along with the rest of the attributes.


static/elements/chromedash-form-field.js
```diff
@@ -54,12 +54,15 @@ export class ChromedashFormField extends LitElement {
 
   render() {
     const fieldProps = ALL_FIELDS[this.name] || {};
+    const labelProp = fieldProps.label;
+    const label = labelProp ? `${labelProp}:` : '';
```
jrobbins, line 58: I think these two lines could be simpler if done like the way you did helpText:
`const label = fieldProps.label || '';`

jrobbins (approved):
LGTM after addressing one comment.

static/elements/chromedash-form-field.js
dlaliberte: We need to add the ':' but only if fieldProps.label is defined.   Django adds the ':', I believe, when it uses the label if provided. So we have to do the same here.

I can duplicate the fieldProps.label expression and drop the labelProp var.

PR #2027, opened 2022-07-15
Fix roadmap page API call when showing unlisted.

jrobbins:
When testing locally signed in to a site admin account I found that the roadmap page would not load.
An exception happed in the JSON request to `/api/v0/features?milestone=...`
`TypeError: Object of type Feature is not JSON serializable`

I believe the root cause was the else clause where one part of `all_features` was copied and assigned to `features_by_type`.  The difference between these two dictionaries is that (1) `features_by_type` is filtered by permissions, and also that (2) it has all python dictionaries ready to be serialized for the JSON response rather than the NDB Feature model instances.   Copying from one dict to the other introduced Feature model instances where dictionaries were expect.   In fact, I don't think that this else-clause was needed because the needed info is already in `features_by_type`.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -886,7 +886,7 @@ class Feature(DictModel):
         ramcache.set(KEY, feature)
 
     return feature
-  
+
   @classmethod
   def filter_unlisted(self, feature_list):
     """Filters a feature list to display only features the user should see."""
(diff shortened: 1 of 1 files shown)


jrobbins:
No worries.   Rather than try to review things to the Nth degree and require complete unit tests for every case, I assume that we will spend some time fixing problems before each release.  Given that that has been the philosophy for the existing code too, when you make changes you need both write some unit tests and also run the app locally and try a few cases that you can think of.

PR #2028, opened 2022-07-15
In notifications, display full email addr rather than nickname.

jrobbins:
This works around a problem identified by Joe in an email thread a few days ago.

He found that some email notifications only give the username of the user who created or edited a feature rather than their full email address.  That is a quirk of GAE user objects that display a nickname, which is just the username for @gmail.com accounts.

To get the full email address all the time, we need to call email(), however that cannot be done in a django template.  Rathe than be stuck on that, I called it in python code before rendering the template.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -63,8 +63,8 @@ def format_email_body(is_update, feature, changes):
 
   body_data = {
       'feature': feature,
-      'created_by_email': feature.created_by.email(),
-      'updated_by_email': feature.updated_by.email(),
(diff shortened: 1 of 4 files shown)


jpmedley:
If I'd known it was that easy, I'd have asked years ago.

dlaliberte (approved):
The names are a bit confusing.  '*_by_email" sounds like an email was exchanged that resulted in creating or updating something.  Instead, the email() function makes the email address data into a fully  qualified address.   Not sure what would be better, however.

Issue #2020, opened 2022-07-13
In feature change notifications, set reply-to to the user who made the change


jrobbins:
**Describe the solution you'd like**
In feature change notifications, set reply-to to the user who made the change.
This will make it easier for interested parties to follow up with the person who made the change.

Only do this for users who could can comment, because other people on the internet who star a feature might not have the same expectations about when it is appropriate to contact the person who made the edit.  They can certainly still do that since that address is in the body of the email.  We just don't want them sending "unsubscribe", vacation, or other off-topic replies to the person who made the change.

jpmedley:
Can the reply show the replier's show the repliers access level. Maybe do something like: "Site editor jmedley@google.com replied to your update."

I'm not in love with that language or what it implies about how this would work. I'm trying to convey that the response should give the feature owner some context about the reply. If I request a change and the recipient of my request knows I'm a site editor, they automatically know that I'm empowered to make requests like that.

jrobbins:
Well, the reply is just a regular email message that you compose yourself.  You could identify yourself as a "site editor" in the body of the message.  There is no email header that the app can set that would control the attribution in a reply that you write.

jrobbins:
This is live now.

PR #2031, opened 2022-07-15
Add Reply-To header to email notifications.

jrobbins:
This should resolve issue #2020.  Basically, he would find it easier to follow up with feature owners who made a change if he could just use the Reply button in gmail to contact the user who made a change.

In this PR:
* In sendemail.py: set the Reply-To header in the outbound email message based on what is set in the task.
* In notifier.py: If the email is being sent to a user who is trusted to reply, then use the address of the user who made the change (and therefore triggered the notification) as the reply_to value in the task.
* Update unit tests

### DIFF
--- a/internals/notifier_test.py
+++ b/internals/notifier_test.py
@@ -162,7 +162,7 @@ class EmailFormattingTest(testing_config.CustomTestCase):
     self.assertIn('reason 2', actual['html'])
 
   def test_convert_reasons_to_task__can_reply(self):
-    """Is the user is allowed to reply, set reply_to to the triggerer."""
+    """If the user is allowed to reply, set reply_to to the triggerer."""
     actual = notifier.convert_reasons_to_task(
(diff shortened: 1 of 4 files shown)


internals/notifier_test.py
```diff
@@ -144,20 +144,35 @@ def test_accumulate_reasons(self):
...
         list(actual.keys()))
     self.assertEqual('addr', actual['to'])
     self.assertEqual('subject', actual['subject'])
+    self.assertEqual(None, actual['reply_to'])  # Lacks perm to reply.
     self.assertIn('html', actual['html'])
     self.assertIn('reason 1', actual['html'])
     self.assertIn('reason 2', actual['html'])
 
+  def test_convert_reasons_to_task__can_reply(self):
+    """Is the user is allowed to reply, set reply_to to the triggerer."""
```
DanielRyanSmith, line 165:
```suggestion
    """If the user is allowed to reply, set reply_to to the triggerer."""
```

DanielRyanSmith (approved):

PR #2033, opened 2022-07-15
Add index that datastore says it needs for can_edit terms.

jrobbins:
When running on staging, there is an error with the request for features in the "Features I can edit" box.

Logs shown an exception with the message:
```
no matching index found. recommended index is:
- kind: Feature
  properties:
  - name: editors
  - name: updated
    direction: desc
```

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -204,8 +204,9 @@ indexes:
   - name: "name"
 - kind: Feature
   properties:
-  - name: owner
   - name: editors
+  - name: updated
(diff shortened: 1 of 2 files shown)


jrobbins:
I am still testing this.  Staging says "index not ready to serve", so far.

jrobbins:
OK it is serving now and that box on the my features page works.

DanielRyanSmith (approved):

PR #2034, opened 2022-07-15
Create new creator field for features

DanielRyanSmith:
This PR adds functionality to reference a new field `creator` on the Feature kind in DataStore. Additionally, a handler that writes the field to all features with the string value of the UserProperty `created_by` is added. This `created_by` field comes with a lot of unnecessary baggage that prevents its reliable use. The new `creator` field allows referencing the email of the original creator of the feature so that this user has permanent edit access to that feature, even if they are removed from the owners and editors list.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -787,9 +787,8 @@ class Feature(DictModel):
           # This includes being an owner, editor, or the original creator
           # of the feature.
           query = query.filter(
-            ndb.OR(Feature.owner == comparator,
-                   ndb.OR(Feature.editors == comparator,
-                          Feature.creator == comparator)))
+            ndb.OR(Feature.owner == comparator, Feature.editors == comparator,
+              Feature.creator == comparator))
         else:
           query = query.filter(getattr(Feature, filter_type) == comparator)
 
--- a/internals/models.py
+++ b/internals/models.py
@@ -634,6 +634,7 @@ class Feature(DictModel):
       }
       d['tags'] = d.pop('search_tags', [])
       d['editors'] = d.pop('editors', [])
+      d['creator'] = d.pop('creator', None)
       d['browsers'] = {
         'chrome': {
           'bug': d.pop('bug_url', None),
@@ -786,7 +787,9 @@ class Feature(DictModel):
           # This includes being an owner, editor, or the original creator
           # of the feature.
           query = query.filter(
-            ndb.OR(Feature.owner == comparator, Feature.editors == comparator))
+            ndb.OR(Feature.owner == comparator,
+                   ndb.OR(Feature.editors == comparator,
+                          Feature.creator == comparator)))
         else:
           query = query.filter(getattr(Feature, filter_type) == comparator)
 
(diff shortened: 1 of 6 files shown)


internals/models.py
```diff
@@ -786,7 +787,9 @@ def get_all(self, limit=None, order='-updated', filterby=None,
           # This includes being an owner, editor, or the original creator
           # of the feature.
           query = query.filter(
-            ndb.OR(Feature.owner == comparator, Feature.editors == comparator))
+            ndb.OR(Feature.owner == comparator,
```
jrobbins, line 790: A single `ndb.OR()` can have any number of arguments, so you don't need to nest.

internals/write_creator.py
```diff
@@ -0,0 +1,15 @@
+from framework.basehandlers import FlaskHandler
+from internals.models import Feature
+
+
+class UpdateCreatorHandler(FlaskHandler):
+  def get_template_data():
+    """Writes string creator field from created_by user field."""
+    q = Feature.query()
+    features = q.fetch()
+    for feature in features:
```
jrobbins, line 10: This would be more robust if could make incremental progress when run repeatedly.   I don't think that we have enough features in datastore to make this timeout, but just in case, let's code it to handle that.

I think writing it this way would allow us to run repeatedly with the code quickly skipping over Features that have already been updated:
```
for feature in features:
  if feature.created_by and not feature.creator:
    feature.creator = feature.created_by.email()
    feature.put()
```
Note that I am calling feature.created_by.email() because str(feature.created_by) ends up calling feature.created_by.nickname() which leaves off "@gmail.com".

internals/models.py
```diff
@@ -786,7 +787,9 @@ def get_all(self, limit=None, order='-updated', filterby=None,
...
+                   ndb.OR(Feature.editors == comparator,
```
jrobbins, line 791: FYI, this might require an update to index.yaml.  ISTR, the docs say that we need an index that exactly matches the set of fields used in each query (including sorting).  However, I have a feeling that an index is needed for each OR clause rather than for the overall query.  

Single field queries use an implicitly created single-field index.  So, if it is by OR clause, we might not need any explicit index.  I am thinking that one of these OR clauses gets combined with the ordering, so that is why we needed to add that one index last time.

  Let's test it on staging and see if we get a helpful error message.

internals/write_creator.py
```diff
@@ -0,0 +1,15 @@
+from framework.basehandlers import FlaskHandler
+from internals.models import Feature
+
+
+class UpdateCreatorHandler(FlaskHandler):
+  def get_template_data():
```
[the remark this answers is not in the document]
jrobbins: This seems fine, it looks like our other cron task handlers.

Style nit: add a blank line after the class definition line.

jrobbins (changes requested):

jrobbins:
I just realized that the `put()` call is going to generate thousands of low-value notification emails.  We should implement a non-notifying mode.

internals/write_creator.py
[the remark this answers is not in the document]
DanielRyanSmith: Changed!

internals/models.py
DanielRyanSmith: Changed!

internals/models.py
DanielRyanSmith: Okay sounds good. I'm not familiar with finding correct indexes, so I'd be interested in knowing your process for index troubleshooting.

jrobbins (approved):

PR #2059, opened 2022-07-25
Set up a partial SPA

kevinshen56714:
This PR sets up a partial single-page app with currently available page components (all except for /guide and /admin pages). Changes include:

1. Installed Page.js for routing.
2. Added `chromedash-app.js` component as the app-level component in which the routing rules are defined.
3. Added `spa.html` as the SPA's entry html file. This file currently highly resembles `_base.html` but I think we can improve it by moving some of the js script into the app component js file in the future.
4. Added a /spa route in `main.py`. This is the temporary entry route to the SPA. The SPA router will redirect the app to /roadmap and take over the routing.

Overall, the SPA version enables a much smoother transitions (no reloads) between pages as excepted :). See the demo video below:

https://user-images.githubusercontent.com/11501902/180865138-751115fc-f4bc-434d-b5ba-903c3dc7d12b.mov

### DIFF
--- a/main.py
+++ b/main.py
@@ -13,9 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from flask import Flask
-import logging
-
 from api import accounts_api
 from api import approvals_api
 from api import channels_api
@@ -183,6 +180,8 @@ page_routes = [
 
     ('/settings', users.SettingsHandler),
     ('/admin/users/new', users.UserListHandler),
+
+    ('/spa', basehandlers.ConstHandler, {'template_path': 'spa.html'}),
 ]
 
 
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -31,8 +31,8 @@ class ChromedashApp extends LitElement {
     this.appTitle = '';
     this.googleSignInClientId = '',
     this.currentPage = '';
-    this.message = '';
-    this.timestamp = null;
+    this.bannerMessage = '';
+    this.bannerTime = null;
     this.pageComponent = null;
     this.contextLink = '/features';
   }
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -40,7 +40,6 @@ class ChromedashApp extends LitElement {
     page('/spa', () => page.redirect('/roadmap'));
     page('/roadmap', () => {
       this.pageComponent = document.createElement('chromedash-roadmap-page');
-      console.log(this.pageComponent);
     });
     page('/myfeatures', () => {
       this.pageComponent = document.createElement('chromedash-myfeatures-page');
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -34,20 +34,24 @@ class ChromedashApp extends LitElement {
 
   setUpRoutes() {
     page.strict(true); // Be precise about trailing slashes in routes.
+    const header = document.querySelector('chromedash-header');
 
     // SPA routing rules.  Note that rules are considered in order.
     // And :var can match any string (including a slash) if there is no slash after it.
     page('/spa', () => page.redirect('/roadmap'));
     page('/roadmap', () => {
       this.pageComponent = document.createElement('chromedash-roadmap-page');
+      header.currentPage = '/roadmap';
     });
     page('/myfeatures', () => {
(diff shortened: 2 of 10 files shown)


static/elements/chromedash-app.js
```diff
@@ -0,0 +1,117 @@
...
+      this.pageComponent = document.createElement('chromedash-settings-page');
+      header.currentPage = ctx.path;
+    });
+    page('/metrics/:type/:view', (ctx) => {
+      this.pageComponent = document.createElement('chromedash-stack-rank-page');
+      this.pageComponent.type = ctx.params.type;
+      this.pageComponent.view = ctx.params.view;
+      header.currentPage = ctx.path;
+    });
+    page('/metrics/:type/timeline/:view/:bucketId', (ctx) => {
```
jrobbins, line 71: Given the comment about :var matching a string that includes slashes, I am surprised that this rule does not need to come before the other metrics routing rule.

static/elements/chromedash-app.js
```diff
@@ -0,0 +1,117 @@
...
+      this.pageComponent = document.createElement('chromedash-timeline-page');
+      this.pageComponent.type = ctx.params.type;
+      this.pageComponent.view = ctx.params.view;
+      this.pageComponent.selectedBucketId = ctx.params.bucketId;
+      header.currentPage = ctx.path;
```
jrobbins, line 76: What are the trade-offs between setting `header.currentPage` vs setting this.currentPage and using a Lit reactive property like `<chromedash-header .currentPage=${this.currentPage}>`?

static/elements/chromedash-app.js
```diff
@@ -0,0 +1,117 @@
...
+
+  render() {
+    // TODO: Create precomiled main css file and import it instead of inlining it here
+    return html`
+      <link rel="stylesheet" href="/static/css/main.css">
+      <div id="app-content-container">
+        <div>
+          <div class="main-toolbar">
+            <div class="toolbar-content">
+              <slot name="header"></slot>
```
jrobbins, line 90: Why use slots instead of putting `<chromedash-header ...>` here?

Slots can keep a component reusable, but this is our top-level app, it is not going to be reused in another context.

main.py
```diff
@@ -183,6 +180,8 @@
 
     ('/settings', users.SettingsHandler),
     ('/admin/users/new', users.UserListHandler),
+
+    ('/spa', basehandlers.ConstHandler, {'template_path': 'spa.html'}),
```
jrobbins, line 184: It's great that this does not even need a handler.

jrobbins (approved):
LGTM after considering some questions.
I'm really excited about trying this out.

static/elements/chromedash-app.js
kevinshen56714: It's because of the Google sign-in button. The current `<chromedash-header>` component insert it to its light DOM to allow the Google Auth library to find it. The `<chromedash-header>`'s light DOM is at document level in the MPA version, which works fine.

But in the SPA version, if `<chromedash-header>` is within `<chromedash-app>`,  `<chromedash-header>`'s light DOM would be in the `<chromedash-app>`'s shadow DOM, which blocks the Auth library's access. That's why I used the slot to make sure the header component is at the document level.

I'd love to not use the slot and there are some things I want to try. But I couldn't make any meaningful changes to the `<chromedas-header>` now because it's still used for the MPA.

static/elements/chromedash-app.js
kevinshen56714: I'd love to use `<chromedash-header .currentPage=${this.currentPage}>`, but because of the Google sign-in button, I couldn't have header component inside the app component. See more explanation below.

static/elements/chromedash-app.js
kevinshen56714: Yes, it's actually working fine.

static/elements/chromedash-app.js
kevinshen56714: This is actually resolved after my changes.

static/elements/chromedash-app.js
kevinshen56714: This is actually resolved after my changes.

kevinshen56714:
@jrobbins the slots for header and banner are removed. So `chromedash-header` is now within `chromedash-app` with currentPage as its reactive property. Please review my changes again. Thanks!

jrobbins (approved):

PR #2066, opened 2022-07-27
Replace django checkbox with chromedash-form-field generated checkbox field.

dlaliberte:
This PR replaces the checkbox field generated by django with one generated by chromedash-form-field.js.

The result is the same except for the missing value and checked attributes.  It's not yet clear how to get the value and checked attributes.


jrobbins:
Should this PR still be a draft?

dlaliberte:
I still need to check that this works on the other pages, and I suspect I will have to make a couple more changes, but this much is good.

dlaliberte:
The "set_stage" field on the stage.html page looks like it will be a problem.  

1. There is no "set_stage" form field declared or used in guideforms.py, and so it is also not yet declared in form-field-specs.js.
2. The "set_stage" field is accessed on the guide.py handler when the form is submitted.
3. The help text for this field is conditional on whether the current stage is "already_on_this_stage".  We don't have a way of testing the current stage yet in the chromedash-form-field component. 
4. The stage_name of the current stage is used within the help text and in the field label.

If we can keep the current way the form field and help text is generated, we still have to escape these values and pass them through attributes.  But this server-side logic will need to move to the client-side if we are going to complete the migration to the client-side.

static/elements/chromedash-form-field.js
```diff
@@ -1,14 +1,22 @@
...
+      field: {type: String},
+      errors: {type: String},
     };
   }
 
-  static get styles() {
+  static get disabledStyles() {
+    // Since we are using light DOM, for now, these styles are
+    // not used.  See styles in forms.scss instead.
+    // TODO: Check whether we can still use these instead.
```
jrobbins, line 18: I'm thinking that these styles could be put into chromedash-form-table.js.  But, that could be in a future PR, if it works.

jrobbins (approved):

static/elements/chromedash-form-field.js
dlaliberte: I'll try that, but while the docs at https://lit.dev/docs/components/styles/#slotted do say that styles added to the component can affect the component's children, since we are rendering into the light DOM, I'm not sure what happens.

dlaliberte:
For the "set_stage" checkbox, I created a spec for it in form-field-specs.js.  Here is what it looks like now:

image

Note that the checkbox is disabled and checked if the current stage is the same as the stage being displayed (as it is in this screenshot).  Otherwise, the checkbox will be enabled and unchecked.  I changed the label to be a bit more generalized, and the help text is the same either way.  

I also think this checkbox might be better as the first item in the table rather than the last.  That way, it will be close to the label that identifies the stage.

jrobbins:
> I also think this checkbox might be better as the first item in the table rather than the last. That way, it will be close 
> to the label that identifies the stage.

I'd prefer to keep it at the bottom where it is now.  I might be eliminating these checkboxes after we have the new schema done and support for a variable number of stages.   I don't want to give users too many changes to the same element, e.g., I don't want them thinking that it got moved again and that they need to find it, if/when it is actually eliminated.

dlaliberte:
For the feature_type on the new.html page, I was able to escape everything that goes in the field attribute, so that works correctly, though it's still a workaround.

PR #2068, opened 2022-07-27
Feature owners and editors receive notification of changes.

DanielRyanSmith:
This is a change listed in the new permissions design where feature editors will receive email notifications of feature changes. Additionally, feature owners have been added to the list of users who receive change notifications.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -152,6 +152,14 @@ def make_email_tasks(feature, is_update=False, changes=[]):
   addr_reasons = collections.defaultdict(list)  # [{email_addr: [reason,...]}]
 
   accumulate_reasons(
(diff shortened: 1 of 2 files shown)


jrobbins (approved):
I'm glad this came out so simple.  I think of notify.py as being complicated, but adding a new notification reason is actually pretty clear when we already have a list of email addresses.

We should consider adding a user preference notify_as_editor that users could uncheck if they don't want notifications for that reason.  That can be a separate PR.

DanielRyanSmith:
> We should consider adding a user preference notify_as_editor that users could uncheck if they don't want notifications for that reason. That can be a separate PR.

This is good - I'll create an issue for this.

PR #2069, opened 2022-07-27
Temporary editing permissions

DanielRyanSmith:
Before the new permissions system is implemented, users who make changes to a feature will be added to the list of feature editors if they are not already an editor, owner, or the creator of the feature. This change should be removed when the new permissions system is officially implemented.

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -569,12 +569,12 @@ class FeatureEditStage(basehandlers.FlaskHandler):
     if self.touched('ongoing_constraints'):
       feature.ongoing_constraints = self.form.get('ongoing_constraints')
     
-    # Add user who updated to list of editors.
+    # Add user who updated to list of editors if not currently an editor.
     # TODO(danielrsmith): This should be removed when enabling new permissions.
-    email = self.get_current_user().email()
-    if (email not in feature.editors and email not in feature.owner and
-        email != feature.creator):
-      feature.editors.append(email)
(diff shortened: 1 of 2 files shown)


pages/guide.py
```diff
@@ -568,6 +568,13 @@ def process_post_data(self, feature_id, stage_id=0):
           'experiment_extension_reason')
     if self.touched('ongoing_constraints'):
       feature.ongoing_constraints = self.form.get('ongoing_constraints')
+    
+    # Add user who updated to list of editors.
+    # TODO(danielrsmith): This should be removed when enabling new permissions.
+    email = self.get_current_user().email()
+    if (email not in feature.editors and email not in feature.owner and
```
jrobbins, line 575: Let's also check that the user is not already a site editor.  Ideally the whole condition would be `if not permissions.can_edit_feature(user, feature.id)`, but that would be calling the version of `can_edit_feature()` that is still in your pending PR #2017 that needs to wait a while because we don't want the new permission rules to be enforced yet.

Maybe you could copy over the `can_edit_feature()` method from your pending PR, so long as you temporarily give it a new name, like `strict_can_edit_feature()`.

jrobbins (approved):
LGTM after you address one more requirement.

pages/guide.py
DanielRyanSmith: Yes, that's a good idea. I've changed it so that admins and site editors will not get added as feature editors if they make changes to a specific feature 👍

PR #2071, opened 2022-08-01
Fix SPA roadmap page and toast bugs

kevinshen56714:
This PR fixes several bugs in the SPA that I found so far, including:

1. Going to the feature page from the roadmap page results in wrong page width. This is fixed by only enabling a full width app when the current page component is specifically `<chromedash-roadmap-page>`)
2. The back button on the feature page would not link to the roadmap page when entering the feature page from roadmap. This is fixed by updating the contextLink for the /roadmap route as well.
3. The toast does not work in SPA because it's in the shadowRoot of the app component, and the document.querySelector cannot find it. This is fixed by putting `<chromedash-toast>` directly in the spa.html. In the future, we can implement the strategy we used for approval dialogs on toast so that it doesn't have to appear in the html.
4. Moving `chromedash-app`-specific css from main.scss to the app component. However these cannot be removed from main.scss yet because it's still used in the MPA version. I think in a later PR, I will merge `main.scss` and `shared.scss`, and we can retire `main.scss`.
5. Removed more unused css (for #site-banner and #container) in main.scss.

### DIFF
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -8,8 +8,41 @@ class ChromedashApp extends LitElement {
     return [
       ...SHARED_STYLES,
       css`
+        .main-toolbar {
+          display: flex;
+          position: relative;
+          padding: 0;
+        }
+
+        #app-content-container {
+          display: flex;
+          flex-direction: column;
+          height: 100%;
+        }
+
+        #content {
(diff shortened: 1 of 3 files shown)


jrobbins (approved):
Nice.  We were missing the roadmap page as a navigational context.  And, its always great to see cruft removed.

PR #2078, opened 2022-08-01
Send reminders to verify accuracy of data before important milestones

DanielRyanSmith:
**NOTE**: This change includes only scheduled reminder emails and is coded to not function until front end changes are implemented, including a new feature view linked in emails for owners to quickly verify data accuracy.


templates/accuracy_notice_email.html
```diff
@@ -0,0 +1,31 @@
...
+    enterprise IT admins who might be affected by web platform changes.
+</p>
+</br>
+<p>
+    [Call to action] We need to know whether your plans are changing or staying
+    the same. <strong>Please click the link below to update and confirm key
+    fields of your feature entry.</strong>
+</p>
+</br>
+<b><a href="https://chromestatus.com/guide/verify_accuracy/{{id}}">
```
DanielRyanSmith, line 29: @jrobbins Does this seem like a good choice for the page to verify feature data accuracy?

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+            'to': owner,
+            'subject': subject,
+            'reply_to': None,
+            'html': html
+          })
+    return email_tasks
+
+  def _send_notifications(self, email_tasks):
+    for email_task in email_tasks:
+      if settings.prod and settings.SEND_EMAIL:
```
DanielRyanSmith, line 347: @jrobbins Do we want to send these emails in the staging environment as well?

templates/accuracy_notice_email.html
jrobbins: Sure, it says what it does.  I like it.

internals/notifier.py
jrobbins: Yes, so just check settings.SEND_EMAIL.  That setting is already True in staging.   Staging does send emails to help us test that email sending works.   However, there is also settings.SEND_ALL_EMAIL_TO which causes staging to send the emails that it generates to the cr-status-staging-emails mailing list.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+    # If reminding 8 weeks in advance, we should only need to notify
+    # for current and (current + 1) milestones.
+    try:
+      resp = requests.get(f'{self.CHROME_RELEASE_SCHEDULE_URL}?mstone=current')
+    except requests.RequestException as e:
+      raise e
+    mstone_info = json.loads(resp.text)
+    mstone = int(mstone_info['mstones'][0]['mstone'])
+    
+    features = models.Feature.get_all()
```
jrobbins, line 298: `get_all()` returns the features as dictionaries with strings for dates.  It would probably be cleaner to use `models.Feature.query(models.Feature.deleted == False).fetch(None)` to get all the feature objects.  That way you can do date and integer comparisons without dealing with strings.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+        accuracy_as_of_date = datetime.strptime(
+          feature['accurate_as_of'], '%y-%m-%d %H:%M:%S.%f')
+      # If the data has been recently verified as accurate, no need for email.
+      if (accuracy_as_of_date is not None and
+          accuracy_as_of_date + accuracy_as_of_delta < now):
+        continue
+      # Check each milestone field and see if it corresponds with
+      # the next 2 milestones.
+      for mstone_field in self.MILESTONE_FIELDS:
+        if mstone_field in feature and feature[mstone_field] == mstone:
```
jrobbins, line 315: You can get attributes of the model instance by doing `feature.getattr(mstone_field)`.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+          feature['notify_mstone'] = str(mstone)  # milestone for email.
```
jrobbins, line 316: Instead of poking temporary info into the feature dict, it could be better to keep one output list of only those features that are being notified because of mstone, and a separate list of features being notified because of mstone + 1.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+          features_to_notify.append(feature)
+          break
+        if mstone_field in feature and feature[mstone_field] == mstone + 1:
+          feature['notify_mstone'] = str(mstone + 1)  # milestone for email.
+          features_to_notify.append(feature)
```
jrobbins, line 321: I think this would allow a case where a feature is added to the list twice (or if you used two lists, added to both).  We should probably check if the feature is already in the list.

Also, the line above would cause such a feature to be notified for mstone +1, however it is probably better for the email to mention mstone since that is coming up sooner.  This problem is avoided if you use two lists and avoid appending if the feature is already in the first list.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+            'to': owner,
+            'subject': subject,
+            'reply_to': None,
+            'html': html
+          })
+    return email_tasks
+
+  def _send_notifications(self, email_tasks):
+    for email_task in email_tasks:
+      if settings.SEND_EMAIL:
```
jrobbins, line 347: This if-statement looks similar to part of FeatureChangeHandler and post_comment_to_mailing_list.   It could be refactored into a shared top-level function.

templates/accuracy_notice_email.html
```diff
@@ -0,0 +1,31 @@
+<p>
+    [Why me?] You are receiving this notification because you are listed as an owner of the following ChromeStatus feature entry:
```
jrobbins, line 2: Sorry, I should have made it more clear in the design docs that the bracketed text was meant just to organize my thoughts about what goes into this message.   Let's leave off the phrases `[Why me?`], `[Why now?]`, `[Background/motivation]`, and `[Call to action]`.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+      for mstone_field in self.MILESTONE_FIELDS:
+        if mstone_field in feature and feature[mstone_field] == mstone:
+          feature['notify_mstone'] = str(mstone)  # milestone for email.
+          features_to_notify.append(feature)
+          break
+        if mstone_field in feature and feature[mstone_field] == mstone + 1:
+          feature['notify_mstone'] = str(mstone + 1)  # milestone for email.
+          features_to_notify.append(feature)
+          break
+    return features_to_notify
```
jrobbins, line 323: If you go with the two-lists approach, you could return:
`{mstone: features_for_current_mstone,
  mstone + 1: features_for_next_mstone}`.
That way you get the mstone and mstone + 1 values returned so that they can be passed into the other method.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+
+  def _build_email_tasks(self, features_to_notify):
+    email_tasks = []
+    for feature in features_to_notify:
+      body_data = {
+        'feature_name': feature['name'],
```
jrobbins, line 329: Rather than a new dictionary with exactly the values needed, let's pass the whole feature dictionary.  If you have a list of feature objects rather than dictionaries, you can construct the dictionary at this point using `format_for_template()`.  So, this line would be:
`'feature': feature.format_for_template(),`

templates/accuracy_notice_email.html
```diff
@@ -0,0 +1,31 @@
...
+<p>
+    <strong>{{feature_name}}</strong>
+</p> 
+</br>
+<p>
+    [Why now?] Your feature is slated to launch soon:
+</p>
+</br>
+<p>
+    <strong>m{{milestone}}</strong>
```
jrobbins, line 14: This was unclear in the design doc, but I was thinking that we should include all the feature milestones, like we do in an issue change notification.  So, it would be:
`{% include "estimated-milestones-table.html" %}`
I suppose the mstone that triggered this notification should me mentioned in a new paragraph or somewhere else in the email body.

templates/accuracy_notice_email.html
jrobbins, line 29: Rather than use `https://chromestatus.com/` here, let's use settings.SITE_URL.  That way the staging emails will have links that work.

Also, I notice that in settings.py SITE_URL uses `http`, we should update that to `https`.

jrobbins (changes requested):
Sorry for so many comments.  Some are because the design doc had gaps.  Others are coding suggestions, with the main one being to try to use Feature model objects more for server-side logic and use the feature dictionaries only for rendering templates.

jpmedley:
Guys,

Thank for this PR. This is a huge improvement to the system. I'm going to spend some time with this PR to understand it. I have one big comment. There are relationships between implementation status and milestones that should trigger a reminder. Specifically:
* Dev Trial, but no dev trial milestones
* Origin Trial, but no origin trial milestones
* Enabled by default, but no shipping milestones. 
* Removal but no shipping milestones.

I just spent an hour updating statuses with these states for 103 and 104. 

In the future we'll also need to include Deprecation, but no deprecation milestones.

jpmedley:
Is there a way for you to bcc me the first few times it runs. I can check for problems that way.

templates/accuracy_notice_email.html
```diff
@@ -0,0 +1,31 @@
+<p>
+    [Why me?] You are receiving this notification because you are listed as an owner of the following ChromeStatus feature entry:
+</p>
+</br>
+<p>
+    <strong>{{feature_name}}</strong>
+</p> 
+</br>
+<p>
+    [Why now?] Your feature is slated to launch soon:
```
jpmedley, line 10: Going with Jason's comment in 14, this should be something like:

>Your feature is listed for [stage] in [milestone]

For example

>Your feature is listed for deprecation in 105.

templates/accuracy_notice_email.html
```diff
@@ -0,0 +1,31 @@
...
+<p>
+    [Background/motivation] Your feature entry is an important resource for
+    cross functional teams that help drive adoption of new features and
+    enterprise IT admins who might be affected by web platform changes.
+</p>
+</br>
+<p>
+    [Call to action] We need to know whether your plans are changing or staying
+    the same. <strong>Please click the link below to update and confirm key
+    fields of your feature entry.</strong>
```
jpmedley, line 26:
```suggestion
    the same. <strong>Please click the link below to update and confirm the
    milestone fields and implementation status of your feature entry.</strong>
```

DanielRyanSmith: Added this change!

templates/accuracy_notice_email.html
DanielRyanSmith: Done!

templates/accuracy_notice_email.html
DanielRyanSmith: I understand. I've added this to add the table to the email.

internals/notifier.py
DanielRyanSmith: Makes sense - done!

internals/notifier.py
DanielRyanSmith: Since we care about beta and stable dates approaching, we actually need to look at 3 milestones. I changed it to use a single list and instead append tuples, `(feature, milestone)`. Let me know if you see anything incorrect about it!

templates/accuracy_notice_email.html
DanielRyanSmith: I see now! removed those.

internals/notifier.py
DanielRyanSmith: Refactored to have these code blocks use the same function. Great suggestion!

internals/notifier.py
DanielRyanSmith: I've made a change to check all the possible starts to find the nearest milestone that would trigger a notification. Maybe you can take a look and see if it's an effective change?

internals/notifier.py
DanielRyanSmith: I changed this to instead append a tuple with the milestone associated with the notification - hopefully this will suffice.

internals/notifier.py
DanielRyanSmith: Yes, I've added this to return feature objects rather than using the dictionary for templating.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+    'ot_milestone_desktop_start',
+    'ot_milestone_webview_start'
+  )
+  TEMPLATE_PATH = 'accuracy_notice_email.html'
+   
+  def get_template_data(self):
+    """Sends notifications to users requesting feature updates for accuracy."""
+
+    # Do not send notifications
+    # TODO(danielrsmith): This should be removed when the new page to
```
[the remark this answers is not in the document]
DanielRyanSmith: Updated the wording.

internals/notifier.py
```diff
@@ -254,6 +254,113 @@ def get_feature_starrers(self, feature_id):
...
+'https://chromiumdash.appspot.com/fetch_milestone_schedule')
+  ACCURACY_AS_OF_WEEKS = 4
+  MILESTONE_FIELDS = (
+    'dt_milestone_android_start',
+    'dt_milestone_desktop_start',
+    'dt_milestone_ios_start',
+    'dt_milestone_webview_start',
+    'ot_milestone_android_start',
+    'ot_milestone_desktop_start',
+    'ot_milestone_webview_start'
```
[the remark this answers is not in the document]
DanielRyanSmith: Yes, I missed these initially when I was looking at staging datastore entries. I've added them now!

internals/models.py
```diff
@@ -1219,7 +1220,8 @@ def __notify_feature_subscribers_of_changes(self, is_update):
     # Diff values to see what properties have changed.
     changed_props = []
     for prop_name, prop in list(self._properties.items()):
-      if prop_name in ('created_by', 'updated_by', 'updated', 'created'):
+      if prop_name in (
+          'created_by', 'updated_by', 'updated','created', 'accurate_as_of'):
```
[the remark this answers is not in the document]
DanielRyanSmith: Got it - added it back 👍

internals/notifier.py
```diff
@@ -254,6 +254,104 @@ def get_feature_starrers(self, feature_id):
...
+    
+    features = models.Feature.query(models.Feature.deleted == False).fetch(None)
+    features_to_notify = []
+    now = datetime.now()
+    accuracy_as_of_delta = timedelta(weeks=self.ACCURACY_AS_OF_WEEKS)
+    for feature in features:
+      accuracy_as_of_date = None
+      # Format accuracy date to compare dates.
+      if feature.accurate_as_of is not None:
+        accuracy_as_of_date = datetime.strptime(
```
jrobbins, line 313: Why do we need to convert to a string and then parse the data back into date time object?  Could we use feature.accurate_as_of directly in the if-statement below?

internals/notifier.py
jrobbins: OK, that works.

jrobbins (approved):

internals/notifier.py
DanielRyanSmith: This was an old code block when I wasn't certain if the `ndb.DateTimeProperty` could be used as a true datetime object. I forgot to refactor it - thanks for pointing it out!

templates/accuracy_notice_email.html
DanielRyanSmith: This is a good addition. Currently, we identify which milestone, but not the stage. The work involved in this might change with the schema changes planned, but I think this could be an upgrade when that time comes.

PR #2079, opened 2022-08-01
Remove roadmap download links.

jrobbins:
This removes the download links on the headers in the roadmap milestone cards.   A user reported as a bug because most of them just link to "#".   The ones that actually go somewhere go to a download page that is not useful enough to our users to justify having the link.


kevinshen56714 (approved):
LGTM!

PR #2082, opened 2022-08-02
Sort pending reviews by review request date via sort= query string param

jrobbins:
A bit of feedback that I have gotten from the API Owners is that the list of features pending review on the my-features page is not sorted in a meaningful way.    It was originally hard-coded to be sorted by the query.order() call done in Approvals.get_approvals(), but that was made moot when I implemented a general default sorting order for all feature search requests.  The recently reviewed features box was likewise hard-coded and also stopped sorting correctly.

Rather than keep hard-coding sort orders for specific queries, I implemented the sort= query string parameter for the feature search API, and specified sort= for those my-feature boxes.

In this PR:
* features_api.py: Get the sort= parameter and pass it to the code in search.py.
* search.py: Don't hard-code sorting for certain queries, instead pass the sort_spec to models.py.
* models.py: Moved big chunks of code around so that references could be in the right order.   Now there is both a QUERIABLE_FIELDS dictionary and also a SORTABLE_FIELDS dictionary that is copied from it with two additional sort options handled by two  new functions.  Implement those two functions to produce total orderings used for sorting.
* chromedash-myfeatures-page.js specify the sorting for those two boxes.
* chromedash-feature-table.js and cs-client.js: pass that sort spec to the server.
* utils.js: add an idiomatic python way to deduplicate a list and call it rather than previous implementations in a couple of different places in the code.


jrobbins:
Sorry that this came out to be a larger PR than I usually like to send.  Hopefully the bullets in the PR description make it manageable.

internals/models_test.py
```diff
@@ -449,22 +552,22 @@ def test_get_approvals(self):
...
+    # actual = models.Approval.get_approvals(
+    #     order=-models.Approval.set_on, limit=1)
+    # self.assertEqual(1, len(actual))
+    # self.assertEqual(models.Approval.APPROVED, actual[0].state)
+
+    # actual = models.Approval.get_approvals(
+    #     order=-models.Approval.set_on, states=[models.Approval.APPROVED],
+    #     limit=10)
+    # self.assertEqual(1, len(actual))
+    # self.assertEqual(models.Approval.APPROVED, actual[0].state)
```
DanielRyanSmith, line 570: Is this worth keeping rather than refactoring or deleting?

static/elements/chromedash-myfeatures-page.js
```diff
@@ -71,14 +71,15 @@ export class ChromedashMyFeaturesPage extends LitElement {
     openApprovalsDialog(this.user.email, e.detail.feature);
   }
 
-  renderBox(title, query, columns, opened=true) {
+  renderBox(title, query, columns, sortSpec='', opened=true) {
```
DanielRyanSmith, line 74: Just to confirm, this functionality only works for the approval/review tables? Is there any reason why this shouldn't also work with the "Features I can edit" or "Features I starred" tables?

static/js-src/cs-client.js
```diff
@@ -256,8 +256,12 @@ class ChromeStatusClient {
       (resp) => resp['features_by_type']);
   }
 
-  searchFeatures(userQuery) {
-    return this.doGet(`/features?q=${userQuery}`);
+  searchFeatures(userQuery, sortSpec) {
+    let url = `/features?q=${userQuery}`;
+    if (sortSpec) {
+      url += '&sort=' + sortSpec;
+    }
```
DanielRyanSmith, line 263: This functionality has no visibility outside of the query string, right? Maybe it would be useful to define a default sort that can be seen on the query string, if only for a little additional visibility.

framework/utils.py
```diff
@@ -148,3 +148,8 @@ def get_banner_time(timestamp):
     return None
   ts = datetime.datetime(*timestamp, tzinfo=_UTC)
   return calendar.timegm(ts.timetuple())
+
+
+def dedupe(list_with_duplicates):
+  """Return a list without duplicates, in the original order."""
+  return list(dict.fromkeys(list_with_duplicates))
```
DanielRyanSmith, line 155: I started to suggest using a set for this, but then saw the filtering needs to be stable. Sets technically aren't guaranteed to be stable, and [dictionaries are as of 3.7](https://www.w3schools.com/python/python_dictionaries.asp#:~:text=Ordered%20or%20Unordered%3F-,As%20of%20Python%20version%203.7%2C%20dictionaries%20are%20ordered.).

DanielRyanSmith (approved):
Approved with some questions, which are mostly to help me understand the requirements. 🙂

static/elements/chromedash-myfeatures-page.js
jrobbins: We could specify a sorting order for any of the feature tables.   The default is to sort by `-created.when`, so newly created feature entries are on top and things never change order.  That seems like a good default for the other boxes.

static/js-src/cs-client.js
jrobbins: These URLs are never seen by the user anyway because they are API calls.   

I think the way to let the user see the sort order would be in a UI element that is near the search filter widgets.  However, I was thinking that that would only be shown on the feature list page.  On the feature list page, sort order would need to be in the URL so that people can bookmark it and share it.  For the my features page, I'd rather keep the page pretty streamlined, without many options cluttering the page.  If users ask for the ability to customize their my features page, we could take that on at that time.

DanielRyanSmith: I see. After looking again, I'm seeing that this is a change to allow receiving the data from the API in a sorted order and doesn't necessary add any user functionality besides having default sort orders for the reviews and approvals section now. Makes sense!

internals/models_test.py
jrobbins: Gah, I forgot to delete these.  The sorting functionality for approvals moved into `sorted_by_pending_request_date()` and `sorted_by_review_date()`, which are tested on lines  169 to 181 of this file.

PR #2090, opened 2022-08-08
Add form for feature owners to verify feature data accuracy

DanielRyanSmith:
Second part of #2078

This change adds a new page that feature owners can use to verify the accuracy of their feature's data. This new page is linked in the notification email that feature owners will see when their feature is reaching an important milestone. The user can confirm that this data is accurate as of the current day, which will update the `accurate_as_of` field to the current day and time.

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -622,15 +622,11 @@ class FeatureVerifyAccuracy(FeatureEditStage):
     f, _ = self.get_feature_and_process(feature_id)
     feature_edit_dict = f.format_for_edit()
 
-    fields_title = "Accuracy last verified at time of creation."
+    forms_title = "Accuracy last verified at time of creation."
     if f.accurate_as_of is not None:
       date = f.accurate_as_of.strftime("%Y-%m-%d")
-      fields_title = f"Accuracy last verified {date}."
-
-    form_section_list = (fields_title, guideforms.Verify_Accuracy),
-    forms = [
-        (section_name, form_class(feature_edit_dict))
-        for section_name, form_class in form_section_list]
+      forms_title = f"Accuracy last verified {date}."
+    forms = [(forms_title, guideforms.Verify_Accuracy(feature_edit_dict))]
     template_data = {
         'feature': f,
         'feature_id': f.key.integer_id(),
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -554,6 +554,8 @@ class FeatureEditStage(basehandlers.FlaskHandler):
 
     if self.touched('standardization'):
       feature.standardization = int(self.form.get('standardization'))
+    if self.form.get('accurate_as_of'):
+      feature.accurate_as_of = datetime.now()
     if self.touched('unlisted'):
       feature.unlisted = self.form.get('unlisted') == 'on'
     if self.touched('comments'):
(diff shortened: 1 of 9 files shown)


templates/guide/verify_accuracy.html
```diff
@@ -0,0 +1,55 @@
...
+        {{ form }}
+      </chromedash-form-table>
+    </section>
+  {% endfor %}
+
+  <section class="final_buttons">
+    <chromedash-form-table>
+      <chromedash-form-field>
+        <span slot="field">
+            <input class="button" type="submit" value="Submit">
```
DanielRyanSmith, line 36: @jrobbins, this is currently implemented to navigate back to the feature edit page after submission like the other edit pages. Will this suffice? Or would it be better to display a message confirming the user has updated the data?

pages/guide.py
```diff
@@ -554,6 +554,8 @@ def process_post_data(self, feature_id, stage_id=0):
 
     if self.touched('standardization'):
       feature.standardization = int(self.form.get('standardization'))
+    if self.form.get('accurate_as_of'):
+      feature.accurate_as_of = datetime.now()
```
DanielRyanSmith, line 558: @jrobbins There are no other ways currently to update the `accurate_as_of` field outside of the new page that is only linked in notification emails. Hopefully, this is okay.

DanielRyanSmith (commented):
No action has been taken to avoid conflicting edits. This is under the assumption that an edit confiict would be a very unusual case. If needed, a change to mitigate this can be added.

pages/guide.py
jrobbins: That's fine for now.

pages/guide.py
```diff
@@ -611,3 +613,27 @@ def get_template_data(self, feature_id):
...
+  def get_template_data(self, feature_id):
+    f, _ = self.get_feature_and_process(feature_id)
+    feature_edit_dict = f.format_for_edit()
+
+    fields_title = "Accuracy last verified at time of creation."
+    if f.accurate_as_of is not None:
+      date = f.accurate_as_of.strftime("%Y-%m-%d")
+      fields_title = f"Accuracy last verified {date}."
+
+    form_section_list = (fields_title, guideforms.Verify_Accuracy),
```
jrobbins, line 630: I think this would be more clear as:
`form_section_list = [(fields_title, guideforms.Verify_Accuracy)]`

pages/guide_test.py
```diff
@@ -449,3 +449,33 @@ def test_html_rendering(self):
         self.template_data, self.full_template_path)
     parser = html5lib.HTMLParser(strict=True)
     document = parser.parse(template_text)
+
+class FeatureEditAllFieldsTemplateTest(TestWithFeature):
```
jrobbins, line 453: This is the same name as the preceding test class.

templates/guide/verify_accuracy.html
jrobbins: Navigating back to the edit page is fine.

Ideally we would display a confirmation as a toast or banner after any form submission.  I'll track an issue for that so that we implement it after these pages are converted to be SPA pages.  https://github.com/GoogleChrome/chromium-dashboard/issues/2092.

pages/guideforms.py
```diff
@@ -916,6 +939,8 @@ def make_display_specs(*shared_field_names):
     'comments',
 ]
 
+DISPLAY_IN_ACCURACY_VERIFICATION = ['accurate_as_of']
```
jrobbins, line 942: Why not include it on line 946 for DISPLAY_FIELDS_IN_STAGES['Metadata']?

jrobbins (changes requested):

pages/guide.py
DanielRyanSmith: I agree, and looking at it further, there were further ways to make it simpler. I've refactored this 👍

pages/guide_test.py
DanielRyanSmith: Wanted to use this as a template for testing the new class, but forgot to change the most important part 😅

pages/guideforms.py
DanielRyanSmith: I was not sure if this would affect any of the forms that currently exist, as we don't want this field showing anywhere else. If it will not affect which fields display elsewhere, I will add it 👍

jrobbins (approved):

PR #2093, opened 2022-08-10
Add OwnersFile models to store the API Owner file

### LINKED ISSUE #2084: Our requests to gitiles are being rate limited, resulting in some 500s on our site
**Describe the bug**
This morning we were seeing some 500s our site.  Logs showed exceptions when we got a 429 or 503 response for our requests to 
`https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/API_OWNERS?format=TEXT`
to see the page in a human-readable format, visit it without the format parameter:
https://chromium.googlesource.com/chromium/src/+/main/third_party/blink/API_OWNERS

**Additional context**
As discussed today: since the content of that file in gitles rarely changes, we should fetch it only once in a while and store the result (either the uudecoded text, or the parsed list of email addresses).   Storing it in datastore could allow all of our GAE instances to access it.  Since datastore does not have TTL to automatically discard that entity after a certain time, we could store a timestamp with it.   E.g.,
```class CachedApprovers(ndb.Model):
  url = ndb.StringProperty()
  content = ndb.TextProperty()
  created_on = ndb.DateTimeProperty(auto_now=True)
```
Then, in internals/approval_defs.py, where we fetch an OWNERS file, we could first query for any CachedApprovers that match the needed URL.  If we find one, and its created_on + one hour > now, then we can use it.  Otherwise, do the request to gitiles and cache the result.

KyleJu:
Fix #2084. 

Migrate `fetch_owners` from Ramcache to NDB and store the API_Owners file to ndb. Currently there should only be one entity for the OwnersFile ndb table.

### DIFF
--- a/internals/models.py
+++ b/internals/models.py
@@ -1740,6 +1740,36 @@ class FeatureOwner(DictModel):
         component_name, remove_as_owner=True)
 
 
+class APIOwner(DictModel):
+  """Describes the properties to store raw API_OWNERS content."""
+  url = ndb.StringProperty()
+  raw_content = ndb.TextProperty()
+  created_on = ndb.DateTimeProperty(auto_now_add=True)
+
+  def put(self):
+    """Add API owner content in ndb and delete all other entities."""
+    # Delete all other entities.
+    ndb.delete_multi(self.query().fetch(keys_only=True))
+    return self.put()
+
+  @classmethod
+  def get_raw_api_owners(cls, url):
+    """Retrieve raw API owner content, if it is created with an hour."""
+    q = cls.query()
+    q = q.filter(cls.url == url)
+    api_owner = q.fetch(1)
+    if not api_owner:
+      logging.info('API_OWNERS content does not exist for URL %s.' % (url))
+      return None
+
+    # Check if it is created within an hour.
+    an_hour_before = datetime.datetime.now() - datetime.timedelta(hours=1)
+    if api_owner.created_on < an_hour_before:
+      return None
+
+    return api_owner[0].raw_content
+
+
 class HistogramModel(ndb.Model):
   """Container for a histogram."""
 
--- a/internals/models.py
+++ b/internals/models.py
@@ -1749,25 +1749,27 @@ class OwnersFile(DictModel):
   def put(self):
     """Add the owner file's content in ndb and delete all other entities."""
     # Delete all other entities.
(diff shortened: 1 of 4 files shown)


internals/models.py
```diff
@@ -1740,6 +1740,36 @@ def remove_as_component_owner(self, component_name):
         component_name, remove_as_owner=True)
 
 
+class APIOwner(DictModel):
+  """Describes the properties to store raw API_OWNERS content."""
+  url = ndb.StringProperty()
```
jrobbins, line 1745: Maybe make url and raw_content required?
I generally don't make a lot of fields required, but in this case the URL is like the key, and the raw_content is pretty much the point of the whole entity.

internals/models.py
```diff
@@ -1740,6 +1740,36 @@ def remove_as_component_owner(self, component_name):
         component_name, remove_as_owner=True)
 
 
+class APIOwner(DictModel):
```
jrobbins, line 1743: It would be good to generalize this a bit to other OWNERS files as well.  E.g., name it something like `OwnersFile`.

internals/models.py
```diff
@@ -1740,6 +1740,36 @@ def remove_as_component_owner(self, component_name):
...
+  """Describes the properties to store raw API_OWNERS content."""
+  url = ndb.StringProperty()
+  raw_content = ndb.TextProperty()
+  created_on = ndb.DateTimeProperty(auto_now_add=True)
+
+  def put(self):
+    """Add API owner content in ndb and delete all other entities."""
+    # Delete all other entities.
+    ndb.delete_multi(self.query().fetch(keys_only=True))
```
jrobbins, line 1752: Deleting all other instances works so long as the API Owners file is the only OWNERS file that we have, but in the future we might have other approvals that parse other OWNERS files from different URLs.   Deleting unneeded entities could still be done like:
`ndb.delete_multi(self.query(OwnersFile.url == url).fetch(keys_only=True))`

jrobbins (changes requested):
I have a few comments to consider.

jrobbins (approved):

KyleJu:
https://github.com/GoogleChrome/chromium-dashboard/runs/7771977381?check_suite_focus=true - the CI is hanging on the test. Let me dig into it

KyleJu:
Tests also hanging locally. The test setup needs to be tweaked

PR #2094, opened 2022-08-10
Draft for discussion about schema revamp

jrobbins:
I don't think that landing this right now would be a good idea, but it shows some of the end goal that I would like to reach by doing the schema revamp work.  Let's discuss this along with the design doc, specifically the source code organization section.


internals/core_enums.py
```diff
@@ -0,0 +1,283 @@
...
+FEATURE_TYPE_INCUBATE_ID = 0
+FEATURE_TYPE_EXISTING_ID = 1
+FEATURE_TYPE_CODE_CHANGE_ID = 2
+FEATURE_TYPE_DEPRECATION_ID = 3
+
+FEATURE_TYPES = {
+  FEATURE_TYPE_INCUBATE_ID: 'New feature incubation',
+  FEATURE_TYPE_EXISTING_ID: 'Existing feature implementation',
+  FEATURE_TYPE_CODE_CHANGE_ID: 'Web developer facing change to existing code',
+  FEATURE_TYPE_DEPRECATION_ID: 'Feature deprecation',
```
jpmedley, line 73:
```suggestion
  FEATURE_TYPE_DEPRECATION_ID: 'Feature deprecation' and removal,
```

internals/core_enums.py
```diff
@@ -0,0 +1,283 @@
...
+INTENT_NONE = 0
+INTENT_INCUBATE = 7  # Start incubating
+INTENT_IMPLEMENT = 1  # Start prototyping
+INTENT_EXPERIMENT = 2  # Dev trials
+INTENT_IMPLEMENT_SHIP = 4  # Eval readiness to ship
+INTENT_EXTEND_TRIAL = 3  # Origin trials
+INTENT_SHIP = 5  # Prepare to ship
+INTENT_REMOVED = 6
+INTENT_SHIPPED = 8
+INTENT_PARKED = 9
```
jpmedley, line 87: Why is there no separate constant for intent to deprecate?

internals/core_enums.py
```diff
@@ -0,0 +1,283 @@
...
+  FEATURE_TYPE_INCUBATE_ID: 'New feature incubation',
+  FEATURE_TYPE_EXISTING_ID: 'Existing feature implementation',
+  FEATURE_TYPE_CODE_CHANGE_ID: 'Web developer facing change to existing code',
+  FEATURE_TYPE_DEPRECATION_ID: 'Feature deprecation',
+}
+
+
+# Intent stages and mapping from stage to stage name.
+INTENT_NONE = 0
+INTENT_INCUBATE = 7  # Start incubating
```
jpmedley, line 79: I would like to understand the naming convention. There's no such thing as an intent to incubate. Same with removed, shipped, and parked.

jpmedley (changes requested):
I noted a few things in enums. It contains other things I don't understand. I don't know if they're things that should be changed or things that have some backward compatibility reason. I hope we'll go over this in a meeting.

internals/core_enums.py
jrobbins: This PR just shoveled the existing enums from models.py into a new file.   This PR is just showing what file they would be located in.  

I would like to redo these enums to give them better names and make them more comprehensive for all our processes.  That will come later, but you are right to point it out so that it does not get forgotten.

internals/review_models.py
```diff
@@ -0,0 +1,152 @@
...
+        feature_id=feature_id, field_id=field_id, states=[cls.REVIEW_REQUESTED])
+    for rr in review_requests:
+      rr.key.delete()
+
+
+class Amendment(ndb.Model):
+  """Comments can log changes to other fields of the feature."""
+  field_name = ndb.StringProperty()  // from QUERIABLE_FIELDS
+  old_value = ndb.TextProperty()
+  new_value = ndb.TextProperty()
```
DanielRyanSmith, line 130: This seems like it would need a timestamp of when the change was made if that doesn't get generated by default.

internals/core_models.py
```diff
@@ -0,0 +1,237 @@
...
+  api_spec = ndb.BooleanProperty(default=False)
+  spec_mentors = ndb.StringProperty(repeated=True)
+  interop_compat_risks = ndb.TextProperty()
+  prefixed = ndb.BooleanProperty()
+  all_platforms = ndb.BooleanProperty()
+  all_platforms_descr = ndb.TextProperty()
+  tag_review = ndb.StringProperty()
+  tag_review_status = ndb.IntegerProperty(default=REVIEW_PENDING)
+  non_oss_deps = ndb.TextProperty()
+  origin_trial_feedback_url = ndb.StringProperty()
```
DanielRyanSmith, line 83: The feedback url might need to live at the stage level, unless all trials associated with a feature assuredly give feedback through the same channel.

DanielRyanSmith (commented):
I wish I had more commentary for this change. I think I need more familiarity with the individual launches and the feature lifecycle (which hopefully I will get more of here 🙂). The biggest thing that I am interested in is how stage-specific fields will be handled. Perhaps this is a good time to keep the Origin Trial integration in mind to discern which fields need to be present to accommodate this and on which entities they need to live.

jpmedley:
>Perhaps this is a good time to keep the Origin Trial integration in mind to discern which fields need to be present to accommodate this and on which entities they need to live.

+100

internals/review_models.py
jrobbins: Every amendment is part of a comment, which has a timestamp.    In cases where the user did not enter any comment content, the comment will be just an empty string.

internals/core_models.py
jrobbins: I am not 100% sure but I think that the origin_trial_feedback_url is supposed to be a document that summarizes the feedback that was given, not a way for users to submit feedback.   If that understanding is correct, then it could be OK to have just one URL for the feature, but I suppose that it would still be OK to have it per-origin-trial.  I'll change it to be in the stage.

jrobbins:
This is superseded by https://github.com/GoogleChrome/chromium-dashboard/pull/2165

PR #2095, opened 2022-08-10
Convert content in guide/new.html into a component

kevinshen56714:
This PR converts the guide/new page into the `chromedash-guide-new-page` component. Changes include:

1. Created chromedash-guide-new-page.js with the forms passed in directly from Django template. This will be changed once the form generation code is migrated to the client side.
2. Created form.css.js and import it to the page component.
3. Moved all the css for chromedash-form-field from form.scss to chromedash-form-field.js itself.


static/elements/chromedash-form-field.js
```diff
@@ -1,8 +1,92 @@
-import {LitElement, html} from 'lit';
+import {LitElement, html, css} from 'lit';
 import {ALL_FIELDS} from './form-field-specs';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
+import {SHARED_STYLES} from '../sass/shared-css.js';
+
 
 export class ChromedashFormField extends LitElement {
+  static get styles() {
```
dlaliberte, line 8: I'm not sure this change to chromedash-form-field.js is correct, or maybe it is simply unnecessary.  Previously I changed the form-field html to render in the light dom, so the styles in this component, which are automatically qualified to apply only to this component, will not apply.  Instead, the same styles were copied to the forms.scss file.

static/elements/chromedash-guide-new-page.js
```diff
@@ -0,0 +1,150 @@
+import {LitElement, css, html} from 'lit';
+import {unsafeHTML} from 'lit/directives/unsafe-html.js';
+import './chromedash-form-table';
+import './chromedash-form-field';
+import {SHARED_STYLES} from '../sass/shared-css.js';
+import {FORM_STYLES} from '../sass/forms-css.js';
```
dlaliberte, line 6: Maybe I needed to add the same form-field styles to forms-css.js.

static/elements/chromedash-form-field.js
kevinshen56714: I see. Thank you for the clarification. I was wondering why the styles are not applied. I will moved these back to form.scss.

kevinshen56714: I moved these (chromedash-form-field related css) back to form.scss, and also put them in form.css.js.

dlaliberte (commented):
Looks good to me.  Are you planning to add some tests for this new component?

kevinshen56714:
> Looks good to me. Are you planning to add some tests for this new component?

Yes, I will add a unit test for this. Also there are some missing css I need to fix.

I will request for reviews once this is ready.

kevinshen56714:
@jrobbins @dlaliberte this PR is ready for review. Some additional changes:

I added a unit test for the page. Also, since we implement a custom createRenderRoot method for <chromedash-form-field> (thus no shadowRoot), the page component style can actually leaks into <chromedash-form-field> elements. So I had to be extra careful with the styles and added a few more css to make sure it looks the same as before.

dlaliberte:
> the page component style can actually leaks into `<chromedash-form-field>` elements

Yes, this is why we should avoid doing the workaround being used by chromedash-form-field.  We are considering how to modify the form-level behavior so that the form fields nested in the shadow DOM of chromedash-form-field will be included in submitted data.  It would be good to make that change before we proceed with migrating other pages that use forms.

dlaliberte (approved):

jrobbins (approved):

PR #2098, opened 2022-08-10
Fix JS lint warning.

jrobbins:
When I was looking at a CI test log for another PR, I noticed that this file has a lint warning:
```
./static/elements/chromedash-app.js

    'Element' is not a valid type for the default converter. Have you considered '{attribute: false}' instead?
    57:  pageComponent: {type: Element},
    no-incompatible-property-type
```

This PR simply applies the suggested change.

### DIFF
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -54,7 +54,7 @@ class ChromedashApp extends LitElement {
       currentPage: {type: String},
       bannerMessage: {type: String},
       bannerTime: {type: Number},
-      pageComponent: {type: Element},
+      pageComponent: {attribute: false},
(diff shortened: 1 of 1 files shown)


kevinshen56714 (approved):
Thank you! Somehow, I didn't notice this.

jrobbins:
I normally don't notice JS lint warnings either.   Lint errors stop the app from running locally, but warnings can slip by.

jrobbins:
> I normally don't notice JS lint warnings either. Lint errors stop the app from running locally, but warnings can slip by.

I guess ideally, these lint warnings would become code review comments in GitHub so that people would need to address them before merging the change.  There might be a way to configure GitHub Actions to do that, but I don't know it.

PR #2099, opened 2022-08-11
Refactor search-related code out of models.py.

jrobbins:
Refactor code that supports search_api out of models.py.  This is because models.py is large and disorganized.  This code is pretty cohesive, so it seems natural to keep it in its on file.  The files search.py and search_queries.py put related code into two reasonably-sized files with corresponding names.


internals/search_queries.py
```diff
@@ -0,0 +1,198 @@
...
+  query = models.Approval.query(
+      models.Approval.state.IN(models.Approval.FINAL_STATES))
+  if descending:
+    query = query.order(-models.Approval.set_on)
+  else:
+    query = query.order(models.Approval.set_on)
+  recent_approvals = query.fetch(projection=['feature_id'])
+
+  feature_ids = utils.dedupe(ra.feature_id for ra in recent_approvals)
+  return feature_ids
```
DanielRyanSmith, line 102: These functions seem very similar. Would there be utility in merging them and making them reusable?

DanielRyanSmith (approved):
Looks like a pretty straightforward refactor - hopefully I didn't miss anything 😅

internals/search_queries.py
jrobbins: Good point.  I made a helper function to avoid repeating similar code.

KyleJu (approved):

internals/search_queries.py
```diff
@@ -74,32 +74,36 @@ def total_order_query_async(sort_spec):
...
+  joined_models = query.fetch(projection=['feature_id'])
+  feature_ids = utils.dedupe(jm.feature_id for jm in joined_models)
   return feature_ids
 
 
+def sorted_by_pending_request_date(descending):
+  """Return feature_ids of pending approvals sorted by request date."""
+  return _sorted_by_joined_model(
+      models.Approval,
+      models.Approval.state == models.Approval.REVIEW_REQUESTED,
```
KyleJu, line 97: `models.Approval.state == models.Approval.REVIEW_REQUESTED` would this part be evaluated automatically to True or. False by Python?

jrobbins: ndb is tricky:  It's not a boolean at all, it evaluates to a `ndb.FilterNode` object.   `models.Approval.state` is an `ndb.IntegerProperty`.  Its  base class `ndb.Property` overrides `__eq__` to return a `ndb.FilterNode` object rather than a boolean.   We make use of that overloaded operator all the time in calls to SomeModelClass.query() and some_query.filter().

KyleJu: ` overrides __eq__ to return a ndb.FilterNode` -> aha interesting choice here

PR #2100, opened 2022-08-11
Refactor metrics-related models out of models.py.

jrobbins:
Refactor a set of model classes related to metrics into a small coherent file named metrics_models.py.

One class is changing its base class from DictModel to regular Model.  I don't believe that the difference in the to_dict() method implementation will make any difference for these classes.


internals/metrics_models.py
```diff
@@ -0,0 +1,58 @@
...
+  day_percentage = ndb.FloatProperty()
+  rolling_percentage = ndb.FloatProperty()
+
+
+class AnimatedProperty(StableInstance):
+  pass
+
+
+class FeatureObserver(StableInstance):
+  pass
```
DanielRyanSmith, line 37: What is the purpose of these empty inheriting classes? Used just for naming clarity?

DanielRyanSmith (approved):
Great change for organization's sake 👍

internals/metrics_models.py
jrobbins: Each ndb.Model subclass creates a different entity kind in datastore.  It is like having separate SQL tables.  This code was here before me, but I think the purpose is just to keep separate types of metrics data in separate entities so that queries done per-entity would retrieve only those entities.

KyleJu (approved):

PR #2103, opened 2022-08-12
Several small tweaks to accuracy verification

jrobbins:
This PR fixes a few small things that I noticed while testing this feature locally:
* The cron job was returning an HTML page rather than JSON with a message.  Our Flask UI handlers typically return dictionary of web page template data from get_template_data().  Returning JSON is a special case for a UI page, it can be done by setting JSONIFY = True.
* The class variable TEMPLATE_PATH is typically used for web page templates.  I renamed it to EMAIL_TEMPLATE_PATH to avoid confusion (and to prevent it from being used to generate a web page in case JSONIFY is ever removed).
* Note that it doesn't really matter what is returned from a /cron handler because nothing looks at the response other than to check that it is status 200, but the above tweaks make it work as intended.
*  Used sentence case for the field name to match our other field names.  Some of the field names have a capital for Chrome or Android and such, but that's because those are proper nouns.
* Tweaked the help text for the accurate_as_of checkbox to make it read more like instructions.
*

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -255,7 +255,7 @@ class FeatureStar(models.DictModel):
 
 
 class FeatureAccuracyHandler(basehandlers.FlaskHandler):
-  SEND_NOTIFICATIONS = False
+  JSONIFY = True
 
   CHROME_RELEASE_SCHEDULE_URL = (
 'https://chromiumdash.appspot.com/fetch_milestone_schedule')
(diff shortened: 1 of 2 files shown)


internals/notifier.py
```diff
@@ -255,7 +255,7 @@ def get_feature_starrers(self, feature_id):
 
 
 class FeatureAccuracyHandler(basehandlers.FlaskHandler):
-  SEND_NOTIFICATIONS = False
+  JSONIFY = True
```
DanielRyanSmith, line 258: Does this need to be used somewhere?

DanielRyanSmith (approved):

internals/notifier.py
DanielRyanSmith: This is handled in the parent class, nevermind!

PR #2106, opened 2022-08-12
Refactor models.py enums into core_enums.py.

jrobbins:
Refactor a bunch of crufty enums into a separate file so that models.py is shorter.  This should make it easier to focus on the parts of models.py that we will need to change when we do the schema revamp.

The enums themselves could probably use a big review and/or revamp, but nothing is forcing a deadline for that yet.


DanielRyanSmith (approved):

internals/core_enums.py
```diff
@@ -0,0 +1,317 @@
...
+FEATURE_TYPE_CODE_CHANGE_ID = 2
+FEATURE_TYPE_DEPRECATION_ID = 3
+
+FEATURE_TYPES = {
+    FEATURE_TYPE_INCUBATE_ID: 'New feature incubation',
+    FEATURE_TYPE_EXISTING_ID: 'Existing feature implementation',
+    FEATURE_TYPE_CODE_CHANGE_ID: 'Web developer facing change to existing code',
+    FEATURE_TYPE_DEPRECATION_ID: 'Feature deprecation',
+}
+
```
KyleJu, line 76: Let's make the spacing consistent. One space between the constant

KyleJu (approved):

internals/core_enums.py
jrobbins: I think they are supposed to be in groups related to each field, with two blink lines between the groups.   In each group there are first the definition of integer values for the enums, and then a dict or OrderedDict that gives display name strings.

PR #2107, opened 2022-08-12
Register our form listener after shoelace.

jrobbins:
Our code in HEAD is back in the situation where a form validation error can be displayed, but then the form submits anyway.

IIRC that was resolved last time by tweaking the order in which JS code was loaded so that our listener gets loaded after the Shoelace form elements add their form listeners.

This PR basically sticks with that strategy but does it via setTimeout.

### DIFF
--- a/static/js-src/shared.js
+++ b/static/js-src/shared.js
@@ -33,4 +33,7 @@ gtag('config', 'UA-179341418-1');
 // our listener tries to programatically submit the form.
 // Ten seconds should be enough time, and there is no rush since
 // XSRF tokens don't expire until much more time has passsed.
-window.setTimeout(window.csClient.addFormSubmitListner, 10 * 1000);
+// TODO(jrobbins): Make this event-driven rather than a timeout.
+window.setTimeout(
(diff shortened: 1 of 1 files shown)


static/js-src/shared.js
```diff
@@ -27,4 +27,10 @@ gtag('config', 'UA-179341418-1');
 // End Google Analytics
 
 
-window.csClient.addFormSubmitListner();
+// Make sure that our form handler for XSRF token updates gets
+// registered after all the Shoelace form listeners are registered
+// so that failed form validation cancels the submit event before
+// our listener tries to programatically submit the form.
+// Ten seconds should be enough time, and there is no rush since
+// XSRF tokens don't expire until much more time has passsed.
```
dlaliberte, line 35: Should probably add a TODO to fix this in a more reliable way that avoids the 10 second timeout.

dlaliberte (approved):

static/js-src/shared.js
jrobbins: Added

PR #2108, opened 2022-08-12
Replace metrics timeline handlers with ConstHandler.

jrobbins:
Thanks to your work on making the metrics pages API-driven, we no longer need to have a handler that gathers data for the menu items.  These handlers were only calculating a template variables that were not used by the django template.   So, we can just use ConstHandler, as you did for the metrics stack rank pages.

Skipping that calculation that the associated datastore query is sure to make the page load a little faster.  I didn't spend much time measuring it but it looks like the speedup is less than 100ms.  So, it was mainly the reduced number of source code lines that were the last little bit of advantage to wrapping up this line of work.


kevinshen56714 (approved):
LGTM! I think the RoadmapHandler, SettingsHandler, MyFeaturesHandler, and NewFeatureListHandler can also be replaced with ConstHandler as those pages no longer rely on tempalte data.

jrobbins:
> I think the RoadmapHandler, SettingsHandler, MyFeaturesHandler, and NewFeatureListHandler can also be replaced with ConstHandler as those pages no longer rely on tempalte data.

Awesome.  I'll send another PR soon.

PR #2110, opened 2022-08-12
Escape field values in HTML form generation.

jrobbins:
This should resolve the problem that I discussed on our chat group this morning.

In this PR:
* One-line fix to escape field values
* Added unit tests

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -560,14 +560,14 @@ class ChromedashForm(forms.Form):
                 row_template = checkbox_row
                 if field.widget.check_test(value):
                     checked = True
-                
+
                 # accurate_as_of field should always be checked, regardless of
                 # the current value. This is only necessary if the feature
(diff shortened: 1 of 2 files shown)


pages/guideforms.py
```diff
@@ -560,14 +560,14 @@ def simple_html_output(self):
                 row_template = checkbox_row
                 if field.widget.check_test(value):
                     checked = True
-                
+
```
DanielRyanSmith, line 563: I'm pretty sure these phantom spacings are from me - I'll try and be more diligent about avoiding this!

DanielRyanSmith (approved):

pages/guideforms.py
dlaliberte: I know I have left a few.  VSCode wasn't cleaning up automatically as I thought it ought to have been.

PR #2115, opened 2022-08-13
Handle form submission for forms in shadowDOMs

kevinshen56714:
The code in cs-client.js that looks for all the `<form>` elements in the document and adds submit event listeners won't work when forms are inside of shadowDOM trees because it's using document.querySelectorAll('input[name=token]'). This PR fixes it by registering and handling form submission event within the components. The changes include:

1. In chromedash-guide-new-page.js and chromedash-guide-metadata.js, the addEventListener('submit') is called for the forms upon `firstUpdated()` or `updated()`. This allows the form's event listener to be added after Shoelace event listeners are attached. And the callback function `handleFormSubmission()` does the xsrf token attachment and expiration check.
2. With this method, the validation works fine for form fields not in the `${unsafeHTML(this.overviewForm)}`. I found that we need an extra timeout after `firstUpdated()` or `updated()` for the Shoelace event listeners within `${unsafeHTML(this.overviewForm)}` to be attached. I believe this would not be needed once the temporary unsafeHTML workaround is removed.
3. Since the `handleFormSubmission()` would attach xsrf token to the hidden input value anyway from window.csClient, I removed the xsrfToken property in the components and stop passing in xsrf token into the components. And the unit tests are modified accordingly.

### DIFF
--- a/static/elements/chromedash-guide-metadata.js
+++ b/static/elements/chromedash-guide-metadata.js
@@ -72,7 +72,6 @@ export class ChromedashGuideMetadata extends LitElement {
       user: {type: Object},
       featureId: {type: Number},
       feature: {type: Object},
-      xsrfToken: {type: String},
       overviewForm: {type: String},
       editing: {type: Boolean},
       loading: {type: Boolean},
@@ -84,7 +83,6 @@ export class ChromedashGuideMetadata extends LitElement {
     this.user = {};
     this.featureId = 0;
     this.feature = {};
-    this.xsrfToken = '';
     this.overviewForm = '';
     this.editing = false;
     this.loading = true;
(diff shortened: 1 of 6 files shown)


static/elements/chromedash-guide-metadata.js
```diff
@@ -95,6 +93,31 @@ export class ChromedashGuideMetadata extends LitElement {
     this.fetchData();
   }
 
+  /* Add the form's event listener after Shoelace event listeners are attached
+   * see more at https://github.com/GoogleChrome/chromium-dashboard/issues/2014 */
+  updated() {
```
jrobbins, line 98: Does updated() get called multiple times?
If each call adds a listener, do we need to remove the old listeners?

jrobbins (approved):
LGTM after considering a couple of questions

static/elements/chromedash-guide-metadata.js
dlaliberte: Yes updated will be called multiple times, so we need to either remove old listeners or not add another if one is already active.

kevinshen56714: I did think about and test this.

The only thing that would trigger updated() after first update currently in this component is the `editing` attribute. However, every time when `editing` changes from false to true, a new form is attached to the DOM because form rendering is toggled by this variable as well:
```js
${this.editing ? this.renderEditForm() : this.renderReadOnlyTable()}
```

I also tested putting console.log() in the handleFormSubmission(), and found that there's always only one active listener (I assume the old listenter is garbage collected with the old form when `editing` changes from true to false).

dlaliberte (approved):

PR #2116, opened 2022-08-13
*chrome_logo* updated

prakrit55:
I updated the chrome logo. The discord link to join chrome community isn't working so, please fix the thing :)


jrobbins:
We are using the one from chrome.com.  It has the slight shadow.  If that website updates the logo, then we will follow it.

Bellisario:
Chrome logo on chrome.com seems it's updated to the latest style.
I think should be good to update all logos stable/dev/canary to match this.

PR #2118, opened 2022-08-15
Convert guide/edit page into a component

kevinshen56714:
This PR transforms content in guide/edit.html to the `<chromedash-guide-edit-page>` component. The changes include:

1. Created chromedash-guide-edit-page.js and chromedash-guide-edit-page_test.js
2. Created ProgressAPI (under the `/features/<int:feature_id>/progress` route) and its unit test. Also added the getFeatureProgress() Get method in cs-client.js.
3. Moved the remaining css in guide.scss (style for #breadcrumbs) to shared.css.js and retired guide.scss
4. Modified chromedash-guide-metadata's attributes because data is now fetched at the page level component and passed down to child components. The loading skeleton rendering is moved to the page component as well.
5. Modified several components' dismissedCues attribute from object type to array type (it is an array).
6. guide/edit.html is now a single component file!


dlaliberte (approved):
LGTM

Jason might have some comments.

jrobbins (approved):

PR #2129, opened 2022-08-15
Generate sl-select fields via chromedash-form-field.js

dlaliberte:
This PR generates sl-select elements and sl-menu-item elements (for the options) instead of the django generated elements.

[Starting with empty PR, publishing first, just to get that part done.
But, surprise, that can't be done since there are no commits.]


static/elements/chromedash-form-field.js
```diff
@@ -33,22 +32,39 @@ export class ChromedashFormField extends LitElement {
...
+      fieldHTML = html`
+        <sl-select
+          name="${this.name}"
+          id="id_${this.name}"
+          value="${this.value}"
+          size="small"
+          ?disabled=${this.disabled}
+        >
+          ${Object.values(choices).map(
+            ([index, label]) => html`
```
jrobbins, line 62: Naming the first value `index` is a little confusing because it seems like it would be the array index that can be used in `map` callbacks.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

It would be better to name it `intValue`, `itemValue`, or something.

static/elements/form-field-models.js
```diff
@@ -0,0 +1,111 @@
+
+// The order of appearance of properties is the order displayed.
```
jrobbins, line 2: Today I learned that order is preserved... so long as you don't also have any numeric keys.

Can we rename this file form-field-enums.js, since it has only enums?

This file starts with a blank line.

jrobbins (changes requested):

static/elements/form-field-models.js
dlaliberte: OK, enums is fine.

static/elements/chromedash-form-field.js
dlaliberte: It is generically any value, but intValue is fine.

jrobbins (approved):

PR #2141, opened 2022-08-17
Refactor user-related models into a separate file.

jrobbins:
This is another step toward being ready for the schema revamp.

In this PR:
* Refactor user-related models into internals/user_models.py
* Update code that refers to those models
* Change their base class to `ndb.Model` rather than a custom base class
* Replace calls to `format_for_template()` with explicit wire formats


api/accounts_api.py
```diff
@@ -13,17 +13,24 @@
...
+
+
+def user_to_json_dict(user):
+  return {
+      'is_admin': user.is_admin,
+      'is_site_editor': user.is_site_editor,
+      'email': user.email,
+      'id': user.key.integer_id(),
+  }
+
```
DanielRyanSmith, line 33: Out of curiosity, what's the motivation to not having this as a method in the `AppUser` class?

DanielRyanSmith (approved):
This will likely require some refactoring in the permissions PR, so I'll have to take care of that soon 🙂

api/accounts_api.py
jrobbins: I feel that the API code should have responsibility for protocol between the client and server, including the serialization format.  The model classes should just have the schema, database-level validation, and basic queries.

PR #2143, opened 2022-08-18
Fix feature creation page category and feature_type fields.

jrobbins:
We got our wires crossed in some recent changes to the /guide/new page and the processing of select fields.  The result was that:
1. The /guide/new page in HEAD is showing a `<select>` instead of the desired radio buttons, and 
2. The category `<select>` had a value of "None" which shows no item selected.  If a user did not select a value, the form POSTs `None` which causes a 500 in form POST processing.

The cause of (1) is that Kevin's version of /guide/new worked when tested by itself, but when combined with Dan's concurrently developed PR to process select fields in JS code caused Kevin's radio buttons to be ignored and a `<select>` element generated.

The cause of (2) was a lack of handling a form field definitions initial value in ChromedashForm.   I'm not sure why this didn't show up before.  It might have simply not been noticed because the /guide/new page is the only page that relies on form field initial values rather than NDB model field default values.

In this PR:
* Change the name of the radio button group to a different name, allowing it to match a new spec in form-field-specs.js that does not have `type: "select"`.  This allows the radio button group HTML markup to be passed through and rendered.
* In ChromedashForm, check for a value of None and use the form field default instead.

### DIFF
--- a/pages/guideforms.py
+++ b/pages/guideforms.py
@@ -553,6 +553,8 @@ class ChromedashForm(forms.Form):
 
             # Get value and checked for the field
             value = field.widget.value_from_datadict(self.data, self.files, self.add_prefix(name))
+            if value is None:
+              value = field.initial
 
             row_template = normal_row
             checked = False
--- a/pages/guideforms_test.py
+++ b/pages/guideforms_test.py
@@ -26,7 +26,7 @@ from internals import models
 
 
 TestForm = guideforms.define_form_class_using_shared_fields(
-    'TestForm', ('name', 'summary'))
+    'TestForm', ('name', 'summary', 'category'))
 
 TEST_TEMPLATE = '''
 <!DOCTYPE html>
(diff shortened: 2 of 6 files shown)


jrobbins:
FYI, I already pushed this change live on staging and prod because some user hit it on the staging server and it was showing up on the GAE Error Reporting page.  I think I got the fix live before anyone hit the problem on prod.    You can still repro the problem locally in HEAD.

kevinshen56714 (approved):
LGTM, but there's a typo in form-field-enums.js for the const `FEATURE_TYPES`
```js
FEATURE_TYPE_CODE_CHANGE_ID: [2, 'Web developer facing change to exi[sting code']
```
should be
```js
FEATURE_TYPE_CODE_CHANGE_ID: [2, 'Web developer facing change to existing code']
```

kevinshen56714 (commented):
Also, line 23 in chromedash-guide-new-page_test.js:
```js
const featureTypeFormField = component.shadowRoot.querySelector(
    'chromedash-form-field[name="feature_type"]');
```
should change to 

```js
const featureTypeFormField = component.shadowRoot.querySelector(
    'chromedash-form-field[name="feature_type_radio_group"]');
```
to fix the unit test failure.

kevinshen56714 (changes requested):
The unit test is failing because of the issue I commented above.

jrobbins:
Kevin, good comments.  I have fixed both.

dlaliberte (approved):

PR #2144, opened 2022-08-18
Use Shoelace elements in the approvals dialog

jrobbins:
This makes this dialog look more like the rest of the app and makes it a little more finger-friendly because the widgets have more padding.

In this PR:
* Switch from native input field elements to the shoelace elements in the approvals dialog
* Use flex layout to make things line up better
* Remove the cancel button to give more space.  Users can click the "X" or anywhere outside the dialog.

One tricky bit: we need to use the `hoist` attribute on `<sl-select>` when it is in a context with `overflow:hidden`.


jrobbins:
Old version:
image

New version:
image

dlaliberte:
> One tricky bit: we need to use the `hoist` attribute on `<sl-select>` when it is in a context with `overflow:hidden`.

This should probably be added as a comment near the use of the 'hoist' attribute.

dlaliberte (approved):
All looks good with a relatively simple change.

I wonder what would happen if the html for approvals dialog happened to be contained inside one of the feature forms.  Does shoelace have a way to selectively exclude its form fields from a containing form?

kevinshen56714 (approved):

jrobbins:
> All looks good with a relatively simple change.

I added the comment about `hoist`.

> I wonder what would happen if the html for approvals dialog happened to be contained inside one of the feature forms. Does shoelace have a way to selectively exclude its form fields from a containing form?

Well, yeah, that was the whole problem that we hit.  Since `chromedash-approvals-dialog` is its own custom element with its own shadow root, its input elements would not be submitted as part of any `<form>` that contains it.   I think that Cory was basically saying that that is the situation and desired behavior in pretty much all cases.  No need to selectively exclude because it should always exclude.

He was missing the opposite case in which some custom element is being used to modularize a group of input elements that should contribute to a `<form>`.   That's the case with our editing forms, but not here.

dlaliberte:
> > I wonder what would happen if the html for approvals dialog happened to be contained inside one of the feature forms. Does shoelace have a way to selectively exclude its form fields from a containing form?
> 
> Well, yeah, that was the whole problem that we hit. Since `chromedash-approvals-dialog` is its own custom element with its own shadow root, its input elements would not be submitted as part of any `<form>` that contains it. 

Right.. I keep forgetting.  In fact, if this component happened to render in light DOM, then the shoelace form fields might be included in the containing form, if any.  But if this component renders these shoelace form fields inside the shadow DOM of yet another shoelace component, are they are hidden again?   Yes!

> I think that Cory was basically saying that that is the situation and desired behavior in pretty much all cases. No need to selectively exclude because it should always exclude.

> He was missing the opposite case in which some custom element is being used to modularize a group of input elements that should contribute to a `<form>`. That's the case with our editing forms, but not here.

I am not accepting that this default behavior is necessarily the desired behavior, nor is the opposite always desired.  Both are useful in different contexts, so both should be supported.  I wouldn't mind if it was easy enough to change from one mode to the other, and clear enough what is going on.   But the current default behavior (hiding form fields that are in shadow  DOM) is rather difficult and confusing to work around (e.g. the CSS is quite different).

PR #2145, opened 2022-08-18
Generate text, url, multi-email, and number inputs through JS

### LINKED ISSUE #2139: Convert single-line input fields to be generated in JS code rather than by django
We're trying to phase our django by doing more and more UI details in custom elements.

Convert all kinds of `<input>` fields to be generated in JS render() elements, similar to how checkboxes and selects are generated now.  That would include setting the needed attributes, such as min, max, pattern, placeholder, and title.

kevinshen56714:
This resolves #2139. All the single-line `<input>` generation is migrated from guideforms.py to chromedash-form-field.js. 

The single line input types include: TextInput, URLInput, MultiEmailField, and NumberInput. These are distinguish by the `input_type` property specified in form-field-specs.js and rendered differently with different title, placeholder, ... etc specified in chromedash-form-field.js.

### DIFF
--- a/static/elements/chromedash-form-field.js
+++ b/static/elements/chromedash-form-field.js
@@ -82,11 +82,15 @@ export class ChromedashFormField extends LitElement {
         </sl-select>
       `;
     } else if (type === 'input') {
-      let inputType = 'text';
+      let inputType = '';
       let title = '';
       let placeholder;
       let pattern;
       switch (fieldProps.input_type) {
+        case 'text':
+          inputType = 'text';
+          break;
+
         case 'url':
           inputType = 'url';
           title = 'Enter a full URL https://...';
@@ -107,6 +111,9 @@ export class ChromedashFormField extends LitElement {
           inputType = 'number';
           placeholder = 'Milestone number';
           break;
+
+        default:
+          console.error(`Invalid input type: ${fieldProps.input_type}`);
       }
       fieldHTML = html`
         <sl-input 
--- a/static/elements/chromedash-form-field.js
+++ b/static/elements/chromedash-form-field.js
@@ -2,7 +2,15 @@ import {LitElement, html} from 'lit';
 import {ALL_FIELDS} from './form-field-specs';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
 
+
+/* Patterns from https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s01.html
+ * Removing single quote ('), backtick (`), and pipe (|) since they are risky unless properly escaped everywhere.
+ * Also removing ! and % because they have special meaning for some older email routing systems. */
+const USER_REGEX = '[A-Za-z0-9_#$&*+/=?{}~^.-]+';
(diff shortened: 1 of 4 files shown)


dlaliberte (approved):
We should consider refactoring chromedash-form-field at some point to move the handling of different field types into their own components.

Not sure I like the 'input_type' property for some subtypes of type: 'input'.  But it's OK for now.

kevinshen56714:
> We should consider refactoring chromedash-form-field at some point to move the handling of different field types into their own components.

Ok, Jason and I had a discussion on this as well, and he thinks in the same way. I can work on creating a `<chromedash-input>` component in the next PR.

>
> Not sure I like the 'input_type' property for some subtypes of type: 'input'. But it's OK for now.

Do you have any suggestions on the naming or different ways to specify the type of the `<input>`? Thank you!

dlaliberte:
> > Not sure I like the 'input_type' property for some subtypes of type: 'input'. But it's OK for now.
> 
> Do you have any suggestions on the naming or different ways to specify the type of the `<input>`? Thank you!

If we have one generic chromedash-input type, and a subclass for each subtype, that might be a good design.  Or it might add more overhead than we want.

A single type, with attributes that modify its behavior, can be equivalent to subtyping.  That's what the current "input" element does with its "type" attribute, and the "pattern" attribute (like each of the other attributes) further modifies the behavior of the type.    So it is a question of what the most expedient packaging of behaviors might be.

static/elements/chromedash-form-field.js
```diff
@@ -65,6 +81,47 @@ export class ChromedashFormField extends LitElement {
           )}
         </sl-select>
       `;
+    } else if (type === 'input') {
+      let inputType = 'text';
+      let title = '';
+      let placeholder;
+      let pattern;
+      switch (fieldProps.input_type) {
```
jrobbins, line 89: I would like every switch statement to have a default clause that logs some kind of an error for unexpected data.  The means you would need to make the `'text'` case explicit as a case in the switch too.   Having a default that flags an error would avoid a lot of time debugging a typo in the spec file.

jrobbins (changes requested):

static/elements/chromedash-form-field.js
jrobbins: I suppose as an alternative, we could write a unit test  to validate the spec file.

kevinshen56714: @jrobbins I made some changes, please review it again.

As you suggested, I added the `'text'` case as well as a default clause that will output an error if the type doesn't match of any listed. I think rather than writing a unit test now, I'd prefer to add it when a specific `<chromedash-input>` component is created.

In addition, I also made a slight change relevant to your recent PR #2143. I found that `value=field.initial` when value=None in guideforms.py does not work properly. I found it because the owners field in guide/new page should have a default value of user.email. I had to change `value=field.initial` to `value=bf.initial` to make it works.

jrobbins (approved):
Awesome.  Thanks for noticing the problem with owner initial value.

PR #2147, opened 2022-08-19
Convert guide/verify_accuracy page into a component

kevinshen56714:
This PR converts content in guide/verify_accuracy.html into a component. The changes include:

1. Created chromedash-guide-verify-accuracy-page.js and chromedash-guide-verify-accuracy-page_test.js.
2. Slightly modified the template data for verify_accuracy.html in guide.py so that the forms can be passed into <chromedash-guide-verify-accuracy-page> as string and then get JSON.parse() when rendering.
3. The guide/verify_accuracy.html is now a single component file.
4. Added loading skeletons for the page.

This component is very similar to the `<chromedash-guide-editall-page>` component. I assume that we will have separate form fields specified in these page components once the form migration is done, thus making these two components more different in the future. But if not, we can consider merging `<chromedash-guide-editall-page>` and `<chromedash-guide-verify-accuracy-page>`. Please let me know your thoughts!

### DIFF
--- a/static/elements/chromedash-guide-editall-page.js
+++ b/static/elements/chromedash-guide-editall-page.js
@@ -135,6 +135,7 @@ export class ChromedashGuideEditallPage extends LitElement {
     return html`
       <form name="feature_form" method="POST" action="/guide/editall/${this.featureId}">
         <input type="hidden" name="token">
+        <input type="hidden" name"form_fields" value="edit_all">
         ${JSON.parse(this.flatForms).map(([sectionName, flatForm]) => html`
           <h3>${sectionName}</h3>
           <section class="flat_form">
--- a/static/elements/chromedash-guide-editall-page.js
+++ b/static/elements/chromedash-guide-editall-page.js
@@ -93,6 +93,15 @@ export class ChromedashGuideEditallPage extends LitElement {
     window.location.href = `/guide/edit/${this.featureId}`;
   }
 
+  // get a comma-spearated list of field names
+  getFormFields() {
+    let fields = '';
+    JSON.parse(this.flatForms).map((form) => {
+      fields += form[2].join();
+    });
+    return fields;
+  }
+
   renderSkeletons() {
     return html`
       <h3><sl-skeleton effect="sheen"></sl-skeleton></h3>
@@ -135,7 +144,7 @@ export class ChromedashGuideEditallPage extends LitElement {
     return html`
       <form name="feature_form" method="POST" action="/guide/editall/${this.featureId}">
         <input type="hidden" name="token">
-        <input type="hidden" name"form_fields" value="edit_all">
+        <input type="hidden" name="form_fields" value=${this.getFormFields()}>
         ${JSON.parse(this.flatForms).map(([sectionName, flatForm]) => html`
           <h3>${sectionName}</h3>
           <section class="flat_form">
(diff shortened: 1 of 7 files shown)


templates/guide/verify_accuracy.html
```diff
@@ -1,54 +1,9 @@
...
-      <iron-icon icon="chromestatus:arrow-back"></iron-icon>
-      Verify feature data for {{ feature.name }}
-    </a>
-  </h2>
-</div>
-{% endblock %}
-
 {% block content %}
-<form name="feature_form" method="POST" action="{{current_path}}">
-  <input type="hidden" name="token" value="{{xsrf_token}}">
```
jrobbins, line 21: I just realized that there is a defect in this existing template.
We need `<input type="hidden" value="">`
It's OK if value is an empty string.  It is used to avoid clearing out other boolean fields.

jrobbins (changes requested):

templates/guide/verify_accuracy.html
kevinshen56714: I see. Should I change this (add `value=""` to the hidden token field) for all the other converted page? I have removed all them because I thought it doesn't matter and the value would be added upon form submission.

jrobbins: Correction: it cannot be empty.  It should be nonempty, but the exact value does not matter, such as:
`<input type="hidden" name="form_fields" value="accurate_as_of">`

jrobbins: The new feature page does not need form_fields because (a) there are no checkboxes on that form and (b) there could not be any True boolean fields that would be inappropriately cleared since the feature is new.

The flat form fields actually does need form_fields because even though the UI says "edit all fields", there may be some fields that are not included in the flat forms.  Any boolean fields that is not part of the flat forms would be cleared in the feature, which would be a surprise to the user since that checkbox was not even offered to them.

form_fields will definitely be needed on the stage editing form.

jrobbins: For the edit_all page, it looks like the existing django template was already missing the form_fields hidden field.  So, you did not introduce a new defect in either that page or this one.  But, this is an opportunity to fix it here for the accuracy page.

kevinshen56714: I see. I have added the hidden name="form_fields" input to the forms in verify_accuracy page and edit_all page. This should fix the defect in both pages.

static/elements/chromedash-guide-editall-page.js
```diff
@@ -135,6 +135,7 @@ export class ChromedashGuideEditallPage extends LitElement {
     return html`
       <form name="feature_form" method="POST" action="/guide/editall/${this.featureId}">
         <input type="hidden" name="token">
+        <input type="hidden" name"form_fields" value="edit_all">
```
jrobbins, line 138: How is this supposed to work?   I was expecting the value to be a comma-separated list of field names.  I suppose you could have `touched()` interpret `"edit_all"` as a special value, but I don't see code for that.

kevinshen56714: My bad, I had some misunderstanding of the code. I have made some changes, please review it again.

I have fixed it by passing the form fields along with the forms from the backend and creating a getFormFields() method to assemble a list of comma-separated list of field names for both verify_accuracy and edit_all pages. The unit tests are also modified to check if the form field values are correct.

This getFormFields() method may seem trivial but in the future it can be renamed as getEditedFormFields() and check what fields have been changed.

jrobbins (approved):
Perfect.  This is what I had in mind.  Thanks for fixing this existing defect.

PR #2149, opened 2022-08-19
Add extra help for feature name and summary.

dlaliberte:
This PR adds the extra_help attribute for a few of the feature form fields based on the [EditingHelp](https://github.com/GoogleChrome/chromium-dashboard/wiki/EditingHelp#feature-name) document.

Note that, for help_text and extra_help, the current style which attempts to remove the margin-top and margin-bottom for the first and last child of those text blocks will fail to do what I expected in the case that there is text not enclosed in an element.  So unless we can find an alternative way to remove the extra margin, we have to unsure that text with multiple blocks should wrap each block with `<p>...</p>` or whatever.

### DIFF
--- a/static/elements/form-field-specs.js
+++ b/static/elements/form-field-specs.js
@@ -28,14 +28,24 @@ export const ALL_FIELDS = {
   'name': {
     label: 'Feature name',
     help_text: html`
-        <p>Capitalize only the first letter and the beginnings of proper nouns.</p>
-        <a target="_blank"
-            href="https://github.com/GoogleChrome/chromium-dashboard/wiki/EditingHelp#feature-name">
-          Learn more</a>.
-        <a target="_blank"
-            href="https://github.com/GoogleChrome/chromium-dashboard/wiki/EditingHelp#feature-name-example">
-          Example</a>.
-      `,
+        <p>Capitalize only the first letter and the beginnings of proper nouns.</p>`,
+    extra_help: html`
+    <p>
+    Each feature should have a unique name that is written as a noun phrase.
+    </p>
+    <ul>
+      <li>Capitalize only the first letter and the beginnings of proper nouns.</li>
+      <li>Avoid using verbs such as "add", "enhance", "deprecate", or "delete". Instead, simply name the feature itself and use the feature type and stage fields to indicate the intent of change.</li>
+      <li>Do not include markup or markdown.</li>
+      <li>Write keywords and identifiers as they would appear to a web developer, not as they are in source code. For example, a method implemented as NewInterface#dostuff would be written as in JavaScript: NewInterface.doStuff().</li>
+    </ul>
+
+    <h4>Examples</h4>
+    <ul>
+      <li>Conversion Measurement API</li>
+      <li>CSS Flexbox: New intrinsic size algorithm</li>
+      <li>Permissions-Policy header</li>
+    </ul>`,
   },
 
   'summary': {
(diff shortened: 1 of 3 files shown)


dlaliberte:
Here is what it looks like with these two extra_help blocks displayed:

https://screenshot.googleplex.com/4hrvGVtGXoRQWgY

static/elements/form-field-specs.js
```diff
@@ -28,14 +28,24 @@ export const ALL_FIELDS = {
...
-      `,
+        <p>Capitalize only the first letter and the beginnings of proper nouns.</p>`,
+    extra_help: html`
+    <p>
+    Each feature should have a unique name that is written as a noun phrase.
+    </p>
+    <ul>
+      <li>Capitalize only the first letter and the beginnings of proper nouns.</li>
+      <li>Avoid using verbs such as "add", "enhance", "deprecate", or "delete". Instead, simply name the feature itself and use the feature type and stage fields to indicate the intent of change.</li>
+      <li>Do not include markup or markdown.</li>
```
jpmedley, line 39:
```suggestion
      <li>Do not include markup or markdown because they will not be rendered.</li>
```

static/elements/form-field-specs.js
```diff
@@ -28,14 +28,24 @@ export const ALL_FIELDS = {
...
+      <li>Write keywords and identifiers as they would appear to a web developer, not as they are in source code. For example, a method implemented as NewInterface#dostuff would be written as in JavaScript: NewInterface.doStuff().</li>
+    </ul>
+
+    <h4>Examples</h4>
+    <ul>
+      <li>Conversion Measurement API</li>
+      <li>CSS Flexbox: New intrinsic size algorithm</li>
```
jpmedley, line 46:
```suggestion
      <li>CSS Flexbox: intrinsic size algorithm</li>
```

static/elements/form-field-specs.js
```diff
@@ -49,14 +59,32 @@ export const ALL_FIELDS = {
...
-            href="https://github.com/GoogleChrome/chromium-dashboard/wiki/EditingHelp#example-1">
-          Example</a>.
       `,
+    extra_help: html`
+    <p>
+    Provide a one sentence description followed by one or two lines explaining how this feature works and how it helps web developers.
+    </p>
+
+    <p>
+    Note: This text communicates with more than just the rest of Chromium development. It's the part most visible to external readers and is used as the basis for the text in the beta release posts.
```
jpmedley, line 69:
```suggestion
    Note: This text communicates with more than just the rest of Chromium development. It's the part most visible to external readers and is used in the beta release announcement, enterprise release notes, and other commuinications.
```

static/elements/form-field-specs.js
```diff
@@ -49,14 +59,32 @@ export const ALL_FIELDS = {
...
+    </p>
+
+    <ul>
+      <li>Write from a web developer's point of view, not a browser developer's</li>
+      <li>Do not use markup or markdown.</li>
```
jpmedley, line 74:
```suggestion
      <li>Do not use markup or markdown because they will not be rendered.</li>
```

static/elements/form-field-specs.js
```diff
@@ -49,14 +59,32 @@ export const ALL_FIELDS = {
...
+      <li>Do not use hard or soft returns.</li>
```
jpmedley, line 75:
```suggestion
      <li>Do not use hard or soft returns  because they will not be rendered.</li>
```

jpmedley (changes requested):
I went ahead and tweaked a few things, so I don't need to remember to fix them later.

static/elements/form-field-specs.js
dlaliberte: done

static/elements/form-field-specs.js
dlaliberte: done

static/elements/form-field-specs.js
dlaliberte: done

static/elements/form-field-specs.js
dlaliberte: done

static/elements/form-field-specs.js
dlaliberte: done

jpmedley (approved):

jpmedley:
Does @jrobbins need to approve this before we merge?

jrobbins (approved):

Issue #2138, opened 2022-08-16
Comment log improvements: deletion and relative date


jrobbins:
**Is your feature request related to a problem? Please describe.**
Sometimes people post a comment and then wish to delete it.  Also, it is useful to see how old a comment is in relative terms (e.g., 'two hours ago', 'yesterday', `6 days go`) when the comment is recent.

**Describe the solution you'd like**
Each comment shown in the comment log in the approvals dialog box should have a "..." menu with a "Delete comment" item.
Only the comment author or site admin can delete a comment, so the "..." menu should only appear in those cases.  Our datastore entities have a field to record who deleted the comment.  The content, author and deleted_by of a deleted comment should not be in the API responses given to other users, otherwise people could snoop.  Once the comment has been deleted, it should be displayed as some kind of indicator like "[Deleted comment]" and here should be an option to undelete.  A deleted comment can only be undeleted by the same user who deleted it or a site admin. 

Each comment shown in the comment log in the approvals dialog box should have a relative time phrase

**Additional context**
I believe there is an NPM package that can do relative dates, otherwise look at how it is done in Monorail.

jrobbins:
Location of relative date and menu:
![image](https://user-images.githubusercontent.com/17365/184979008-9a855d1e-57f5-4951-8895-19747d580e45.png)

Example from Monorail:
![image](https://user-images.githubusercontent.com/17365/184979062-04bcc57c-ec19-49b5-9ffc-d2385e28c6d8.png)

PR #2161, opened 2022-08-23
Add relative time since comment creation to comment log

DanielRyanSmith:
Part of #2138 

Add relative date information to comments similar to Monorail's implementation.

Adjusted to mimic [the monorail logic](https://source.chromium.org/chromium/infra/infra/+/main:appengine/monorail/framework/timestr.py;l=94).

Screen Shot 2022-08-22 at 4 57 10 PM


jrobbins (approved):
I like it.

However, I realize now that we could reuse a Shoelace component.  Before merging, consider pros and cons of using this
https://shoelace.style/components/relative-time

DanielRyanSmith:
> I realize now that we could reuse a Shoelace component

Great idea! Changed to use the component.

PR #2165, opened 2022-08-24
Initial version of new schema model classes.

jrobbins:
This PR adds all the new schema model classes.  They are not used yet, but having the code in place will help us write the copier functions and double-write logic.

### DIFF
--- a/internals/core_models.py
+++ b/internals/core_models.py
@@ -954,3 +954,206 @@ class Feature(DictModel):
   anticipated_spec_changes = ndb.StringProperty()
 
   finch_url = ndb.StringProperty()
+
+
+# Note: This class is not used yet.
+class FeatureEntry(ndb.Model):  # Copy from Feature
+  """This is the main representation of a feature that we are tracking."""
+
+  # Metadata: Creation and updates.
+  created = ndb.DateTimeProperty(auto_now_add=True)
+  updated = ndb.DateTimeProperty()
+  accurate_as_of = ndb.DateTimeProperty()
+  creator = ndb.StringProperty()
+  updater = ndb.StringProperty()
+
+  # Metadata: Access controls
+  owners = ndb.StringProperty(repeated=True)  # copy from owner
+  editors = ndb.StringProperty(repeated=True)
+  unlisted = ndb.BooleanProperty(default=False)
+  deleted = ndb.BooleanProperty(default=False)
+
+  # Descriptive info.
+  name = ndb.StringProperty(required=True)
+  summary = ndb.TextProperty(required=True)
+  category = ndb.IntegerProperty(required=True)
+  blink_components = ndb.StringProperty(repeated=True)
+  star_count = ndb.IntegerProperty(default=0)
+  search_tags = ndb.StringProperty(repeated=True)
+  feature_notes = ndb.TextProperty()  # copy from comments
+
+  # Metadata: Process information
+  feature_type = ndb.IntegerProperty(default=FEATURE_TYPE_INCUBATE_ID)
+  intent_stage = ndb.IntegerProperty(default=0)
+  bug_url = ndb.StringProperty()
+  launch_bug_url = ndb.StringProperty()
+
+  # Implementation in Chrome
+  impl_status_chrome = ndb.IntegerProperty(required=True)
+  flag_name = ndb.StringProperty()
+  ongoing_constraints = ndb.TextProperty()
+
+  # Gate: Adoption
+  motivation = ndb.TextProperty()
+  devtrial_instructions = ndb.TextProperty()
+  activation_risks = ndb.TextProperty()
+  measurement = ndb.TextProperty()
+
+  # Gate: Standardization & Interop
+  initial_public_proposal_url = ndb.StringProperty()
+  explainer_links = ndb.StringProperty(repeated=True)
+  requires_embedder_support = ndb.BooleanProperty(default=False)
+  standard_maturity = ndb.IntegerProperty(required=True, default=UNSET_STD)
+  spec_link = ndb.StringProperty()
+  api_spec = ndb.BooleanProperty(default=False)
+  spec_mentors = ndb.StringProperty(repeated=True)
+  interop_compat_risks = ndb.TextProperty()
+  prefixed = ndb.BooleanProperty()
+  all_platforms = ndb.BooleanProperty()
+  all_platforms_descr = ndb.TextProperty()
+  tag_review = ndb.StringProperty()
+  tag_review_status = ndb.IntegerProperty(default=REVIEW_PENDING)
+  non_oss_deps = ndb.TextProperty()
+  anticipated_spec_changes = ndb.TextProperty()
+
+  ff_views = ndb.IntegerProperty(required=True, default=NO_PUBLIC_SIGNALS)
+  safari_views = ndb.IntegerProperty(required=True, default=NO_PUBLIC_SIGNALS)
+  web_dev_views = ndb.IntegerProperty(required=True)
+  ff_views_link = ndb.StringProperty()
+  safari_views_link = ndb.StringProperty()
+  web_dev_views_link = ndb.StringProperty()
+  ff_views_notes = ndb.StringProperty()
+  safari_views_notes = ndb.TextProperty()
+  web_dev_views_notes = ndb.TextProperty()
+  other_views_notes = ndb.TextProperty()
+
+  # Gate: Security & Privacy
+  security_risks = ndb.TextProperty()
+  security_review_status = ndb.IntegerProperty(default=REVIEW_PENDING)
+  privacy_review_status = ndb.IntegerProperty(default=REVIEW_PENDING)
+
+  # Gate: Testing / Regressions
+  ergonomics_risks = ndb.TextProperty()
+  wpt = ndb.BooleanProperty()
+  wpt_descr = ndb.TextProperty()
+  webview_risks = ndb.TextProperty()
+
+  # Gate: Devrel & Docs
+  devrel = ndb.StringProperty(repeated=True)
+  debuggability = ndb.TextProperty()
+  doc_links = ndb.StringProperty(repeated=True)
+  sample_links = ndb.StringProperty(repeated=True)
+
+  DEFAULT_CACHE_KEY = 'FeatureEntries'
+
+  def __init__(self, *args, **kwargs):
+    # Initialise Feature.blink_components with a default value.  If
+    # name is present in kwargs then it would mean constructor is
+    # being called for creating a new feature rather than for fetching
+    # an existing feature.
+    if 'name' in kwargs:
+      if 'blink_components' not in kwargs:
+        kwargs['blink_components'] = [BlinkComponent.DEFAULT_COMPONENT]
+
+    super(FeatureEntry, self).__init__(*args, **kwargs)
+
+
+  @classmethod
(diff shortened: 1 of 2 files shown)


internals/review_models.py
```diff
@@ -188,3 +188,131 @@ def get_raw_owner_file(cls, url):
...
+
+  # Note: clear_request() moved to Gate
+
+
+# Note: This class is not used yet.
+class Amendment(ndb.Model):
+  """ReviewComments can log changes to other fields of the feature."""
+  field_name = ndb.StringProperty()  # from QUERIABLE_FIELDS
+  old_value = ndb.TextProperty()
+  new_value = ndb.TextProperty()
```
DanielRyanSmith, line 293: Is there any merit in noting the user who made an amendment?

DanielRyanSmith (approved):

internals/review_models.py
DanielRyanSmith: Will the author always be the one making an amendment? I would think so, and if not, it would be a site admin. I imagine the case is so rare that it's just not worth keeping track of. Thanks for answering my question!

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  category = ndb.IntegerProperty(required=True)
+  blink_components = ndb.StringProperty(repeated=True)
+  star_count = ndb.IntegerProperty(default=0)
+  search_tags = ndb.StringProperty(repeated=True)
+  feature_notes = ndb.TextProperty()  # copy from comments
+
+  # Metadata: Process information
+  feature_type = ndb.IntegerProperty(default=FEATURE_TYPE_INCUBATE_ID)
+  intent_stage = ndb.IntegerProperty(default=0)
+  bug_url = ndb.StringProperty()
```
jpmedley, line 988: Should this be `tracking_bug_url`? That's what it's called in the interface.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  launch_bug_url = ndb.StringProperty()
+
+  # Implementation in Chrome
+  impl_status_chrome = ndb.IntegerProperty(required=True)
+  flag_name = ndb.StringProperty()
```
jpmedley, line 993: This should be developer flag name or something similar. I have to frequently tell people feature owners that this does NOT apply to finch flags.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  ongoing_constraints = ndb.TextProperty()
+
+  # Gate: Adoption
+  motivation = ndb.TextProperty()
```
jpmedley, line 997: What's your implementation timeline? I'm still trying to work out whether this should be merged wit the summary or renamed in some way.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  intent_stage = ndb.IntegerProperty(default=0)
+  bug_url = ndb.StringProperty()
+  launch_bug_url = ndb.StringProperty()
+
+  # Implementation in Chrome
+  impl_status_chrome = ndb.IntegerProperty(required=True)
+  flag_name = ndb.StringProperty()
+  ongoing_constraints = ndb.TextProperty()
+
+  # Gate: Adoption
```
jpmedley, line 996: 'Gate'

Is this intended to imply the approvals-centered approach you've been touting? If so, then the summary field is also a gate.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  explainer_links = ndb.StringProperty(repeated=True)
+  requires_embedder_support = ndb.BooleanProperty(default=False)
+  standard_maturity = ndb.IntegerProperty(required=True, default=UNSET_STD)
+  spec_link = ndb.StringProperty()
+  api_spec = ndb.BooleanProperty(default=False)
+  spec_mentors = ndb.StringProperty(repeated=True)
+  interop_compat_risks = ndb.TextProperty()
+  prefixed = ndb.BooleanProperty()
+  all_platforms = ndb.BooleanProperty()
+  all_platforms_descr = ndb.TextProperty()
```
jpmedley, line 1013: The description, when it is present, is almost always something like 'Windows only' or 'mac and Android only'. 

What if `all_platforms` became `platforms` with an enum including 'all' (default), and `all_platforms_descr`  becomes `platform_notes`, which can take existing descriptions for backwards compatibility.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  tag_review = ndb.StringProperty()
+  tag_review_status = ndb.IntegerProperty(default=REVIEW_PENDING)
+  non_oss_deps = ndb.TextProperty()
```
jpmedley, line 1016: Here's [the only feature](https://chromestatus.com/feature/5759984304390144) where this has a value other than blank or 'none'. I can't believe there aren't more than that. 

We should start a document that lists who uses every field in the system to keep track of what's still needed and what isn't.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  anticipated_spec_changes = ndb.TextProperty()
```
jpmedley, line 1017: This is only used 5 times and of those only two contain explanatory text. The rest of links to GitHub issues.

internals/review_models.py
```diff
@@ -188,3 +188,131 @@ def get_raw_owner_file(cls, url):
...
+  REVIEW_STARTED = 3
+  NEED_INFO = 4
+  APPROVED = 5
+  NOT_APPROVED = 6
+  VOTE_VALUES = {
+      # Not used: NEEDS_REVIEW: 'needs_review',
+      NA: 'na',
+      REVIEW_REQUESTED: 'review_requested',
+      REVIEW_STARTED: 'review_started',
+      NEED_INFO: 'need_info',
```
jpmedley, line 226: If DevRel and product are going to approve summaries, we need a vote value called `NEED_REVISION`.

Or maybe a value called `NEED_CHANGE` could apply to both.

internals/review_models.py
```diff
@@ -188,3 +188,131 @@ def get_raw_owner_file(cls, url):
...
+      APPROVED: 'approved',
+      NOT_APPROVED: 'not_approved',
+  }
+
+  FINAL_STATES = [NA, APPROVED, NOT_APPROVED]
+
+  feature_id = ndb.IntegerProperty(required=True)
+  gate_id = ndb.IntegerProperty(required=True)
+  state = ndb.IntegerProperty(required=True)
```
jpmedley, line 235: What's 'state'?

internals/review_models.py
jpmedley: I think we're going to want to know who made a particular change auditing purposes. If I read this correctly, I can tell the changes made by a particular author, but I can't tell the author of a particular change.

jrobbins: I'm thinking that any time that a user edits a feature, the app will generate a set of Amendments just like the way that we generate diffs for the notifications now.  Rather than being used only for email, those Amendments would also be put into a ReviewComment.  That ReviewComment would have the author set to the email address of the person who made the change, and the content of the comment would be empty.

The ID of the ReviewComment entity could be passed to notifier.py to generate the notification email.  And, that same ReviewComment could be displayed somewhere on the feature detail page (regardless of gate ID), and in the approvals dialog box (only for the relevant gate ID).  

I'll write that up as an issue.

internals/core_models.py
jrobbins: I was thinking the same thing.   I'll make that change in a follow-up PR.

internals/core_models.py
jrobbins: I need to better understand the distinction.  I'll ask this question on webstatus@.

internals/review_models.py
jrobbins: Since our users will familiar with Monorail FLT and will also be using go/launch at the same time, we need to follow the conventions used in those tools.  Basically `NEED_INFO` means that something is needed from the feature owner, it could be info or a revision.

internals/review_models.py
jrobbins: State will be one of the enum values listed above.

internals/core_models.py
jrobbins: I started making that follow-up PR and I realized that a fair number of details in the code need to change for every field that we rename.  So, even though clarifying the source code is one of the goals of this effort, I think we need to be selective about how many things we rename.  bug_url is not as good as tracking_bug_url, but I think it can be clarified in the help text and a comment on the field name rather than changing this particular field name.

internals/review_models.py
jrobbins: I looked at go/launch again and they use the term "Needs Work", so we will go with that.  I think that is consistent with a reviewer asking for changes or information.  Monorail FLT uses "Needs info", however going forward our users will be concurrently using go/launch, so aligning with that tool is more important.  I am sending a separate small PR.

internals/core_models.py
jpmedley: Finch flags are controlled by Chrome exclusively. Having a Finch flag does NOT constitute a developer trial. developer trial flags are set on chrome://flags or via a command line. Beyond programming a default value, Chrome has NO control over how they're set. With Finch we can turn features on or off for all Chrome via in-house infrastructure. To change a dev trial default, we have to release a new version of Chrome.

internals/core_models.py
jpmedley: That's fair.

internals/core_models.py
jpmedley: Ping!

internals/review_models.py
jpmedley: 👍

internals/review_models.py
jpmedley: I think I'm getting a bit outside my swim lane here. I beg your indulgence.

'State' is not very domain specific. By that I mean that implementation status and process stage are both states. I asked the question thinking this was one of those and wondering which one. (If I had looked at line 231, I would have figured this out.)

internals/core_models.py
jrobbins: We might get to this part of the data migration in the next 3 or 4 weeks.  We can always make changes after that, but it would be better to make changes before if possible.

internals/core_models.py
```diff
@@ -954,3 +954,206 @@ def put(self, notify=True, **kwargs):
...
+  DEFAULT_CACHE_KEY = 'FeatureEntries'
+
+  def __init__(self, *args, **kwargs):
+    # Initialise Feature.blink_components with a default value.  If
+    # name is present in kwargs then it would mean constructor is
+    # being called for creating a new feature rather than for fetching
+    # an existing feature.
+    if 'name' in kwargs:
+      if 'blink_components' not in kwargs:
+        kwargs['blink_components'] = [BlinkComponent.DEFAULT_COMPONENT]
```
KyleJu, line 1056: BlinkComponent is defined in user_models. I see an undefined variable error on my lint

jrobbins: Thanks for catching this.
I sent a follow-up PR: https://github.com/GoogleChrome/chromium-dashboard/pull/new/20220831-fix-reference-to-DEFAULT_COMPONENT

PR #2166, opened 2022-08-24
Comment log improvements (delete and undelete comments)

### LINKED ISSUE #2138: Comment log improvements: deletion and relative date

DanielRyanSmith:
fixes #2138 

This change allows for deleting and undeleting comments as described in the issue.
Some changes were made for passing user info in django so that we can check whether the user is a site admin in the comment section.

---
### Screenshots
User's can see a `...` menu at the top right of comments if it is a comment they wrote or the user is a site admin.
Screen Shot 2022-08-24 at 10 16 41 AM

Clicking the menu will display an option to delete the comment.
Screen Shot 2022-08-24 at 10 16 50 AM

After selecting to delete the comment, the request will process and after receiving a response, the comment will be updated with text noting its deletion. The user will also now see the option to undelete the comment from the menu. **NOTE:** Only the user who deleted the comment or a site admin can undelete the comment. The comment author will not have the option to undelete the comment if the comment was deleted by a site admin.
Screen Shot 2022-08-24 at 10 17 08 AM

To users who cannot undelete the comment, the original content of the comment will not be displayed (and is not passed to the client).
Screen Shot 2022-08-24 at 10 17 29 AM


DanielRyanSmith:
@jrobbins I am not a CSS wizard, so if you see anything that doesn't have the correct look, feel, or implementation, let me know!

internals/review_models.py
```diff
@@ -137,7 +137,7 @@ class Comment(ndb.Model):
   """A review comment on a feature."""
   feature_id = ndb.IntegerProperty(required=True)
   field_id = ndb.IntegerProperty()  # The approval field_id, or general comment.
-  created = ndb.DateTimeProperty(auto_now=True)
+  created = ndb.DateTimeProperty()
```
jrobbins, line 140: I agree that auto_now was wrong.  Why not use auto_now_add=True?

api/comments_api.py
```diff
@@ -90,4 +102,19 @@ def do_post(self, feature_id, field_id=None):
     # Callers don't use the JSON response for this API call.
     return {'message': 'Done'}
 
-  # TODO(jrobbins): do_patch to soft-delete comments
+  def do_patch(self, feature_id):
```
jrobbins, line 105: Nice.  We are real REST programmers now :).

api/comments_api.py
```diff
@@ -90,4 +102,19 @@ def do_post(self, feature_id, field_id=None):
...
+
+    user = self.get_current_user(required=True)
+    if not permissions.can_admin_site(user) and user.email() != comment.author:
+      self.abort(403, msg='User does not have comment edit permissions')
+
+    is_undelete = self.get_param('isUndelete', required=True)
+    if is_undelete:
+      comment.deleted_by = None
+    else:
+      comment.deleted_by = self.get_current_user().email()
```
jrobbins, line 117: This could be user.email()

static/elements/chromedash-approvals-dialog.js
```diff
@@ -379,26 +392,92 @@ class ChromedashApprovalsDialog extends LitElement {
...
+    // by site admins or the user who deleted it.
+    if (comment.deleted_by) {
+      return this.user.is_admin || comment.deleted_by === this.user.email;
+    }
+    // If the comment is not deleted, site admins or the author can edit.
+    return this.user.email === comment.author || this.user.is_admin;
+  }
+
+  // Format a message to show on a deleted comment the user can edit.
+  formatCommentDeletedPreface(comment) {
```
jrobbins, line 407: I think that the conventional verb to use in these method names would be 'render'.  So, this would be `renderCommentDeletedPreface`.

api/comments_api.py
```diff
@@ -22,13 +23,21 @@
...
 
 
-def comment_to_json_dict(comment):
+def comment_to_json_dict(comment, user_email=None, show_deleted=False):
+  content = comment.content
+  # If the comment is deleted and the current user can't view it,
+  # don't send the comment content to the client.
+  if (comment.deleted_by is not None and
+      user_email != comment.deleted_by and not show_deleted):
+    content = "[Deleted]"
```
jrobbins, line 32: Thanks for including he screenshots, that makes it much easier to think about the UX in addition to reviewing the code.

I guess I didn't think of this before, but now I think that the last screenshot you posted (the case where a comment is deleted and the user lacks perms) should just not be in the API response at all.  That could be done before line 57 by filtering the list of comments before converting them to json_dicts.  The filtering could use the same condition that you use here.

jrobbins (changes requested):

api/comments_api.py
DanielRyanSmith: >:)

internals/review_models.py
DanielRyanSmith: Yes, that's what should be used - I was not aware of the difference, but now it's been changed

api/comments_api.py
DanielRyanSmith: Yes, missed this!

static/elements/chromedash-approvals-dialog.js
DanielRyanSmith: Name changed

api/comments_api.py
DanielRyanSmith: Good idea, I reverted the json_dicts function and added a filter beforehand!

api/comments_api.py
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime, timezone
```
jrobbins, line 16: Do you need timezone?

api/comments_api.py
```diff
@@ -39,10 +42,22 @@ class CommentsAPI(basehandlers.APIHandler):
   """Users may see the list of comments on one of the approvals of a feature,
    and add their own, if allowed."""
 
+  def _should_show_comment(self, comment, email, is_admin):
+    # If the comment is deleted and the current user can't view it,
```
jrobbins, line 46: This can be a docstring rather than a comment.  Ideally it would be one line, e.g.:
`"""Only send comments that the user can view."""`

static/elements/chromedash-approvals-dialog.js
```diff
@@ -36,10 +36,10 @@ const APPROVAL_DEFS = [
 
 let approvalDialogEl;
 
-export async function openApprovalsDialog(signedInUser, feature) {
+export async function openApprovalsDialog(user, feature) {
```
jrobbins, line 39: Thanks for making this consistent with some of our newer elements.

jrobbins (approved):
LGTM after considering a couple nits.

api/comments_api.py
DanielRyanSmith: I changed up the wording a bit that is hopefully sufficient.

`"""Check whether a comment should be visible to the user."""`

api/comments_api.py
DanielRyanSmith: No, nor `datetime`! It's a remnant from the way I was previously adding the value for `created`. It's removed now.

PR #2167, opened 2022-08-24
Autolinking

### LINKED ISSUE #2137: Autolinking of short links in comment and text fields
When users enter text into ChromeStatus, they often refer to other records in other tools or websites.  We should make those references into clickable links.

We would like to offer auto linking that is similar to what monorail offers:
* URLs
* Short links like `go/something`
* Bug or issue links like `b/12345`, `issue 1234`, `issue project:1234`, and `issue 1234#c9`
* Other numeric links like `omg/123`, `cl/123456`, `cr/123456`, and others supported by monorail.
* Git hashes

We can defer any API integration with any of those backend systems.  I.e., we do not need to check monorail to see if an issue has been closed to draw a cross-through, check viewing permissions, or show an issue summary on hover.

**Additional context**
We currently have a function `autolink()` in static/elements/utils.js.  It uses this library to link URLs:
https://www.npmjs.com/package/autolinker
 
We want to keep support for linking URLs, but also add support for the short-links that monorail supports.
Monorail's implementation is here:
https://chromium.googlesource.com/infra/infra/+/refs/heads/main/appengine/monorail/static_src/autolink.js

We can copy-and-modify monorail's auto linker instead of the autolink NPM package.

DanielRyanSmith:
Fixes #2137 

Copies over part of the [monorail autolinking implementation](https://chromium.googlesource.com/infra/infra/+/refs/heads/main/appengine/monorail/static_src/autolink.js) and refactors for use in WebStatus text entries.


static/elements/utils.js
```diff
@@ -2,18 +2,15 @@
 
 import {html} from 'lit';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
-import Autolinker from 'autolinker';
+import {markupAutolinks} from './autolink.js';
```
jrobbins, line 5: So, can we remove the `autolinker` NPM dependency from package.json now?

jrobbins (approved):

static/elements/utils.js
DanielRyanSmith: Yes, we should be able to remove this. Is it worth trying to generate a new `package-lock.json`? Or just simply remove from `package.json`?

DanielRyanSmith: I'm having a lot of trouble successfully updating package-lock here. This is the correct approach, but I'll handle this in a separate PR.

PR #2171, opened 2022-08-25
Generate milestone tables via f-strings instead of Django templates

kevinshen56714:
This PR is the first step towards using python strings and f-strings to generate email body html (instead of via Django templates). 

Specifically, the `estimated-milestones-table.html` is removed and instead generated through a new python function in `internals/html_templates.py`. A corresponding unit test is also created to practice the new html generation method. The generated html strings are now passed to other templates as a template data.


jrobbins (commented):
Again, sorry to have set you on the wrong path.  See my comment about using jinja2 on #2172.

PR #2172, opened 2022-08-25
Create new and updated feature emails via f-strings

kevinshen56714:
This PR is the second part of transforming the way we generate email html. Specifically, this gets rid of the need of Django templates - `templates/update-feature-email.html` and `templates/new-feature-email.html`. 

Two new functions in `internals/html_templates.py` are now used to generate the email body html via f-strings. New unit tests are also added to practice these functions. The old `format_email_body()` in notifier.py is no longer needed thus removed, and the notifier_test.py is changed accordingly.


jrobbins (commented):
Thanks for doing this.    However, I think I failed to do my research before assigning you this task.  I apologize for having you do work that won't be used.

The issue is that we need auto-escaping.  According to Google security guidelines, escaping expressions by calling `escape()` as needed is not good enough because it is too error-prone both in this initial implementation and in the future as changes are made.  Even if we get it right in this PR, there is a policy against it because there is a long, documented history of other projects having security problems because of inconsistent escaping.  Instead, we need to use a template library that will automatically escape everything.

I want to phase out django, including django templates, so that we use flask as our only server-side framework.   Flask includes jinja2 templates, which are similar to django templates.  So, maybe what we could do is convert the existing django template to jinja2 templates and change from django's `render_to_string(path, data)` to `flask.render_template(path, **data)`.

jrobbins:
Also, I realize that this was a bit of a side-track from your work on the SPA.  Thanks for exploring this approach, seeing it worked out is what allowed me to realize that it was not the best choice, which is itself a contribution.  It's OK with me if you get back to the SPA work and leave the django to jinja2 conversion for me to do later.

jrobbins (changes requested):
For some reason, GitHub thinks that reviews are optional on this PR.

kevinshen56714:
> Thanks for doing this. However, I think I failed to do my research before assigning you this task. I apologize for having you do work that won't be used.

No worries. I didn't really spend a lot of time on this.

> The issue is that we need auto-escaping. According to Google security guidelines, escaping expressions by calling `escape()` as needed is not good enough because it is too error-prone both in this initial implementation and in the future as changes are made. Even if we get it right in this PR, there is a policy against it because there is a long, documented history of other projects having security problems because of inconsistent escaping. Instead, we need to use a template library that will automatically escape everything.

Do you know any sources of the security guidelines? I would love to learn more (or maybe we can discuss next week in the meeting)
 
> I want to phase out django, including django templates, so that we use flask as our only server-side framework. Flask includes jinja2 templates, which are similar to django templates. So, maybe what we could do is convert the existing django template to jinja2 templates and change from django's `render_to_string(path, data)` to `flask.render_template(path, **data)`.

I will probably work on the SPA next. Should I close this and the previous PR #2171?

jrobbins:
> Do you know any sources of the security guidelines? I would love to learn more (or maybe we can discuss next week in the meeting)

Excellent question.  I was looking at an internal set of guidelines, but a quick google search  turned up this public facing document that we apparently use when outsourcing.
"One way to do this consistently throughout the application is to use an auto-escaping templating language (as opposed to manually escaping in code). A few examples for different programming languages are listed below."
https://partner-security.withgoogle.com/docs/webapp_requirements.html

The internal guidelines go further to say that context-sensitive escaping is needed, and there is no python templating system that does context-sensitive escaping.  So, most python web 1.0 apps cannot meet those guidelines.  Lit has context-sensitive escaping.   So, a python web 2.0 app that uses APIs and custom elements rather than server-side template (like what you have been helping us build) can meet the guidelines.    I would think that just always escaping quotes, angle brackets, and ampersands would be good enough, but apparently there is more to it than that.

Google also has a library that basically uses python objects to build a structure like a DOM on the server, and carefully validates every attribute and content string, and then that whole DOM tree is used to generate HTML rather than using a template at all.  I don't know if that is used very much because it seems pretty inconvenient, and it is not open source.   

My suggested django to jinja2 change should be basically a lateral move security-wise.  I feel that it is OK for this app, in part because the users who have write access (and can thus supply strings that would end up in these emails) are a fairly narrow set of people that we have some kind of known relationship with.  The f-string approach would be a step down security-wise because there some string could get through with no escaping at all.  While they probably won't hack us, they are very likely to include some markup in some of the field values.

> Yeah I'm glad that the PRs still have some values. I will probably work on the SPA next. Should I close this and the previous PR #2171?

Yes please, go ahead and close these two.

jrobbins:
> Lit has context-sensitive escaping.

In fact, that explains a lot of the strict rules of lit templates:
https://lit.dev/docs/templates/expressions/#expression-locations

One detail relevant to your work on textarea:
```<!-- BEWARE -->
<textarea>${content}</textarea>

<!-- OK -->
<textarea .value=${content}></textarea>```

PR #2173, opened 2022-08-26
Fix approval row author comparison.

jrobbins:
This should resolve a problem I found while testing the release on staging.

In your #2166, you upgraded the approvals dialog from using a property `this.signedInUser` to using a property `this.user`, but  some occurances were missed.   This resulted in a comparison to `undefined`, which caused the current user's previous approval values to render as read-only as if they were some other user's approval values.  Also, for the row where a user is offered the chance to enter their first own approval value, the user name was blank.


DanielRyanSmith (approved):
Not sure how i missed that! Thanks for the fix.

PR #2182, opened 2022-08-29
Fix passing user object from features.html to chromedash-featurelist.

jrobbins:
This fixes I problem that I did not catch in #2166.  The features.html template was changed to pass the `user` object to chromedash-featurelist.  However, that did not work because django will only expand strings and other simple datatypes when the name of a variable is used as a django template expression, it won't automatically convert a python dict to a JSON string.  So, I added a template variable to make that JSON string available.  From there, lit can accept a JSON string in an attribute and parse it into a JS object and assign that to a property.

### DIFF
--- a/static/elements/chromedash-roadmap-milestone-card.js
+++ b/static/elements/chromedash-roadmap-milestone-card.js
@@ -1,6 +1,6 @@
 import {LitElement, html, nothing} from 'lit';
 import {ROADMAP_MILESTONE_CARD_CSS} from
-  '../sass/elements/chromedash-roadmap-milestone-card-css.js';
+'../sass/elements/chromedash-roadmap-milestone-card-css.js';
 
 const REMOVED_STATUS = ['Removed'];
 const DEPRECATED_STATUS = ['Deprecated', 'No longer pursuing'];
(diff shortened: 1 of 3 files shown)


static/elements/chromedash-roadmap-milestone-card.js
```diff
@@ -1,6 +1,6 @@
 import {LitElement, html, nothing} from 'lit';
 import {ROADMAP_MILESTONE_CARD_CSS} from
-  '../sass/elements/chromedash-roadmap-milestone-card-css.js';
+'../sass/elements/chromedash-roadmap-milestone-card-css.js';
```
DanielRyanSmith, line 3: This auto-formatting continues to happen to me as well. Maybe it's fine to just keep this change?

DanielRyanSmith (approved):
Thanks for this. I'll try and vet my changes more thoroughly in the future. It seems like this one could have been found by me earlier.

static/elements/chromedash-roadmap-milestone-card.js
jrobbins: Yeah, I'll merge it and then it should not cause diffs in the future.

PR #2184, opened 2022-08-30
Rename ReviewComment to Activity.

jrobbins:
Rename ReviewComment to Activity because a lot of these items will just be field changes without any comment content and not necessarily part of a review, so I am thinking of labeling that part of the UI "Activity log".

### DIFF
--- a/internals/review_models.py
+++ b/internals/review_models.py
@@ -287,7 +287,7 @@ class Vote(ndb.Model):  # copy from Approval
 
 # Note: This class is not used yet.
 class Amendment(ndb.Model):
-  """ReviewComments can log changes to other fields of the feature."""
+  """Activity log entries can record changes to fields."""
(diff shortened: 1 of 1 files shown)


internals/review_models.py
```diff
@@ -296,8 +296,8 @@ class Amendment(ndb.Model):
 # Note: This class is not used yet.
 # TODO(jrobbins): Decide on either copying to this new class or adding
 # and removing fields from the existing Comment class.
-class ReviewComment(ndb.Model):  # copy from Comment
-  """A review comment on a gate or a general comment on a feature."""
+class Activity(ndb.Model):  # copy from Comment
+  """An activity log entry (comment + amentments) on a gate or feature."""
```
DanielRyanSmith, line 300:
```suggestion
  """An activity log entry (comment + amendments) on a gate or feature."""
```

DanielRyanSmith (approved):

PR #2185, opened 2022-08-30
Refactor comment display into a new element.

jrobbins:
There is already a fair amount of code for displaying comments, and we might be doing more with them in the future.  Also, review comments will appear in this dialog, but we will also want to show general comments directly on the feature list page in the future.

In this PR:
* Refactor methods relevant to displaying and delete/undelete of comments into a new Lit element.

### DIFF
--- a/static/elements/chromedash-activity-log_test.js
+++ b/static/elements/chromedash-activity-log_test.js
@@ -1,0 +1,155 @@
+import {html} from 'lit';
+import {assert, fixture} from '@open-wc/testing';
+import {ChromedashActivityLog} from './chromedash-activity-log';
+import '../js-src/cs-client';
+import sinon from 'sinon';
+
+describe('chromedash-activity-log', () => {
+  const nonAdminUser = {
+    can_approve: false,
+    can_create_feature: true,
+    can_edit_all: true,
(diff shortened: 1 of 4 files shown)


jrobbins:
I'll request review after adding unit tests.

jrobbins:
Since I expect to reuse this component on the feature detail page to show an activity log, I added support for viewing by a signed-out user by just adding a case to the logic to check if the edit menu should be offered.

static/elements/chromedash-activity-log_test.js
```diff
@@ -0,0 +1,155 @@
...
+    window.csClient = new ChromeStatusClient('fake_token', 1);
+    sinon.stub(window.csClient, 'deleteComment');
+    window.csClient.deleteComment.returns(Promise.resolve({message: 'Done'}));
+
+    const doomedComment = {
+      comment_id: 4,
+      author: 'non-admin@google.com',
+      created: '2022-08-30 12:34:45.567',
+      content: 'something off the cuff',
+    };
```
DanielRyanSmith, line 109: Good name choice 🙂

DanielRyanSmith (approved):

PR #2187, opened 2022-08-31
Populate blink components form field options via an API

kevinshen56714:
This PR migrates the "Blink component" select form field definition to the frontend but with the choice options fetched via API. 

Specifically, in chromedash-form-field.js, if the name of the field is "blink_components", a list of blink components will be fetched via the new BlinkComponentsAPI (URI: /blinkcomponents) and then the select field options get populated. During fetching, the select field is disabled.

### DIFF
--- a/static/elements/chromedash-form-field.js
+++ b/static/elements/chromedash-form-field.js
@@ -49,10 +49,10 @@ export class ChromedashFormField extends LitElement {
     if (this.name !== 'blink_components') return;
 
     // get the choice values from API when the field is blink component select field
-    this.loading =true;
+    this.loading = true;
     window.csClient.getBlinkComponents().then((componentChoices) => {
       this.componentChoices = componentChoices;
-      this.loading =false;
+      this.loading = false;
     }).catch(() => {
       showToastMessage('Some errors occurred. Please refresh the page or try again later.');
     });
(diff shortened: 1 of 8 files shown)


dlaliberte:
Seems fine to me, though I wonder if this pattern of asynchronous, dynamic loading of values could be abstracted out and reused more simply.

dlaliberte (approved):

static/elements/chromedash-form-field.js
```diff
@@ -27,9 +27,37 @@ export class ChromedashFormField extends LitElement {
...
+    this.loading = false;
+    this.componentChoices = {};
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    if (this.name !== 'blink_components') return;
+
+    // get the choice values from API when the field is blink component select field
+    this.loading =true;
```
jrobbins, line 52: nit: space after assignment

static/elements/chromedash-form-field.js
```diff
@@ -27,9 +27,37 @@ export class ChromedashFormField extends LitElement {
...
+    window.csClient.getBlinkComponents().then((componentChoices) => {
+      this.componentChoices = componentChoices;
+      this.loading =false;
```
jrobbins, line 55: same

jrobbins (approved):
LGTM after considering some nits.

static/elements/chromedash-form-field.js
kevinshen56714: Fixed!

static/elements/chromedash-form-field.js
kevinshen56714: Fixed!

Issue #2186, opened 2022-08-30
Relative time not displayed in Safari and FF


jrobbins:
**Describe the bug**
The relative dates on comments are only showing up in Chrome (and probably Edge).  Nothing shows in Safari or FF.

**To Reproduce**
1. Go to approvals dialog
2. View a comment relative time

**Expected behavior**
Works in all browsers

The root cause seems to be that the date format used in the call to `new Date()` is one that Chrome supports but Safari and FF do not.  Maybe we need a "Z" in it or something.

jrobbins:
Correction: We need a "T" in the middle.  The "Z" at the end is optional.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

PR #2193, opened 2022-08-31
Show relative timestamp on Safari and Firefox

DanielRyanSmith:
fixes #2185

### DIFF
--- a/static/elements/chromedash-approvals-dialog.js
+++ b/static/elements/chromedash-approvals-dialog.js
@@ -455,7 +455,7 @@ class ChromedashApprovalsDialog extends LitElement {
 
   // Display how long ago the comment was created compared to now.
   formatCommentRelativeDate(comment) {
-    const commentDate = new Date(`${comment.created} UTC`);
+    const commentDate = new Date(`${comment.created}Z`);
     if (isNaN(commentDate)) {
       return nothing;
     }


DanielRyanSmith:
Whoops, this change needs to be re-added after the comment refactor.

static/elements/chromedash-activity-log.js
```diff
@@ -128,7 +128,7 @@ export class ChromedashActivityLog extends LitElement {
 
   // Display how long ago the comment was created compared to now.
   formatCommentRelativeDate(comment) {
-    const commentDate = new Date(`${comment.created} UTC`);
+    const commentDate = new Date(`${comment.created}Z`);
```
jrobbins, line 131: Does that actually work?
When I look at the value of a comment's created field in devtools, I see a string like
 "2022-01-07 20:21:06.334990"
but the format expected is this
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date
I was wrong when I said add a "Z", that is optional.
But, it does require a "T" in the middle.

jrobbins: If you write a unit test for this, our `npm run wettest` target should execute the tests on both chrome and firefox, so we'll know for sure if it works.

jrobbins (changes requested):

jrobbins:
> fixes #2185

 I think this PR addresses issue #2186

static/elements/chromedash-activity-log.js
DanielRyanSmith: I wasn't sure if this change would work, but it does fix the issue on Chrome and Firefox. However, I went back and checked after your comment and it does not fix the issue on Safari, which I should have checked (I saw Firefox and Chrome working and assumed the issue was fixed).

Replacing the space with a "T" does fix the issue in all browsers as far as I can tell, and I'll write a test as well.

static/elements/chromedash-activity-log.js
```diff
@@ -128,7 +128,11 @@ export class ChromedashActivityLog extends LitElement {
 
   // Display how long ago the comment was created compared to now.
   formatCommentRelativeDate(comment) {
-    const commentDate = new Date(`${comment.created} UTC`);
+    // format date to "YYYY-MM-DDTHH:mm:ss.sssZ" to represent UTC.
+    if (comment.created) {
+      comment.created = comment.created.replace(' ', 'T');
```
jrobbins, line 133: It seems odd to modify the passed in object.  I think there might be a case where the value with the "T" in it would subsequently get displayed to the user.  It would be better to use a local variable.

jrobbins (changes requested):

static/elements/chromedash-activity-log.js
DanielRyanSmith: I agree with this. updated!

jrobbins (approved):

Issue #2135, opened 2022-08-16
Phase out core_enums.STANDARD_MATURITY_BACKFILL


jrobbins:
**Describe the bug**
When we added the `standard_maturity` field, we did not migrate values from the old `standardization` field.  That work still needs to be done so that we can phase out some python code and drop the `standardization` field when we migrate to using the new feature models.

**Plan**
* Write a cron task to copy data from the `standardization` field to the `standard_maturity` field using `core_enums.STANDARD_MATURITY_BACKFILL` as the value mapping.
* Deploy and run that task on staging and prod
* Delete `STANDARD_MATURITY_BACKFILL` and the logic that references it from `format_for_template()`
* Delete the copying code

DanielRyanSmith:
All existing features have had their `standard_maturity` fields backfilled 👍

PR #2195, opened 2022-08-31
Populate standard_maturity field with deprecated standardization field

DanielRyanSmith:
Part of #2135 

This is a temporary cron task that will check if any feature has a valid `standardization` field and does not have a valid `standard_maturity` field. If so, the `standard_maturity` field will be inferred based on a mapping of the `standardization` value.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -19,4 +19,4 @@ cron:
   schedule: every monday 09:00
 - description: Copy over deprecated standardization field
   url: /cron/write_standard_maturity
-  schedule: every day 09:00
+  schedule: 1 of jan 00:00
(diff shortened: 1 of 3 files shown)


cron.yaml
```diff
@@ -17,3 +17,6 @@ cron:
 - description: Send reminders to verify the accuracy of feature data.
   url: /cron/send_accuracy_notifications
   schedule: every monday 09:00
+- description: Copy over deprecated standardization field
+  url: /cron/write_standard_maturity
+  schedule: every day 09:00
```
jrobbins, line 22: Since we will run this manually by pressing the button on the Cloud Console, how about a date that will not match?  E.g.,
schedule: 1 of jan 00:00
See https://cloud.google.com/appengine/docs/standard/python3/scheduling-jobs-with-cron-yaml#formatting_the_schedule

jrobbins (changes requested):

cron.yaml
DanielRyanSmith: Good suggestion - updated 🙂

jrobbins (approved):

PR #2197, opened 2022-09-01
Refactor <chromedash-activity> out of activity log.

jrobbins:
We'll be doing more with the display of activities, so I simplified our current code a bit.  Instead of a single element for the entire activity log, there are now separate elements for each activity.  This modularizes the code a bit, and it also avoids the need to have ID numbers in the element ID names, and reduces a bunch of method names and arguments.


DanielRyanSmith (approved):
Might have to deal with some merge conflicts with the update to the Firefox/Safari relative date changes, but looks good nonetheless 🙂

PR #2198, opened 2022-09-01
Hack the layout of the checkbox to set impl status

jrobbins:
I believe that the old version of this more-or-less worked back when we were using shadowDOM for the form fields.   However, once we switched to lightDOM, the slots stopped working and we didn't notice the problem.  Also, there was a bug with a string  === a number which is solved by using == instead.  And, the field was accessed using the wrong expression.

### DIFF
--- a/static/elements/chromedash-guide-stage-page.js
+++ b/static/elements/chromedash-guide-stage-page.js
@@ -216,40 +216,44 @@ export class ChromedashGuideStagePage extends LitElement {
   }
 
   renderImplStatusFormSection() {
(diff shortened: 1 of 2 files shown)


jrobbins:
Before:
image
image


After:
image
image

kevinshen56714 (approved):

kevinshen56714:
Sorry, I didn't notice this at the time.

jrobbins:
> Sorry, I didn't notice this at the time.

No worries. I consider 24 hour turn-around on code reviews to be full speed.  24-48 is also common in practice.

PR #2199, opened 2022-09-01
Refactor input form fields by specifying attributes

kevinshen56714:
This PR refines the way we handle input form fields in chromedash-form-field.js. Previously (in #2145), we differentiate different input field behaviors by the `input_type` subclass and a switch statement. However, I think that was a bit hard-coded and not scalable.

In this PR, all input field types are given one of the four constant `attrs` objects. And the attrs are appended as the element attributes upon rendering by the new `updateAttributes()` method. I put this method as Lit's `ref` callback function, which gets called when the element is rendered.


jrobbins (approved):
This turned out well.
I didn't notice that ref() had the option for a callback.  That seems useful.

PR #2200, opened 2022-09-01
Remove field and error unsafeHTMLs and add radios field type in `<chromedash-form-field>`

kevinshen56714:
This PR finishes form field migration to the frontend and refines `<chromedash-form-field>`.  The changes include:
1. Added a `radios` field type and removed the hard coded fieldHTML for feature type radio group in chromedash-guide-new-page. Feature type descriptions are added to the form-field-enums.js.
2. Removed all the unsafeHTMLs in chromedash-form-field as we have migrated all fields to JS and no longer need Django-created htmls. This also includes removing the error messages (see the before and after snapshots below).
3. Not render `<tr>` for empty field labels. This makes the margin smaller for fields with no label (see the before and after snapshots below).
4. Moved all the submit buttons out of the slots of chromedash-form-table.
5. Removed unused properties (field, error, stage) in chromedash-form-field.
6. Updated the chromedash-form-field unit tests to practice html generation of all 5 field types.

Before | After
--- | --- 
 | 

Before | After
--- | --- 
 |

### DIFF
--- a/static/elements/chromedash-form-field.js
+++ b/static/elements/chromedash-form-field.js
@@ -1,21 +1,7 @@
 import {LitElement, html} from 'lit';
 import {ALL_FIELDS} from './form-field-specs';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
-
-
-/* Patterns from https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s01.html
- * Removing single quote ('), backtick (`), and pipe (|) since they are risky unless properly escaped everywhere.
- * Also removing ! and % because they have special meaning for some older email routing systems. */
-const USER_REGEX = '[A-Za-z0-9_#$&*+/=?{}~^.-]+';
-const DOMAIN_REGEX = String.raw`(([A-Za-z0-9-]+\.)+[A-Za-z]{2,6})`;
-
-const EMAIL_ADDRESS_REGEX = USER_REGEX + '@' + DOMAIN_REGEX;
-const EMAIL_ADDRESSES_REGEX = EMAIL_ADDRESS_REGEX + '([ ]*,[ ]*' + EMAIL_ADDRESS_REGEX + ')*';
-
-// Simple http URLs
-const PORTNUM_REGEX = '(:[0-9]+)?';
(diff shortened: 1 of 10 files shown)


dlaliberte:
Can you show what the radio button group looks like now?  If it's the same before and after, no worries.

kevinshen56714:
> Can you show what the radio button group looks like now? If it's the same before and after, no worries.

Yes, there's no visual difference before and after.

Before | After
--- | --- 
 |

static/elements/chromedash-form-field.js
```diff
@@ -68,14 +62,10 @@ export class ChromedashFormField extends LitElement {
     return this;
   }
 
-  render() {
-    const label = this.fieldProps.label;
-    const helpText = this.fieldProps.help_text || '';
-    const extraHelpText = this.fieldProps.extra_help || '';
+  renderField() {
```
jrobbins, line 65: Since the whole thing is the field, maybe this should be named `renderWidgets()`.

static/elements/chromedash-guide-editall-page.js
```diff
@@ -142,15 +141,9 @@ export class ChromedashGuideEditallPage extends LitElement {
...
-          <chromedash-form-table>
-            <chromedash-form-field>
-              <span slot="field">
-                <input class="button" type="submit" value="Submit">
-                <button id="cancel-button" type="reset"
-                  @click=${this.handleCancelClick}>Cancel</button>
-              </span>
-            </chromedash-form-field>
-          </chromedash-form-table>
+          <input class="button" type="submit" value="Submit">
```
jrobbins, line 144: In this or a follow-up PR, it would be good to convert these to sl-buttons.

jrobbins (approved):

static/elements/chromedash-form-field.js
kevinshen56714: Changed!

static/elements/chromedash-guide-editall-page.js
kevinshen56714: I'll do these conversions in a follow-up PR.

PR #2201, opened 2022-09-02
Start migrating form definitions and render guide/edit form without unsafeHTML

kevinshen56714:
This PR starts the migration of form definitions to the frontend, and specifically, the metadata form fields are migrated. This allows the form rendering of guide edit page to be based on API-fetched feature values and a form field constant instead of the unsafeHTML. The changes include:

1. Created `form-definition.js` which will contain all the form definition constants currently in guideforms.py.
2. Created a `format_for_edit` dict within core_models.Feature.render_for_template dict. This is needed to (1) separate view and edit values (for example, `d['feature_type']` is the feature type string, whereas `d['format_for_edit']['feature_type']` is the feature type enum number value) and (2) return post-processed values just for edit (for example, `d['editors']` is the editor list, whereas `d['format_for_edit']['editors']` is the comma-separated list string)
3. Removed the `unsafeHTML` in `<chromedash-guide-metadata>` and render the form based on form field constant defined in form-definition.js. The `setTimeout` in updated() can also be removed now that we don't have unsafeHTML.
4. Redesigned the way we register form submission handler. I found that awaiting updateComplete of `<chromedash-form-table>` seems to be equivalent to awaiting updateComplete of shoelace elements in it; thus making it a good solution. I put the new await updateComplete logic in a `ref` callback of chromedash-form-table which gets called when the element is mounted. Overall, I like this method better than our previous method (`setTimeout()` within `updated()`)
5. Removed unused template_data in guide.py for the guide/edit page, and the unit tests are also updated.

### DIFF
--- a/internals/core_models.py
+++ b/internals/core_models.py
@@ -209,6 +209,7 @@ class Feature(DictModel):
       else:
         d['id'] = None
       d['category'] = FEATURE_CATEGORIES[self.category]
+      d['category_int'] = self.category
       if self.feature_type is not None:
         d['feature_type'] = FEATURE_TYPES[self.feature_type]
         d['feature_type_int'] = self.feature_type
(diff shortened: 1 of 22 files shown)


internals/core_models.py
```diff
@@ -314,6 +314,25 @@ def format_for_template(self, version=2):
       else:
         d['browsers']['chrome']['status']['milestone_str'] = d['browsers']['chrome']['status']['text']
 
+      # The followings are for the guide forms
+      d['format_for_edit'] = {
```
jrobbins, line 318: Can this be done in JS using the existing fields?
Usually an API should be aligned with the data models rather than being biased toward any one particular UX.   Adding a lot of editing-specific fields will increase the transfer time for other use cases.

Since logic to format these fields for editing would likely be used in several elements, it could be located in utils.js or one of the forms-related JS files.

jrobbins (changes requested):

internals/core_models.py
kevinshen56714: OK. That makes sense. I will work on this!

kevinshen56714: I've made the requested changes. Please review it again.

I "un-popped" a lot of fields in format_for_template(), this makes formatting for edit on the frontend easier but adds redundant fields.

The other approach is to still pop these fields but in JS, do more mappings between feature objects for view and for edit. For example:
```js
formattedFeature['bug_url'] = feature['browsers']['chrome']['bug']
formattedFeature['blink_components'] = feature['browsers']['chrome']['blink_components']
formattedFeature['devrel'] = feature['browsers']['chrome']['devrel']
formattedFeature['owner'] = feature['browsers']['chrome']['owners']
```

This would affect many following PRs as well. Please let me know what you think!

jrobbins (approved):
I see what you mean about unpopping.   The editing fields need to get a value from a field of the feature, and all the nested value paths would not match the field names.

PR #2204, opened 2022-09-06
Migrate more form definitions and render guide/verify_accuracy form without unsafeHTML 

kevinshen56714:
This PR migrates of the verify_accuracy form definition to the frontend. This allows us to remove the unsafeHTML in the guide verify_accuracy page and to render the form based on API-fetched feature values and a form field constant. The changes include:

1. Moved `VERIFY_ACCURACY_FORM_FIELDS` from pages/guideforms.py to static/elements/form-definition.js.
2. Removed the unsafeHTML in <chromedash-guide-verify-accuracy-page> and render the form based on form field constant defined in form-definition.js. Also updated the unit tests of that page based on the changes.
3. Use `ref` callback to register form submission handler (same way as in #2201)
4. In #2201, I removed a lot of `self.pop` in render_for_template() method of the Feature model. However, I think this would create too many redundant fields in the returned feature objects. Here, instead of removing the self.pops in render_for_template(), I specify how to get editing field values from a field of the feature, and all the nested value paths would not match the field names in form-definition.js.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -17,3 +17,6 @@ cron:
 - description: Send reminders to verify the accuracy of feature data.
   url: /cron/send_accuracy_notifications
   schedule: every monday 09:00
+- description: Copy over deprecated standardization field
+  url: /cron/write_standard_maturity
+  schedule: 1 of jan 00:00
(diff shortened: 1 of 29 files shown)


jrobbins (approved):
Awesome, another form taken care of, and less data sent for each response to the feature API.

PR #2209, opened 2022-09-06
Fix blink component form field

### LINKED ISSUE #2207: Blink components sl-select takes 20 to 30 seconds to open
**Describe the bug**
In HEAD, the blink_components sl-select element is populated by an API call now.   The API call returns JSON in only 200ms, but the sl-select makes the user wait 20 to 30 seconds after a click until the items are shown.  This latency is not new: it happened before the switch to being API-driven.

**To Reproduce**
1. Go to your local server
2. Sign in
3. Edit some feature's metadata or use the feature creation form
4. Click the blink component widget.

**Expected behavior**
The menu should open almost instantly, certainly in less than a second.

**Additional context**
Maybe we can find a way to make sl-select render the menu faster.
Or, maybe this should be a button that opens a dialog that lists out the choices... kind of like a home-made menu.
Or, maybe it should be a regular text field with a datalist of choices (like we choose metrics buckets)

kevinshen56714:
This PR fixes #2207. I tried using datalist, but it didn't work that well. Instead, I created a sl-dialog and a button to toggle the display. This PR is to provide a quick fix. If the solution looks good, I will create a custom element (e.g. `<chromedash-blink-component-dialog>`) for this in a follow-up PR.

## Demo video

https://user-images.githubusercontent.com/11501902/188759546-239d83f9-d5cc-4bb4-b723-e511c0b2009c.mov

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -125,7 +125,9 @@ class FeatureNew(basehandlers.FlaskHandler):
     owners = self.split_emails('owner')
     editors = self.split_emails('editors')
 
-    blink_components = (self.split_input('blink_components', delim=',') or [])
+    blink_components = (
+        self.split_input('blink_components', delim=',') or
+        [settings.DEFAULT_COMPONENT])
 
     # TODO(jrobbins): Validate input, even though it is done on client.
 
@@ -597,5 +599,18 @@ class FeatureVerifyAccuracy(FeatureEditStage):
     if redirect_resp:
       return redirect_resp
 
-    template_data = {'feature_id': feature_id}
+    f, _ = self.get_feature_and_process(feature_id)
+    feature_edit_dict = f.format_for_edit()
+
+    forms_title = "Accuracy last verified at time of creation."
+    if f.accurate_as_of is not None:
+      date = f.accurate_as_of.strftime("%Y-%m-%d")
+      forms_title = f"Accuracy last verified {date}."
+    form = guideforms.Verify_Accuracy(feature_edit_dict)
+    forms = [(forms_title, str(form), list(form.fields))]
+    template_data = {
+        'feature': f,
+        'feature_id': f.key.integer_id(),
+        'forms': json.dumps(forms),
+    }
     return template_data
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -599,18 +599,5 @@ class FeatureVerifyAccuracy(FeatureEditStage):
     if redirect_resp:
       return redirect_resp
 
-    f, _ = self.get_feature_and_process(feature_id)
-    feature_edit_dict = f.format_for_edit()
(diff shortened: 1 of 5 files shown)


jrobbins:
Thanks for quickly exploring this.  The speed looks good, but I guess this approach would lose the ability to have keyboard input.

What went wrong with datalist?

Also, how about using a plain HTML <select>?

kevinshen56714:
> What went wrong with datalist?

Because the default component is "Blink", so the first time you click on the datalist, it would just show options start with Blink (e.g. `Blink>XXX>XXX`; and things like `Internal>XXX>XXX` are not shown). The entire list won't be revealed until you click the cross symbol to clear the input. I think the users might not know there are more options other than `Blink>XXX>XXX`.

> Also, how about using a plain HTML ?

Do you mean to not use shoelace at all?

jrobbins:
> > What went wrong with datalist?
> 
> Because the default component is "Blink", so the first time you click on the datalist, it would just show options start with Blink (e.g. `Blink>XXX>XXX`; and things like `Internal>XXX>XXX` are not shown). The entire list won't be revealed until you click the cross symbol to clear the input. I think the users might not know there are more options other than `Blink>XXX>XXX`.

What if we don't set any initial value?  We could have it start blank, but make it required.


> > Also, how about using a plain HTML ?
> 
> Do you mean to not use shoelace elements at all?

Yeah.  We could have this one widget be a normal HTML widget, because I suspect that a normal HTML <select> will handle 900 items easily.   We could use a little CSS to make the <select> look sort of like the shoelace widgets, but it doesn't need to be exact.

kevinshen56714:
> > > What went wrong with datalist?
> > 
> > 
> > Because the default component is "Blink", so the first time you click on the datalist, it would just show options start with Blink (e.g. `Blink>XXX>XXX`; and things like `Internal>XXX>XXX` are not shown). The entire list won't be revealed until you click the cross symbol to clear the input. I think the users might not know there are more options other than `Blink>XXX>XXX`.
> 
> What if we don't set any initial value? We could have it start blank, but make it required.

Cool! This is a great idea, I will make a datalist version real quick!

kevinshen56714:
@jrobbins please review the changes again. I've modified it into a datalist. I made it a required field with an empty initial value as you suggested. Also styled it like the other sl-inputs. See the following demo video:


https://user-images.githubusercontent.com/11501902/188773062-d9256a1a-d637-4b04-b7c9-9a6276dfe41d.mov

jrobbins (approved):
That works great.  Thanks for doing the styling.
Your video didn't play, so I just checked out your branch and ran it locally.

jrobbins:
Oh, your video played this time.  I don't know what the problem was before.

PR #2210, opened 2022-09-07
Migrate flat forms to JS to render editall page form

kevinshen56714:
This PR migrates the flat form definitions to the frontend. This allows us to remove the unsafeHTML in the guide editall page and to render the form based on API-fetched feature values and a form field constant. The changes include:

1. Moved all flat form constants from pages/guideforms.py to static/elements/form-definition.js.
2. Removed the unsafeHTML in and render the form based on form field constant defined in form-definition.js. Also updated the unit tests of that page based on the changes.
3. Use ref callback to register form submission handler as in previous PRs.
4. Return 3 additional values from Feature API data: tag_review_status_int, security_review_status_int, and privacy_review_status_int.


jrobbins (approved):
Awesome.   Thanks for all the work needed to get to this point.

dlaliberte (approved):

PR #2214, opened 2022-09-07
Partially migrate Ramcache usage to Redis

KyleJu:
Taken from #2146. Migrate partial Ramcache usage to Redis, except feature-related usage.

- Remove instances of `flush_all`, except in testing, and replace it with individual cache delete
- Remove the cache limit and cache validation logic in the code

### Testing
- [Staging testing](https://a166bab-dot-cr-status-staging.appspot.com/features): verified through the Redis monitoring

### Performance improvement
This PR significantly speeds up first-time page load for parts migrated

### DIFF
--- a/framework/rediscache.py
+++ b/framework/rediscache.py
@@ -52,6 +52,9 @@ def set(key, value, time=86400):
   if redis_client is None:
     return
 
+  if value is None:
+    return
+
   cache_key = add_gae_prefix(key)
   if time:
     redis_client.set(cache_key, pickle.dumps(value), ex=time)
@@ -101,6 +104,9 @@ def set_multi(entries, time=86400):
 
   data_entries = {}
   for key in entries:
+    if entries[key] is None:
+      continue
+
     # gae prefix is needed for mset.
     cache_key = add_gae_prefix(key)
     data_entries[cache_key] = pickle.dumps(entries[key])
--- a/framework/rediscache.py
+++ b/framework/rediscache.py
@@ -52,9 +52,6 @@ def set(key, value, time=86400):
   if redis_client is None:
     return
 
-  if value is None:
-    return
-
   cache_key = add_gae_prefix(key)
   if time:
     redis_client.set(cache_key, pickle.dumps(value), ex=time)
@@ -104,9 +101,6 @@ def set_multi(entries, time=86400):
 
   data_entries = {}
   for key in entries:
-    if entries[key] is None:
-      continue
-
     # gae prefix is needed for mset.
     cache_key = add_gae_prefix(key)
     data_entries[cache_key] = pickle.dumps(entries[key])
--- a/internals/fetchmetrics.py
+++ b/internals/fetchmetrics.py
@@ -24,7 +24,6 @@ from google.auth.transport import requests as reqs
 from google.oauth2 import id_token
 
 from framework import basehandlers
-from framework import ramcache
 from framework import utils
 from internals import metrics_models
 from internals import user_models
@@ -212,7 +211,6 @@ class YesterdayHandler(basehandlers.FlaskHandler):
                 'WebStatusAlert-1: Failed to get metrics even after 2 days')
           return error_message, 500
 
-    ramcache.flush_all()
     return 'Success'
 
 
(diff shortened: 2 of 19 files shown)


framework/rediscache.py
```diff
@@ -52,6 +52,9 @@ def set(key, value, time=86400):
   if redis_client is None:
     return
 
+  if value is None:
```
jrobbins, line 55: Why not allow setting a value of None?

I think None would work: calling `pickle.dumps(None)` gives a binary string, and `pickle.loads(pickle.dumps(None))` gives None.    It could be useful to cache a None in some case, although I don't think we do that currently.

internals/fetchmetrics.py
```diff
@@ -212,7 +211,6 @@ def get_template_data(self, today=None):
                 'WebStatusAlert-1: Failed to get metrics even after 2 days')
           return error_message, 500
 
-    ramcache.flush_all()
```
jrobbins, line 215: I think we need a `flush_all()` or some other kind of invalidation here.  I should have documented the reason with a comment.

This code calls `FetchAndSaveData()` which calls `_SaveData()` which calls `put()` a bunch of times to add a new entity for each metrics datapoint.     Separately, when a request comes in get get metrics data, the file api/metricsdata.py does a query on those datapoints and caches the result.   If we don't invalidate when we add datapoints, the cached query result will be lacking the new datapoints.

I'm not too worried about doing a `flushall()` in this case, because the fetch metrics cron job only runs once every 6 hours.

pages/featurelist_test.py
```diff
@@ -13,14 +13,14 @@
 # limitations under the License.
 
 
+from framework import ramcache
```
jrobbins, line 16: Can this be put down on line 23 with the other imports for our app code?
You are proving that the comment on the next line is not strictly true.

jrobbins (changes requested):
This is good progress, but I think I spotted one main problem.  
Thanks for catching all these unused imports.

framework/rediscache.py
KyleJu: Yea None indeed works. There were testing errors and I suspected it was due to None. Maybe unrelated

KyleJu: Okie I just remembered.  `pickle.loads(None)` in `get` throws errors when a `key` doesn't exist

internals/fetchmetrics.py
KyleJu: Added `flushall` and your comment above. I think `flushall` is fine for now until we don't cache lots stuff in this PR, but we need to come up with a better way to invalidate once we start caching features

KyleJu: Actually could you elaborate on the cron job?  How is the cron job used?

pages/featurelist_test.py
KyleJu: Done

internals/fetchmetrics.py
jrobbins: Our metrics data comes from the uma-export app.   We have a cron job configured to run every 6 hours that checks uma-export for new metrics data.  When it finds metrics data, it stores those new datapoints in datastore via `_SaveDatas()`.  That's the code in the fetchmetrics.py file.

When a user views metrics data, that runs code in api/metricsdata.py which can load datapoints from the cache (or on a cache miss it loads from datastore and then puts them in the cache).

If we have a cached value, that value would keep being used until it expires even if we have fresher datapoints in datastore.  E.g., if the cache was written at 11am, but then the cron job ran at 12 noon, we would still be displaying the stale data until 11am the next day, meaning that it was stale by up to 23 hours in that example.

I suppose there are actually two solution approaches for this:
1. We could invalidate the cache somehow, such as calling `flush_all()`.  That would cause users to see new metrics data as soon as we get it.  But, if we flush everything rather than flushing just the metrics data then users viewing other parts of the site will still see some cache misses.
2. Some staleness in metrics data could be acceptable.  23 hours of staleness is probably too much, but if we reduced the expiration time to 2, 3, or 6 hours, users probably would not notice.   A shorter expiration time for this data means that there will be more cache misses for metrics but it would not cause cache misses for other kinds of cached data.

KyleJu: That makes sense! Thanks!

jrobbins (approved):

PR #2217, opened 2022-09-07
Migrate all stage and implementation status page forms to JS

kevinshen56714:
This PR migrates the all the stage and implementation status forms to the frontend. This allows us to remove the unsafeHTML in the guide stage page and to render the form based on API-fetched feature values and form field constants. The changes include:

1. Moved `STAGE_FORMS` and `IMPL_STATUS_FORMS` constants from pages/guideforms.py to static/elements/form-definition.js.
2. Removed the unsafeHTML in and render the form based on form field constants defined in form-definition.js. Also updated the unit tests of that page based on the changes.
3. Use ref callback to register form submission handler as in previous PRs.
4. Added a name field for a few fields in form-field-specs.js because some fields in the deprecation forms post data into a different DB field name. For example, `deprecation_motivation` field sets the `motivation` field in DB, `r4dt_url` field sets the `intent_to_experiment_url` field, ...etc.


jrobbins (approved):
Whew.  It is like a great weight of unsafeHTML() has been lifted :).

PR #2219, opened 2022-09-08
Move display field defs to JS and remove fielddefs API

kevinshen56714:
This PR migrates the `DISPLAY_FIELDS_IN_STAGES` constant from guideform.py to form-field-specs.js. With this constant available on the frontend, we no longer need the fielddefs API. To this point, we are almost ready to remove guideforms.py! I will submit another PR for that :).

 The changes in this PR include:
1. Removed api/fielddefs_api.py, api/fielddefs_api_test.py, and its related code in main.py and cs-client.js.
2. Moved `DISPLAY_FIELDS_IN_STAGES` constant and its related functions to form-field-specs.js.
3. In `<chromedash-feature-detail>`, render each field based on the imported DISPLAY_FIELDS_IN_STAGES constant. I also renamed field type `urllist` to `multi-url`, and `bool` to `checkbox` so that it matches our new js field specs naming style better.
4. I found that `anticipated_spec_changes` is a multi-url field but was treated as regular text field on the backend. So I modified two places on the backend: 

In core_model.js:
```js
anticipated_spec_changes = ndb.StringProperty(repeated=True)
``` 
and in guide.py
```js
if self.touched('anticipated_spec_changes'):
    feature.anticipated_spec_changes = self.parse_links('anticipated_spec_changes')
```

The existing data with anticipated_spec_changes value as a string in DB should still render fine based on my tests. However, for previous data with no anticipated_spec_changes value, it now produces `['']` when fetched from ndb. So I modify the del_none() method in core_model.js to filter it out.


internals/core_models.py
```diff
@@ -973,7 +974,7 @@ def put(self, notify=True, **kwargs):
   experiment_extension_reason = ndb.StringProperty()
   ongoing_constraints = ndb.StringProperty()
   origin_trial_feedback_url = ndb.StringProperty()
-  anticipated_spec_changes = ndb.StringProperty()
+  anticipated_spec_changes = ndb.StringProperty(repeated=True)
```
jrobbins, line 977: Let's keep this as a single string value rather than a list of URLs.   The help text asks the user to enter a list of issues, which could be URLs or it could just be an English description.  In other words, I think the existing defect was with the widget type rather than with the DB or POST processing.

jrobbins (changes requested):

internals/core_models.py
kevinshen56714: I have changed it back to a string value so the core_model.py basically remains unchanged in this PR. I also removed the MULTI_URL_FIELD_ATTRS for the `anticipated_spec_changes` field in form-field-spec.js so the field gets rendered as a regular textarea. Please review the changes again, thank you!

jrobbins (approved):
I am happy to see this.

PR #2224, opened 2022-09-09
Launch the SPA

kevinshen56714:
This PR sets up the SPA routes, permission checks, and launches the single-page app.

## Routes
1. All the page routes (except for /features, /admin, and /omaha_data) are now handled via a new `SPAHandler` whose template_path is spa.html and doesn't require any page-specific data. The `SPAHandler` is slightly different than the `ConstHandler` in that it has more permission checks (more about permissions below), so I think it's better to separate the two.
2. I separated the GET and POST routes for SPA. The GET routes are added to the Flask app as usual, and the POST routes are provided separately as a list to `basehandlers.FlaskApplication`. These POST routes have the methods=['POST'] and are added ahead of the GET routes so they don't get handled by `SPAHandler`.
3. The POST routes are either handled by the `FeatureCreateHandler` or `FeatureEditHandler` in guide.py (they are just renamed from previous `FeatureNew` and `FeatureEditStage` classes).
4. Some routes were redirected by the server (via `Redirector`), for example, `'/' -> '/roadmap'` and `'/metrics' -> '/metrics/css/popularity'`. These redirections are now handled by the client router in chromedash-app element.

## Permissions
1. In `SPAHandler`, there are three levels of permissions checks - `require_signin`, `require_create_feature`, and `require_edit_feature` that can be specified in the "defaults" dict argument.
2. The /myfeatures and /settings pages have `require_signin` set to True
3. The /guide/new page have `require_create_feature` set to True. The create feature permission is checked by a new method `validate_feature_create_permission` in framework/permissions.py. I had to create this new function because although the current `require_create_feature` function does the same thing, it has to be used as a decorator.
4. The /guide/edit, /guide/editall, /guide/verify_accuracy, and guide/stage pages have `require_edit_feature` set to True. Users need the feature-specific permission to view these pages.
5. Users will be redirected if they don't have the required permissions.

## Misc
1. User object is now fetched through API in `<chromedash-app>`. The user object is now only passed down to the header element, but in a follow-up PR, I will pass it down to all the page elements that fetch user data.
2. `guide.ProcessOverview`, `guide.FeatureEditAllFields`, `guide.FeatureVerifyAccuracy` classes and their unit tests are removed because they don't handle GET request anymore and their POST requests can all be handled the same way via the new `FeatureEditHandler`
3. If this PR looks good. In a follow-up PR, I will remove all the unused HTML files!

### DIFF
--- a/main.py
+++ b/main.py
@@ -37,7 +37,6 @@ from internals import notifier
 from internals import data_backup
 from internals import deprecate_field
 from pages import blink_handler
-from pages import featuredetail
 from pages import featurelist
 from pages import guide
 from pages import intentpreview
--- a/main.py
+++ b/main.py
@@ -144,7 +144,7 @@ spa_pages = [
 
 spa_page_routes = []
 for route in spa_pages:
-  page_defaults ={}
+  page_defaults = {}
   if isinstance(route, tuple):
     route, additional_defaults = route
     page_defaults.update(additional_defaults)
--- a/main.py
+++ b/main.py
@@ -118,12 +118,56 @@ api_routes = [
     # (API_BASE + '/metrics/<str:kind>/<int:bucket_id>', TODO),
 ]
 
+spa_pages = [
+  '/',
+  '/roadmap',
+  ('/myfeatures', {'require_signin': True}),
+  '/newfeatures',
+  '/feature/<int:feature_id>',
+  ('/guide/new', {'require_create_feature': True}),
+  ('/guide/edit/<int:feature_id>', {'require_edit_feature': True}),
+  ('/guide/stage/<int:feature_id>/<int:stage_id>', {'require_edit_feature': True}),
+  ('/guide/editall/<int:feature_id>', {'require_edit_feature': True}),
+  ('/guide/verify_accuracy/<int:feature_id>', {'require_edit_feature': True}),
+  '/metrics',
+  '/metrics/css',
+  '/metrics/css/popularity',
+  '/metrics/css/animated',
+  '/metrics/css/timeline/popularity',
+  '/metrics/css/timeline/popularity/<int:bucket_id>',
+  '/metrics/css/timeline/animated',
+  '/metrics/css/timeline/animated/<int:bucket_id>',
+  '/metrics/feature/popularity',
+  '/metrics/feature/timeline/popularity',
+  '/metrics/feature/timeline/popularity/<int:bucket_id>',
+  ('/settings', {'require_signin': True}),
+]
 
-page_routes = [
+spa_page_routes = []
+for route in spa_pages:
+  page_defaults ={}
+  if isinstance(route, tuple):
+    route, additional_defaults = route
+    page_defaults.update(additional_defaults)
+  spa_page_routes.append((route, basehandlers.SPAHandler, page_defaults))
+
+spa_page_post_routes = [
+  ('/guide/new', guide.FeatureCreateHandler),
+  ('/guide/edit/<int:feature_id>', guide.FeatureEditHandler),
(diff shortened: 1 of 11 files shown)


static/elements/chromedash-header.js
```diff
@@ -182,11 +182,22 @@ export class ChromedashHeader extends LitElement {
...
+    // user is passed in from chromedash-app
+    if (this.user && this.user.email) return;
+
+    // user is passed in from chromedash-app, but the user is not logged in
+    if (!this.user) {
+      this.initializeGoogleSignIn();
+      return;
+    };
+
+    // user is not passed in from anywhere, i.e. this.user is still {}
```
jrobbins, line 200: Is this for MPA pages?  It would be more clear if the comment mentioned "MPA".

jrobbins (commented):
This code all looks good.
I'm going to hold off on the LGTM until tomorrow because I want to test it interactively myself, and because I deployed a new version this morning and I want to see if there are any problems that need a quick fix before landing new changes.

jrobbins:
I did manual testing on Chrome, Safari, and Firefox and everything looks good, except one small problem:
I can't star a feature on the feature detail page.  I think the cause is that the feature ID is being passed as a string of digits rather than a JS number. 

Also, this is not a blocker, but there's also some glitches with browser navigation among the metrics pages.  If I start with the stack rank, then drill into a timeline page, I can't get back to the stack rank by pressing the back button once.  I need to hit it twice fast.  And, once I get back, I can't navigate forward via cmd-right-arrow to get back to the timeline.

jrobbins:
I deployed a version to our staging server, so you can see the metrics pages there:
https://d05fbf4-tainted-dot-cr-status-staging.appspot.com/metrics/feature/timeline/popularity/2089

main.py
```diff
@@ -118,83 +117,71 @@
...
+  '/metrics/css/timeline/animated/<int:bucket_id>',
+  '/metrics/feature/popularity',
+  '/metrics/feature/timeline/popularity',
+  '/metrics/feature/timeline/popularity/<int:bucket_id>',
+  ('/settings', {'require_signin': True}),
+]
+
+spa_page_routes = []
+for route in spa_pages:
+  page_defaults ={}
```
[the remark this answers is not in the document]
kevinshen56714: Fixed!

static/elements/chromedash-header.js
kevinshen56714: Added!

kevinshen56714:
> I did manual testing on Chrome, Safari, and Firefox and everything looks good, except one small problem:
> I can't star a feature on the feature detail page. I think the cause is that the feature ID is being passed as a string of digits rather than a JS number.

Good catch, this is fixed by parseInt the featureId string before passing into the feature detail page component.

> Also, this is not a blocker, but there's also some glitches with browser navigation among the metrics pages. If I start with the stack rank, then drill into a timeline page, I can't get back to the stack rank by pressing the back button once. I need to hit it twice fast. And, once I get back, I can't navigate forward via cmd-right-arrow to get back to the timeline.

This is also fixed. I found that we do history.pushState even when the url already has the correct bucketId, so we end up adding two identical entries to the browser's session history stack. And this happens when a bucketId is provided through url, which is the case when you drill into a timeline page from the stack rank page.

Please review the changes again. Really appreciate it for testing out the SPA!

jrobbins (approved):
Cue the fireworks!

kevinshen56714:
🎉 🎉 🎉

jstenback:
Exciting! I can't wait to see this live on chromestatus.com!

PR #2225, opened 2022-09-09
Update test coverage commands and configs

kevinshen56714:
The existing python test coverage report includes all the python files in /cs-env  and all the test files themselves. This PR modifies the `.coveragerc` so that those files are excluded in the coverage calculation.

Meanwhile, I added a new command `npm run webtest-coverage` to generate the JS coverage report. The report is generated to the ./coverage folder so I add that directory in .gitignore.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,8 @@ include =
     ./*.py
 omit =
     ./lib/*
+    ./cs-env/*
+    ./*_test.py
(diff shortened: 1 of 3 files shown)


jrobbins (approved):
I'm glad that this turned out to be just a simple command.
I look forward to seeing where we stand

PR #2226, opened 2022-09-09
Use hoist to make all sl-select elements faster

jrobbins:
Earlier this week I discovered that the blink component field took a really long time to open, and that adding hoist is a work-around.  Today I realized that all our sl-selects are kind of slow (but not as bad as that one was), so using hoist is an improvement.

See also:
https://github.com/shoelace-style/shoelace/discussions/902


kevinshen56714 (approved):
I saw your CodePen. This is a surprisingly easy way to improve the speed.

PR #2227, opened 2022-09-09
Add stack rank and timeline page unit tests

kevinshen56714:
This PR adds unit tests for `<chromedash-stack-rank-page>` and `<chromedash-timeline-page>` that I forgot to add when creating the page elements. The tests are pretty standard except that in the timeline test, I had to add the following to avoid warnings about undefined load and setOnLoadCallback methods from window.google.charts.
```js
window.google = {charts: {load: () => {}, setOnLoadCallback: () => {}}};
```


dlaliberte:
I'm surprised that all the time I have been working on ChromeStatus, I missed the fact that Google Charts was being used.

jrobbins (approved):

PR #2234, opened 2022-09-13
Completely migrate feature-related cache off ramcache

KyleJu:
Partially taken from #2146. Completely migrate Ramcache.

- Replace the use of `flush_all` with `scan` and `delete`. `Scan` the feature key prefix, `features|` for all feature query results, and `delete` all keys that start with that prefix
- Remove the cache limit and cache validation logic

### Concurrency
`scan` only guarantees that it retrieves all the elements that were present in the collection from the start to the end of a full iteration. If an element is not constantly present in the collection during a full iteration, may be returned or not: it is undefined. In other words, if a feature is added or deleted during an iteration, it may or may not be returned by `scan`

### Performance improvement
This PR speeds up first-time page load for feature page


pages/guide.py
```diff
@@ -72,9 +72,9 @@ def process_post_data(self):
         created_by=signed_in_user,
         updated_by=signed_in_user)
     key = feature.put()
-
-    # TODO(jrobbins): enumerate and remove only the relevant keys.
-    ramcache.flush_all()
+
```
KyleJu, line 75:
```suggestion

```

framework/rediscache.py
```diff
@@ -118,6 +117,23 @@ def delete(key):
...
+  if redis_client is None:
+    return
+
+  prefix = add_gae_prefix(pattern)
+  # https://redis.io/commands/scan/
+  pos, keys = redis_client.scan(cursor=0, match=prefix)
+  target = keys
+  while pos != 0:
+    pos, keys = redis_client.scan(cursor=pos, match=prefix)
+    target.append(keys)
```
jrobbins, line 131: Shouldn't that `append()` be `extend()`?

internals/core_models.py
```diff
@@ -1090,14 +1097,14 @@ def __init__(self, *args, **kwargs):
   @classmethod
   def get_feature_entry(self, feature_id, update_cache=False):
     KEY = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, feature_id)
```
jrobbins, line 1099: This could use your new method `feature_cache_prefix()`.

internals/core_models.py
```diff
@@ -1126,7 +1133,7 @@ def get_by_ids(self, entry_ids, update_cache=False):
 
     for fe_id in entry_ids:
       lookup_key = '%s|%s' % (FeatureEntry.DEFAULT_CACHE_KEY, fe_id)
```
jrobbins, line 1135: Same.

internals/core_models.py
```diff
@@ -1137,7 +1144,7 @@ def get_by_ids(self, entry_ids, update_cache=False):
       if entry and not entry.deleted:
         store_key = '%s|%s' % (
```
jrobbins, line 1145: Same.

internals/core_models.py
```diff
@@ -845,12 +844,20 @@ def put(self, notify=True, **kwargs):
...
-    # Invalidate ramcache for the individual feature view.
-    cache_key = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, self.key.integer_id())
-    ramcache.delete(cache_key)
+    # Invalidate rediscache for the individual feature view.
+    cache_key = Feature.feature_cache_key(self.key.integer_id())
+    rediscache.delete(cache_key)
 
     return key
 
+  @classmethod
```
jrobbins, line 853: Since these methods don't actually use `css`, they could be top-level functions rather than class methods.   That would make them available for both `Feature` and `FeatureEntry` and shorten the calling code because it would not need the class name.

internals/core_models.py
```diff
@@ -1096,7 +1094,7 @@ def __init__(self, *args, **kwargs):
 
   @classmethod
   def get_feature_entry(self, feature_id, update_cache=False):
-    KEY = '%s|%s' % (Feature.DEFAULT_CACHE_KEY, feature_id)
```
KyleJu, line 1092: @jrobbins should be this `FeatureEntry.DEFAULT_CACHE_KEY`? I think I have to pass the cache_key in as well

jrobbins: Good point.  It should be different because there will be a time period in which we have both models used at the same time.

jrobbins (approved):

PR #2235, opened 2022-09-13
Flush metrics specific data in Redis

KyleJu:
Address comments in #2214. Only flush metrics specific data when running uma-export app.


api/metricsdata.py
```diff
@@ -92,7 +92,7 @@ def get_template_data(self):
 
 class PopularityTimelineHandler(TimelineHandler):
 
-  CACHE_KEY = 'css_pop_timeline'
+  CACHE_KEY = 'metrics|css_pop_timeline'
```
jrobbins, line 95: We should make `'metrics'` a constant, like the way that it is done in the Feature class.

jrobbins (approved):

PR #2236, opened 2022-09-13
Remove unused html files and change app title dynamically

kevinshen56714:
As the SPA has been launched, a lot of the previous MPA html files are no longer used thus can be removed. Some of these html files update the title of the document, so I added code to update document.title from js dynamically. The title would be changed back via disconnectedCallback once these page components are unmounted.

Google search engines should be able to pick up these titles as well since JavaScript generated content also gets indexed nowadays.


static/elements/chromedash-guide-verify-accuracy-page.js
```diff
@@ -51,6 +56,11 @@ export class ChromedashGuideVerifyAccuracyPage extends LitElement {
     });
   }
 
+  disconnectedCallback() {
```
jrobbins, line 59: I feel that this is too repetitive.
How about defining a base class for all our page classes?

jrobbins (approved):
LGTM after considering one suggestion.

static/elements/chromedash-guide-verify-accuracy-page.js
kevinshen56714: Yeah, I think it can be good, especially the form handling methods are shared among multiple pages too. I'll probably do it a follow-up PR.

Also, just FYI, I also deleted pages/featuredetails.py and pages/featuredetails_test.py because these are no longer used too and the unit test was complaining there's no feature.html.

PR #2238, opened 2022-09-13
Share user data among multiple SPA pages

kevinshen56714:
There are 5 SPA pages that require user data, and this PR passes down the user object from `<chromedash-app>` to these page components, so that basically, all the SPA pages share the same user object. This saves one API request for each of the 5 pages. The user data mostly only changes when user log in or log out. In those cases, the page will reload and force the `<chromedash-app>` to reset the user data.

### DIFF
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -100,22 +100,26 @@ class ChromedashApp extends LitElement {
     page('/', () => page.redirect('/roadmap'));
     page('/roadmap', (ctx) => {
       this.pageComponent = document.createElement('chromedash-roadmap-page');
+      this.pageComponent.user = this.user;
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
     });
     page('/myfeatures', (ctx) => {
       this.pageComponent = document.createElement('chromedash-myfeatures-page');
+      this.pageComponent.user = this.user;
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
     });
     page('/newfeatures', (ctx) => {
       this.pageComponent = document.createElement('chromedash-all-features-page');
+      this.pageComponent.user = this.user;
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
     });
     page('/feature/:featureId', (ctx) => {
       this.pageComponent = document.createElement('chromedash-feature-page');
       this.pageComponent.featureId = parseInt(ctx.params.featureId);
+      this.pageComponent.user = this.user;
       this.pageComponent.contextLink = this.contextLink;
       this.pageComponent.appTitle = this.appTitle;
     });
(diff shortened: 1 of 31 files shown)


static/elements/chromedash-app.js
```diff
@@ -100,22 +100,26 @@ class ChromedashApp extends LitElement {
     page('/', () => page.redirect('/roadmap'));
     page('/roadmap', (ctx) => {
       this.pageComponent = document.createElement('chromedash-roadmap-page');
+      this.pageComponent.user = this.user;
```
jrobbins, line 103: Sometimes user permissions do change, but I think that just loading the same user permissions once is OK because: (a) the server checks updated permissions on POSTs, and (b) users who request new permissions can be expected to reload.

jrobbins (approved):

static/elements/chromedash-guide-edit-page.js
```diff
@@ -60,13 +60,11 @@ export class ChromedashGuideEditPage extends LitElement {
...
     this.loading = true;
     Promise.all([
-      window.csClient.getPermissions(),
       window.csClient.getFeature(this.featureId),
       window.csClient.getFeatureProcess(this.featureId),
       window.csClient.getFeatureProgress(this.featureId),
       window.csClient.getDismissedCues(),
-    ]).then(([user, feature, process, progress, dismissedCues]) => {
-      this.user = user;
+    ]).then(([feature, process, progress, dismissedCues]) => {
```
jrobbins, line 67: Actually, let's think through one other scenario, imagine that a user visits the site, opens a second tab on the site, signs out in that tab, then absent-mindedly tries to continue using the first tab.    The user would still see themselves as signed in, until they try to submit an edit, they would get an error at that time, and their form values would be lost.  That's not a great experience.    How about we keep reloading the user permissions whenever the user navigates into an editing page?

jrobbins (changes requested):

static/elements/chromedash-guide-edit-page.js
kevinshen56714: I like it. The guide-edit-page is like the "home page" for all the editing pages, so it's nice to re-check user permission before users enter any form pages. Also, it feels great now we have the flexibility to determine when to fetch data and when not to. 

I have made the changes. Now the user data is always refetched upon entering the edit page. I also tried the scenario you mentioned and added the following code. If the user has no permissions and somehow sneak into the edit page, a page reload would be forced. And this would triggers server-side check and redirect the user to the all features page.

```js 
// re-check permission every time users visit the edit page
const canEdit = (this.user &&
(this.user.can_edit_all || this.user.editable_features.includes(this.featureId)));
if (!canEdit) location.reload();
```

jrobbins (approved):

PR #2244, opened 2022-09-14
Script to migrate Comment entities to Activity entities

DanielRyanSmith:
This introduces a new script to copy over each Comment entity to a new Activity entity. This is made to be idempotent, so no comments can be copied multiple times erroneously.

### DIFF
--- a/internals/review_models.py
+++ b/internals/review_models.py
@@ -141,6 +141,7 @@ class Comment(ndb.Model):
   author = ndb.StringProperty()
   content = ndb.StringProperty()
   deleted_by = ndb.StringProperty()
+  migrated = ndb.BooleanProperty()
   # If the user set an approval value, we capture that here so that we can
   # display a change log.  This could be generalized to a list of separate
   # Amendment entities, but that complexity is not needed yet.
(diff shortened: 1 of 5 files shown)


internals/review_models.py
```diff
@@ -141,6 +141,7 @@ class Comment(ndb.Model):
   author = ndb.StringProperty()
   content = ndb.StringProperty()
   deleted_by = ndb.StringProperty()
+  migrated = ndb.BooleanProperty()
   # If the user set an approval value, we capture that here so that we can
   # display a change log.  This could be generalized to a list of separate
   # Amendment entities, but that complexity is not needed yet.
```
DanielRyanSmith, line 147: Comment entities also contain an `old_approval_state` and `new_approval_state`. Should we ignore these or do they need to be migrated over as well?

jrobbins: Good point.  Ideally, we would migrate over that early approach to storing history to Amendments. 

The amendments could have these values:
* field_name: just use the string "vote"
* old_value: the string from Approval.APPROVAL_VALUES[old_approval_state]
* new_value: the string from Approval.APPROVAL_VALUES[new_approval_state]

internals/review_models.py
```diff
@@ -300,7 +301,7 @@ class Activity(ndb.Model):  # copy from Comment
   """An activity log entry (comment + amendments) on a gate or feature."""
   feature_id = ndb.IntegerProperty(required=True)
   gate_id = ndb.IntegerProperty()  # The gate commented on, or general comment.
-  created = ndb.DateTimeProperty(auto_now=True)
+  created = ndb.DateTimeProperty(auto_now_add=True)
```
jrobbins, line 304: Good catch, otherwise setting deleted_by would change the created date.

jrobbins (changes requested):

internals/review_models.py
jrobbins: It turns out that these values are currently all null, so no need to convert.

jrobbins (approved):

PR #2256, opened 2022-09-20
Avoid reload loop due to comparing str to int.

jrobbins:
When visiting the `/guide/edit` page as a feature editor, the page kept reloading in a loop.   The reload logic is meant to check that the current user still has permission to edit a feature.  A feature editor does have permission to edit the feature that lists them, which is why the server does not give a 403 on each reload.   However, the check in JS code fails because it is comparing `this.featureId` (a string) to `this.user.editable_features` (an int).

In this PR:
* Have the app always convert strings of digits from the URL into JS number values.

### DIFF
--- a/static/elements/chromedash-app.js
+++ b/static/elements/chromedash-app.js
@@ -129,23 +129,23 @@ class ChromedashApp extends LitElement {
     });
     page('/guide/edit/:featureId', (ctx) => {
       this.pageComponent = document.createElement('chromedash-guide-edit-page');
-      this.pageComponent.featureId = ctx.params.featureId;
+      this.pageComponent.featureId = parseInt(ctx.params.featureId);
       this.pageComponent.appTitle = this.appTitle;
     });
     page('/guide/editall/:featureId', (ctx) => {
       this.pageComponent = document.createElement('chromedash-guide-editall-page');
-      this.pageComponent.featureId = ctx.params.featureId;
+      this.pageComponent.featureId = parseInt(ctx.params.featureId);
       this.pageComponent.appTitle = this.appTitle;
     });
     page('/guide/verify_accuracy/:featureId', (ctx) => {
       this.pageComponent = document.createElement('chromedash-guide-verify-accuracy-page');
-      this.pageComponent.featureId = ctx.params.featureId;
+      this.pageComponent.featureId = parseInt(ctx.params.featureId);
       this.pageComponent.appTitle = this.appTitle;
     });
     page('/guide/stage/:featureId/:stageId', (ctx) => {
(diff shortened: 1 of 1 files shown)


kevinshen56714 (approved):

KyleJu (approved):

static/elements/chromedash-app.js
```diff
@@ -129,23 +129,23 @@ class ChromedashApp extends LitElement {
     });
     page('/guide/edit/:featureId', (ctx) => {
```
jcscottiii, line 130: Do we want to change this to `'/guide/edit/:featureId(-?\d+)'`. 

I think this *should* work but I didn't try this out personally. There's talk about it on this [issue](https://github.com/visionmedia/page.js/issues/251). That issue redirects to this express test router [page](http://forbeslindesay.github.io/express-route-tester/) (since PageJS is inspired by Express)

| Scenario      | Screenshot | parseInt result |
| ----------- | ----------- | ----------- |
| positive numbers using (-?\d+) regex `'/guide/edit/2'`    | image      |  2 ✅ |
| negative numbers using (-?\d+) regex `'/guide/edit/-2'`    | image      |  -2 ✅ |
| 0 using (-?\d+) regex `'/guide/edit/0'`    | image     |  0 ✅ |
| non integer using (-?\d+) regex `'/guide/edit/hi'`    | image   |  N/A (Doesn't match this route)  ✅ |
| non integer without using (-?\d+) regex `'/guide/edit/hi'`    | image   |  NaN  ❌ |

Why change it to this:
It will give us better safety for the route. Without it, a user could go to `/guide/edit/somethingThatsNotAInteger` and parseInt will return NaN.

*Note this regex also matches `-0` but that is okay because parseInt interprets that as a number too*

jcscottiii (commented):
Thanks for this PR! I was experiencing this redirect loop while testing my PR this morning. I had a quick suggestion @jrobbins . Not sure if it would work but wanted to run it by you.

static/elements/chromedash-app.js
jcscottiii: Similar changes for the other routes too if we go with this.

jrobbins: Good suggestion.  I was expected to get a NaN with the existing code as you said above, but somehow I couldn't get it to do it.  I went ahead and added the regexes because I think that they helot make our expectations about the URLs more explicit.  I left off the dash because all our IDs seem to be positive.

Also, I learned that while testing the SPA, one needs to look at the Network tab filtered by doc requests to verify that no new full page loads are happening.

jcscottiii (approved):

PR #2262, opened 2022-09-21
Refactored reminders and added prepublication notice.

jrobbins:
This implements a new kind of reminder email that was discussed at our "data hygiene" meetings over the past several weeks.

The goal is to give feature owners a concrete look at how their feature will be presented in something like the beta blog post or release notes.   The emails are meant to go out two weeks before the feature's shipping milestone reaches the beta channel.  The cron runs every week, but roughly 3 out of 4 times it does nothing.

To reuse the existing logic for the accurate-as-of reminder, I refactored it into a new reminders.py file and defined an abstract base class with specific methods and constants for subclasses to override.

In this PR:
* main.py: New route registered for the handler, and accuracy handler reference changed
* notifier.py and notifier_test.py: removed accuracy handler and some imports
* reminders.py and reminders_test.py: added refactored accuracy handler and wrote new prepublication handler
* prepublication-notice-email.html: new template for the prepublication handler


internals/reminders.py
```diff
@@ -0,0 +1,179 @@
...
+      'shipped_milestone',
+      'shipped_webview_milestone')
+
+  def prefilter_features(self, current_milestone_info, features):
+    now = datetime.now()
+    prefiltered_features = [
+        feature for feature in features
+        # It needs review if never reviewed, or if grace period has passed.
+        if (feature.accurate_as_of is None or
+            feature.accurate_as_of + self.ACCURACY_GRACE_PERIOD < now)]
```
jrobbins, line 148: Note that I reversed the sense of this condition (now it evaluates to True for features that should notify whereas you had it it True to skip a a feature), but I kept the comparison as less-than. When looking at this code for this PR, I thought that the original less than was actually a defect. Please double check that I have this right now.

internals/reminders.py
```diff
@@ -0,0 +1,179 @@
...
+
+  def filter_by_milestones(self, current_milestone_info, features):
+    """Return [(feature, milestone)] for features with a milestone in range."""
+    # 'current' milestone is the next stable milestone that hasn't landed.
+    # We send notifications to any feature planned for beta or stable launch
+    # in the next N weeks. E.g., beta for (current + 2) starts in roughly 8
+    # weeks.  So we check if any features have launches in these 3 milestones
+    # (current, current + 1, and current + 2).
+    min_mstone = int(current_milestone_info['mstone'])
+    max_mstone = min_mstone + self.FUTURE_MILESTONES_TO_CONSIDER
```
DanielRyanSmith, line 98: This comment is no longer accurate for this code block.

internals/reminders.py
```diff
@@ -0,0 +1,179 @@
...
+      })
+  return email_tasks
+
+
+class AbstractReminderHandler(basehandlers.FlaskHandler):
+  JSONIFY = True
+  SUBJECT_FORMAT = '%s'
+  EMAIL_TEMPLATE_PATH = None  # Subsclasses must override
+  FUTURE_MILESTONES_TO_CONSIDER = 0
+  MILESTONE_FIELDS = None  # Subsclasses must override
```
DanielRyanSmith, line 72:
```suggestion
  EMAIL_TEMPLATE_PATH = None  # Subclasses must override
  FUTURE_MILESTONES_TO_CONSIDER = 0
  MILESTONE_FIELDS = None  # Subclasses must override
```

DanielRyanSmith (approved):
Great job making the code reusable 😊

internals/reminders.py
jrobbins: Updated

Issue #2267, opened 2022-09-23
Our metrics pages are too slow


jrobbins:
**Describe the bug**
Our metrics pages still seem slow, even though other pages in our app (except the old feature list page) are pretty fast.

**Additional context**
Things to consider:
* HTTP caching
* Pagination with a smaller initial page size
* A more compact, less redundant wire format
* Client-side caching and prefetching

Our current HTTP cache header might be ignored because something also sets a `vary: cookie` header, and flask sets an updated session cookie value on each response, meaning that items cached using the old cookie are not supposed to be reused.    Since the metrics data is never private, maybe the client could make these requests without any credentials, thus allowing a previously HTTP cached version to be reused.

Regarding pagination or wire format changes: we need to keep the current (legacy) metrics data feeds (/data/.*) as-is, but if we add new API handlers (/api/v0/.*), we could do whatever we want with the response and with new request parameters (for pagination).

PR #2268, opened 2022-09-23
Omit cookies for metrics data requests because of Vary: Cookie

jrobbins:
In researching the problem as I wrote up issue #2267, I thought that I found an easy solution to the HTTP caching problem.  I think this PR solves the HTTP caching problem, but there is still a need for other kinds of speedups outlined in that issue.

In this PR:
* settings.py: allow cachable HTTP content to be used for one hour instead of one minute
* chromedash-timeline-page.js and chromedash-stack-rank-page.js: omit credentials so that the `vary: cookie` response header does not cause cached responses to be unusable.

### DIFF
--- a/api/metricsdata.py
+++ b/api/metricsdata.py
@@ -220,6 +220,7 @@ class FeatureObserverPopularityHandler(FeatureHandler):
 
 
 class FeatureBucketsHandler(basehandlers.FlaskHandler):
+  HTTP_CACHE_TYPE = 'private'
   JSONIFY = True
 
   def get_template_data(self, prop_type):
--- a/settings.py
+++ b/settings.py
@@ -115,7 +115,7 @@ SECRET_KEY = os.environ['DJANGO_SECRET']
 
 RSS_FEED_LIMIT = 15
 
-DEFAULT_CACHE_TIME = 60 # seconds
+DEFAULT_CACHE_TIME = 3600 # seconds
 
 USE_I18N = False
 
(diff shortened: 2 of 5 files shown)


jrobbins:
You can try this out on staging:
https://84cbd90-tainted-dot-cr-status-staging.appspot.com/metrics/css/popularity
In the network tab, I see "(disk cache)" showing up for these requests if I navigate away and back to these pages.

The CSS and CSS animated properties stack rank pages load pretty fast with this change.  However, the JS/HTML stack rank page is still kind of slow, maybe because it has a lot more items to parse and render.

KyleJu:
Checking all requests, I see `https://84cbd90-tainted-dot-cr-status-staging.appspot.com/data/csspopularity` is cached properly with the `cache-control: private, max-age=3600` header (great work!). But looking at the network tab, I see `/metrics/css/popularity`
Screen Shot 2022-09-26 at 12 13 16 AM (2)

`cache-control: private` is specified without a max-age, then it would fall back to `expires`, which is basically no-cache in this case. 300ms for this request isn't a big deal but perhaps we can add the `max-age` directive here

KyleJu:
A different question: would `vary: cookie` be an issue? Unless a user logs in or logs out, I don't think cache will be unusable. I guess using the same cache, regardless of user's log-in status, is also acceptable for these two URLs.

jrobbins:
> `cache-control: private` is specified without a max-age, then it would fall back to `expires`, which is basically no-cache in this case. 300ms for this request isn't a big deal but perhaps we can add the `max-age` directive here

That's the HTML for the page, so it's out of scope for what I was intending here.   

However, it raises an interesting point: our MPA pages used to be personalized for each user with a bunch of different user-specific values.   Now that we have the SPA, each page is less personalized, but we still include a nonce, app version, and some other details that should not be cached.   So, we can't make this page caching.  We could try to get to a point where we can serve all our client-side stuff without any personalization.  But, I think that could be a big project with pros and cons.

On the bright side, you are only seeing this page load at all because you are testing reloading the page.   A user navigating around the SPA (except for the all features page) would avoid this page load, so caching it is moot.

KyleJu (approved):

PR #2273, opened 2022-09-26
Write new Activity entities when creating Comments

DanielRyanSmith:
When a new Comment entity is created, an Activity entity will also be created with the same ID. Additionally, deleting or undeleting a comment will also delete/undelete an Activity with the same ID.

### DIFF
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -96,6 +96,12 @@ class CommentsAPI(basehandlers.APIHandler):
           new_approval_state=new_state)
       comment.put()
 
+      # Schema migration double-write.
+      comment_activity = review_models.Activity(
+        id=comment.key.integer_id(), feature_id=feature_id, gate_id=field_id,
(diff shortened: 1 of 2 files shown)


api/comments_api.py
```diff
@@ -96,6 +96,12 @@ def do_post(self, feature_id, field_id=None):
           new_approval_state=new_state)
       comment.put()
 
+      # Schema migration double-write.
+      comment_activity = review_models.Activity(
+        id=comment.key.integer_id(), feature_id=feature_id, gate_id=field_id,
```
jrobbins, line 101: Nice.  I thought it would require creating a `Key` object rather than just setting the `id` directly.

jrobbins (approved):

api/comments_api.py
DanielRyanSmith: Yeah, I did too due to needing Key objects for filtering, but it seems it's easier than expected. 😊

PR #2275, opened 2022-09-27
Migrate Feature entities to FeatureEntry entities

DanielRyanSmith:

This script writes all existing features to new FeatureEntry entities. *Note*: This script only copies over fields on the Feature kind that should exist on any FeatureEntry entities. Additional work will be necessary to move some Feature fields that are not part of the schema of different kinds.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -26,8 +26,11 @@ cron:
 - description: Removes any users that have been inactive for 9 months.
   url: /cron/remove_inactive_users
   schedule: 1st monday of month 9:00
-- description: Copy over comment entities to new activity entities
+- description: Copy over Comment entities to new Activity entities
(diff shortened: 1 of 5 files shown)


DanielRyanSmith:
@jrobbins This might need clarification and refactoring, as I only copied over known fields and their new formats or named. Let me know if anything need to also be migrated at the same time, or I've missed something here. 🙂

DanielRyanSmith:
@jrobbins, I believe this is ready for review now - sorry for the hasty review request 😅

jrobbins:
I assume you are are going to simplify this after landing your other PR.

DanielRyanSmith:
This PR should now be refactored and ready for review 🙂

jrobbins (approved):
I like how much simpler this is with the generic function.

PR #2276, opened 2022-09-27
Replace django templates for email with jinja2 templates

### LINKED ISSUE #2030: Replace django templates for email with jinja2 templates
We currently use two python web app frameworks (django and flask) and we are trying to use just flask.   We'd like to phase out our use of django, including where it is used to generate the body of email messages.

While it would be tempting to get really simple by using python formatted string literals like `f'Hello {name}'`, we must follow a security guideline of automatically escaping all user-controlled values.

Flask includes jinja2 templates, which are similar to django templates.  So, maybe what we could do is convert the existing django template to jinja2 templates and change from django's `render_to_string(path, data)` to `flask.render_template(path, **data)`.

We would need to upgrade the email templates from the django template syntax to the jinja2 syntax: `templates/*-email.html` and `templates/estimated-milestones-table.html`.

There should be no user-visible impact of this change.  It is a project health issue to reduce our excessive library dependencies.

jcscottiii:
This commit converts the email templates from django templates to jinja2 templates.

It also creates a new helper function: flask_compat_get_template_path

This function uses the existing configuration for django templates and uses it to return a string used to locate the templates for jinja2.

Fixes #2030

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -19,6 +19,19 @@ TEMPLATES = [
   },
 ]
 
+def flask_compat_get_template_path() -> str:
+  """Returns a path to the templates.
+  Leverages the existing TEMPLATES variable used for Django templates and
+  returns a path usable by Flask/Jinja2.
+
+  This is useful because by default flask.render_template will look for a
+  template folder relative to the module.
+
+  Once all Django templates are completely converted to Jinja2, a simpler
+  variable TEMPLATES could be used to hold information about the templates.
+  """
+  return TEMPLATES[0]['DIRS'][0]
+
 # This is necessary to override django templates.
 FORM_RENDERER = 'django.forms.renderers.TemplatesSetting'
 
--- a/templates/update-feature-email.html
+++ b/templates/update-feature-email.html
@@ -7,7 +7,7 @@
         >{{feature.name}}</a></b>
 </p>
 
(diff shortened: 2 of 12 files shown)


templates/update-feature-email.html
```diff
@@ -7,7 +7,7 @@
         >{{feature.name}}</a></b>
 </p>
 
-<p><b>Components</b>: {{feature.blink_components|join:", "}}</p>
+<p><b>Components</b>: {{feature.blink_components|join(", ")}}</p>
```
jrobbins, line 10: It's great that so few changes were needed to the templates themselves.

settings.py
```diff
@@ -19,6 +19,19 @@
   },
 ]
 
+def flask_compat_get_template_path() -> str:
+  """Returns a path to the templates.
+  Leverages the existing TEMPLATES variable used for Django templates and
+  returns a path usable by Flask/Jinja2.
+
+  This is useful because by default flask.render_template will look for a
```
jrobbins, line 27: Thanks for this clear explanation.

jrobbins (approved):

jcscottiii:
I added rendered templates to the tests.

jrobbins (approved):

PR #2281, opened 2022-09-28
Reusable migration function

DanielRyanSmith:

Much of these migration scripts were mostly similar, so this change writes a generic function that can be reused for most kinds.

### DIFF
--- a/internals/schema_migration.py
+++ b/internals/schema_migration.py
@@ -18,7 +18,8 @@ from google.cloud import ndb
 from framework.basehandlers import FlaskHandler
 from internals.review_models import Activity, Comment
 
-def handle_migration(original_cls, new_cls, migration_cls, kwarg_mapping):
+def handle_migration(original_cls, new_cls, kwarg_mapping,
+    special_handler=None):
   originals = original_cls.query().fetch()
   new_keys = new_cls.query().fetch(keys_only=True)
   new_ids = set(key.integer_id() for key in new_keys)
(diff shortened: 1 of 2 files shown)


internals/schema_migration.py
```diff
@@ -18,6 +18,33 @@
...
+  new_keys = new_cls.query().fetch(keys_only=True)
+  new_ids = set(key.integer_id() for key in new_keys)
+  migration_count = 0
+  for original in originals:
+    # Check if a new entity with the same ID has already been created.
+    # If so, do not create the entity again.
+    if original.key.integer_id() in new_ids:
+      continue
+
+    kwargs = {k : getattr(original, v) for (k, v) in kwarg_mapping}
```
jrobbins, line 32: The variables k and v could be more clear.  E.g., `(new_field, old_field)`.

internals/schema_migration.py
```diff
@@ -18,6 +18,33 @@
...
+    kwargs['id'] = original.key.integer_id()
+
+    # If any fields need special mapping, handle them in the class method.
+    if hasattr(migration_cls, 'handle_special_fields'):
```
jrobbins, line 36: Why pass a class instead of passing a callable?
If you pass in a callable, it could just be None when not needed.

jrobbins (approved):
Clever.  This will make other migration code cleaner.

internals/schema_migration.py
DanielRyanSmith: True, fixed.

internals/schema_migration.py
DanielRyanSmith: Yes, that's a better approach. I've refactored it with this change!

PR #2280, opened 2022-09-28
Write and edit Vote entities when creating Approval entities

DanielRyanSmith:
Part of the schema migration work.

This change implements creation and editing of Vote entities while creating or editing a corresponding Approval entity. Creating, editing, and deleting is handled in class methods in `review_models.py`.

PR #2283, opened 2022-09-30
Migrate ApprovalConfig entities to Gate entities

DanielRyanSmith:

This script writes all existing ApprovalConfig to new Gate entities.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -35,6 +35,9 @@ cron:
 - description: Copy over Feature entities to new FeatureEntry entities
   url: /cron/schema_migration_feature_featureentry
   schedule: 1 of jan 00:00
+- description: Copy over ApprovalConfig entities to new Gate entities
+  url: /cron/schema_migration_approvalconfig_gate
+  schedule: 1 of jan 00:00
 - description: Copy over deprecated standardization field
   url: /cron/write_standard_maturity
   schedule: 1 of jan 00:00
--- a/internals/schema_migration.py
+++ b/internals/schema_migration.py
@@ -17,7 +17,7 @@ from google.cloud import ndb
 
 from framework.basehandlers import FlaskHandler
 from internals.core_models import Feature, FeatureEntry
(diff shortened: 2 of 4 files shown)


jrobbins:
Let's hold on to this PR.  We might not need it because there are actually zero ApprovalConfig entities in our prod datastore.

Maybe instead, we could write the application code to:
1. Create Gates lazily only when one of the Gate fields is updated.  The most common update would be setting a Vote that changes a Gate state.   
2. When displaying Gates: load the ones that exist, but display default data for any defined in internals/processes.py that don't have an entity.

I'll take a hack at a PR to backfill Gates based on existing Approvals (or Votes).

Vladisvob (changes requested):
#2280

Vladisvob (changes requested):
Duplicate of #

Vladisvob (approved):
#2284

Vladisvob (approved):
#2285 #2283

Vladisvob (commented):
#2283

Vladisvob:
> Part of the schema migration work.
> 
> This script writes all existing ApprovalConfig to new Gate entities.

Vladisvob:
> Let's hold on to this PR. We might not need it because there are actually zero ApprovalConfig entities in our prod datastore.
> 
> Maybe instead, we could write the application code to:
> 
> 1. Create Gates lazily only when one of the Gate fields is updated.  The most common update would be setting a Vote that changes a Gate state.
> 2. When displaying Gates: load the ones that exist, but display default data for any defined in internals/processes.py that don't have an entity.
> 
> I'll take a hack at a PR to backfill Gates based on existing Approvals (or Votes).

Vladisvob (approved):

Vladisvob:
#2283

DanielRyanSmith:
Because no ApprovalConfig entities currently exist in production, this migration script should not be necessary. Instead, the Stage entity creation script will be upgraded to create gates for each stage that requires intent approval. Closing this.

PR #2288, opened 2022-10-03
Fix FeatureEntry editing

DanielRyanSmith:
Quick fix for error in staging causing issues with updating features.


pages/guide.py
```diff
@@ -452,7 +452,7 @@ def process_post_data(self, feature_id, stage_id=0):
           int(self.form.get('web_dev_views'))))
     if self.touched('web_dev_views'):
```
DanielRyanSmith, line 453: Should this be `if self.touched('web_dev_views_link'):` ?

jrobbins (approved):
Good catch.

PR #2298, opened 2022-10-04
Indexes for new entities

DanielRyanSmith:

This change adds new indexes based on the fields in the new kinds. Additionally, the redundant quotes are removed from older fields.


jrobbins (approved):
Thanks for the quote cleanup.

PR #2301, opened 2022-10-05
Fix Activity logs to not expand enums

DanielRyanSmith:
When making changes to a feature and notifying subscribed users of changes, fields represented by enums are sent in their expanded readable form. However, Activity entities were erroneously logging this expanded form as well. This change addresses that error.

### DIFF
--- a/internals/core_enums.py
+++ b/internals/core_enums.py
@@ -310,8 +310,11 @@ PROPERTY_NAMES_TO_ENUM_DICTS = {
 
 def convert_enum_int_to_string(property_name, value):
   """If the property is an enum, return human-readable string, else value."""
-  if type(value) != int:
+  # Check if the value is an int or can be converted to an int.
+  try:
+    int_val = int(value)
+  except Exception:
     return value
   enum_dict = PROPERTY_NAMES_TO_ENUM_DICTS.get(property_name, {})
-  converted_value = enum_dict.get(value, value)
+  converted_value = enum_dict.get(int_val, value)
   return converted_value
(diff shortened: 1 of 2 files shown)


internals/core_models.py
```diff
@@ -845,8 +845,6 @@ def _get_changes_as_amendments(self):
       if new_val != old_val:
         if (new_val == '' or new_val == False) and old_val is None:
           continue
-        new_val = convert_enum_int_to_string(prop_name, new_val)
-        old_val = convert_enum_int_to_string(prop_name, old_val)
         amendments.append(
             review_models.Amendment(field_name=prop_name,
             old_value=str(old_val), new_value=str(new_val)))
```
DanielRyanSmith, line 850: Amendment field values are all converted to strings, so when we want to reference them for converting enums to readable form, we need to consider this.

internals/core_enums.py
```diff
@@ -310,8 +310,11 @@
...
-  if type(value) != int:
+  # Check if the value is an int or can be converted to an int.
+  try:
+    int_val = int(value)
+  except Exception:
     return value
   enum_dict = PROPERTY_NAMES_TO_ENUM_DICTS.get(property_name, {})
-  converted_value = enum_dict.get(value, value)
+  converted_value = enum_dict.get(int_val, value)
   return converted_value
```
DanielRyanSmith, line 320: This implementation allows enums to be expanded even if they are passed in as string format (which is how Amendments carry them).

jrobbins (approved):

PR #2308, opened 2022-10-08
Add show all button to metrics page

KyleJu:
Add a Show All button - only show the first 30 entries unless user clicks on `Show All`.  This change shouldn't affect the search field.

### DIFF
--- a/static/elements/chromedash-stack-rank-page.js
+++ b/static/elements/chromedash-stack-rank-page.js
@@ -39,6 +39,7 @@ export class ChromedashStackRankPage extends LitElement {
       type: {type: String}, // "css" or "feature"
       view: {type: String}, // "popularity" or "animated"
       viewList: {attribute: false},
+      topList: {attribute: false},
     };
   }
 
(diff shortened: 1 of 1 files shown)


KyleJu:
@jrobbins does polymer3 still work in Lit? I am using `iron-collapse` because we include it as a dependency in https://github.com/GoogleChrome/chromium-dashboard/blob/56129b4eab24c7217bf996a54a08dfded11ac7cb/package.json#L95, but it is not rendering when I test it locally with fake data

jrobbins:
polymer components should still work.  In fact, chromestatus is using iron-icon.

Is iron-collapse supposed to render anything when closed?  I thought that you need to implement your own "More" button that sets an attribute to make it show its contents.

KyleJu:
Screen Shot 2022-10-13 at 10 06 19 AM

jrobbins (approved):

KyleJu:
Tested on staging - https://b1464fc-dot-cr-status-staging.appspot.com/roadmap

PR #2312, opened 2022-10-10
Handle Stage data when changing feature_type

DanielRyanSmith:
NOTE: These changes is built on top of #2311 and will require that PR to be merged prior to merging this.

This change adds handling of changing the `feature_type` field on the FeatureEntry kind. Because a FeatureEntry entity will have specific Stage entity types based on its `feature_type`, special procedures are needed to prepare Stages with matching `stage_type` fields that correspond to the given `feature_type`. Now changing `feature_type`, new Stage entities will be created as needed and data from older stage fields will be migrated to the new stage entities.

### DIFF
--- a/internals/core_enums.py
+++ b/internals/core_enums.py
@@ -208,6 +208,41 @@ STAGE_TYPES_BY_FIELD_MAPPING = {
     'dt_milestone_webview_start': STAGE_TYPES_DEV_TRIAL
   }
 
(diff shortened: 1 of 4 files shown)


DanielRyanSmith:
To handle this more appropriately and avoid possible data loss as stages are changed in and out, we're making `feature_type` immutable. Closing this to focus on the new approach.

PR #2319, opened 2022-10-11
Some introductory type hinting and refactoring

DanielRyanSmith:
A bit of type hinting to introduce to the repository, as well as small pieces of refactoring.

### DIFF
--- a/internals/user_models.py
+++ b/internals/user_models.py
@@ -113,7 +113,7 @@ class AppUser(ndb.Model):
     rediscache.delete(cache_key)
 
   @classmethod
-  def get_app_user(cls, email) -> users.User:
+  def get_app_user(cls, email: str) -> users.User:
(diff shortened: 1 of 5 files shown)


internals/user_models.py
```diff
@@ -113,7 +113,7 @@ def delete(self, **kwargs):
     rediscache.delete(cache_key)
 
   @classmethod
-  def get_app_user(cls, email):
+  def get_app_user(cls, email) -> users.User:
```
jrobbins, line 116: email should be a str.

jrobbins (approved):
Fun.  But, we need to also set up `pytype` to check it, right?

DanielRyanSmith:
> Fun. But, we need to also set up `pytype` to check it, right?

I'd like to set up type verification, so I'll look into this 🙂

PR #2320, opened 2022-10-11
Remove ability to update feature_type field

DanielRyanSmith:

This change removes the ability to change the `feature_type` field after the feature has been created, as well as adding a warning to users that this feature cannot be updated after creation.

### DIFF
--- a/static/elements/form-field-specs.js
+++ b/static/elements/form-field-specs.js
@@ -209,15 +209,6 @@ export const ALL_FIELDS = {
         Select the most specific category. If unsure, leave as "Miscellaneous".`,
   },
 
-  'feature_type': {
-    type: 'select',
-    choices: FEATURE_TYPES,
-    initial: FEATURE_TYPES.FEATURE_TYPE_INCUBATE_ID[0],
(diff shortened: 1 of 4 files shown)


static/elements/form-field-specs.js
```diff
@@ -1127,9 +1122,7 @@ function makeDisplaySpecs(fields) {
 };
 
 export const DISPLAY_FIELDS_IN_STAGES = {
-  'Metadata': makeDisplaySpecs([
-    'category', 'feature_type', 'intent_stage', 'accurate_as_of',
-  ]),
+  'Metadata': makeDisplaySpecs(['category', 'intent_stage', 'accurate_as_of']),
```
jrobbins, line 1125: I think we should still display the feature_type.

jrobbins (approved):
LGTM after one suggested change.

static/elements/form-field-specs.js
DanielRyanSmith: Agreed. I've added  a change to display it as a disabled field. 👍

PR #2322, opened 2022-10-11
Add helper methods and constants to Gate class

DanielRyanSmith:

This change adds some addtional methods and constants to the Gate entity class. Some of these are redundant with methods and constants found in `approval_defs.py`, but are arguably a better fit in the Gate class. Some more work will be required to change the code to use these methods rather than the functions in `approval_defs.py`.

### DIFF
--- a/internals/review_models.py
+++ b/internals/review_models.py
@@ -12,6 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Import needed to reference a class within its own class method.
+# https://stackoverflow.com/a/33533514
+from __future__ import annotations
+import collections
+
 import datetime
 import logging
 from google.cloud import ndb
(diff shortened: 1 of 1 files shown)


jrobbins (commented):
This logic looks OK, but I think that it would be better to keep it in in approvals_def.py.   There could be a lot new approvals logic and configuration added over time, and putting it here would make the actual model code harder to understand.

DanielRyanSmith:
> This logic looks OK, but I think that it would be better to keep it in in approvals_def.py

I added this and change and then went through and saw the number of places using the logic in `approval_defs.py` that would need to be changed - it's a lot of refactoring work for little gain, and I agree with you now. I'll try and vet these changes a bit more before requesting a PR review in the future. Thanks!
Closing this.

PR #2328, opened 2022-10-13
Fix prepublication earliest-beta calculation.

jrobbins:
Today I checked if the prepublication emails were going out from the prod server.  I think I found a bug in that the calculation for which milestones are relevant always starts by retrieving a current milestone using ?mstone=current in the URL.  That's not right for the prepublication emails because ?mstone=current returns the milestone that is about to go onto the stable channel.  The release on the beta channel has already happened for that milestone.  Instead, we need to fetch the milestone that is about to go onto the beta channel, which can be fetched with ?mstone=beta.

In this PR:
* Add a new constant to configure which milestone the calculation starts with.
* Add some logging and make some minor code improvements

### DIFF
--- a/internals/reminders.py
+++ b/internals/reminders.py
@@ -29,10 +29,10 @@ CHROME_RELEASE_SCHEDULE_URL = (
     'https://chromiumdash.appspot.com/fetch_milestone_schedule')
 
 
-def get_current_milestone_info():
-  """Return a dict of info about the next milestone reaching beta."""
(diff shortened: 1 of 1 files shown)


DanielRyanSmith (approved):
I've found the milestone requesting to be a little confusing myself. I think it's correct, but I wouldn't be surprised if the notifications to verify feature accuracy are slightly off. I'll have to keep an eye on it.

PR #2329, opened 2022-10-13
Script to run all tests

DanielRyanSmith:
Very simplistic script to run most tests checked by the CI. This might be useful to run just before submitting a PR.


jrobbins (approved):
You could even call it `presubmit`.

DanielRyanSmith:
> You could even call it presubmit.

Much better name! I'll make the change.

PR #2331, opened 2022-10-13
Move approval constants into core_enums.py

DanielRyanSmith:
This change is required for additional work on writing and editing Stage, Gate, and Vote entities. Many files were importing `approval_defs.py` to reference a set of constants referencing approvals constants representing the value of the `feature_type` field, or the `field_type` in Approval entities. These constants are needed for methods to be implemented in the Gate and Vote classes, but would require importing `approval_defs.py` into `review_models.py`. This would cause a circular import. Adding these values to `core_enums.py` seemed like the logical conclusion, and made for a fix that is easy to understand.

### DIFF
--- a/api/permissions_api.py
+++ b/api/permissions_api.py
@@ -17,6 +17,7 @@
 from framework import basehandlers
 from framework import permissions
 from internals import approval_defs
+from internals import core_enums
 
 
 class PermissionsAPI(basehandlers.APIHandler):
@@ -32,7 +33,7 @@ class PermissionsAPI(basehandlers.APIHandler):
     # get user permission data if signed in
     user = self.get_current_user()
     if user:
-      field_id = approval_defs.ShipApproval.field_id
+      field_id = core_enums.SHIP_APPROVAL.field_id
       approvers = approval_defs.get_approvers(field_id)
       user_data = {
         'can_create_feature': permissions.can_create_feature(user),
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -34,6 +34,7 @@ from framework import users
 from framework import utils
 from framework import xsrf
 from internals import approval_defs
+from internals import core_enums
 from internals import core_models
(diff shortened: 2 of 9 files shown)


jrobbins:
> These constants are needed for methods to be implemented in the Gate and Vote classes, but would require importing `approval_defs.py` into `review_models.py`. This would cause a circular import. 

What about taking the methods that you want to add to the model classes and writing them as functions in approval_deps.py instead?  E.g., there are already functions `is_approved()` and `is_resolved()` there.

DanielRyanSmith:
> What about taking the methods that you want to add to the model classes and writing them as functions in approval_deps.py instead? E.g., there are already functions is_approved() and is_resolved() there.

I can do this. However, there is a lot of functionality that Gate entities bring that simplifies some of the actions in `approval_defs.py`. For example, evaluating `is_approved()` and `is_resolved()` become a matter of checking the attribute on the Gate entity rather than tallying votes each time approval needs to be checked. It's possible to write these functions in `approval_defs.py` more generally by explicitly passing the Gate object into the function, but there are multiple methods that would need to be implemented in this way, passing only Gate objects in to read/mutate them. At that point it seems like it would make more sense for these functions to live as methods on the Gate object.

Is there a reason to leave these variables in `approval_defs.py`? My understanding is that these are constant values, and are the single source of truth to which gate types map to which intents. I originally had a little difficulty finding exactly which `gate_type` (previously `field_id`) mapped to which intent type because I wasn't aware they lived in this file. My initial thought is that this information would typically be grouped with the `core_enums`.

That's my reasoning, but I will defer to your expertise on whichever approach your think is best 🙂

DanielRyanSmith:
> What about taking the methods that you want to add to the model classes and writing them as functions in approval_deps.py instead?

I'm going to go ahead and close this PR in favor of the workaround at #2341 that implements the change this way 🙂

PR #2341, opened 2022-10-14
Refactor Gate and Vote to update Gate state when setting Votes

DanielRyanSmith:
The original problem this solves involves setting vote values using the new schema. These steps were originally followed:
1. Vote value is changed in the UI.
2. Vote.set_vote() is called.
3. The Gate that corresponds to the vote needs to be checked to see if the value needs to be updated (Approved, Not approved, etc.).

This is where the original problem arose. Checking this requires access to the approval constants/enums so that the approval state can be evaluated. The constants live in `approval_defs.py`, which imports classes from `review_models.py`, therefore we cannot call a function in `approval_defs.py` from `review_models.Vote.set_vote()`, so we cannot update the state of the gate while also setting the vote value.

This change is a workaround to that, which moves `Vote.set_vote()` into `approval_defs.py` and implements a gate state update function alongside it.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -160,7 +160,8 @@ def is_resolved(approval_values, field_id):
 
   return False
 
-def set_vote(feature_id, gate_id, new_state, set_by_email):
+def set_vote(
+    feature_id: int, gate_id: int, new_state: int, set_by_email: str) -> None:
   """Add or update an approval value."""
   if not Vote.is_valid_state(new_state):
     raise ValueError('Invalid approval state')
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -174,12 +174,10 @@ def set_vote(
     existing.set_on = now
     existing.state = new_state
     existing.put()
-    return
-
-  new_vote = Vote(
-      feature_id=feature_id, gate_id=gate_id, state=new_state,
-      set_on=now, set_by=set_by_email)
-  new_vote.put()
+  else:
+    new_vote = Vote(feature_id=feature_id, gate_id=gate_id, state=new_state,
+        set_on=now, set_by=set_by_email)
+    new_vote.put()
 
   # TODO(danielrsmith): As of today, there is only 1 gate per
   # gate type and feature. Passing the gate ID will be required when adding
@@ -203,7 +201,7 @@ def _calc_gate_status(gate: Gate) -> int:
   if ((afd.rule == ONE_LGTM and approvals >= 1) or
       (afd.rule == THREE_LGTM and approvals >= 3)):
     return Vote.APPROVED
-  return Vote.NA
+  return gate.state
 
 def update_gate_approval_state(gate: Gate) -> int:
   """Change the Gate state if it has changed."""
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -15,11 +15,13 @@
 
 import base64
 import collections
+import datetime
 import logging
 import requests
 
 from framework import permissions
-from internals import review_models
+from internals import core_enums
+from internals.review_models import Approval, Gate, OwnersFile, Vote
 import settings
 
 CACHE_EXPIRATION = 60 * 60  # One hour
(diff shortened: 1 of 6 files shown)


internals/approval_defs.py
```diff
@@ -153,7 +155,61 @@ def is_resolved(approval_values, field_id):
...
+
+  now = datetime.datetime.now()
+  existing_list: list[Vote] = Vote.get_votes(
+      feature_id=feature_id, gate_id=gate_id, set_by=set_by_email)
+  if existing_list:
+    existing = existing_list[0]
+    existing.set_on = now
+    existing.state = new_state
+    existing.put()
+    return
```
jrobbins, line 177: I think you need an `else` here instead of `return`.   With `return` the code might not reach the next section of code that updates the gate.

internals/approval_defs.py
```diff
@@ -153,7 +155,61 @@ def is_resolved(approval_values, field_id):
...
+  # gate type and feature. Passing the gate ID will be required when adding
+  # UI functionality for multiple versions of the same stage/gate.
+  gates: list[Gate] = Gate.query(
+      Gate.feature_id == feature_id, Gate.gate_type == gate_id).fetch()
+  if len(gates) > 0:
+    update_gate_approval_state(gates[0])
+
+def _calc_gate_status(gate: Gate) -> int:
+  """Evaluates the current state that this gate should have."""
+  votes = Vote.get_votes(gate_id=gate.key.integer_id())
```
jrobbins, line 194: I think we need `read_consistency=ndb.STRONG` here.

internals/approval_defs.py
```diff
@@ -153,7 +155,61 @@ def is_resolved(approval_values, field_id):
...
+    if vote.state in (Vote.APPROVED, Vote.NA):
+      approvals += 1
+    elif vote.state == Vote.NOT_APPROVED:
+      return Vote.NOT_APPROVED
+  afd = APPROVAL_FIELDS_BY_ID[gate.gate_type]
+
+  if ((afd.rule == ONE_LGTM and approvals >= 1) or
+      (afd.rule == THREE_LGTM and approvals >= 3)):
+    return Vote.APPROVED
+  return Vote.NA
```
jrobbins, line 206: For the default case, let's return `gate.state` meaning that the state of the gate does not change.   I think a new gate instance would start with state NEEDS_WORK or REVEW_REQUESTED, but there will be a way for a user to say that this gate in not applicable for this feature and set it to NA.  So, we want to keep that state if there is not a reason to change it.

internals/approval_defs.py
```diff
@@ -153,7 +155,61 @@ def is_resolved(approval_values, field_id):
...
+
+def update_gate_approval_state(gate: Gate) -> int:
+  """Change the Gate state if it has changed."""
+  was_not_resolved = not gate.is_resolved()
+  gate.state = _calc_gate_status(gate)
+  gate.put()
+  if was_not_resolved and gate.is_resolved():
+    gate.clear_request()
```
jrobbins, line 214: FYI, I'd like to stop deleting the review requests at some point soon.  That can come in a later PR, but just as a heads up: We should keep the review requests because they are part of the feature's history.  To decide which features to show in the "features pending my review" box, we will eventually query for Gate state rather than presence of review requests.

Alternatively (and maybe this is simpler), we could eliminate the concept of a Vote for a review request.  When the user pressed the Request review button, we could record a requester email and timestamp in the Gate and set the Gate state to indicate that a review is requested.  It seemed a little odd that a request is a special kind of vote.

jrobbins (approved):

internals/approval_defs.py
DanielRyanSmith: This is true - I made this change in a previous PR, but I think it got lost in translation up to this point. Fixed!

internals/approval_defs.py
DanielRyanSmith: `Vote.get_votes()` is implemented to always specify this 👍

internals/approval_defs.py
DanielRyanSmith: I've changed it. Perhaps soon it would be useful to map out exactly what circumstance every state should map to, so that we use them consistently, but I think it is not the highest priority.

PR #2342, opened 2022-10-14
Change existing handler classes to properly override request methods

DanielRyanSmith:
This change refactors methods in APIHandler and FlaskHandler subclasses to properly override their parent method counterparts. Technically, the subclass methods previously did not match the signatures of the methods in their parent classes.

APIHandler methods changed:
- do_get
- do_post
- do_patch
- do_delete

FlaskHandler methods changed:
- process_post_data
- get_template_data


api/comments_api.py
```diff
@@ -45,8 +45,9 @@ def _should_show_comment(
     """Check whether a comment should be visible to the user."""
     return comment.deleted_by is None or email == comment.deleted_by or is_admin
 
-  def do_get(self, feature_id: int,
-      field_id: Optional[int]=None) -> dict[str, dict[str, Any]]:
+  def do_get(self, **kwargs) -> dict[str, dict[str, Any]]:
```
jrobbins, line 48: I think the return type should be something more like
dict[str, list[dict[str, Any]]]

jrobbins (approved):
OK, this looks like what I was expecting.
I like that it keeps the abstract methods that are in the base class.

Does pylint complain about all the unused kwargs parameters?  Actually, I just checked and it does, but it complains about a million things, so that is probably a whole separate KR.

api/comments_api.py
DanielRyanSmith: You're right - I think this is part of the original change I made to introduce type hinting, but with no evaluation of whether or not it was correct with mypy - this should be changed in the mypy PR as well

PR #2348, opened 2022-10-17
Remove inline_file. Use external links for js/css

### LINKED ISSUE #2326: Remove embedding of babel generated js/css. Move to serving with <script src> and <link rel="stylesheet" href>
Right now, we embed the css and js on some templates using a custom tag [inline](https://github.com/GoogleChrome/chromium-dashboard/blob/b6bb10195a9eca3935375a995c938beb1e2a74f2/customtags/templatetags/inline_file.py). 

This came up as part of my work in #2303. After I moved from Django to Flask, all the template tests failed after I changed the inline to jinja2 [include](https://jinja.palletsprojects.com/en/3.1.x/templates/#include). The failure was when using this: `{% include "/static/css/main.css" %}` , I get `jinja2.exceptions.TemplateNotFound: /static/css/main.css`. I get similar results for `{% include "../static/css/main.css" %}` 

The cause of the problem: for `include` to work, it looks within the template_folder. template_folder is assigned to `templates`. The code that becomes embedded all resides in a folder called `static`. That folder is a sibling of the `templates` folder.

Snippet of the current file system. You will see a mix of both static and generated content in the `static` folder
```sh
.
├── static
│   ├── components.js
│   ├── css # <--- generated
│   ├── dist # <--- generated
│   ├── elements
│   ├── img
│   ├── js # <--- generated
│   ├── js-src
│   ├── robots.txt
│   ├── sass
│   └── shoelace
├── templates
│   ├── _base.html
│   ├── accuracy_notice_email.html
│   ├── admin
│   ├── blink
│   ├── estimated-milestones-table.html
│   ├── farewell-rss.xml
│   ├── features.html
│   ├── inactive_user_email.html
│   ├── index.html
│   ├── new-feature-email.html
│   ├── prepublication-notice-email.html
│   ├── review-comment-email.html
│   ├── spa.html
│   ├── test_template.html
│   └── update-feature-email.html
```

Accomplish this 3 steps

## Step 1.

Instead, we should serve the javascript and css with <script src> and <link rel="stylesheet" href="">

This will prevent runtime errors of a

Benefits of continuing with inline
- It simplifies the number of HTTP requests
- **Is necessary for email templates (luckily, our email templates do not use inline)**

Benefits of using external scripts
- Can add separate caching policies between html, js and css.
- After the initial load, the browser could be configured to cache the js and css separately. Thus the html downloaded will be smaller since js and css won't be inlined.

Other reasons why to move to serving externally:
- https://stackoverflow.com/a/8284398

## Step 2.

Separated generated babel files and actually served static files (e.g. robots.txt) into a `public` folder

jcscottiii:
Refer to #2326 fo reasoning

Changes in this commit:

- Remove inline_file function
- For js: use script src=""
  - Due to this change, there were some minor tweaks: - create `loadFeatureLegendViews` in feature-page.js. This will allow feature.html to load inlined VIEWS variable without using window global.
- For css: use link rel="stylesheet" href=""
- Add helper function "get_static_path" to get the path to the static folder

Part 1 of 2 to fix #2326


framework/basehandlers.py
```diff
@@ -552,6 +552,7 @@ def FlaskApplication(import_name, routes, post_routes, pattern_base='', debug=Fa
   """Make a Flask app and add routes and handlers that work like webapp2."""
 
   app = flask.Flask(import_name,
+    static_folder=settings.get_static_path(),
```
jrobbins, line 555: Doesn't it default to 'static'?

templates/spa.html
```diff
@@ -44,7 +43,7 @@
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
   <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&amp;display=swap" rel="stylesheet">
 
-  <style>{% inline_file "/static/css/main.css" %}</style>
+  <link rel="stylesheet" href="/static/css/main.css">
```
jrobbins, line 46: Let's add `?v={{app_version}}` to each of these URLs.  

We use that in existing client-side includes to cause the client to load the newest version of these assets whenever we deploy a new version to GAE.

jrobbins (changes requested):

framework/basehandlers.py
jcscottiii: good catch! fixed. 

Also, as a result of you bringing this up, I did some more digging for the existing template_folder parameter I added earlier. I added static_folder here because I did it for template_folder. It turns out, I only need  to set the template_folder for the unit tests. I will be removing it here in #2323

jrobbins (approved):
Awesome, this is progress.

framework/basehandlers.py
jcscottiii: Correction, I will be doing this in the PR that removes django completely.

PR #2351, opened 2022-10-17
Move source for generated frontend to client-src

### LINKED ISSUE #2326: Remove embedding of babel generated js/css. Move to serving with <script src> and <link rel="stylesheet" href>

jcscottiii:
Fixes #2326 

Any source used to generate final static files is moved to a folder named client-src.
This PR moves those files and updates any references.


jrobbins (approved):
Nice

Issue #2343, opened 2022-10-14
Implement pagination in feature search API and newfeatures UI


jrobbins:
**Is your feature request related to a problem? Please describe.**
We are building a new feature list page that will be faster and more conventional, but it is still a work-in-progress.

**Describe the solution you'd like**
Implement pagination in the search API and the newfeatures page UI:
- [x] Accept optional `start` and `num` URL parameters in api/features_api.py `do_search()` and pass them to `process_query()`.
- [x] Add unit tests for pagination to search_test.py and features_api.py
- [x] Add pagination widgets to chromedash-feature-table
   * They should have the count, next, and prev buttons something like https://material.io/components/data-tables#anatomy
   * Skip the rows-per-page option for now and always use implicit num=100
   * Locate the pagination widgets at the top of the table because the table could be long 
- [x] When navigating to a new pagination page:
   * [x] show some kind of loading animation or skeletons
   * [x] update the start= query string parameter in the URL

There should be some properties that control the behavior of chromedash-feature-table:
- [x] `start` defaults to zero
- [x] `num` defaults to 100
- [x] `alwaysOffePaginate` if true, offer the pagination widgets
- [x] `usequerystring` if true, initialize `start` and `num` from the query string, and reflect any changes back to the query string  by pushing a new navigation state (allowing browser back button to undo pagination).

We will set `userquerystring` on the feature list, but not when this element is used for boxes on the myfeatures page.

PR #2352, opened 2022-10-17
Add pagination parmeters to feature search API.

jrobbins:
This adds pagination parameters `start=` and `num=` to the search functionality of features_api.
These were already supported in the underlying feature search functionality.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -55,8 +55,20 @@ class FeaturesAPI(basehandlers.APIHandler):
 
     user_query = self.request.args.get('q', '')
     sort_spec = self.request.args.get('sort')
+    try:
+      start = int(self.request.args.get('start', 0))
+      num = int(self.request.args.get(
+          'num', search.DEFAULT_RESULTS_PER_PAGE))
+    except ValueError:
+      self.abort(400, msg='Invalid start or num')
+    if start < 0:
+      self.abort(400, msg='Parameter out of range: start')
+    if num < 0:
+      self.abort(400, msg='Parameter out of range: num')
+
     features_on_page, total_count = search.process_query(
-        user_query, sort_spec=sort_spec, show_unlisted=show_unlisted_features)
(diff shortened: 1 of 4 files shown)


internals/search.py
```diff
@@ -194,7 +195,7 @@ def _sort_by_total_order(result_id_list, total_order_ids):
 
 def process_query(
     user_query, sort_spec=None, show_unlisted=False, show_deleted=False,
-    start=0, num=100):
+    start=0, num=DEFAULT_RESULTS_PER_PAGE):
```
KyleJu, line 198: now that user can decide `start` and `num`, we should check array out of index in  `paginated_id_list = sorted_id_list[start : start + num]`, at line 241. e.g.  sorted_id_list[min(start, len(paginated_id_list))  : min(start + num, len(paginated_id_list))

api/features_api.py
```diff
@@ -55,8 +55,20 @@ def do_search(self) -> dict[str, Any]:
...
     sort_spec = self.request.args.get('sort')
+    try:
+      start = int(self.request.args.get('start', 0))
+      num = int(self.request.args.get(
+          'num', search.DEFAULT_RESULTS_PER_PAGE))
+    except ValueError:
+      self.abort(400, msg='Invalid start or num')
+    if start < 0:
+      self.abort(400, msg='Parameter out of range: start')
+    if num < 0:
```
KyleJu, line 66:
```suggestion
    if num <= 0:
```
Make it at least 1?

jrobbins: I considered that, but it turns out that requesting with num=0 is potentially useful because the response includes the total number of results.

internals/search.py
jrobbins: Python does the right thing for us automatically.  If these values exceed the length of the list, the expression returns only up to the end of the list, or an empty list if appropriate.

KyleJu (approved):

jrobbins:
This is part of #2343.

PR #2354, opened 2022-10-18
Add num query param for metrics data

KyleJu:
Backend for fixing https://github.com/GoogleChrome/chromium-dashboard/issues/2267. Add the **num** query param for metrics data APIs and add testing for the metricsdata.FeatureHandler class


jrobbins (approved):
LGTM after considering two comments

PR #2356, opened 2022-10-18
Add assertions for the templates in #2355 (Part 2 of 2)

jcscottiii:
This is a split of https://github.com/GoogleChrome/chromium-dashboard/pull/2323. In order to make that PR readable, I split that up. This commit contains the assertions. Depends on #2355 

- Move the TESTDATA functionality to a testing_config to be a test helper. In this new helper, users can still retreive from it like a dictionary. In addition, there's a helper function to help make goldens that can be used to regenerate the golden templates. Add commented out code to use it.
- Add some clean ups to individual tests or test cases. This was needed because some of the existing rendering tests would pull from the database. But since the tests only asserted the existence of some values, it did not matter. Now that we assert a complete template, everything in the database matters. (In the future, we can add mocks for all of those)
  - These cleanups also help for more than the rendering tests. Before these fixes, there were ~15 non-rendering tests that would fail if a developer ran the full suite with a clean db, then re-ran a test indivdually without resetting the database. There are still 2 non-rendering tests that fail without clearing the database first. But this first iteartion got a lot of them.
- Add a missing test for the PrepublicationHandler to increase coverage

**WHY merge this PR**: Prior to removing Django, it would be nice to have a baseline of exactly how everything renders

### DIFF
--- a/internals/notifier_test.py
+++ b/internals/notifier_test.py
@@ -14,10 +14,8 @@
 
 import collections
 import json
-import os
 import testing_config  # Must be imported before the module under test.
 from datetime import datetime
-from pathlib import Path
 
 import flask
 from unittest import mock
@@ -37,9 +35,7 @@ test_app = flask.Flask(__name__,
   template_folder=settings.flask_compat_get_template_path())
 
 # Load testdata to be used across all of the CustomTestCases
-TESTDATA = testing_config.Testdata(
-  os.path.abspath(os.path.dirname(__file__)),
-  Path(__file__).stem)
+TESTDATA = testing_config.Testdata(__file__)
 
 class EmailFormattingTest(testing_config.CustomTestCase):
 
@@ -501,10 +497,7 @@ class FeatureStarTest(testing_config.CustomTestCase):
 
     actual = notifier.FeatureStar.get_user_stars(email)
     expected_ids = [feature_1_id, feature_2_id, feature_3_id]
-    self.assertEqual(
-        sorted(expected_ids,
-                      reverse=True),
-        actual)
+    self.assertEqual(sorted(expected_ids, reverse=True), actual)
     # Cleanup
     for feature_id in expected_ids:
       notifier.FeatureStar.get_star(email, feature_id).key.delete()
--- a/internals/notifier_test.py
+++ b/internals/notifier_test.py
@@ -37,15 +37,9 @@ test_app = flask.Flask(__name__,
   template_folder=settings.flask_compat_get_template_path())
 
 # Load testdata to be used across all of the CustomTestCases
-TESTDATA = {}
-testdata_dir = os.path.join(
+TESTDATA = testing_config.Testdata(
   os.path.abspath(os.path.dirname(__file__)),
-  'testdata',
-  Path(__file__).stem
-  )
-for filename in os.listdir(testdata_dir):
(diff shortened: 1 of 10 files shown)


internals/notifier_test.py
```diff
@@ -496,10 +500,14 @@ def test_get_user_stars__some_stars(self):
     notifier.FeatureStar.set_star(email, feature_2_id)
 
     actual = notifier.FeatureStar.get_user_stars(email)
+    expected_ids = [feature_1_id, feature_2_id, feature_3_id]
     self.assertEqual(
-        sorted([feature_1_id, feature_2_id, feature_3_id],
+        sorted(expected_ids,
```
jrobbins, line 505: nit: Since this expression is shorter now, it can fit on one line.

pages/users_test.py
```diff
@@ -16,30 +16,52 @@
...
 
 from internals import user_models
 from pages import users
 
 test_app = flask.Flask(__name__)
 
 
+# Load testdata to be used across all of the CustomTestCases
+TESTDATA = testing_config.Testdata(
+  os.path.abspath(os.path.dirname(__file__)),
```
jrobbins, line 33: Nit: This repeated code could probably be simpler it it only passed `__file__` and the Testdata constructor did all the OS path stuff.

pages/users_test.py
```diff
@@ -16,30 +16,52 @@
...
 
     with test_app.test_request_context('/request_path'):
       self.template_data = self.handler.get_template_data()
+      self.template_data.update(self.handler.get_common_data())
+      self.template_data['nonce'] = 'fake nonce'
+      self.template_data['xsrf_token'] = ''
+      self.template_data['xsrf_token_expires'] = 0
     self.full_template_path = self.handler.get_template_path(self.template_data)
+    self.maxDiff = None
+  def tearDown(self):
```
jrobbins, line 55: Style: need blank line before a method definition.

testing_config.py
```diff
@@ -89,3 +89,31 @@ def run(self, result=None):
...
+    Common pattern to place the testdata in the following format:
+
+    Given a test file, atest_test.py, and it is located at
+    /some/module/atest_test.py.
+
+    The testdata should be located at /some/module/testdata/atest_test/
+    """
+    self.testdata = {}
+    self.testdata_dir = os.path.join(
+    test_file_abs_directory, 'testdata', test_file_name)
```
jrobbins, line 106: Style: This continued line should be indented 4 spaces from the line above.  Maybe it is, but it does not look that way on GitHub.

internals/reminders_test.py
```diff
@@ -46,29 +55,47 @@ def make_test_features():
...
+    self.feature_template.key = ndb.Key('Feature', 123)
+    # This test does not require saving to the database.
+
+    self.maxDiff = None
 
   def tearDown(self):
-    self.feature_1.key.delete()
-    self.feature_2.key.delete()
-    self.feature_3.key.delete()
+    pass
```
jrobbins, line 70: If there is nothing to tear down, this method can be simply deleted.

jrobbins (approved):
LGTM after considering several style-related comments.

Thanks for making tests pass individually.  I have not been running them that way, but now that we have a lot more tests, I can see how it will be a time-saver to be able to run them individually and repeatedly without the DB emulator reset.

PR #2357, opened 2022-10-18
Unify int arg parsing

KyleJu:
Unify int arg parsing. Will change the usage in a follow-up


jrobbins (approved):
Perfect.  I was thinking of writing almost the same code.

PR #2358, opened 2022-10-18
Remove all references to django

jcscottiii:
Replace any functionality with flask


templates/_base.html
```diff
@@ -74,8 +74,7 @@
 </head>
 
 <body class="loading" data-path="{{current_path}}">
-  {% comment %}
-  {% endcomment %}
+  {# #}
```
jrobbins, line 77: This could be deleted.

framework/utils.py
```diff
@@ -23,7 +23,7 @@
 from framework import users
 import settings
 
-from django.utils import feedgenerator
+from feedgenerator.django.utils import feedgenerator
```
jrobbins, line 26: Thanks for figuring out how to make this still work without needing all of django.  However, the only code that uses this is no longer reachable.  In a follow-up PR we can remove it, and FeatureListXMLHandler and render_atom_feed.

jrobbins (approved):
Woo-hoo!  Our dependencies are getting lighter.

framework/utils.py
jcscottiii: Created https://github.com/GoogleChrome/chromium-dashboard/issues/2366

PR #2360, opened 2022-10-19
Upgrade filter widgets to shoelace.

jrobbins:
In the feature filter element (which appears near the top of the new feature list page), replace basic HTML widgets with shoelace widgets.  This makes the look and feel consistent with other shoelace widgets on other pages, and it simplifies our code.

In this PR:
* Replace `<input>` with `<sl-input>` and `<select>` with `<sl-select>` and make associated changes.
* Replace the previous `#searchbar #searchbox` elements with a single `<sl-select>` that has prefix and suffix icons.
* Replace `iron-icon` with `sl-icon-button` and add the needed icons under static/shoelace/assets/icons.
* Using `sl-select` for the menu to insert a row now needs a little more explicit logic to reset it for the next use and to avoid processing that value change like a user-initiated value change.
* Clarified that the "contains" operator (when we implement support for it) will be word search, not arbitrary substrings.


KyleJu (approved):

KyleJu:
> Old screenshot: image
> 
> New screenshot: image

The colors for `Select field name` is lighter than `Name` or `Updated`?

client-src/elements/chromedash-feature-filter.js
```diff
@@ -215,35 +215,21 @@ class ChromedashFeatureFilter extends LitElement {
...
-        --iron-icon-width: 18px;
+      sl-icon-button {
+        font-size: 1.6rem;
+        margin: 0 !important;
       }
-      button.searchbutton {
-        padding: 2px 8px;
-        font-size: inherit;
+      sl-input sl-icon-button {
+        font-size: 1.4rem;
```
KyleJu, line 223: font-size is also set at 1.6rem above?

client-src/elements/chromedash-feature-filter.js
```diff
@@ -252,16 +238,16 @@ class ChromedashFeatureFilter extends LitElement {
...
       .cond-op-menu {
-        width: 80px;
+        width: 150px;
       }
       .filterrow label {
         padding-right: 4px;
         align-self: center;
       }
-      .filterrow input {
+      .filterrow sl-input {
```
KyleJu, line 250: merge the `sl-input` to the `sl-input` in line 222

client-src/elements/chromedash-feature-filter.js
jrobbins: These are two different sizes for two different contexts.  The icons that stand alone are a little bigger, whereas the ones that are part of the search input field are a little smaller to be more similar to the size of the text in the input field.

jrobbins:
> The color for `Select field name` is lighter than `Name` or `Updated`?

Yeah, it is.  It's placeholder text that tells the user to select an actual value from that sl-select.   Whereas, the rows above are actual values.   Knowing that explanation, I think it makes sense.  But, you're right that it could strike some users as odd and distracting.  I think that they will still be able to accomplish their goals.   Let's just keep it in mind for now.  If any users actually give that feedback, I can make the placeholder text use the non-placeholder color.

client-src/elements/chromedash-feature-filter.js
jrobbins: They are very different CSS selectors.  The one on line 222 is for icons inside the search query input field, whereas this one is for query values.  CSS can be so confusing sometimes.  I am glad that Lit allows us to deal with small pieces of it.

client-src/elements/chromedash-feature-filter.js
jrobbins: These are two different sizes for two different contexts. The icons that stand alone are a little bigger, whereas the ones that are part of the search input field are a little smaller to be more similar to the size of the text in the input field.

jrobbins: These are two different sizes for two different contexts. The icons that stand alone are a little bigger, whereas the ones that are part of the search input field are a little smaller to be more similar to the size of the text in the input field.

PR #2364, opened 2022-10-19
Allow word-wrap in pending-my-approval highlights row

jrobbins:
This resolves a problem where the API owners could not see the approval icon because it was pushed too far off to the side.


DanielRyanSmith (approved):
Looks like the problem has been solved to me. 👍 
Screen Shot 2022-10-19 at 10 48 50 AM

PR #2369, opened 2022-10-19
Schema migration staging fixes 2022-10-19

DanielRyanSmith:
- Fixed approval_defs.set_vote() to write `Vote.gate_id` properly. Previously was erroneously writing `gate_type` to `gate_id`, as well as passing in `gate_type` in a number of places with the name of `gate_id`.
- Created Gate entities in tests to allow Vote entities to be properly written with a corresponding `gate_id`.
- Incremented stage counts in schema migration script properly reflect the amount of stages created.
- Removed Vote double-write call in Approval.set_approval(). This is no longer needed as approval_defs.set_vote() is being called in the places it is necessary.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -166,7 +166,7 @@ def set_vote(
     gate_id: Optional[int]=None) -> None:
   """Add or update an approval value."""
   gate: Optional[Gate] = None
-  if not gate_id:
+  if gate_id is None:
     gate = get_gate_by_type(feature_id, gate_type)
     # If a vote is being set, a corresponding gate should already exist.
     if not gate:
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -17,6 +17,7 @@ import base64
 import collections
 import datetime
 import logging
+from typing import Optional
 import requests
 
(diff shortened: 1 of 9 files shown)


internals/approval_defs.py
```diff
@@ -161,8 +162,21 @@ def is_resolved(approval_values, field_id):
   return False
 
 def set_vote(
-    feature_id: int, gate_id: int, new_state: int, set_by_email: str) -> None:
+    feature_id: int,  gate_type: int, new_state: int, set_by_email: str,
+    gate_id: Optional[int]=None) -> None:
   """Add or update an approval value."""
+  gate: Optional[Gate] = None
+  if not gate_id:
```
jrobbins, line 169: This would be more clear if you reversed the condition and swapped the then and else branches.

jrobbins (approved):
LGTM after considering one style suggestion

internals/approval_defs.py
DanielRyanSmith: This went through some iterations and some style got lost in translation here - I've updated this line to
```
if gate_id is None:
  ...
```
which is hopefully more legible.

PR #2371, opened 2022-10-19
Upgrade feature-table icons to shoelace.

jrobbins:
This is related to something I tried and failed to do in PR #2364:  many of the iron-icon elements on our pages are not tab-stops, which means that a user who cannot use a mouse will have a hard time accessing them.    Some of the iron-icon elements are wrapped in `<a href="#">` elements, which does make them keyboard accessible, but using an icon that has that can cause the page to scroll to the top when an empty anchor (`#`) is added to the end of the URL.

The sl-icon-button element seems to work better.  It is keyboard accessible, and it does not alter the URL or scroll to the top.


jrobbins:
Old screenshot (with weird-looking keyboard focus indicator):
image

New screenshot:
image

client-src/elements/chromedash-feature-table.js
```diff
@@ -207,34 +205,32 @@ class ChromedashFeatureTable extends LitElement {
 
   renderApprovalsIcon(feature) {
     return html`
-      <a class="tooltip"
+      <sl-icon-button
         @click="${() => this.openApprovalsDialog(feature)}"
-        title="Review approvals">
-        <iron-icon icon="chromestatus:approval"></iron-icon>
+        title="Review approvals"
+        name="approval" ></sl-icon-button>
```
KyleJu, line 211:
```suggestion
        name="approval"></sl-icon-button>
```

KyleJu (approved):

jrobbins:
After working on #2379 I realized that the material design icons are better than the bootstrap icons.  The bootstrap icons have lines that are just too thin.  I copied some a change in components.js from that PR here, so I will hang on to this PR until the overlapping code has been reviewed.

KyleJu (approved):

PR #2375, opened 2022-10-20
Implement new feature list pagination UI.

jrobbins:
This is progress on #2343.

In this PR:
* Rename some properties to match the server-side pagination parameter names `start` and `num`.
* Add a new property to `chromedash-feature-table` for `alwaysShowPagination` (see below)
* Render the pagination counts and buttons in the cases where they are needed
* Implement a click handler that requests the desired pagination page
* Increase the size of the "My features" boxes to 25 rows to make pagination less likely to show there

I think that the pagination info should always be shown on the new "All features" page.   The count of features is in the pagination area, and that is useful information even if everything fits onto one pagination page.  It is useful because a user is typing in a query that is specific to some task that the user is doing at that time.

We also use `chromedash-feature-table` for each box on the "My features" page.  I don't want to show pagination widgets there unless needed.  That is because the pagination widgets would take up space in each box and they are less likely to be useful.   The main use cases for the "My features" page are for users to quickly access a small set of features that they are already familiar with or that they need to take action on.   On this page, the pagination widgets are only shown in the fairly rare case where there are too many features to fit in a box (who's size is limited to avoid pushing other boxes too far down the page).


jrobbins:
Screenshot of "My features" page with as-needed pagination (click to zoom the screenshot):
image

Screenshot of "All features" page.  I temporarily set num=25 to demo this with limited test data that I have:

client-src/elements/chromedash-feature-table.js
```diff
@@ -338,6 +355,45 @@ class ChromedashFeatureTable extends LitElement {
     return nothing;
   }
 
+  shiftStartBy(delta) {
```
KyleJu, line 358: The function description could be more accurate than `shiftStartBy`. It reserves the space and load the feature for the next/previous page. How about something like "populateCurrentPage"?

client-src/elements/chromedash-feature-table.js
```diff
@@ -338,6 +355,45 @@ class ChromedashFeatureTable extends LitElement {
...
+      // On MyFeatures page, don't always show pagination.  Omit it if
+      // results fit in each box (the most common case).
+      if (this.loading || (first == 1 && last == this.totalCount)) {
+        return nothing;
+      }
+    }
+
+    return html`
+      <div class="pagination">
+        <span>${first} - ${last} of ${this.totalCount}</span>
```
KyleJu, line 384: This can simply be `this.features.length`

client-src/elements/chromedash-feature-table.js
```diff
@@ -338,6 +355,45 @@ class ChromedashFeatureTable extends LitElement {
...
   }
 
+  shiftStartBy(delta) {
+    const proposedStart = this.start + delta;
+    this.start = clamp(proposedStart, 0, this.totalCount - 1);
+    this.fetchFeatures();
+  }
+
+  renderPagination() {
+    const first = this.start + 1;
```
KyleJu, line 365: Nit: `first` and `last` variable names can be more descriptive. `OfferLeft` and `offerRight` can be `hasPreviousPage` and `hasNextPage`

KyleJu: Also why do we need to `+ 1` here?

client-src/elements/chromedash-feature-table.js
```diff
@@ -338,6 +355,45 @@ class ChromedashFeatureTable extends LitElement {
     return nothing;
   }
 
+  shiftStartBy(delta) {
+    const proposedStart = this.start + delta;
+    this.start = clamp(proposedStart, 0, this.totalCount - 1);
+    this.fetchFeatures();
+  }
+
+  renderPagination() {
```
KyleJu, line 364: In the case when `this.loading=true` and `return html`<div class="pagination"></div>`;` (line382) is rendered, what data binding will trigger `renderPagination` to reload and show line 382?

client-src/elements/chromedash-feature-table.js
jrobbins: I see your point.  I changed it to `loadNewPagination()` page.

client-src/elements/chromedash-feature-table.js
jrobbins: I renamed the variables.

The `+1` is to make the index one-based rather than zero-based.  Humans prefer to see one-based index numbers in the UI.   There is no `+ 1` for the last index because the index of the last element of an array is normally `length - 1`, so it would be `length -1 + 1`, so I just left those off.

client-src/elements/chromedash-feature-table.js
jrobbins: `this.loading` is a property of `ChromedashFeatureTable`.  'this.loading' is set true by `fetchFeatures()` when a request to the server is initiated and then set to false when the response has been processed.  Any change to a Lit element property causes `render()` to eventually be called, which in turn calls `renderPagination()`.

KyleJu (approved):

PR #2392, opened 2022-10-25
Implement simple full-text search using NDB

jrobbins:
Our new feature list page needs server-side full-text search.   Rather than add a new dependency on full-text search engine that would complicate our operational set-up, I have made a simple full-text search engine using NDB.  The basic concept is the FeatureWords model that has a list of words.  Each word can be searched for, which can give us the feature_ids of the matching features.

In this PR:
* search_fulltext: new FeatureWords model class and methods to index and search for features.  Also a couple of handlers to reindex all features and to list the 100 most common words (in case we want to make some of them stop words)
* search: use search_fulltext rather than say that full-text terms are unsupported.
* main: add URLs for the admin handlers
* guide: add a call to index features when they are created or edited


jrobbins:
I know that I still need to add unit tests, but the bones of the new functionality are there for you to review now.

jrobbins:
FYI, you can try it out at https://cr-status-staging.appspot.com/newfeatures

internals/search_fulltext.py
```diff
@@ -0,0 +1,216 @@
...
+
+def search_fulltext(textterm):
+  """Return IDs of features that have some word(s) from phrase."""
+  search_words = parse_words([textterm])
+  if not search_words:
+    logging.warn('Cannot process fulltext term: %r', textterm)
+    return []  # user is searching for stop words.
+
+  logging.info('looking for words: %r', search_words)
+  query = FeatureWords.query(FeatureWords.words.IN(search_words))
```
DanielRyanSmith, line 163: What is the result of this? Does it return any FeatureWords entity that contains every word specified in the search?

internals/search_fulltext.py
```diff
@@ -0,0 +1,216 @@
...
+    if not s: continue
+    s = s.lower().replace("'", "")
+    words.update(WORD_RE.findall(s))
+
+  words.difference_update(STOP_WORDS)
+  return words
+
+
+def batch_index_features(fe_list, existing_fw_list):
+  """XXX."""
```
DanielRyanSmith: Docstring needs filled out here 🙂

internals/search_fulltext.py
```diff
@@ -0,0 +1,216 @@
...
+  return strings
+
+
+def parse_words(strings):
+  """Return a list of all searchable words in the given feature entry."""
+  words = set()
+  for s in strings:
+    # skip missing fields and eliminate apostrophes
+    if not s: continue
+    s = s.lower().replace("'", "")
```
DanielRyanSmith, line 120: Would it be worth adding a `strip()` invocation here? Is there any chance a whitespace character gets here? (newline, etc.)

internals/search_fulltext.py
jrobbins: It will match any FeatureWords entity that has any word in search_words.
Which is wrong... we want only those that have every word.  I will change that.

internals/search_fulltext.py
```diff
@@ -0,0 +1,216 @@
...
+        all_feature_entries, all_feature_words)
+    # Don't actually store them, we just want to report stats.
+
+    counts = collections.Counter()
+    for fw in updated_fw_list:
+      counts.update(fw.words)
+
+    logging.info('100 most common words')
+    for word, count in counts.most_common(100):
+      logging.info('%6d: %s', count, word)
```
DanielRyanSmith, line 215: I haven't used a Counter since interview practice 😆 Great time to use it!

main.py
```diff
@@ -200,9 +201,11 @@
   ('/cron/schema_migration_comment_activity', schema_migration.MigrateCommentsToActivities),
   ('/cron/schema_migration_entities', schema_migration.MigrateEntities),
   ('/cron/write_standard_maturity', deprecate_field.WriteStandardMaturityHandler),
+  ('/cron/reindex_all', search_fulltext.ReindexAllFeatures),
 
-  ('/tasks/email-subscribers', notifier.FeatureChangeHandler),
+  ('/admin/find_stop_words', search_fulltext.FindStopWords),
```
DanielRyanSmith, line 206: `/admin/find_stop_words` requires the cron header - does that affect executing this?

internals/search_fulltext.py
jrobbins: Filled in now.

internals/search_fulltext.py
jrobbins: There could certainly be some whitespace before, between, and after any words in the string.  The entire string could also be whitespace.  All of that should work out fine when we call `findall()` on the next line.  I'll add some test cases to verify that.

internals/search_fulltext.py
jrobbins: In my next PR, I'll show how to sort millions of unsigned short ints in O(n) time :).

main.py
jrobbins: I discovered that require_cron_header() allows the handler to run if the user is a site admin.  So, I am basically using that to prevent random visitors from hitting that URL to burn our CPU time.   It also means that the various migration scripts that we want to run in GAE don't need cron.yaml entries to be run, so long as you are a site admin.  In fact, maybe their URLs should start with /admin.

jrobbins:
I believe that I have responded to your first round of comments.
I have fleshed out the tests and fixed some defects that the tests found.
There are still some TODO comments there, but I think that they can be addressed in follow-up PRs to keep the scope of this PR reasonable.
Also, I have not added type specifications yet, but I will do that in a follow-up PR that focuses only on that.

internals/search_fulltext.py
```diff
@@ -0,0 +1,222 @@
...
+  """Return IDs of features that have some word(s) from phrase."""
+  search_words = parse_words([textterm])
+  if not search_words:
+    logging.warning('Cannot process fulltext term: %r', textterm)
+    return None  # user is searching for stop words.
+
+  logging.info('looking for words: %r', search_words)
+  query = FeatureWords.query()
+  for sw in search_words:
+    query = query.filter(FeatureWords.words == sw)
```
DanielRyanSmith, line 166: I think you need to filter out stop words that the user has searched for as well here. For example on staging, if I search "old busted", I see the result for "Old and busted feature", but if I search for "old and busted" I get no results.

jrobbins: I believe that I fixe that in a commit yesterday.  I just deployed to staging again and the "old and busted" query works now.

DanielRyanSmith: Yes, I can confirm it works now 👍

DanielRyanSmith (approved):
Great addition!

Issue #2377, opened 2022-10-21
Refactor format_for_template() into new file for legacy data feeds


jrobbins:
**Is your feature request related to a problem? Please describe.**
We need to support the legacy data feed that we host at /features.json and /features_v2.json, but we don't want it to constrain or complicate the API code that we use for our UI and or future API efforts.

**Describe the solution you'd like**
- [ ] Define a new file for legacy data feeds and move the handler for /features.json and /features_v2.json there (it is just one handler).
- [ ] Move format_for_template() to that new file
- [ ] Change our caching logic to always cache the NDB model objects, not the output of format_for_template().  
- [ ] Change the return types of many of our functions from a "formatted feature" to a NDB Feature model object.
- [ ] Call that legacy format_for_template() function from our API code for now, but note that we could change the JSON representation of features soon.

jrobbins:
I updated the issue description to cover everything that we discussed today.
I'll send you the PR that I have for the first checkbox (delete version=1) today.

PR #2394, opened 2022-10-25
Remove unused feature json version=1

jrobbins:
This is progress on #2377.  We are removing the version=1 format for our JSON responses because it is unused.

We have handlers for URLs /features.json and /features_v2.json, but they both serve the version=2 JSON and have been that way for at least a year.  There was a path through the code with a query string parameter to request the version number, but our app does not use it  and I don't see any outside use of it in the logs.

In this PR:
* Delete the version=1 logic in format_for_template() and update all callers


DanielRyanSmith:
Per the mypy error, we could update the `DictModel.format_for_template()` implementation to add the `id` field under the same circumstances that `Feature.format_for_template()` does, which is by checking if `is_saved()` returns True.

DanielRyanSmith (approved):

jrobbins:
> Per the mypy error, we could update the `DictModel.format_for_template()` implementation to no longer take an additional argument and to instead add the `id` field under the same circumstances that `Feature.format_for_template()` does, which is by checking if `self.is_saved()` returns True.

It turns out that DictModel.format_for_template() was never called.

PR #2399, opened 2022-10-26
Rename NOT_APPROVED to DENIED.

jrobbins:
Monorail FLT used to offer an approval value "Not approved", but the internal launch tool uses "Denied".
Our users will have most recently used the launch tool, and they will continue to do use it in some cases, so we want to align our terminology with that tool.

This is not a schema change because the integer value store is the same as it has been, it is just a code and UI strings change.


KyleJu (approved):

KyleJu:
+1 to "This is not a schema change because the integer value store is the same as it has been, it is just a code and UI strings change." I see the benefit of using integer value in schema

PR #2405, opened 2022-10-26
Add type annotations to search-related files.

jrobbins:
I started off adding types to the full-text file, but then I went wild.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -16,6 +16,9 @@
 import datetime
 import logging
 import re
+from typing import Any, Union
+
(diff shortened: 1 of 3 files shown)


DanielRyanSmith (approved):
Love it! I don't know about you, but VSCode picking up the types here is really making it clear what I'm working with and is helping write unambiguous, clean code. 🙂

Whenever we move to Py3.10, we have [a new way of expressing Optional/Union](https://docs.python.org/3/library/typing.html#typing.Union).

Looks great!

PR #2406, opened 2022-10-26
Adjust intent parsing to cope with more variations.

jrobbins:
This intent slipped past our intent-detection logic because the email sender wrote their own section header name:
https://groups.google.com/a/chromium.org/g/blink-dev/c/JNOQvsTxecI

They wrote "Chrome Platform Status page:", which is reasonable, so we might as well try to accept it.

### DIFF
--- a/internals/detect_intent_test.py
+++ b/internals/detect_intent_test.py
@@ -162,6 +162,17 @@ class FunctionTest(testing_config.CustomTestCase):
         5144822362931200,
         detect_intent.detect_feature_id(body))
 
+  def test_detect_feature_id__alternative(self):
+    """We can parse the feature ID from another common link."""
+    body = (
+        'blah blah blah\n'
+        'Entry on the feature dashboard\n'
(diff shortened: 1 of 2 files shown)


internals/detect_intent_test.py
```diff
@@ -162,6 +162,17 @@ def test_detect_feature_id__alternative(self):
         5144822362931200,
         detect_intent.detect_feature_id(body))
 
+  def test_detect_feature_id__alternative(self):
```
KyleJu, line 165: `test_detect_feature_id__alternative` already exists

KyleJu (approved):
LGMT with a comment

internals/detect_intent_test.py
jrobbins: D'oh.   Editing error on my part.  That test method is an exact copy of the one above it.  The new code is below.

KyleJu (approved):
LGTM again on the tests

PR #2407, opened 2022-10-26
Move Feature methods to helper files

DanielRyanSmith:
Part of #2377

Changes:
- Many Feature class methods moved to new `feature_helpers.py`, similar to `approval_defs.py`, to simplify the Feature class and remove possibility of circular imports (required for refactoring `Feature.format_for_template()`).
- Move `Feature.__notify_feature_subscribers_of_changes()` to new helper file `notifier_helpers.py`.
- `Feature.format_for_template()` is now `feature_helpers.format_for_template(feature)`.
- Refactor `BaseHandler.get_specified_feature()` to no longer pass in `required` param and assume that required is always true (All invocations of this method did not specify this param and always defaulted to true).

This looks like a behemoth of changes, but it's mostly just moving Feature methods to a new file, so it's hopefuly not that bad 😅


api/converters.py
```diff
@@ -0,0 +1,208 @@
...
+  Delete dict keys with None values, and empty lists, recursively.
+  """
+  for key, value in list(d.items()):
+    if value is None or (isinstance(value, list) and len(value) == 0):
+      del d[key]
+    elif isinstance(value, dict):
+      del_none(value)
+  return d
+
+def migrate_views(f: Feature) -> None:
```
jrobbins, line 63: TODO: In a follow-up PR, we should do another migration script to change all these values so that this function can be deleted.

api/converters.py
```diff
@@ -0,0 +1,208 @@
...
+    f.safari_views = OPPOSED
+
+def feature_to_legacy_json(f: Feature) -> dict[str, Any]:
+  migrate_views(f)
+  d: dict[str, Any] = to_dict(f)
+  is_released = f.impl_status_chrome in RELEASE_IMPL_STATES
+  d['is_released'] = is_released
+
+  standard_maturity_val = f.standard_maturity
+  if (standard_maturity_val == UNSET_STD and
```
jrobbins, line 87: Can this be deleted now?
#2195 landed back in Sept.

internals/notifier_helpers.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc.
```
jrobbins, line 2: 2022

jrobbins (approved):

api/converters.py
DanielRyanSmith: Added a TODO for this 👍

api/converters.py
DanielRyanSmith: Yes! removed.

PR #2410, opened 2022-10-27
Verbose FeatureEntry to JSON function

DanielRyanSmith:
This change adds a new function `feature_entry_to_json_verbose(feature_entry)`, which functions in a mostly identical way to `feature_to_legacy_json(feature)`, with the exception of using the new FeatureEntry entities. Many fields needed to mimic the functionality of the legacy version are located on Stage and Gate entities associated with the FeatureEntry entity, so this requires querying for those stages and gates.

Based on initial testing of each of these functions, the new implementation is noticeably slower due to the additional db querying involved. It is likely not viable for processing basic feature information in bulk to the client.
```
--- 0.0009918212890625 seconds for LEGACY ---
--- 0.0118560791015625 seconds for NEW    ---
```

Some new props exist on the new dictionary that are not in the legacy version, and vice versa
| Property | exists on Legacy | exists on New |
|---|---|---|
| comments | ✅ | ❌ |
| feature_notes | ❌ | ✅ |
| visibility | ✅ | ❌ |
| stages | ❌ |  ✅ |
| gates | ❌ |  ✅ |

### DIFF
--- a/api/converters.py
+++ b/api/converters.py
@@ -252,7 +252,7 @@ BASIC_FE_FIELDS = [
     "wpt",
     "wpt_descr"]
 
-def prep_stage_gate_info(
+def _prep_stage_gate_info(
     fe: FeatureEntry, d: dict) -> dict[str, Optional[Stage]]:
   """Adds stage and gate info to the dict and returns major stage info."""
   proto_type = STAGE_TYPES_PROTOTYPE[fe.feature_type]
@@ -267,7 +267,8 @@ def prep_stage_gate_info(
       'proto': None,
       'dev_trial': None,
       'ot': None,
-      'extend': None}
+      'extend': None,
+      'ship': None}
 
   # Write a collection of stages and gates associated with the feature,
   # sorted by type.
@@ -276,6 +277,7 @@ def prep_stage_gate_info(
   # Stages and gates are given as a dictionary, with the type as the key,
   # and a list of entity IDs as the value.
   for s in stages:
+    # Keep major stages for referencing additional fields.
     if s.stage_type == proto_type:
       major_stages['proto'] = s
     elif s.stage_type == dev_trial_type:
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
@jrobbins I'm curious if you have an opinion on this. Is it worth keeping this for serving information on a single feature, but creating a more basic function for bulk feature data as you suggested?

api/converters.py
```diff
@@ -201,5 +202,252 @@ def feature_to_legacy_json(f: Feature) -> dict[str, Any]:
...
+  return major_stages
+
+def feature_entry_to_json_verbose(fe: FeatureEntry) -> dict[str, Any]:
+  """Returns a verbose dictionary with all info about a feature."""
+  d: dict[str, Any] = {}
+  for field in BASIC_FE_FIELDS:
+    d[field] = str(getattr(fe, field))
+
+  if not fe.key:
+    return {}
```
jrobbins: Should this return `d`?

api/converters.py
```diff
@@ -201,5 +202,252 @@ def feature_to_legacy_json(f: Feature) -> dict[str, Any]:
...
+    d['stages'][s.stage_type] = s.key.integer_id()
+  for g in gates:
+    d['gates'][g.gate_type].append(g.key.integer_id())
+  
+  return major_stages
+
+def feature_entry_to_json_verbose(fe: FeatureEntry) -> dict[str, Any]:
+  """Returns a verbose dictionary with all info about a feature."""
+  d: dict[str, Any] = {}
+  for field in BASIC_FE_FIELDS:
```
jrobbins, line 304: Rather than maintaining BASIC_FE_FIELDS as we add fields, could we use ndb's `fe._to_dict()` and then overwrite the items that have a format that we don't like?

jrobbins (approved):
I am glad to see that this will make it possible to use FeatureEntries in the frontend with modest code changes. 

I am not concerned about the performance because the new time is fine and I am thinking that this would really only be used for the single feature case.    We should write a non-verbose one to power list views.

api/converters.py
DanielRyanSmith: I moved this to the beginning of the function before `d` is initialized. This function should not be used if the feature is not saved.

api/converters.py
DanielRyanSmith: Agreed. Refactored to do this. 👍

PR #2413, opened 2022-10-28
Display a "Comments & activity" log at the bottom of the feature detail page.

### LINKED ISSUE #2332: Display an activity log of changes on each feature
There should be a new section at the bottom of the feature detail page that lists all the Activity entities that are associated with that feature.  Those can be a combination of user comments and field changes.

**Additional context**
I am thinking that all Activity entities should be shown, even the ones that relate to a specific Gate.  Those should be labeled with the Stage and Gate that they relate to.

jrobbins:
This should mostly resolve #2332.

We want to make it possible for users to see what has changed in a feature entry over time.  The records of what changed are called Amendments.  One or more Amendments be present any Activity entity.  The Activities and Amendments already exist because we have been capturing changes for the past few months.

This PR adds a new section to the bottom of the feature detail page called "Comments and activity".  That section shows a list of activity bubbles.  Each activity has an author, date, and comment content as it did before, but it can now also show one line for each amendment.  We clean up and omit some field values that are not very informative.  Clicking on the one-line display of the amendment expands it to show both the old and new values.

In this PR:
* comments_api.py: Add logic to convert amendments to JSON.  Accept a query string argument to control whether the response should have only comments or both comments and activities.
* basehandlers.py: Add a handy method to get a boolean query string parameter.
* chromedash-activity-log.js: Add another element for displaying amendments.  Change the look of the comments to match the UXD's mocks, basically each whole comment is in a light gray background bubble.
* chromedash-approvals-dialog.js: put a little dialog-specific formatting here so that chromedash-activity-log is reusable.
* chromedash-feature-details.js: add an accordion section for "Comments & activity"


jrobbins:
Here's a screenshot of what it looks like after the user expands the "Comments & activity" section and expands one of the amendments:
image

api/comments_api.py
```diff
@@ -19,19 +19,31 @@
...
+      'field_name': amendment.field_name,
+      'old_value': amendment.old_value.strip('[]'),
+      'new_value': amendment.new_value.strip('[]'),
+      }
+
+
+def activity_to_json_dict(comment: Activity) -> dict[str, Any]:
+  amendments_json = [
+      amendment_to_json_dict(amnd) for amnd in comment.amendments
+      if amnd.old_value != 'None' or amnd.new_value != '[]']
```
KyleJu, line 37: why is this 'None'?

api/comments_api.py
```diff
@@ -48,17 +60,19 @@ def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
...
-    
+
     # Filter deleted comments the user can't see.
+    user_email = user.email() if user else None
     comments = list(filter(
-      lambda c: self._should_show_comment(c, user.email(), is_admin), comments))
+      lambda c: self._should_show_comment(c, user_email, is_admin), comments))
 
-    dicts = [comment_to_json_dict(c) for c in comments]
+    dicts = [activity_to_json_dict(c) for c in comments]
```
KyleJu, line 75: maybe call it activities_json. dicts is too close to the type `dict`

client-src/elements/chromedash-activity-log.js
```diff
@@ -193,29 +266,23 @@ export class ChromedashActivityLog extends LitElement {
...
-      css`
-        .comment_section {
-          max-height: 250px;
-          overflow-y: scroll;
-        }
-      `];
-  }
+  render() {
+    if (this.loading) {
+      return 'loading...';
```
KyleJu, line 271: html tag around this?

api/comments_api.py
jrobbins: It is actually the string "None" in the Amendment.old_value field.   When a user makes a change to a feature entry, we compute the Amendment object by converting old and new FeatureEntry field values to strings.  That conversion creates some "None" and "[]" Amendment values.   In the API, I choose to omit the Amendments that are "None" -> "[]" because those are not useful to users.  

I think that those get created simply because there are two ways of representing a missing value for repeated fields. We should actually stop creating them.  I filed issue #2424.

api/comments_api.py
jrobbins: I am using the word `dict` here to indicate that it is a python `dict`.  It is a dict that will be used to generate json.  If the name just said `json`, that would seem to indicate a string in json format.

client-src/elements/chromedash-activity-log.js
jrobbins: Done.

KyleJu:
> Here's a screenshot of what it looks like after the user expands the "Comments & activity" section and expands one of the amendments: image

The left alignment between shipped_milesotne, old and new could be better. Maybe make sure that they all align to the left?

KyleJu (approved):
LGTM with UI adjustments

jrobbins:
OK added some padding-left.  Here is what it looks like now
image

PR #2416, opened 2022-10-30
Upgrade search to use FeatureEntry

jrobbins:
In this PR:
* Upgrade search.py and search_queries.py to use FeatureEntry objects rather than Features.  The final objects returned to the APIs are still FeatureObjects.  Search just uses FeatureEntry objects while computing the list of feature IDs.
* Redo how the "me" predefined queries are processed.  This is more efficient because a future is returned rather than blocking on loading actual features early in the process, and it decouples from the get_all() function which still uses Feature objects.

With this PR, we lose the ability to search for fields that are in the Stages objects.   I will add that back in the next PR in this series.  It is OK to lose that functionality for now because it is not used by any part of the site that users can navigate to.  (The roadmap queries features by milestone using different code.)

### DIFF
--- a/api/features_api_test.py
+++ b/api/features_api_test.py
@@ -96,30 +96,49 @@ class FeaturesAPITestDelete(testing_config.CustomTestCase):
     self.assertFalse(revised_feature.deleted)
 
 
-class FeaturesAPITestGet(testing_config.CustomTestCase):
+class FeaturesAPITestGet_NewSchema(testing_config.CustomTestCase):
 
   def setUp(self):
     self.feature_1 = core_models.Feature(
         name='feature one', summary='sum Z',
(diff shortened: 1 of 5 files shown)


DanielRyanSmith (approved):
Hopefully querying for fields on separate entities does not increase the computation time by too much - I'm looking into this myself for using FeatureEntry entities on the front-end now.

jrobbins:
> Hopefully querying for fields on separate entities does not increase the computation time by too much - I'm looking into this myself for using FeatureEntry entities on the front-end now.

The idea that makes it practical is that these async queries pretty much all run in parallel.  There is a final phase that loads the features that are on the current pagination page, but that is only up to 100 features.

Issue #2333, opened 2022-10-13
Show Gate chips in list views (at least in the pending approvals box)


jrobbins:
**Describe the solution you'd like**
Replace the stamp icon with a set of Gate chips.  Each Gate chip will show the state of an approval Gate, and clicking on one will open the approvals column focused on that Gate.

Each Gate chip is a rounded-rectangle with a color and icon indicating Gate state and text indicating the name of the Gate.

jrobbins:
The last step is to make the chips open the gate column with a regular click instead of shift-click.

PR #2421, opened 2022-10-31
Implement chromedash-gate-chip.

jrobbins:
This is is progress on #2333.

A "gate chip" is a UI chip (a pill-shaped button with a pastel background) that shows the state of a review gate.  We will display these in the "Features pending my review" box and in the stage sections of the feature detail page.  Eventually, clicking on one of these chips will open the approvals column dialog box focused on that gate.   This PR just implements the appearance of the gate chip element: pill-shape, colors, icons, and labels.   PRs in the coming weeks will add logic to display these on a page and a click handler.

In this PR:
* Define chromedash-gate-chip element, including icon names and styles that reference theme colors
* Add the logical colors in theme.scss
* Add the relevant material design icons
* Add python logic to pass the team name and gate name in the gate_api, and update tests


jrobbins:
This PR does not place these chips on a page yet.  But, here is a screenshot showing what it might look like if one feature had all different kinds of gate chips:
image

KyleJu (approved):

PR #2426, opened 2022-11-02
Add ability to search Features for fields that occur in Stages

jrobbins:
When using the old schema, everything about a feature was in the Feature model class.  So, processing a search term like `browsers.chrome.ot.desktop.start=99` could be done by mapping that to `Feature.dt_milestone_desktop_start` and making an ndb query filter with the 99 in the condition.

Our new schema allows one FeatureEntry to have many Stages of different types.  So, that same `browsers.chrome.ot.desktop.start=99` user query term needs to generate an ndb query filter on `Stage.milestones.desktop_first` being 99, but it also needs a join condition that checks that we are looking at an origin trial stage rather than a shipping or dev trial stage.

In this PR:
* search_queries.py:
  * Look up supported user query terms in two dictionaries: one for direct FeatureEntry fields and one for Stage fields.
  * When searching Stage fields, add in the appropriate join condition.
  * When starting the query: use keys_only to get the ID of FeatureEntries, but fetch the feature_id field for Stages. 
* search.py:
  * After the async query has complete, extract the feature ID from either the key or feature_id of the projected Stage object.
* index.yaml:
  * Now that we are querying Stage on a combination of two fields, we need to define those indexes.


internals/search_queries.py
```diff
@@ -23,26 +23,38 @@
...
+    query = FeatureEntry.query()
+    field = QUERIABLE_FIELDS.get(field_name)
+  elif field_name in STAGE_QUERIABLE_FIELDS:
+    # It is a query on a field in Stage.
+    query = Stage.query()
+    stage_type_dict = STAGE_TYPES_BY_QUERY_FIELD[field_name]
+    # Only consider the appropriate stage type(s).
+    stage_types = [st for st in stage_type_dict.values() if st is not None]
+    query = query.filter(Stage.stage_type.IN(stage_types))
+    field = STAGE_QUERIABLE_FIELDS.get(field_name)
```
DanielRyanSmith, line 48:
```suggestion
  if field_name in QUERIABLE_FIELDS:
    # It is a query on a field in FeatureEntry.
    query = FeatureEntry.query()
    field = QUERIABLE_FIELDS[field_name]
  elif field_name in STAGE_QUERIABLE_FIELDS:
    # It is a query on a field in Stage.
    query = Stage.query()
    stage_type_dict = STAGE_TYPES_BY_QUERY_FIELD[field_name]
    # Only consider the appropriate stage type(s).
    stage_types = [st for st in stage_type_dict.values() if st is not None]
    query = query.filter(Stage.stage_type.IN(stage_types))
    field = STAGE_QUERIABLE_FIELDS[field_name]
```

This should fix the mypy errors, and should be safe based on the if statement check.

index.yaml
```diff
@@ -391,3 +391,43 @@ indexes:
...
+- kind: Stage
+  properties:
+  - name: stage_type
+  - name: milestones.desktop_last
+  - name: feature_id
+- kind: Stage
+  properties:
+  - name: stage_type
+  - name: milestones.android_first
+  - name: feature_id
```
DanielRyanSmith, line 408: Nice, I didn't know we could reference the subproperties like this 🙂

DanielRyanSmith (approved):

index.yaml
jrobbins: I didn't either.  I ran it on staging and went with the suggested indexes that were logged on failures, then repeatedly pasted and edited it.

KyleJu (approved):

PR #2427, opened 2022-11-02
Fix double-write for ready_for_trial_url.

jrobbins:
When testing searching for Stage fields, I noticed that ready_for_trial_url was not being written to a dev trail stage intent_thread_url.


pages/guide.py
```diff
@@ -623,7 +623,7 @@ def update_stage_fields(self, feature_id: int, feature_type: int,
         stage.milestones = milestoneset_entity
       # If the field starts with "intent_", it should modify the
       # more general "intent_thread_url" field.
-      elif field.startswith('intent_'):
+      elif field.startswith(('intent_', 'ready_for_trial_url')):
```
DanielRyanSmith, line 626: We'll need to change this in the migration script as well. Currently, this field is saved on Dev Trial stage entities as `announcement_url`. I'm not sure if this was a mistake on my part, but I'll create a PR changing this.

DanielRyanSmith (approved):

pages/guide.py
jrobbins: Oh, announcement_url sounds like the better place to put it.  I forgot what we had previously said about that field.  I'll change this PR.

jrobbins: Please check that I have it right this time.

pages/guide.py
```diff
@@ -241,7 +241,7 @@ def process_post_data(self, **kwargs):
     if self.touched('ready_for_trial_url'):
       feature.ready_for_trial_url = self.parse_link(
           'ready_for_trial_url')
-      update_items.append(('ready_for_trial_url',
+      stage_update_items.append(('ready_for_trial_url',
           self.parse_link('ready_for_trial_url')))
```
DanielRyanSmith, line 245:
```suggestion
      stage_update_items.append(('announcement_url',
          self.parse_link('ready_for_trial_url')))
```
The first item of this tuple represents the field that should be changed, and the second item is the value. This change can be made and the `elif` statement also added on this file can be removed.

DanielRyanSmith (changes requested):

pages/guide.py
DanielRyanSmith: Actually, this is a bit more complex, and you'll need to add a key value pair to `core_enums.STAGE_TYPES_BY_FIELD_MAPPING` to ensure this is saved to the right stage as well.

```
'announcement_url': STAGE_TYPES_DEV_TRIAL,
```

jrobbins: Good point. Changed.

DanielRyanSmith (approved):

PR #2428, opened 2022-11-02
Do double-writes of star_count.

jrobbins:
When handling starring requests, update both Feature and FeatureEntry.


DanielRyanSmith (approved):
Thanks for taking care of this! This is one that I missed.

Issue #2347, opened 2022-10-14
Support negation for feature search


jrobbins:
**Is your feature request related to a problem? Please describe.**
We currently have a simple user language for specifying feature queries, but it does not support negation.

**Describe the solution you'd like**
Allow the user to write a minus sign in front of a search term to negate it.  For example `owner=jrobbins@chromium.org -memory` to find features that have that owner but that do not include the word "memory" anywhere.  

For numeric fields the user could just use `!=` or reverse an inequality.   The real usefulness of negation is when negating a full-text search term.

PR #2431, opened 2022-11-02
[1] Support negation in structured query search

KyleJu:
#2347. Support logical operators parsing  **- (negation), OR** and implement **-** in structured query search. By default, there is an implied **AND** logic amongst all query terms. When there is a **-** operator, the negation logical operator is translated into the inverse of field name operators, e.g.

- NOT A=B is equivalent to A != B
- B=C NOT A=B is equivalent to B=C AND A!=B


internals/search.py
```diff
@@ -125,40 +126,46 @@ def process_queriable_field(
...
-        TEXT_PATTERN),
+    '(?P<logical>%s)?\s*(?P<field>%s)(?P<op>%s)(?P<val>%s)\s+|(?P<textterm>%s)\s+' % (
+        LOGICAL_OPERATORS_PATTERN, FIELD_NAME_PATTERN, OPERATORS_PATTERN,
+        VALUE_PATTERN, TEXT_PATTERN),
     re.I)
 
 
 def process_query_term(
-    field_name: str, op_str: str, val_str: str) -> Future:
+    logical_op: str, field_name: str, op_str: str, val_str: str) -> Future:
```
jrobbins, line 144: I think that only negation can be handled at this level of the code (in `process_query_term()` and `process_queriable_field()`).  Logical OR needs to be handled in `process_query()`.  If there is only the possibility of negation here, it would make more sense to pass it as a boolean keyword like `negate=False`.   But, you might not need it at all, see next comment.

internals/search.py
```diff
@@ -125,40 +126,46 @@ def process_queriable_field(
...
-    return search_queries.handle_me_query_async('owner')
-  if query_term == 'editor:me':
-    return search_queries.handle_me_query_async('editor')
-  if query_term == 'can_edit:me':
-    return search_queries.handle_can_edit_me_query_async()
-  if query_term == 'cc:me':
-    return search_queries.handle_me_query_async('cc')
+
+  if logical_op != 'NOT':
+    # TODO(kyleju): support negation for the following queries.
```
jrobbins, line 148: Another approach to negation would be to run the original query as-is, but then negate the result.  That could be done by getting all FeatureEntry IDs and then using a set difference operation between that and the original query in `process_query()`.   That approach should not slow down the overall query process too much so long as it is done as an async query and only done once.

internals/search.py
```diff
@@ -125,40 +126,46 @@ def process_queriable_field(
 # A value that a feature field can be compared against.  It can be
 # a single word or a quoted string.
 VALUE_PATTERN = r'[^" ]+|"[^"]+"'
+# Logical operators.
+# TODO(kyleju): support 'OR' logic
+LOGICAL_OPERATORS_PATTERN = r'AND|OR|NOT'
```
jrobbins, line 131: I should have specified this before:
* For logical OR, the user should enter the keyword `OR`
* Logical AND is implicit, so users don't need to enter that.  If they do enter `AND` it would already be treated as a full-text stop word which return None which does not affect search results.
* For negation, the user should precede the query term with a leading minus sign, e.g.:
    *  `-starred_by:me` 
    * `-owner=user@example.com`
    *  `-word` 
    *  `-"phrase of words"`

jrobbins (changes requested):

internals/search.py
KyleJu: I agree and done!

internals/search.py
KyleJu: I refactored the code and will implement that for both predefined query and full-text search

internals/search.py
```diff
@@ -204,20 +222,30 @@ def process_query(
...
       future = search_fulltext.search_fulltext(textterm)
+    elif is_predefined_query_term(field_name, op_str, val_str):
+      future = process_predefined_query_term(field_name, op_str, val_str)
     else:
-      future = process_query_term(field_name, op_str, val_str)
+      future = process_query_term(is_negation, field_name, op_str, val_str)
+      is_normal_query = True
+
+    if is_negation and is_normal_query:
+      # TODO: Use set different from all FeatureEntry IDs.
```
jrobbins, line 246: Spelling: different -> difference.

internals/search.py
```diff
@@ -125,20 +115,43 @@ def process_queriable_field(
 # A value that a feature field can be compared against.  It can be
 # a single word or a quoted string.
 VALUE_PATTERN = r'[^" ]+|"[^"]+"'
+# Logical operators.
+# TODO(kyleju): support 'OR' logic
+LOGICAL_OPERATORS_PATTERN = r'OR|-'
```
jrobbins, line 120: I think that allowing `OR` here would capture the first two letters of a field name that begins with 'or'.   E.g., if we had a field called 'origin', then it would parse as ('or', 'igin').

jrobbins (approved):
This is progress.  LGTM after addressing two comments.

internals/search.py
KyleJu: I will remove OR for now and think more carefully

PR #2433, opened 2022-11-03
Use FeatureEntry for most client-side use cases

DanielRyanSmith:
This change replaces most uses of the Feature kind with the FeatureEntry kind. Many files are touched, replacing references of `core_models.Feature` with `core_models.FeatureEntry`. FeatureEntry entity use is not yet implemented for the features star, process, and detect intent APIs.

To help navigate, here is an explanation of major changes by file:
- `api/converters.py`: Add a new function `feature_entry_to_json_basic()` which returns a more basic dictionary of feature information for uses as list items, etc.
- `client-src/elements/form-definition.js`: remove deprecated field `standardization` from `formatFeatureForEdit()`
- `framework/basehandlers.py`: refactor `BaseHandler.get_specified_feature()` to return a FeatureEntry object.
- `framework/permissions.py`: refactor functions to use FeatureEntry entities.
- `internals/core_models.py`: Remove some FeatureEntry methods that exist as helper functions in `feature_helpers.py`. Also implement most code for writing amendments and notifications for FeatureEntry in `FeatureEntry.put()` and `notifier_helpers.py`, but continue to use Feature for this functionality for now (see TODO on `FeatureEntry.put()`).
- `internals/feature_helpers.py`: Rewrite helper functions `get_all()`, `get_feature()`, and `get_by_ids()` to use FeatureEntry entities, and add "_legacy" suffix to old functions that use Feature.
- `internals/notifier_helpers.py`: Move more Feature methods out of the Feature class and rework as helper functions with FeatureEntry compatibility.

### Important Note!
This change makes it so that fields that should not be visible based on the feature type are shown as BLANK for the user, and can be edited and submitted, but will never display on the front end.
For example, a feature designated as type "Web developer facing change to existing code", will always display `origin_trial_feedback_url` as blank, as this field should not be shown for the given feature type.
With these changes, front end adjustments to remove the display of these fields that are irrelevant to certain feature types should be prioritized.

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -14,16 +14,48 @@
 # limitations under the License.
 
 import logging
+import re
 from typing import Any, Optional
 from google.cloud import ndb  # type: ignore
 
-from api.converters import feature_to_legacy_json
+from api import converters
 from framework import rediscache
 from framework import users
 from internals.core_enums import *
-from internals.core_models import Feature
+from internals.core_models import Feature, FeatureEntry
+import settings
 
-def get_by_ids(feature_ids: list[int],
+
+def _crbug_number(bug_url: Optional[str]) -> Optional[str]:
+  if bug_url is None:
+    return None
+  m = re.search(r'[\/|?id=]([0-9]+)$', bug_url)
+  if m:
+    return m.group(1)
+  return None
+
+def _new_crbug_url(blink_components: Optional[list[str]],
+    bug_url: Optional[str], impl_status_chrome: int,
+    owner_emails: list[str]=list()) -> str:
+  url = 'https://bugs.chromium.org/p/chromium/issues/entry'
+  if blink_components and len(blink_components) > 0:
+    params = ['components=' + blink_components[0]]
+  else:
+    params = ['components=' + settings.DEFAULT_COMPONENT]
+  crbug_number = _crbug_number(bug_url)
+  if crbug_number and impl_status_chrome in (
+      NO_ACTIVE_DEV,
+      PROPOSED,
+      IN_DEVELOPMENT,
+      BEHIND_A_FLAG,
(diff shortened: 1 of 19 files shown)


DanielRyanSmith:
Sorry @jrobbins, this change turned out bigger than I anticipated (again 😅). Hopefully the list of changes helps with the review.

internals/feature_helpers.py
```diff
@@ -61,7 +95,7 @@ def get_by_ids(feature_ids: list[int],
       if feature_id in result_dict]
   return result_list
 
-def get_all(limit=None, order='-updated', filterby=None,
+def get_all_legacy(limit=None, order='-updated', filterby=None,
```
jrobbins, line 98: This can be for a follow-up PR: I don't see any callers of `get_all()` or `get_all_legacy()` that pass `keys_only`, so support for that could probably be deleted.

internals/feature_helpers.py
```diff
@@ -423,3 +460,123 @@ def get_all_with_statuses(statuses, update_cache=False):
...
+    feature_list = query.fetch(limit, keys_only=keys_only)
+    if not keys_only:
+      feature_list = [
+          converters.feature_entry_to_json_basic(f) for f in feature_list]
+
+    rediscache.set(KEY, feature_list)
+
+  return feature_list
+
+def get_feature(
```
jrobbins, line 511: I don't see any callers of this function.

jrobbins (approved):
It was a lot of files, but the changes look pretty straightforward.

internals/feature_helpers.py
DanielRyanSmith: I thought that as well, but I wasn't 100% on that, so I just kept it in. Maybe this is worth removing in the follow-up PR you mentioned above. 🙂

Issue #2346, opened 2022-10-14
Implement logical OR operator for feature search


jrobbins:
**Is your feature request related to a problem? Please describe.**
We currently allow users to use a query language in which a query consists of a series of terms that are implicitly combined with a logical AND.  That is what is needed most of the time, but sometimes users want to do a logical OR.

**Describe the solution you'd like**
We should support logical OR at two levels:
- [ ] Quick-OR by using comma-separate value lists.  A convenient way specify matching multiple values is to use a comma-separated list.  For example, `owner=jrobbins@google.com,jrobbins@chromium.org`.
- [x] A top-level OR operator that goes between search terms.  For example `owner=jrobbins@google.com OR owner=jrobbins@chromium.org`.

**Additional context**
I don't think that we need to handle parenthesis, but if we did that would be cool.

Processing quick-OR should be pretty easy.  Basically when we have a list of values, we use the NDB `IN` operator instead of `==`.  We don't need to support quick-OR with inequalities.

Processing top-level OR is basically like processing separate queries.  They need to be combined before sorting and pagination.

PR #2434, opened 2022-11-03
Implement top-level OR operator

KyleJu:
A step closer for https://github.com/GoogleChrome/chromium-dashboard/issues/2346


jrobbins (changes requested):
Sorry, I should have been more clear about what I was looking for when I wrote up the issue.

I don't think that you can treat `OR` like a unary operator that only affects the term that immediately follow it.  Instead `OR` needs to be treated like a binary operator that divides the overall query into OR-clauses.  Each OR-clause needs to be processed independently, like a whole query, and then the results of the OR-clauses need to be combined with a set union.

What you have now work for a query like `category=1 OR category=2` only because there is just one term after the `OR`.  It will not be correct for `category=1 OR category=2 cc:me` because it will treat it like `(category=1 OR category=2) AND cc:me.  The parens and `AND` are just for my explanation, we don't need to support that.  Specifically, a feature that has category=1 but does not CC me should be returned in the results.

I think we need to:
* Parse the `OR` keyword as a separate query term.  It would be its own item in the `terms` list.
* Use the `terms` list to construct a nested `clauses` list, with each item being a list of terms between the `OR` terms.   The `OR` terms themselves do not need to be kept.  E.g., the query `category=1 OR category=2 cc:me` would have `clauses` equal to `[[m1] [m2 m3]]` for the match tuples for `category=1`, `category=3`, and `cc:me`.
* Change the loop over terms (line 234) into a nested loop (or a loop that calls a `process_clause()` function that has a loop over the terms in a clause).

PR #2435, opened 2022-11-03
[4] Implement Quick-Or in structured query term

### LINKED ISSUE #2346: Implement logical OR operator for feature search

KyleJu:
Fix #2346


jrobbins (approved):
The main part of this change looks good, but there are some other diffs mixed in.  Be careful when merging.

KyleJu:
Will send out a new pr

PR #2436, opened 2022-11-03
[3] Implement negation for predefined query term and full-text search

KyleJu:
Fix https://github.com/GoogleChrome/chromium-dashboard/issues/2347.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -242,15 +242,13 @@ def process_query(
       future = process_query_term(is_negation, field_name, op_str, val_str)
       is_normal_query = True
 
-    if is_negation and not is_normal_query:
-      # TODO: Use set different from all FeatureEntry IDs.
-      pass
+    if future is None:
+      continue
 
-    if future is not None:
-      if logical_op == 'OR':
-        feature_id_future_ops.append((True, future))
-      else:
-        feature_id_future_ops.append((False, future))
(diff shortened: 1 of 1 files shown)


internals/search.py
```diff
@@ -259,16 +257,25 @@ def process_query(
...
     feature_ids = _resolve_promise_to_id_list(future)
     if result_id_set is None:
       logging.info('first term yields %r', feature_ids)
-      result_id_set = set(feature_ids)
-      continue
+      if logical_op == '-':
+        # If a negation is the first term, interpret it as
+        # all feature ids - the negated feature ids. Thus adding
+        # all feature ids here initially.
+        all_feature_entries = core_models.FeatureEntry.query().fetch()
```
jrobbins, line 268: I think this is correct, but it would be faster if you use `fetch(keys_only=True)` and then get loop over those keys on the next line.

Another small speed up: there is no need to construct the list object before passing it to the set() constructor.   Instead of `set([x for y in z])` it could be `set(x for y in z)`, or even python3 set constructor syntax: `{x for y in z}`.

It could be made even faster by doing it via `fetch_async` and calling it earlier in the process, but that can be left as a TODO.

jrobbins (approved):

KyleJu:
I will send out a new PR for this implementation

PR #2437, opened 2022-11-03
Upgrade our app to use python 3.10.

jrobbins:
Let's upgrade to python3.10, because our workstations stopped fully supporting 3.9.

Apparently, we will get better error messages and the ability to use `|` instead of `Union`.
https://docs.python.org/3/whatsnew/3.10.html


jcscottiii:
Before this is merged, could you change it here to `python 3.10`: https://github.com/GoogleChrome/chromium-dashboard/blob/7069f4e4f961c7141ed73064fdec13bb51b49899/.devcontainer/devcontainer.json#L58

and here to `3.10-bullseye`:
https://github.com/GoogleChrome/chromium-dashboard/blob/7069f4e4f961c7141ed73064fdec13bb51b49899/.devcontainer/docker-compose.yml#L12

This will keep the devcontainer in sync too 🎉

requirements.txt
```diff
@@ -6,7 +6,7 @@ appengine-python-standard>=1.0.0
 html5lib==1.1
 funcsigs==1.0.2
 google-api-python-client==2.47.0
-google-python-cloud-debugger==2.18
+# Only suppoted for py3.9: google-python-cloud-debugger==2.18
```
jcscottiii, line 9: Could you add an issue to revisit this in the future when 3.10 is supported. https://github.com/GoogleCloudPlatform/cloud-debug-python/issues/33

jcscottiii (approved):
LGTM

KyleJu (approved):

PR #2440, opened 2022-11-05
Calculate better gate states from votes.

jrobbins:
Redo the calculation of gate state based on vote state.  This changes the logic a little bit to introduce a new Gate.PENDING state for when the is no review request yet.  Also, while the review is on-going, more priority is given to the most recent vote, which allows for explicit re-requests to show up if we decide to offer that.

In this PR:
* Rewrite `_calc_gate_state()` and simplify the arguments for easier testing.
* Write a bunch of new tests, and moved existing tests to the most appropriate file.
* Stop clearing the REVIEW_REQUESTED Vote because we will be able to query for active gates directly.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -161,6 +161,7 @@ def is_resolved(approval_values, field_id):
 
   return False
 
+
 def set_vote(feature_id: int,  gate_type: int, new_state: int,
     set_by_email: str, gate_id: Optional[int]=None) -> None:
   """Add or update an approval value."""
@@ -195,6 +196,7 @@ def set_vote(feature_id: int,  gate_type: int, new_state: int,
   if gate:
     update_gate_approval_state(gate)
 
+
 def get_gate_by_type(feature_id: int, gate_type: int):
   """Return a single gate based on the feature and gate type."""
   # TODO(danielrsmith): As of today, there is only 1 gate per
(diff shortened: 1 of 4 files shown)


internals/approval_defs.py
```diff
@@ -206,27 +208,39 @@ def get_gate_by_type(feature_id: int, gate_type: int):
...
-    elif vote.state == Vote.DENIED:
-      return Vote.DENIED
-  afd = APPROVAL_FIELDS_BY_ID[gate.gate_type]
 
-  if ((afd.rule == ONE_LGTM and approvals >= 1) or
-      (afd.rule == THREE_LGTM and approvals >= 3)):
-    return Vote.APPROVED
-  return gate.state
+def _calc_gate_state(votes: list[Vote], rule: str) -> int:
+  """Returns the state that a gate should have based on its votes."""
```
DanielRyanSmith, line 213: Thanks for taking a look at this. The old implementation would not revert to a non-approved state if one of the API owners approved but then set their state back to undecided, so my plan was to revisit this.

internals/review_models_test.py
```diff
@@ -172,86 +172,8 @@ def test_get_activities__specific_fields(self):
...
-    # Gate 1 should evaluate to not approved after updating.
-    self.assertEqual(approval_defs.update_gate_approval_state(self.gate_1), 6)
-    self.assertEqual(self.gate_1.state, 6)
-
-  def test_update_approval_state__no_change(self):
-    """Gate's approval state does not change unless it needs to."""
-    # Gate 2 is already marked as approved and should not change.
-    self.assertEqual(approval_defs.update_gate_approval_state(self.gate_2), 5)
-    self.assertEqual(self.gate_2.state, 5)
+  # TODO(jrobbins): Add tests for is_resolved, is_approved, get_feature_gates.
```
DanielRyanSmith, line 175: I know it would be a worthwhile endeavor to add more tests for many of the uses of the new entities and for all the API refactors. Perhaps this is a good next step after we're ahead of schedule.

DanielRyanSmith (approved):

PR #2446, opened 2022-11-08
Check for undefined MilestoneSet.

jrobbins:
The value of stage.milestone is initially null, so we just need to check for that.
This resolves a problem with newly created issues no being viewable.


DanielRyanSmith (approved):
This is a flaw in my implementation here. all stages that might need to use milestone fields should have milestone entities as a property even if all fields are null, so something like this doesn't happen. I'll work to make this change.

PR #2448, opened 2022-11-08
create MilestoneSets as needed for Stage entities

DanielRyanSmith:
Follow-up from #2446


jrobbins:
This code looks good, but what if we just allowed milestones to be None in cases where the stage does not require milestones?  It means that we need to code defensively when accessing milestones, but that is not too unusual.

PR #2451, opened 2022-11-08
[3] Implement the negation logic for predefined and full-text queries

KyleJu:
Fix https://github.com/GoogleChrome/chromium-dashboard/issues/2347. Preprocess the negation queries so that it doesn't only affect the term that immediately before it.
- Update the logical operator parser for both structured and full-text queries, to parse `OR ` (with trailing space) and '-' (without trailing space)
- Turn all negation operations into a AND operation

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -315,6 +315,6 @@ def process_negation_operations(feature_id_future_ops):
 
 def fetch_all_feature_ids_set():
   """Fetch all FeatureEntry ids. """
-  all_feature_entries = core_models.FeatureEntry.query().fetch(keys_only=True)
-  feature_ids_set = set(fe.key.integer_id() for fe in all_feature_entries)
+  all_feature_keys = core_models.FeatureEntry.query().fetch(keys_only=True)
+  feature_ids_set = set(key.integer_id() for key in all_feature_keys)
   return feature_ids_set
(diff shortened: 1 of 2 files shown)


internals/search.py
```diff
@@ -279,3 +287,34 @@ def process_query(
...
+
+    if negated_id_set is None:
+        negated_id_set = fetch_all_feature_ids_set()
+
+    feature_ids = _resolve_promise_to_id_list(future)
+    negated_id_set.difference_update(feature_ids)
+
+  # Append the result of negation operations to the end.
+  if negated_id_set is not None:
+    new_future_ops.append(('', list(negated_id_set)))
```
jrobbins, line 312: Could we append `('', negated_id_set)` without converting to a list?

internals/search_test.py
```diff
@@ -55,14 +55,14 @@ def test_structured_query_terms__complex(self):
         [('-', 'field', '=', 'value', '')],
         search.TERM_RE.findall('-field=value '))
     self.assertEqual(
-        [('-', 'field', '>', 'value', '')],
-        search.TERM_RE.findall('-field>value '))
+        [('OR ', 'field', '>', 'value', '')],
```
jrobbins, line 58: You have some `OR` unit test cases mixed into this negation PR.

jrobbins (approved):

internals/search.py
KyleJu: Yup we could!

internals/search_test.py
KyleJu: I also added `OR ` parsing in this PR.

internals/search.py
KyleJu: I got too many complaints from mypy. I will change it in the next PR, along with the OR logic

Issue #2452, opened 2022-11-08
Creating new features locally causes errors 


DanielRyanSmith:
This is a result of sorting query results using the `FeatureEntry.updated` field when it is not initialized.

**To Reproduce**
1. Clear local db.
2. Log in.
3. Create a new feature.
4. The next page should display errors. Additionally, this server-side error appears:

```
File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 161, in check_success
    raise self._exception
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 334, in _advance_tasklet
    yielded = self.generator.throw(type(error), error, traceback)
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/_datastore_query.py", line 113, in fetch
    while (yield results.has_next_async()):
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/tasklets.py", line 338, in _advance_tasklet
    yielded = self.generator.send(send_value)
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/_datastore_query.py", line 674, in has_next_async
    if value < min_value:
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/_datastore_query.py", line 771, in __lt__
    return self._compare(other) == -1
  File "/Users/danielrsmith/projects/chromium-dashboard/cs-env/lib/python3.10/site-packages/google/cloud/ndb/_datastore_query.py", line 827, in _compare
    if this_value < other_value:
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
```

DanielRyanSmith:
Fixed with #2453

PR #2453, opened 2022-11-08
Add default value to updated field

DanielRyanSmith:
Addresses #2452

This change initializes the updated field of the FeatureEntry entity to initialize the value of the `updated` field upon creation, as well as adding a script to backfill any missing values that currently exist in the database.

### DIFF
--- a/main.py
+++ b/main.py
@@ -211,7 +211,7 @@ internals_routes: list[tuple] = [
   ('/admin/schema_migration_write_entities', schema_migration.MigrateEntities),
   ('/admin/schema_migration_approval_vote', schema_migration.MigrateApprovalsToVotes),
   ('/admin/schema_migration_gate_status', schema_migration.EvaluateGateStatus)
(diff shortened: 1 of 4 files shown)


main.py
```diff
@@ -211,6 +211,7 @@
   ('/admin/schema_migration_write_entities', schema_migration.MigrateEntities),
   ('/admin/schema_migration_approval_vote', schema_migration.MigrateApprovalsToVotes),
   ('/admin/schema_migration_gate_status', schema_migration.EvaluateGateStatus)
+  ('/admin/schema_migration_updated_field', schema_migration.AddUpdatedField)
```
jrobbins: It's always better to end these lines with commas.  One is needed above.

jrobbins (approved):
LGTM after one fix

main.py
DanielRyanSmith: Good catch, thanks!

PR #2461, opened 2022-11-09
FeatureEntry edit bug fix

DanielRyanSmith:
Fixes two issues where emails cannot be properly saved after editing, and occasionally old field names were references on FeatureEntry entities rather than the new field names.


DanielRyanSmith:
Sorry about this @jrobbins, somehow I missed adding 'emails' as one of the input types. 😅

jrobbins (approved):

PR #2463, opened 2022-11-10
Correct permission terms query logic

KyleJu:
Correct permission terms query logic. The permission queries should apply to the result of user queries, instead of only the last OR query clause. 

- When the user query is empty, it should return the result of permission queries.

### DIFF
--- a/internals/search_test.py
+++ b/internals/search_test.py
@@ -125,6 +125,16 @@ class SearchFunctionsTest(testing_config.CustomTestCase):
     self.featureentry_2.put()
     notifier.FeatureStar.set_star(
         'starrer@example.com', self.feature_1.key.integer_id())
+    self.feature_3 = core_models.Feature(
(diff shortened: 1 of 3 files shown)


internals/search_test.py
```diff
@@ -344,6 +344,11 @@ def test_process_query__negated_predefined(
 
   def test_process_query__single_field(self):
     """We can can run single-field queries."""
+    actual, tc = search.process_query('')
```
jrobbins, line 347: We need some test cases with Features (and FeatureEntries) that have unlisted=true and deleted=true.

jrobbins (approved):
Thanks! The algorithm looks right.
LGTM after you add a couple more unit tests.

internals/search_test.py
KyleJu: Added!

Issue #2361, opened 2022-10-19
Edit page for "Write up motivation" stage loses all existing data


davidben:
**Describe the bug**
The "Update" or "Revisit" button under the "Write up motivation" goes to a page that

**To Reproduce**
1. Go to a launch
2. Click on "Update" or "Revisit" next to "Write up motivation"
3. See the page is empty, instead of filling in the existing contents.

**Expected behavior**
Existing contents should not be lost.

**Additional context**
Example URL: https://chromestatus.com/guide/stage/6678134168485888/1

The "edit all fields" button seems not to have this bug.

PR #2464, opened 2022-11-10
Show motivation value in deprecation features

DanielRyanSmith:
Addresses #2361

This change fixes a bug that caused the value of the "motivation" and "comments" field to display without their existing values. It appears there were plans to change the motivation field to map to "deprecation_motivation" for deprecation trials at some point, but did not coalesce.


jrobbins (approved):
This can be an immediate fix to unblock other developers today.
But, I thought that we did have deprecation_motivation working, so let's try to follow up with a PR that gets it working again.

PR #2466, opened 2022-11-10
Map deprecation_motivaton to motivation both ways.

jrobbins:
The should fix the display of the motivation field on deprecation features.

There is only one motivation field in a feature, however the client-side forms can reference it by the name `motivation` or `deprecation_motivation` to choose one field spec or the other.   The field specs are different in that they provide different on-page help text.   The `deprecation_motivation` field spec says `name: 'motivation'` to allow for a mapping from the form element name `deprecation_motivation` back to the field name `motivation`.

The problem was that that name mapping was being used only when rendering the NAME `<textarea name=NAME>` to be used in the form submission POST.   It was not being used to map the name used to look up the value to be displayed when the page loads.  The net effect was that the text that the user entered was being saved, but then it is shown as blank when the user visits the editing form again.

This PR checks for a mapping of in the form field spec.  If there is one,  it is used when looking up the value to be displayed in the texture.  If there is not a mapping, the original field name is used.

### DIFF
--- a/client-src/elements/chromedash-guide-stage-page.js
+++ b/client-src/elements/chromedash-guide-stage-page.js
@@ -5,6 +5,7 @@ import './chromedash-form-table';
 import './chromedash-form-field';
 import {formatFeatureForEdit, STAGE_FORMS, IMPL_STATUS_FORMS} from './form-definition';
 import {IMPLEMENTATION_STATUS} from './form-field-enums';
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Thanks for making a more accurate fix 🙂

haaami01:
Thank you!👌

PR #2475, opened 2022-11-14
Add dev API to write and delete test data

DanielRyanSmith:
This change adds two new API endpoints that add or delete basic test data to the local dev environment:
- `dev/clear_entities`: Deletes most entities from the local db.
- `dev/write_dev_data`: Writes some test Feature, FeatureEntry, Stage, and Gate entities.

These endpoints can only be used in local development.

A json file has been created that holds the basic information to create these features. Ideally, this data will be expanded to create approvals data like Vote entities, as well as other useful data that mimics common scenarios (e.g. recently created features with many empty fields, features with multiple versions of the same Stage entity, etc.)

Screen Shot 2022-11-14 at 12 26 04 PM


api/dev_api.py
```diff
@@ -0,0 +1,97 @@
...
+    kinds: list[ndb.Model] = [Feature, FeatureEntry, MilestoneSet,
+        Stage, Activity, Approval, Comment, Gate, Vote]
+    
+    for kind in kinds:
+      for entity in kind.query():
+        entity.key.delete()
+    
+    return 'Entities cleared.'
+
+class WriteDevData(APIHandler):
```
jrobbins, line 42: Style: Two blank lines between top-level declarations.

main.py
```diff
@@ -232,12 +232,17 @@ class Route:
       schema_migration.WriteUpdatedField),
 ]
 
-
+dev_routes: list[Route] = []
+if settings.DEV_MODE:
+  dev_routes = [
+    Route('/dev/clear_entities', dev_api.ClearEntities),
```
jrobbins, line 238: I feel that this needs an extra layer of protection.  The settings.DEV_MODE and and your new settings.LOCAL_MODE depend on certain strings in the environment variables.  I can imagine that over time, there could be a variation in those strings... which would put the whole app in to DEV mode, but it might not be noticed at first.

How about commenting out these lines and having a comment saying that developers can uncomment them when needed?

jrobbins (changes requested):

main.py
jrobbins: Alternatively, what about writing these as scripts under the scripts/ directory rather than being part of the app?

DanielRyanSmith: > Alternatively, what about writing these as scripts under the scripts/ directory rather than being part of the app?

Is there a good way to do this without having the local server running? I tried writing the script and then realized it would not be of much use in its current form.

> How about commenting out these lines and having a comment saying that developers can uncomment them when needed?

I went ahead with this approach for now.

jrobbins: > > Alternatively, what about writing these as scripts under the scripts/ directory rather than being part of the app?
> 
> Is there a good way to do this without having the local server running? I tried writing the script and then realized it would not be of much use in its current form.

Good point.  I don't think that it wold need to have the flask app running, but it would need the middleware for Cloud NDB.  And, it would need to do the steps for starting the cloud NDB emulator and stopping it after.  So, we could explore it sometime, but no need to do it now.

jrobbins (approved):

PR #2478, opened 2022-11-15
Reference feature stages as possibly plural

DanielRyanSmith:
This change adjusts most parts of code that reference stages to assume that it is possible to have multiple instances of the same stage type. Additionally, some Stage class methods have been moved to `internals/stage_helpers.py` to mimic the same structure as the Feature/FeatureEntry kinds.

### DIFF
--- a/internals/stage_helpers.py
+++ b/internals/stage_helpers.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


internals/stage_helpers.py
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import defaultdict
+
+from internals.core_models import Stage
+
+def get_feature_stages(feature_id: int) -> dict[int, list[Stage]]:
```
jrobbins, line 20: Style: Two blank lines between top-level declarations.

jrobbins (approved):

internals/stage_helpers.py
DanielRyanSmith: One day I'll get this right!

PR #2479, opened 2022-11-15
Fix editing of unlisted checkbox on metadata form.

jrobbins:
This fixes two problems with editing feature metadata:
* The form did not have a form_fields hidden field, so all checkboxes were considered touched, even if they were not present on the form, which caused other boolean fields in the issue to be cleared whenever a user used the metadata form.
* guide.py's list of fields to parse did not include unlisted, so the user checking or unchecking the box had no effect.


DanielRyanSmith (approved):
Thanks for the catch and the fix!

PR #2480, opened 2022-11-15
Use FeatureEntry entities to display roadmap information

DanielRyanSmith:
This change updates the `get_in_milestone()` function to use FeatureEntry entities rather than the old Feature entities. This requires that Stage entities be queried for first to obtain feature IDs relevant to the given milestone, and then query for those features by IDs.

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -340,58 +340,58 @@ def get_in_milestone(milestone: int,
     # but no desktop milestone.
     q = Stage.query(Stage.milestones.android_first == milestone,
         Stage.milestones.desktop_first == None,
-        Stage.stage_type.IN(STAGE_BLINK_SHIPPING, STAGE_PSA_SHIPPING,
-            STAGE_FAST_SHIPPING, STAGE_DEP_SHIPPING))
+        Stage.stage_type.IN((STAGE_BLINK_SHIPPING, STAGE_PSA_SHIPPING,
+            STAGE_FAST_SHIPPING, STAGE_DEP_SHIPPING)))
     android_only_shipping_future = q.fetch_async()
 
     # Origin trial stages (Desktop) in this milestone.
     q = Stage.query(Stage.milestones.desktop_first == milestone,
-        Stage.stage_type.IN(STAGE_BLINK_ORIGIN_TRIAL, STAGE_FAST_ORIGIN_TRIAL,
-            STAGE_DEP_DEPRECATION_TRIAL))
+        Stage.stage_type.IN((STAGE_BLINK_ORIGIN_TRIAL, STAGE_FAST_ORIGIN_TRIAL,
+            STAGE_DEP_DEPRECATION_TRIAL)))
     desktop_origin_trial_future = q.fetch_async()
 
     # Origin trial stages (Android) in this milestone.
     q = Stage.query(Stage.milestones.android_first == milestone,
         Stage.milestones.desktop_first == None,
-        Stage.stage_type.IN(STAGE_BLINK_ORIGIN_TRIAL, STAGE_FAST_ORIGIN_TRIAL,
-            STAGE_DEP_DEPRECATION_TRIAL))
+        Stage.stage_type.IN((STAGE_BLINK_ORIGIN_TRIAL, STAGE_FAST_ORIGIN_TRIAL,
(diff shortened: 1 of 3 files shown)


internals/feature_helpers.py
```diff
@@ -312,63 +312,149 @@ def get_in_milestone(milestone: int,
...
-        desktop_origin_trial_features_future.result())
-    android_origin_trial_features = (
-        android_origin_trial_features_future.result())
-    webview_origin_trial_features = (
-        webview_origin_trial_features_future.result())
-    desktop_dev_trial_features = desktop_dev_trial_features_future.result()
-    android_dev_trial_features = android_dev_trial_features_future.result()
+    # Shipping stages with a matching desktop milestone.
+    q = Stage.query(Stage.milestones.desktop_first == milestone,
+        ndb.OR(Stage.stage_type == STAGE_BLINK_SHIPPING,
```
jrobbins, line 318: Instead of ndb.OR, could this be simpler as the following?
Stage.stage_type.IN(STAGE_BLINK_SHIPPING, STAGE_PSA_SHIPPING, ...)

internals/feature_helpers.py
```diff
@@ -312,63 +312,149 @@ def get_in_milestone(milestone: int,
...
+        Stage.milestones.desktop_first == None,
+        ndb.OR(Stage.stage_type == STAGE_BLINK_DEV_TRIAL,
+            Stage.stage_type == STAGE_PSA_DEV_TRIAL,
+            Stage.stage_type == STAGE_FAST_DEV_TRIAL,
+            Stage.stage_type == STAGE_DEP_DEV_TRIAL))
+    android_dev_trial_future = q.fetch_async(None)
+
+    # Wait for all futures to complete and collect unique feature IDs.
+    desktop_shipping_ids = list(set(
+        [s.feature_id for s in desktop_shipping_future.result()]))
```
jrobbins, line 377: You can save a few characters of syntax by using python3 set creation syntax:
list({
   s.feature_id for s in desktop_shipping_future.result()})

internals/feature_helpers.py
```diff
@@ -312,63 +312,149 @@ def get_in_milestone(milestone: int,
...
+        [s.feature_id for s in webview_origin_trial_future.result()]))
+    desktop_dev_trials_ids = list(set(
+        [s.feature_id for s in desktop_dev_trial_future.result()]))
+    android_dev_trials_ids = list(set(
+        [s.feature_id for s in android_dev_trial_future.result()]))
+
+    # Query for FeatureEntry entities that match the stage feature IDs.
+    # Querying with an empty list will raise an error, so check if each
+    # list is not empty first.
+    if desktop_shipping_ids:
```
jrobbins, line 394: You could save some lines by introducing a tiny function that returns None or a future.  It would be called like:
desktop_shipping_future = _get_entries_by_id_async(desktop_shipping_ids)

internals/feature_helpers.py
```diff
@@ -312,63 +312,149 @@ def get_in_milestone(milestone: int,
...
+      desktop_dev_trial_future = None
+    if android_dev_trials_ids:
+      q = FeatureEntry.query(FeatureEntry.key.IN(
+          [ndb.Key('FeatureEntry', id) for id in android_dev_trials_ids]))
+      q = q.order(FeatureEntry.name)
+      android_dev_trial_future = q.fetch_async()
+    else:
+      android_dev_trial_future = None
+
+    desktop_shipping_features = ([] if desktop_shipping_future is None
```
jrobbins, line 444: Likewise, a tiny function could help here too.  It would return [] or the future result. E.g.,
desktop_shipping_features = _resolve_future(desktop_shipping_future)

jrobbins (approved):
LGTM after considering some comments

internals/feature_helpers.py
DanielRyanSmith: I tried this, but then realized that this format will try to hash the list rather than unpack it, so it requires a `*` to preface the list. I think it's a bit harder to read what is happening this way, but I went with the change.

internals/feature_helpers.py
DanielRyanSmith: I went with this approach, and I agree that it's much more readable and less repetitive. 🙂

PR #2482, opened 2022-11-16
Add enterprise feature and breaking change support

yanndago:
This allows users to create a feature with the Enterprise type, and mark a feature as breaking


api/processes_api.py
```diff
@@ -34,6 +35,10 @@ def do_get(self, **kwargs):
     feature_process = processes.ALL_PROCESSES.get(
       f.feature_type, processes.BLINK_LAUNCH_PROCESS)
 
+    
+    if feature_id != core_enums.FEATURE_TYPE_ENTERPRISE_ID and f.breaking_change:
```
jcscottiii, line 39: Shouldn't this be `if f.feature_type != core_enums.FEATURE_TYPE_ENTERPRISE_ID and f.breaking_change:` (change feature_id to f.feature_type)

jcscottiii: Can you add another test case to https://github.com/GoogleChrome/chromium-dashboard/blob/main/api/processes_api_test.py that checks this branch?

client-src/elements/chromedash-form-field.js
```diff
@@ -107,6 +106,27 @@ export class ChromedashFormField extends LitElement {
           )}
         </sl-select>
       `;
+    } else if (type === 'multiselect') {
```
jcscottiii, line 109: Add a test to support this new type to this file https://github.com/GoogleChrome/chromium-dashboard/blob/main/client-src/elements/chromedash-form-field_test.js#L56

You can see other tests for the other types.

jcscottiii (changes requested):

api/processes_api.py
yanndago: > 

We actually want to insert the feature rollout stage in non enterprise features if they are breaking, we do not want to change the feature type altogether.

yanndago: I misunderstood. Fixed it.

client-src/elements/chromedash-form-field.js
yanndago: Done.

pages/guide.py
```diff
@@ -206,7 +209,12 @@ class FeatureEditHandler(basehandlers.FlaskHandler):
...
       ('experiment_risks', 'str'),
-      ('experiment_extension_reason', 'str')]
+      ('experiment_extension_reason', 'str'),
+      ('rollout_milestone', 'int'),
+      ('rollout_platforms', 'split_str'),
+      ('rollout_details', 'str'),
+      ('enterprise_policies', 'split_str'),
+      ]
 
   CHECKBOX_FIELDS: frozenset[str] = frozenset([
```
jrobbins, line 219: Isn't breaking_change also needed here?

jrobbins (commented):
This is an impressive PR.  I skimmed through it today and will try to take another look tomorrow.

client-src/elements/form-field-enums.js
```diff
@@ -42,6 +64,9 @@ export const FEATURE_TYPES = {
     'This type of feature entry can be referenced from a PSA immediately.'],
   FEATURE_TYPE_DEPRECATION_ID: [3, 'Feature deprecation',
     'Deprecate and remove an old feature.'],
+  FEATURE_TYPE_ENTERPRISE_ID: [4, 'New Feature or removal affecting enterprises',
+    'For features or changes that need to be communicated to enterprises or schools',
+    'enterprisefeature'],
```
jrobbins, line 69: Where is this used?

client-src/elements/form-field-specs.js
```diff
@@ -1134,6 +1141,51 @@ export const ALL_FIELDS = {
     help_text: '',
   },
 
+  'enterprise_policies': {
+    type: 'input',
+    attrs: MULTI_STRING_FIELD_ATTRS,
+    required: false,
+    label: 'Enterprise policies',
+    help_text: html`
+      List of policies that control the feature, if any.`,
```
jrobbins, line 1150: What is the syntax of each item in the list?  Can we give an example?

client-src/elements/form-field-specs.js
```diff
@@ -1134,6 +1141,51 @@ export const ALL_FIELDS = {
...
+      selected platforms. Include any controls admins have to test it 
+      (e.g. flags) and control it (e.g. an enterprise policy). Write in the 
+      present tense.`,
+  },
+
+  'breaking_change': {
+    type: 'checkbox',
+    label: 'Breaking change',
+    initial: false,
+    help_text: 'This is a breaking change (customers/developers must take action)',
```
jrobbins, line 1187: I think that we can do better with this help text, but let's leave that for a follow-up PR.

internals/core_enums.py
```diff
@@ -175,43 +200,50 @@
...
   }
 
 # Prototype stage types for every feature type.
 STAGE_TYPES_PROTOTYPE: dict[int, Optional[int]] = {
     FEATURE_TYPE_INCUBATE_ID: STAGE_BLINK_PROTOTYPE,
     FEATURE_TYPE_EXISTING_ID: STAGE_FAST_PROTOTYPE,
     FEATURE_TYPE_CODE_CHANGE_ID: None,
-    FEATURE_TYPE_DEPRECATION_ID: None
+    FEATURE_TYPE_DEPRECATION_ID: None,
+    FEATURE_TYPE_ENTERPRISE_ID: None
```
jrobbins, line 214: Let's use final trailing commas on all these list items.  That will make future diffs easier to read.
https://google.github.io/styleguide/jsguide.html#features-arrays-trailing-comma

internals/core_models.py
```diff
@@ -486,3 +488,9 @@ class Stage(ndb.Model):
   announcement_url = ndb.StringProperty()
   # Origin trial stage id that this stage extends, if trial extension stage.
   ot_stage_id = ndb.IntegerProperty()
+
+  #Enterprise
+  rollout_milestone = ndb.IntegerProperty()
+  rollout_platforms = ndb.StringProperty(repeated=True)
+  rollout_details = ndb.StringProperty()
```
jrobbins, line 495: If this can be a paragraph of text, it would be better to make it a ndb.TextProperty().  The problem with StringProperty is that it is limited to only 1500 bytes.

internals/processes.py
```diff
@@ -478,18 +482,67 @@ def process_to_dict(process):
   ]
 
 
+FEATURE_ROLLOUT_STAGE = ProcessStage(
```
jrobbins, line 485: Please add a comment to FEATURE_ROLLOUT_STAGE to explain that this is a stage that can be inserted into the stages for non-enterprise features that are marked as breaking.

internals/processes.py
```diff
@@ -478,18 +482,67 @@ def process_to_dict(process):
...
+      [PI_ROLLOUT_MILESTONE,
+       PI_ROLLOUT_PLATFORMS,
+       PI_ROLLOUT_DETAILS,
+       PI_ENTERPRISE_POLICIES,
+      ],
+      [],
+      [],
+      core_enums.INTENT_SHIP, core_enums.INTENT_SHIP)
+
+ENTERPRISE_STAGES = [
```
jrobbins, line 500: Please add a comment here saying that these are the stages for a feature that has the  enterprise feature_type.

pages/guide.py
jrobbins: To be more clear, I think it is needed here.  If found that your code worked without including breaking_change in CHECKBOX_FIELDS, that is because of a different defect that was fixed a few days ago.

jrobbins (changes requested):

client-src/elements/chromedash-form-field.js
```diff
@@ -107,6 +106,27 @@ export class ChromedashFormField extends LitElement {
...
+      const valueArray = fieldValue.split(',');
+      console.log(valueArray);
```
jrobbins, line 111: Do we still need this logging statement?

jrobbins (changes requested):

client-src/elements/form-field-enums.js
```diff
@@ -28,6 +28,28 @@ export const FEATURE_CATEGORIES = {
...
+  PLATFORM_FUCHSIA: [8, 'Fuchsia'],
+};
+
+export const PLATFORMS_DISPLAYNAME = {
+  1: 'Android', // PLATFORM_ANDROID
+  2: 'iOS', // PLATFORM_IOS
+  3: 'Chrome OS', // PLATFORM_CHROMEOS
+  4: 'LaCrOS', // PLATFORM_LACROS
+  5: 'Linux', // PLATFORM_LINUX
+  6: 'Mac', // PLATFORM_MAC
```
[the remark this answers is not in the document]
yanndago: In crbug and launch it says Mac only, I think it is better to keep it Mac for consistency.
What do you think ?

client-src/elements/form-field-enums.js
yanndago: Good catch, it is a remnant of a test I made.

client-src/elements/form-field-specs.js
yanndago: MULTI_STRING_FIELD_ATTRS I put an example in the placeholder. But it is a list comma separated of pascal case policy names.
Would the placeholder example be enough ? Or should I repeat it in the help text ?

client-src/elements/form-field-specs.js
yanndago: Ack.

client-src/elements/form-field-enums.js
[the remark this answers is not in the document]
jrobbins: Fine with me.

client-src/elements/form-field-specs.js
jrobbins: OK, I saw that but I didn't know if that was an example of the syntax expected or more of a meta-syntactic variable.   If the feature owners will recognize it, that is fine.

jrobbins (approved):

pages/guide.py
```diff
@@ -259,7 +267,9 @@ def _get_field_val(self, field: str, field_type: str) -> Any:
       return self.form.get(field)
     elif field_type == 'split_str':
       val = self.split_input(field, delim=',')
-      if field == 'blink_components' and len(val) == 0:
+      if field == 'rollout_platforms':
+        val = self.form.getlist(field)
```
jcscottiii, line 271: I have question for my own understanding: why use self.form.getlist instead of [split_input](https://github.com/GoogleChrome/chromium-dashboard/blob/d331c4540eb0a9c37e091021f92c53942770f04c/framework/basehandlers.py#L457) that was already used above? It uses the form already too.

jcscottiii (changes requested):

jrobbins:
For that failed lint test, my first thought is that `.value="[]"` needs to be `.value=${[]}`.

pages/guide.py
yanndago: this is because we use the select input here, and it stores data into an array and not a string separated list

jcscottiii:
@yanndago could you rebase your changes with the latest main branch?

jcscottiii (approved):

PR #2484, opened 2022-11-16
Update and handle deprecated views fields

DanielRyanSmith:
This PR adds a script to migrate all deprecated views fields to their respective new values, as well as handles cases when the field is an unexpected value.

### DIFF
--- a/internals/schema_migration.py
+++ b/internals/schema_migration.py
@@ -440,3 +440,40 @@ class WriteUpdatedField(FlaskHandler):
         count += 1
     
     return f'{count} FeatureEntry entities given updated field values.'
+
+class UpdateDeprecatedViews(FlaskHandler):
+
+  def get_template_data(self, **kwargs):
+    """Migrates deprecated feature views fields to their new values"""
+    self.require_cron_header()
+
+    for fe in FeatureEntry.query():
+      if fe.ff_views == MIXED_SIGNALS:
+        fe.ff_views = NO_PUBLIC_SIGNALS
+      if fe.ff_views == PUBLIC_SKEPTICISM:
+        fe.ff_views = OPPOSED
+
+      if fe.safari_views == MIXED_SIGNALS:
+        fe.safari_views = NO_PUBLIC_SIGNALS
+      if fe.safari_views == PUBLIC_SKEPTICISM:
+        fe.safari_views = OPPOSED
+      fe.put()
+    
+    for f in Feature.query():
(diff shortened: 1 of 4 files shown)


internals/schema_migration.py
```diff
@@ -440,3 +440,40 @@ def get_template_data(self, **kwargs) -> str:
...
+      if fe.ff_views == MIXED_SIGNALS:
+        fe.ff_views = NO_PUBLIC_SIGNALS
+      if fe.ff_views == PUBLIC_SKEPTICISM:
+        fe.ff_views = OPPOSED
+
+      if fe.safari_views == MIXED_SIGNALS:
+        fe.safari_views = NO_PUBLIC_SIGNALS
+      if fe.safari_views == PUBLIC_SKEPTICISM:
+        fe.safari_views = OPPOSED
+      fe.put()
```
jrobbins, line 460: It would be better to only call put() on a FeatureEntry that is actually changed by one of the conditions above.   Having that logic will make this go faster, but also greatly reduce the risk of stomping on a change to an entry by some human user at the same time.

internals/schema_migration.py
```diff
@@ -440,3 +440,40 @@ def get_template_data(self, **kwargs) -> str:
...
+      if f.ie_views == MIXED_SIGNALS:
+        f.ie_views = NO_PUBLIC_SIGNALS
+      if f.ie_views == PUBLIC_SKEPTICISM:
+        f.ie_views = OPPOSED
+
+      if f.safari_views == MIXED_SIGNALS:
+        f.safari_views = NO_PUBLIC_SIGNALS
+      if f.safari_views == PUBLIC_SKEPTICISM:
+        f.safari_views = OPPOSED
+      f.put()
```
jrobbins, line 477: Same.

jrobbins (approved):
LGTM after addressing comments

internals/schema_migration.py
```diff
@@ -440,3 +440,40 @@ def get_template_data(self, **kwargs) -> str:
         count += 1
     
     return f'{count} FeatureEntry entities given updated field values.'
+
+class UpdateDeprecatedViews(FlaskHandler):
```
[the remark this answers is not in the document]
DanielRyanSmith: Fixed. I realize now that I misunderstood "top-level" as being at the top of the file. I'll adjust moving forward.

internals/schema_migration.py
DanielRyanSmith: Agreed. Changed 🙂

Issue #2344, opened 2022-10-14
Implement support for enums in the search API and query builder


jrobbins:
**Is your feature request related to a problem? Please describe.**
We have a new search API that can accept "field operator value" search terms, however the values of enums would need to be specified as numbers, which would make no sense to users.

**Describe the solution you'd like**
On the server side:
- [ ] When parsing a query term on an enum field, check if the value is a string that is one of the enum choices.  If so, convert it to the appropriate integer.  If it is not one of the enum choices, treat that as an error.

On the client side:
- [ ] Change chromedash-feature-filter.py to use form-field-specs.js and form-field-enums.js instead of defining QUERIABLE_FIELDS.
- [ ] When the field type is 'select' use an sl-select widget to offer the user choices of values to query
- [ ] When the user is filtering on a enum field, generate a query term like field_name="enum value string"

KyleJu:
For the task, **Change chromedash-feature-filter.py to use form-field-specs.js and form-field-enums.js instead of defining QUERIABLE_FIELDS**, we need to consolidate the constants on the frontend and the backend. Plan to take a stab at it Q1 next year

jrobbins:
I changed the title of this issue to reflect the remaining work.

PR #2498, opened 2022-11-23
Implement enum support in feature search & query builder

KyleJu:
#2344.

Server side:
- When parsing a query term on an enum field, check if the value is a string that is one of the enum choices. If so, convert it to the appropriate integer. If it is not one of the enum choices, treat that as an error.

Client side:
- Add sl-select widget to offer the user choices of enum values
- Generate a query term like field_name="enum value string" for enum fields


client-src/elements/chromedash-feature-filter.js
```diff
@@ -144,6 +156,57 @@ const QUERIABLE_FIELDS = [
...
+    type: ENUM_TYPE},
+  {name: 'privacy_review_status',
+    display: 'Privacy review status',
+    choices: REVIEW_STATUS_CHOICES,
+    type: ENUM_TYPE},
+  {name: 'tag_review_status',
+    display: 'TAG Specification Review Status',
+    choices: REVIEW_STATUS_CHOICES,
+    type: ENUM_TYPE},
+  {name: 'safari_views',
```
jrobbins, line 198: These should use the dotted property names like the ones commented out on lines 123-128.   And then those commented out lines can be deleted.

internals/core_enums.py
```diff
@@ -506,3 +506,17 @@ def convert_enum_int_to_string(property_name, value):
...
+  """If the property is an enum, return its enum value, else -1."""
+  enum_dict = PROPERTY_NAMES_TO_ENUM_DICTS.get(property_name, {})
+  for index, enum_str in enumerate(enum_dict):
+    if enum_str == value:
+      return index
+  return -1
+
+
+def is_enum_field(property_name):
+  enum_dict = PROPERTY_NAMES_TO_ENUM_DICTS.get(property_name, {})
```
jrobbins, line 521: This could be: `return property_name in PROPERTY_NAMES_TO_ENUM_DICTS`

internals/search_queries.py
```diff
@@ -257,6 +256,20 @@ def sorted_by_review_date(descending: bool) -> list[int]:
...
+    # Enum fields
+    'category': FeatureEntry.category,
+    'intent_stage': FeatureEntry.intent_stage,
+    'impl_status_chrome': FeatureEntry.impl_status_chrome,
+    'security_review_status': FeatureEntry.security_review_status,
+    'privacy_review_status': FeatureEntry.privacy_review_status,
+    'standards.maturity': FeatureEntry.standard_maturity,
+    'standardization': FeatureEntry.standardization,
+    'browsers.ff.view': FeatureEntry.ff_views,
+    'ie_views': FeatureEntry.ie_views,
```
jrobbins, line 269: Likewise, dotted names for these.

KyleJu:
> There's a test failure referencing `standardization`. That field was deprecated in Feature so I did not add it to FeatureEntry. Instead there is `standard_maturity`.

'rollout_platforms' seems to be a field in stage. I will add it separately as it is not supported yet

KyleJu:
Seeing a bug for "feature type" search. I will send out a fix in a follow-up PR. Otherwise this PR has been tested on staging https://33c0dce-dot-cr-status-staging.appspot.com/newfeatures

PR #2499, opened 2022-11-23
Reducing the font size for Queriable Fields to fit inside the drop down column : Fixes #2412 

### LINKED ISSUE #2412: Font size for feature search in dropdown menus
We could reduce the font size to display the search fields better. Currently the texts in dropdown menus are cut off. Or we can increase the width of the field names

Screen Shot 2022-10-28 at 4 13 44 PM

Anuragkillswitch:
This should reduce the size of the text in the search fields and fix #2412

### DIFF
--- a/client-src/elements/chromedash-feature-filter.js
+++ b/client-src/elements/chromedash-feature-filter.js
@@ -1,24 +1,26 @@
-import {LitElement, css, html, nothing} from 'lit';
-import {SHARED_STYLES} from '../sass/shared-css.js';
+import { LitElement, css, html, nothing } from "lit";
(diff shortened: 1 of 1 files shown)


jrobbins:
It looks like you ran the file through a source code formatter.
Maybe it is mixed in with the other changes but I don't see where the font size itself is changed.
Also, I am not sure that just making the font smaller is the right solution.

Anuragkillswitch:
@jrobbins I see, I will fix the formatting issues and update this and discuss my solution then.

jrobbins:
This was resolved by
https://github.com/GoogleChrome/chromium-dashboard/pull/2522

PR #2503, opened 2022-11-23
Show useful error on feature not found

### LINKED ISSUE #2257: When viewing a non-existent feature, the UI just hangs
**Describe the bug**
If the user tries to visit a non-existent feature (e.g., a deleted one or a random number for the ID), then the page just shows a shimmering skeleton forever.  It should say "404" or "Feature not found".

**To Reproduce**
1. Go to https://cr-status-staging.appspot.com/feature/12345

**Additional context**
There might be something shown in the toast, but the toast is not being displayed due to a different error that I have a pending PR for.

The message that the toast is set to show is "Some errors occurred.  Please refresh the page or try again later."  But, that is not helpful in the case of a 404.

jcscottiii:
Also, stop showing skeleton upon feature not being found.

Add two new Errors to cs-client.js:
- ChromeStatusHttpError
- FeatureNotFoundError

ChromeStatusHttpError is meant to be returned at the doFetch level FeatureNotFoundError is meant at the getFeature level. getFeature interprets ChromeStatusHttpError and returns as necessary.

Add isFeatureLoaded to chromedash-feature-page.
Allows us to detect the case that we have finished loading but nothing was loaded.

Use that function to detect when to show main feature components or show nothing.

In the event a feature is not found. A blank page with the normal header is shown. A toast appears and says: `Feature not found.`

Fixes #2257

### DIFF
--- a/client-src/elements/chromedash-feature-page.js
+++ b/client-src/elements/chromedash-feature-page.js
@@ -144,7 +144,6 @@ export class ChromedashFeaturePage extends LitElement {
     }).catch((error) => {
       if (error instanceof FeatureNotFoundError) {
         this.loading = false;
-        showToastMessage('Feature not found.');
       } else {
         showToastMessage('Some errors occurred. Please refresh the page or try again later.');
       }
@@ -464,7 +463,7 @@ export class ChromedashFeaturePage extends LitElement {
     }
     // If after loading, the feature did not load, render nothing.
     if (!this.isFeatureLoaded()) {
-      return html ``;
+      return html `Feature not found.`;
     }
     // At this point, the feature has loaded successfully, render the components.
     return html`
--- a/client-src/elements/chromedash-feature-page.js
+++ b/client-src/elements/chromedash-feature-page.js
@@ -465,19 +465,22 @@ export class ChromedashFeaturePage extends LitElement {
   render() {
     // TODO: create another element - chromedash-feature-highlights
     // for all the content of the <div id="feature"> part of the page
+    // If loading, only render the skeletons.
+    if (this.loading) {
+      return this.renderSkeletons();
+    }
+    // If after loading, the feature did not load, render nothing.
+    if (!this.isFeatureLoaded()) {
+      return html ``;
+    }
+    // At this point, the feature has loaded successfully, render the components.
     return html`
-      ${this.loading ?
-        this.renderSkeletons() :
-        this.isFeatureLoaded() ?
-          html`
-            ${this.renderSubHeader()}
-            <div id="feature">
-              ${this.renderFeatureContent()}
-              ${this.renderFeatureStatus()}
-            </div>
-            ${this.renderFeatureDetails()}
-            ` :
-          html``}
+      ${this.renderSubHeader()}
+      <div id="feature">
+        ${this.renderFeatureContent()}
+        ${this.renderFeatureStatus()}
+      </div>
+      ${this.renderFeatureDetails()}
     `;
   }
 }
--- a/client-src/elements/chromedash-feature-page.js
+++ b/client-src/elements/chromedash-feature-page.js
@@ -116,6 +116,10 @@ export class ChromedashFeaturePage extends LitElement {
     this.fetchData();
   }
 
+  isFeatureLoaded() {
+    return this.feature && Object.keys(this.feature).length !== 0;
+  }
+
   fetchData() {
     this.loading = true;
     Promise.all([
@@ -137,8 +141,20 @@ export class ChromedashFeaturePage extends LitElement {
         document.title = `${this.feature.name} - ${this.appTitle}`;
       }
       this.loading = false;
-    }).catch(() => {
-      showToastMessage('Some errors occurred. Please refresh the page or try again later.');
+    }).catch((error) => {
+      // Cannot import the type currently with rollup when doing:
+      // import {FeatureNotFoundError} from '../js-src/cs-client';
+      // Will get:
+      //    RollupError: "FeatureNotFoundError" is not exported by
+      //    "client-src/js-src/cs-client.js", imported by
+      //    "client-src/elements/chromedash-feature-page.js"
+      // For now, compare the name.
+      if (error.name && error.name === 'FeatureNotFoundError') {
+        this.loading = false;
+        showToastMessage('Feature not found.');
+      } else {
+        showToastMessage('Some errors occurred. Please refresh the page or try again later.');
+      }
     });
   }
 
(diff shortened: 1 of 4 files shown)


jcscottiii:
I just saw this [message](https://github.com/GoogleChrome/chromium-dashboard/issues/2257#issuecomment-1252700901) about creating a base class. Will need to incorporate that.

client-src/elements/chromedash-feature-page.js
```diff
@@ -452,14 +468,16 @@ export class ChromedashFeaturePage extends LitElement {
...
         this.renderSkeletons() :
-        html`
-          ${this.renderSubHeader()}
-          <div id="feature">
-            ${this.renderFeatureContent()}
-            ${this.renderFeatureStatus()}
-          </div>
-          ${this.renderFeatureDetails()}
-      `}
+        this.isFeatureLoaded() ?
```
jrobbins, line 471: These conditionals would probably be more clear as regular if-statements with return html statements for each case.  E.g.:
if (this.loading) {
  return this.renderSkeletons();
}
if (this.isFeatureLoaded()) {
  return html...
 ;
}
return '';

jcscottiii:
When adding a new test to client-src/elements/chromedash-feature-page_test.js, I ran into an interesting problem. Turns out when retrieving ChromedashToast via document.querySelector('chromedash-toast'); gets a new toast but the document uses the first toast rendered. As a result, any assertions will fail for N+1 tests because showMessage is only affecting the first one. 

For me, there was the existing toast test that checked for `Some errors occurred. Please refresh the page or try again later.` When I copied the same test over and changed the Error Promise and the expected text, the local `toastEl` saw no changes in state. (this.open never changed and my callback for that specific test never was triggered.)

<details>
  <summary>Click this to expand experiment</summary>
  
  In order to test what was going on, I added a new property to `ChromedashToast` called setCallback and everywhere this.open was changed, I called this.setCallback afterwards.

  ###  Code
  ```js
  it('renders with no data', async () => {
    console.log('starting existing test');
    toastEl.setCallback = function() {
      console.log('hello from existing test');
    }
    assert.isEmpty(toastEl.msg);
    assert.isFalse(toastEl.open);

    const invalidFeaturePromise = Promise.reject(new Error('Got error response from server'));
    window.csClient.getFeature.withArgs(0).returns(invalidFeaturePromise);

    const component = await fixture(
      html`<chromedash-feature-page></chromedash-feature-page>`);
    assert.exists(component);
    assert.instanceOf(component, ChromedashFeaturePage);

    const toastMsgSpan = toastEl.shadowRoot.querySelector('span#msg');
    const expectedToastMsg = 'Some errors occurred. Please refresh the page or try again later.';
    assert.isTrue(toastEl.open);
    assert.equal(toastEl.msg, expectedToastMsg);
    assert.include(toastMsgSpan.innerHTML, expectedToastMsg);
  });

  it('renders with "No Feature Found" when feature not found', async () => {
    console.log('starting new test');
    toastEl.setCallback = function() {
      console.log('hello from new test'); // SHOULD PRINT BUT DIDN'T
    }
    assert.isEmpty(toastEl.msg);
    assert.isFalse(toastEl.open);

    const featureNotFoundPromise = Promise.reject(new FeatureNotFoundError(12345));
    window.csClient.getFeature.withArgs(0).returns(featureNotFoundPromise);

    const component = await fixture(
      html`<chromedash-feature-page></chromedash-feature-page>`);
    assert.exists(component);
    assert.instanceOf(component, ChromedashFeaturePage);

    // If a request for a feature returns stating that it is not found, it
    // would trigger the toast to show message.
    const toastMsgSpan = toastEl.shadowRoot.querySelector('span#msg');
    const expectedToastMsg = 'Feature not found.';
    assert.isTrue(toastEl.open);
    assert.equal(toastEl.msg, expectedToastMsg);
    assert.include(toastMsgSpan.innerHTML, expectedToastMsg);
  });
  ```

  ## Test Run
It shows that my new callback never got called.

  ```sh
$ npx web-test-runner "client-src/elements/chromedash-feature-page_test.js" --node-resolve --playwright --browsers chromium 

client-src/elements/chromedash-feature-page_test.js:

 🚧 Browser logs:
      starting existing test
      hello from existing test
      starting new test
      hello from existing test

 ❌ chromedash-feature-page > renders with "No Feature Found" when feature not found
      AssertionError: expected false to be true
      + expected - actual
      
      -false
      +true
      
      at o.<anonymous> (client-src/elements/chromedash-feature-page_test.js:188:12)
  ```
</details>

After that experiment, I decided to add a reset function to the toast class and keep a single instance of toastEl. This fixed the tests.

jcscottiii:
> I just saw this [message](https://github.com/GoogleChrome/chromium-dashboard/issues/2257#issuecomment-1252700901) about creating a base class. Will need to incorporate that.

@jrobbins thoughts about splitting that issue up? I can write the new issue about having generic error pages across the board in the future. This PR will address the current issue. If you're okay with that plan, this PR is ready for review.

client-src/elements/chromedash-feature-page.js
```diff
@@ -137,8 +141,13 @@ export class ChromedashFeaturePage extends LitElement {
         document.title = `${this.feature.name} - ${this.appTitle}`;
       }
       this.loading = false;
-    }).catch(() => {
-      showToastMessage('Some errors occurred. Please refresh the page or try again later.');
+    }).catch((error) => {
+      if (error instanceof FeatureNotFoundError) {
+        this.loading = false;
+        showToastMessage('Feature not found.');
```
jrobbins, line 147: The toast displays for 7 seconds by default, but in this case I think we should have the message stay there until the user navigates to another page.

We could pass -1 for optDuration to make the toast stay, but then we need to add logic to make it disappear once the user navigates away.

I'm thinking that it would be better to show a message in the body of the SPA page.  Basically, on line 467, return an error message to display rather than using a toast for this situation at all.  In the issue, I mentioned the toast because it was showing the wrong message, but we're not bound to make the right message show in a toast.

jrobbins (changes requested):
Thanks for doing this.  I always prefer smaller PRs, so deferring creating a base class sounds good.  We only have a couple of pages where we would want to display a "Thing not found" message, so I'm not sure that a base class is needed just for this, but it might have other benefits, so we can keep it in mind.

client-src/elements/chromedash-feature-page.js
jcscottiii: I changed it. It doesn't have any styling. But I think that could be for the new issue for the [common  error page](https://github.com/GoogleChrome/chromium-dashboard/issues/2257#issuecomment-1252700901)
image

jrobbins (approved):
Unstyled is fine so long as the user sees something that explains what went wrong.

jcscottiii:
Thanks @jrobbins! I started this [issue](https://github.com/GoogleChrome/chromium-dashboard/issues/2533) to discuss the long term strategy more.

PR #2515, opened 2022-11-29
Make pending review and recent review queries use Gate.state

jrobbins:
This is a step toward using the new schema for the FLT functionality.  These queries are used to populate the pending-reviews-by-me and recent-reviews boxes on the myfeatures page.

In this PR:
* Have the frontend boxes request a search using sort spec `gate.requested_on` or `gate.reviewed_on`.
* Implement those sort specs on the back end
* Reorder the classes in review_models so that Gate can refer to the constants defined in Vote
* Add Gate.requested_on field
*  Remove _get_referenced_feature_ids: it is no longer needed because we return a future
* Add an index to index.yaml because we search on two NDB model fields


jrobbins:
Ah, I just realized that we should continue to support the old `approval.*` sort specs on the server for a couple of weeks.  That's for any users who have our SPA already open in a browser tab and continue to use it after this change is deployed.  I'll add those back but route them to the new logic.

jrobbins:
> Ah, I just realized that we should continue to support the old `approval.*` sort specs...
That part is done.

DanielRyanSmith (approved):
Thanks for taking care of this 🙂

PR #2517, opened 2022-11-29
Update reminders to use new entities

DanielRyanSmith:
This change updates the email reminders methods to use new FeatureEntry and Stage entities in calculating who and when to send reminder emails to.

Note that the matching Feature entity is still used to render the reminder email due to the `estimated-milestones-table.html` file using the old Feature milestone fields, and will require a larger refactor to update to use Stage milestone fields.

### DIFF
--- a/internals/reminders_test.py
+++ b/internals/reminders_test.py
@@ -100,7 +100,8 @@ class FunctionTest(testing_config.CustomTestCase):
     self.maxDiff = None
   
   def tearDown(self) -> None:
-    for kind in [Feature, FeatureEntry, Stage]:
+    kinds: list[ndb.Model] = [Feature, FeatureEntry, Stage]
+    for kind in kinds:
       for entity in kind.query():
         entity.key.delete()
 
--- a/internals/reminders_test.py
+++ b/internals/reminders_test.py
@@ -18,7 +18,7 @@ import settings
 from datetime import datetime
 from unittest import mock
 
-from internals import core_models
+from internals.core_models import Feature, FeatureEntry, MilestoneSet, Stage
 from internals import reminders
 
 from google.cloud import ndb  # type: ignore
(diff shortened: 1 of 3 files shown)


jrobbins (approved):

internals/reminders_test.py
```diff
@@ -37,18 +37,39 @@ def __init__(self, status_code=200, text='{}'):
 
 
 def make_test_features():
-  feature_1 = core_models.Feature(
-      name='feature one', summary='sum', owner=['feature_owner@example.com'],
-      category=1, ot_milestone_desktop_start=100)
+  feature_1 = FeatureEntry(
```
jcscottiii, line 40: feature_1 -> feature_entry_1
similar below for the other features that were changed to FeatureEntry

internals/reminders_test.py
```diff
@@ -37,18 +37,39 @@ def __init__(self, status_code=200, text='{}'):
...
+      name='feature one', summary='sum',
+      owner_emails=['feature_owner@example.com'],
+      category=1, feature_type=0)
   feature_1.put()
-  feature_2 = core_models.Feature(
+  stages = [110, 120, 130, 140, 150, 151, 160]
+  for stage_type in stages:
+    stage = Stage(feature_id=feature_1.key.integer_id(), stage_type=stage_type)
+    if stage_type == 150:
```
jcscottiii, line 48: Can you add a comment about this conditional?

internals/reminders_test.py
```diff
@@ -37,18 +37,39 @@ def __init__(self, status_code=200, text='{}'):
...
-  feature_1 = core_models.Feature(
-      name='feature one', summary='sum', owner=['feature_owner@example.com'],
-      category=1, ot_milestone_desktop_start=100)
+  feature_1 = FeatureEntry(
+      name='feature one', summary='sum',
+      owner_emails=['feature_owner@example.com'],
+      category=1, feature_type=0)
   feature_1.put()
-  feature_2 = core_models.Feature(
+  stages = [110, 120, 130, 140, 150, 151, 160]
```
jcscottiii, line 45: Should we use the core_enum variables here instead of `[110, 120, 130, 140, 150, 151, 160]`?

Also is there a reason to use the incubating stages here vs feature_entry 2 uses the existing id stages?

Optional suggestion:
- `[incubatingStage[0] for incubatingStage in STAGES_AND_GATES_BY_FEATURE_TYPE[FEATURE_TYPE_INCUBATE_ID]]` instead of `[110, 120, 130, 140, 150, 151, 160]`
- Why: I had to look to do a global lookup of 110 to find it comes from core_enums. It'll make it easier to track down valid values of this.

internals/reminders_test.py
```diff
@@ -58,13 +79,31 @@ def setUp(self):
     self.current_milestone_info = {
         'earliest_beta': '2022-09-21T12:34:56',
     }
-    self.feature_template = core_models.Feature(
-      name='feature one', summary='sum', owner=['feature_owner@example.com'],
-      category=1, ot_milestone_desktop_start=100)
-    self.feature_template.key = ndb.Key('Feature', 123)
-    # This test does not require saving to the database.
+
+    self.old_feature_template = Feature(id=123,
```
jcscottiii, line 83: Where is `old_feature_template` used?

jcscottiii (commented):
A few questions  and suggestions @DanielRyanSmith

internals/reminders_test.py
```diff
@@ -58,13 +79,31 @@ def setUp(self):
...
+      if stage_type == 150:
+        stage.milestones = MilestoneSet(desktop_first=100)
+      stage.put()
+    
+    self.old_feature_template.put()
+    self.feature_template.put()
 
     self.maxDiff = None
+  
+  def tearDown(self) -> None:
```
jcscottiii, line 102: I like this clean up. I wonder if we should do this more often and also do it during the setup as well.

internals/reminders_test.py
jcscottiii: Similarly for the other stage loops below.

internals/reminders_test.py
DanielRyanSmith: I agree with this, and I will do this in the future. The hope is to replace the old Feature kind entirely, so eventually they will phase out. However, they should be differentiated while they're still in the code together.

internals/reminders_test.py
DanielRyanSmith: Will do - I've done this in a couple of places and they should be explained each place (this is this ID associated with the origin trial stage type of this feature type)

internals/reminders_test.py
DanielRyanSmith: I agree - we've had scenarios where unrelated tests cause other tests to fail because the entities are not properly cleared out. I think this makes it a bit safer. 🙂

internals/reminders_test.py
jcscottiii: Oh. I didn't know `Feature` is going away eventually. Thanks for that. I only mentioned this change because the class changed from `Feature` to `FeatureEntry`. But if FeatureEntry will eventually become the Feature, we can keep it.

internals/reminders_test.py
jcscottiii: Ahh. I understand now.
I think this would be better by being `self.feature_template` and the other one being `self.feature_entry_template`. When I saw `old`, I thought it was a mistake since I saw they both had 123 as the id.  
However, this [thread](https://github.com/GoogleChrome/chromium-dashboard/pull/2517/files#r1035108814) and your TODO gave me context around FeatureEntry becoming the new way.

PR #2522, opened 2022-11-30
Adjust select and menu size

### LINKED ISSUE #2412: Font size for feature search in dropdown menus

KyleJu:
Fix #2412. Increase the size of select and dropdown menu:
- Increase the sl-select to 240px, the smallest width to accommodate the largest text, Webview Shipping Milestone
- Set enum drop-down menu flex 1 1 auto, aligned with other input fields


KyleJu:
The screenshot after adjustment:
Screen Shot 2022-11-29 at 4 17 21 PM (2)

KyleJu:
What ENUM drop-down menu looks like:
Screen Shot 2022-11-29 at 4 47 31 PM (2)

jrobbins (approved):

KyleJu:
After lots experiments, I had trouble override sl-menu-item's css properties (worked if using Chrome devtools). I will only keep the size change for the sl-select

PR #2526, opened 2022-11-30
Remove bad input from rollout_platforms

DanielRyanSmith:
For enterprise features, occasionally, the multi-select form field for the rollout_platforms field would give an additional empty string as a user-selected value. This was causing incorrect values and views.

Screen Shot 2022-11-30 at 11 04 25 AM

This is a simple workaround to this issue (although I do not yet know the root cause).

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -275,7 +275,7 @@ class FeatureEditHandler(basehandlers.FlaskHandler):
         val = self.form.getlist(field)
         # Occasionally, input will give an empty string as the first element.
         # It needs to be removed.
-        if val and len(val[0]) == 0:
+        if val and val[0] == '':
           val = val[1:]
         return val
       elif field == 'blink_components' and len(val) == 0:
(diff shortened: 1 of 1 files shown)


pages/guide.py
```diff
@@ -271,12 +271,16 @@ def _get_field_val(self, field: str, field_type: str) -> Any:
     elif field_type == 'str':
       return self.form.get(field)
     elif field_type == 'split_str':
-      val = self.split_input(field, delim=',')
       if field == 'rollout_platforms':
         val = self.form.getlist(field)
+        # Occasionally, input will give an empty string as the first element.
+        # It needs to be removed.
+        if val and len(val[0]) == 0:
```
jrobbins, line 278: This would be more clear as `val and val[0] == ''`.

jrobbins (approved):

pages/guide.py
DanielRyanSmith: Good idea - adjusted 🙂

PR #2531, opened 2022-12-01
Exclude enterprise features from search results by default.

jrobbins:
We want enterprise features to only show up when the user is explicitly asking for enterprise features.
This PR makes feature search default to excluding enterprise features, unless the user's query includes a mention of feature_type.

In this PR:
* features_api.py: pass show_enterprise as True only if the user query mentions feature_type.
* core_enums.py: allow enums fields to be specified by enum string or numeric string, because sometimes using the numbers is more convenient.
* search.py:
  *  Accept a new show_enterprise=False parameter and treat it like the way we treat permissions terms so that it applies to all OR-clauses of the user's query.
  * Re-group some lines under comments more concisely.
  * Process the permission terms more like we treat the regular terms: allow OR clauses even through we currently don't use them because that allows the following fix
  * In process_or_terms(), treat zero terms as unconstrained, so it returns the universe of all feature IDs

That change to process_or_terms() fixes a bug and an incorrect existing unit test that was previously testing that an empty query and no permission checks returned an empty list rather than a list of all features.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -294,7 +294,7 @@ def create_future_operations_from_queries(terms):
 
 def process_or_operations(or_clauses):
   """Process OR operations for all id sets."""
-  # If there were no conditions, then all features match.
+  # If there were no conditions, all features match.
   if not or_clauses:
     return fetch_all_feature_ids_set()
 
--- a/internals/search.py
+++ b/internals/search.py
@@ -24,6 +24,7 @@ from google.cloud.ndb.tasklets import Future  # for type checking only
 from framework import users
 from framework import utils
 from internals import approval_defs
+from internals import core_enums
 from internals import core_models
 from internals import feature_helpers
 from internals import notifier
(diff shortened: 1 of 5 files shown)


internals/search.py
```diff
@@ -233,30 +238,22 @@ def process_query(
...
-  permission_ids = process_and_operations(permissions_future_ops)
-  if len(permission_ids) == 1:
-    if terms == []:
-      # Empty search terms.
-      result_id_set = permission_ids[0]
-    else:
-      result_id_set.intersection_update(permission_ids[0])
+  # 3b. Process all permission ops, then interesect to apply permisisons.
+  permission_clauses = process_and_operations(permissions_future_ops)
+  permission_ids = process_or_operations(permission_clauses)
```
KyleJu, line 248: In the case of permission clauses, OR clauses only apply within permission terms, rather than the whole query?

jrobbins: Yes, I suppose that is true:  If there was ever an 'OR' in the permissions clauses that 'OR' would apply only to the permissions clauses. I think that makes sense, however I don't foresee that we would use 'OR'.  The main reason to make this call here is to convert from clauses to IDs in the same way that is done for the user search terms.

KyleJu (approved):

PR #2535, opened 2022-12-02
Create Stages API

DanielRyanSmith:
This PR creates a new Stages API that allows CRUD functionality for feature stages.

These new endpoints are available for `api/v0/features/<feature_id>/stages/<stage_id>`:
- `GET`: List the details of a given stage based on the stage ID.
- `POST`: Create a new stage.
- `PATCH`: Edit an existing stage.
- `DELETE`: Archive an existing stage.

### DIFF
--- a/api/stages_api.py
+++ b/api/stages_api.py
@@ -1,0 +1,261 @@
+# -*- coding: utf-8 -*-
+# Copyright 2022 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from datetime import datetime
+import logging
+from typing import Any, Optional
+from google.cloud import ndb  # type: ignore
+
+# Appengine imports.
+from framework import rediscache
+
+from framework import basehandlers
+from framework import permissions
+from internals import core_enums, notifier_helpers
+from internals import stage_helpers
+from internals.core_models import FeatureEntry, MilestoneSet, Stage
+from internals.review_models import Gate
+import settings
+
+
+
+
+class StagesAPI(basehandlers.APIHandler):
+
+  # Categorized field names of the Stage kind.
+  GENERAL_FIELDS: tuple[str] = (
+      'stage_type',
+      'browser',
+      'pm_emails',
+      'tl_emails',
+      'ux_emails',
+      'te_emails',
+      'intent_thread_url')
+
+  MILESTONE_FIELDS: tuple[str] = (
+      'desktop_first',
+      'desktop_last',
+      'android_first',
+      'android_last',
+      'ios_first',
+      'ios_last',
+      'webview_first',
+      'webview_last')
+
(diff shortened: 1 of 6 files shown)


api/stages_api.py
```diff
@@ -0,0 +1,251 @@
...
+
+from framework import basehandlers
+from framework import permissions
+from internals import core_enums
+from internals.core_models import FeatureEntry, MilestoneSet, Stage
+from internals.review_models import Gate
+
+
+
+
```
jrobbins, line 24: Style: only 2 blank lines between top-level declarations.

api/stages_api.py
```diff
@@ -0,0 +1,251 @@
...
+      if field in kwargs:
+        setattr(stage, field, kwargs[field])
+
+  def _update_stage_vals(self, stage: Stage, feature_type: int,
+      kwargs: dict, create_gate: bool=False) -> None:
+    """Check for relevant stage fields and update stage as needed."""
+    if 'stage_type' in kwargs:
+      stage.stage_type = kwargs['stage_type']
+    s_type = stage.stage_type
+
```
jrobbins, line 119: Style: Only one blank line in a row inside a method of class.

api/stages_api.py
```diff
@@ -0,0 +1,251 @@
...
+  def _create_gate_for_stage(
+      self, feature_id: int, stage_id: int, gate_type: int) -> None:
+    """Create a Gate entity for the given stage type."""
+    gate = Gate(feature_id=feature_id, stage_id=stage_id, gate_type=gate_type,
+        state=Gate.PREPARING)
+    gate.put()
+
+  def _add_given_stage_vals(self,
+      stage: Stage, kwargs: dict, fields: list[str]) -> None:
```
jrobbins, line 107: Usually `kwargs` is for keyword arguments of the current method.  Maybe you could give it a name that explains its purpose, like values_dict or something.

internals/core_models.py
```diff
@@ -494,3 +494,5 @@ class Stage(ndb.Model):
   rollout_platforms = ndb.StringProperty(repeated=True)
   rollout_details = ndb.TextProperty()
   enterprise_policies = ndb.StringProperty(repeated=True)
+
+  deleted = ndb.BooleanProperty(default=False)
```
jrobbins, line 498: Let's call this `archived`.

api/stages_api.py
```diff
@@ -0,0 +1,251 @@
...
+    if redirect_resp:
+      return redirect_resp
+
+    # The stage_type field should not be able to change.
+    if 'stage_type' in kwargs:
+      self.abort(404, msg='stage_type field cannot be mutated.')
+
+    # Update specified fields. No need to create a gate for existing stage.
+    self._update_stage_vals(
+        stage, feature.feature_type, kwargs, create_gate=False)
```
jrobbins, line 228: The kwargs here is just the URL path parameters parsed by Flask.  I think you want the JSON content in the body of the POST request.  You can get that with:
`json_body = self.request.get_json(force=True, silent=True) or {}`

Because that will be needed in a couple of different places, you could define a one-line method `get_json_param_dict()` in BaseHandler.

jrobbins (changes requested):
This is off to a good start, but needs some changes.

api/stages_api.py
DanielRyanSmith: Renamed 👍

internals/core_models.py
DanielRyanSmith: Renamed 👍

api/stages_api.py
DanielRyanSmith: Yeah, my approach was wrong here and needed a good refactoring, but it should be correct now. 🙂

jrobbins (approved):

PR #2537, opened 2022-12-05
Add rollout stage to the edit_all page

yanndago:
- Show the rollout stage in the edit all page for all non-enterprise feature types
- Show the rollout stage in the edit page if it has been modified or if the feature is a breaking change
- Use STAGE_ENT_ROLLOUT as the id for the piggybacking on the Shipping stage.


yanndago:
For testing use https://3d01d64-dot-cr-status-staging.appspot.com/roadmap.
Test cases are the following:
1. Create a non-enterprise feature
2. Go to the edit all page -> There should be the "Start rollout" section
3. If any edit is made on that section, it should appear on that feature's page.

1. Create a non-enterprise feature and check "breaking change"
2. The Start rollout section should be shown on that feature page, and also on the edit all page

1. Create an enterprise feature
2. The Start rollout section should be there by default.

jrobbins:
I was thinking of a different UX:  we'll be implementing an "Add stage" button that will have some menu for the kinds of stages that a user could add, so "Enterprise rollout" would be one option in that menu.   Choosing that would add the stage to the feature, causing the stage to appear on the feature detail page and the edit_all page for that feature.  I think that is better than always showing blank rollout fields on the edit_all page because those fields would not be relevant for most non-enterprise features.

yanndago:
I have updated the CL to move the enterprise rollout to its own stage id, and to only add it on the edit all page for breaking changes that do not yet have it. It should be a good stop gap until the "add phase" is ready.
Available on staging at https://3a70acd-dot-cr-status-staging.appspot.com

Testing should be the following steps:
1. Create a non-enterprise feature
2. Go to the edit all page, there should not be a "rollout" phase
3. Mark it as breaking change and save it. There should be a rollout stage added on the edit all page and on the feature page.
4. Create an enterprise feature and verify that only one rollout phase is visible whether it is marked as breaking or not.

DanielRyanSmith (approved):
I think this is fine to go ahead and land. There might be some radical changes to this approach in the next couple of weeks as we allow for adding specific stages, but this should take care of things for now 👍 
Thanks @yanndago !

DanielRyanSmith (approved):
I went ahead and fixed a merge conflict associated with a block of code I recently merged.
I'm not able to test this on the staging URL because I can't log in with that URL, but I assume you've tested this well locally - thanks again!

PR #2538, opened 2022-12-05
Display feature stages rather than process stages

jrobbins:
The exiting code for displaying stages on the feature detail page iterated over the stages defined in the process.  However, a given feature could have additional stages added to it.  So, instead of using the process as the main list of stages, the PR makes the page use the feature's stages and it only references the process stages to get the name and description of each stage.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -361,8 +361,8 @@ class ChromedashFeatureDetail extends LitElement {
         ${this.renderControls()}
       </h2>
       ${this.renderMetadataSection()}
-      ${Object.values(this.feature.stages).map(feStageList =>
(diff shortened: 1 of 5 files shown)


DanielRyanSmith:
@jrobbins This change is very close to the one that I've been working on last week, and approaches it the same way (adding the stage_type equivalent enum to each Process dataclass), so I'm glad to see we're on the same page 😊

> I'm don't recall why the feature_api JSON returns a dict of list of stages rather than a flat stage list.

It is a dict with the keys being stage types and values being a list of stages associated with the feature that are of that stage type.
e.g. stages[160] would have a value of all of the stage entities associated with the feature that have a shipping stage type (160)for the incubation feature type. I hope that's a clear example 😅

DanielRyanSmith (approved):

PR #2544, opened 2022-12-05
Use feature stages in UI

DanielRyanSmith:
This change updates pieces of the UI to display and edit existing feature stages rather than display the same single stages for each feature type. This will allow for adding UI functionality to create multiple stages of the same type.

Major changes by file:
- converters.py: Change stages field to be a simple list of dictionaries representing stages associated with the feature, sorted by stage type and containing `stage_id`, `stage_type`, and `intent_stage` fields. Remove gates field from the feature converter, as it is currently not used.
- chromedash-app.js: Change the stage editing path to include the stage ID.
- chromedash-feature-detail.js: Generate stages shown based on stages associated with the feature.
- chromedash-guide-stage-page.js: Get stage data based on stage ID in path. Check if the new field `FeatureEntry.active_stage_id` matches the given stage ID. If so, denote this stage as active in the UI.
- chromedash-process-overview.js: Display process stages based on existing feature stages.
- form-definition.js: Remove `intent_stage` form field. This field does not change its options dynamically based on feature type even now, so it can allow the user to choose irrelevant stages as options. Additionally, the stages now will need to be set based on the stage ID. This field can still be set by checking the box to make the stage the active stage on the edit stage page.
- cs-client.js: Add new GET endpoint for stage data based on stage ID.
- stage_helpers.py: Add new function `get_feature_stage_ids_list()` that creates a list of dictionaries describing basic information about stages associated with the feature.
- main.py: Add new routes that allow passing the stage ID to feature edit forms.
- guide.py: Update `process_post_data()` to check if a `stage_id` param is used. If so, this denotes that all changes are to a single stage, and a new method, `update_single_stage()`, can be used to quickly update this stage.


jrobbins:
I looked through this code and it generally looks good.   However, I would prefer it in smaller PRs that are (a) easier to carefully review, (b) allow us to merge the active_stage_id changes after doing the needed migration scripts, and (c) allow us to merge a lot of this work while holding off on the editing forms until we have edit_all working.

I suggest PRs for:
1. The converters.py and stage_helpers.py changes to the JSONification of feature stages, plus the updates to chromedash-feature-detail.js that makes use of the new format for this data.  Maybe some of the chromedash-process-overview.js fit in this too.
2. Add the stage_id to the URLs and add the new parameter to the URL routes, but make that parameter optional for now and don't use it in the guide POST handler yet.   This will give time for users to load the new version of the app that generates the new URLs.
3. form-definition.js: Remove intent_stage form field.
4. Blocked on migration script for active_stage_id: add the active_stage_id_field, use that field in the guide elements to indicate the active stage, add getStage() to cs-client.js and use it to display the stage name.  At this point the stage_id in the URL routes would become required.
5. Blocked on a solution for edit_all: the remaining changes to the guide elements and guide.py.

Maybe we could get 1, 2, and 3 merged and deployed this week so that users are visiting the new stage editing URLs.

DanielRyanSmith:
Closing this and breaking changes into smaller PRs.

Issue #2539, opened 2022-12-05
Old feature list is returning zero results when searching by milestone


jrobbins:
E.g.,
https://chromestatus.com/features#milestone%3D59

jrobbins:
This fix was deployed last Tuesday.

PR #2545, opened 2022-12-06
Milestone search fix

DanielRyanSmith:
Addresses #2539

This change adds a `milestone` field to feature data passed to the main page feature list, so that the features can be filtered by milestone.


jrobbins (approved):
Yep, that batch fetch approach should perform OK.
It could be even better if you use NDB fetch_async() when you create the other futures.

jrobbins (approved):

Issue #2521, opened 2022-11-29
Accept and reflect search parameters in the URL


jrobbins:
**Is your feature request related to a problem? Please describe.**
Our new feature list page has search widgets that allow the user to do a search, but that is not reflected in the URL, so a search is not bookmark able.

**Describe the solution you'd like**
* [ ] Parse search parameters in the URL to populate the search widgets
* [ ] When the user does a search, reflect the state of the search widgets in the URL

Our search widgets cannot handle negation or logical-OR, so for now if those are present in the URL, just put the entire user query into the text box rather than the individual field search widgets.

PR #2547, opened 2022-12-06
Populate query from URL

KyleJu:
A part of #2521. 

- Parse search parameters "q" in the URL to populate the search widgets
- Show the default search results based on the URL
- q's value should be encoded

### Test
Go to /newfeatures?q=abc. Query abc should be populated in the searchbar and its result is shown.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -113,6 +113,7 @@ class ChromedashApp extends LitElement {
     page('/newfeatures', (ctx) => {
       this.pageComponent = document.createElement('chromedash-all-features-page');
       this.pageComponent.user = this.user;
+      this.pageComponent.rawQuery = window.csClient.parseRawQuery(ctx.querystring);
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
     });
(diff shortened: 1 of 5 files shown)


client-src/js-src/cs-client.js
```diff
@@ -325,6 +325,20 @@ class ChromeStatusClient {
   getSpecifiedChannels(start, end) {
     return this.doGet(`/channels?start=${start}&end=${end}`);
   }
+
+  parseRawQuery(rawQuery) {
```
jrobbins, line 329: This could use a docstring.

client-src/elements/chromedash-app.js
```diff
@@ -113,6 +113,7 @@ class ChromedashApp extends LitElement {
     page('/newfeatures', (ctx) => {
       this.pageComponent = document.createElement('chromedash-all-features-page');
       this.pageComponent.user = this.user;
+      this.pageComponent.rawQuery = window.csClient.parseRawQuery(ctx.querystring);
```
jrobbins, line 116: I'm happy to see that this part worked out pretty cleanly.   I am curious to see if this code is executed when you change the URL (as part of your upcoming PR).

jrobbins (approved):

Issue #2335, opened 2022-10-13
Combine the feature details page with the feature editing page (A.K.A. process overview page)


jrobbins:
**Is your feature request related to a problem? Please describe.**
Our current UI for editing features uses a pencil icon on the feature detail page and in each row of a list view.  Clicking the pencil icon takes the user to /guide/edit.html which gives an overview of the process for the current type of feature. That page has links to forms to edit fields in each stage of the process.  The user can navigate back and forth between the feature detail page and the feature editing page.

**Describe the solution you'd like**
Combine the feature editing page and feature editing page (process overview).   These two pages already have a parallel structure and are somewhat redundant in that they both list out stages.  The reasons for doing this are to simplify our site map, save the user a click, and make our UX more similar to the UX of an internal tool that.

The new feature detail page will have:
* a 2/3 width main column and a 1/3 width right-hand column
* roughly the same feature overview content displayed at the top of those two columns
* separate accordion sections for each stage rather than the current single large "additional details" section.
    * these sections will have a button that takes the user into the stage editing forms
* a new accordion section for activities (comments and changes)
* gate chips that open an approvals column dialog instead of the current stamp icon and modal approvals dialog

PR #2549, opened 2022-12-07
Display gate chips in feature details.

jrobbins:
This PR implements the last checkbox item of issue #2335.
Basically, we show gate chips on the feature detail page inside the accordion sections for the stages (that have approval gates).

In this PR:
* In chromedash-feature-page.js: get the gates data and pass it to chromedash-feature-detail, also accept the event that is fired when the user clicks on a gate chip.
* In chromedash-feature-detail.js: accept the passed in gates data and render them in the accordion section for the appropriate stage.
* chromedash-gate-chip.js and theme.js: Define the colors and icons for gates in the PREPARING state.  This was never needed before because only active gates are shown on the myfeatures page whereas all gates are shown here.

### DIFF
--- a/client-src/elements/chromedash-feature-page.js
+++ b/client-src/elements/chromedash-feature-page.js
@@ -87,6 +87,7 @@ export class ChromedashFeaturePage extends LitElement {
       user: {type: Object},
       featureId: {type: Number},
       feature: {type: Object},
+      gates: {type: Array},
       comments: {type: Array},
       process: {type: Object},
       dismissedCues: {type: Array},
@@ -102,6 +103,7 @@ export class ChromedashFeaturePage extends LitElement {
     this.user = {};
     this.featureId = 0;
     this.feature = {};
+    this.gates = {};
     this.comments = {};
     this.process = {};
     this.dismissedCues = [];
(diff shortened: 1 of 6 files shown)


jrobbins:
And, this is how each gate looks initially, before the user requests a review or sends in intent email to blink-dev:
image

client-src/elements/chromedash-feature-page.js
```diff
@@ -447,9 +459,12 @@ export class ChromedashFeaturePage extends LitElement {
         .user=${this.user}
         ?canEdit=${this.userCanEdit()}
         .feature=${this.feature}
+        .gates=${this.gates}
         .comments=${this.comments}
         .process=${this.process}
-        .dismissedCues=${this.dismissedCues}>
+        .dismissedCues=${this.dismissedCues}
+        @open-approvals-event=${this.handleOpenApprovals}
```
KyleJu, line 466: @open-approvals-event=${this.handleOpenApprovals} where is it being used?

jrobbins: When the user clicks on a chromedash-gate-chip element, it generates an "open-approvals" event.  Those events are caught here and handleOpenApprovals() is called, which opens the approvals dialog.  Basically, dialog state is managed at the page level rather than having lower-level elements open a dialog directly.

KyleJu: So this is an EventListener?

KyleJu (approved):

client-src/elements/chromedash-feature-page.js
jrobbins: Yep, the @ syntax is pretty handy in lit.
https://lit.dev/docs/components/events/#adding-event-listeners-in-the-element-template

KyleJu: Aha thanks!

PR #2550, opened 2022-12-07
Encode url when search

### LINKED ISSUE #2521: Accept and reflect search parameters in the URL

KyleJu:
Fix #2521.

TODO:
Errors when pressing go back

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -116,6 +116,7 @@ class ChromedashApp extends LitElement {
       this.pageComponent.rawQuery = window.csClient.parseRawQuery(ctx.querystring);
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
+      this.pageComponent.addEventListener('search', this.handleSearchQuery.bind(this));
     });
     page('/feature/:featureId(\\d+)', (ctx) => {
       this.pageComponent = document.createElement('chromedash-feature-page');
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-app.js
```diff
@@ -187,6 +188,53 @@ class ChromedashApp extends LitElement {
...
+ * Update window.locaton with new query params.
+ * @param {string} key is the key of the query param.
+ * @param {string} val is the unencoded value of the query param.
+ */
+  updateURLParams(key, val) {
+    // Update the query param object.
+    const rawQuery = window.csClient.parseRawQuery(window.location.search);
+    rawQuery[key] = encodeURIComponent(val);
+
+    // Aseemble the new URL.
```
jrobbins, line 205: Typo: Aseemble -> Assemble.

jrobbins (changes requested):
This looks pretty good, but I tried it out and I think that I found a bug:
Repeatedly pushing the "Search" button causes the structured query terms to be added to the URL and then to the unstructured search box over and over, once for each push.
image
Users would normally press that button more than once when they are refining their query.

One way to resolve that would be to parse the URL query into structured terms and update the state of the widgets, but I was hoping that we could defer doing that much work.
Another way to handle it might be to only parse the URL query on the first time that the SPA visits that virtual page.  So, following a link would cause the URL to be parsed, but clicking the Search button would cause the URL to be updated but not parsed again.  That might cause the browser's back arrow navigation to correctly set the URL to an earlier location, but the SPA would not parse it so the user would see no effect.

Another approach could be to avoid adding redundant terms in chromedash-feature-filter.js computeQuery().  That might be the best option.

jrobbins (approved):
OK, that approach should work.

PR #2555, opened 2022-12-07
Add routes for intent stage and stage ID to guide

DanielRyanSmith:
These new routes will exist but remain unused for the moment.


jrobbins (approved):
I was looking for the part that made those optional, but now I see that they are just additional rules and the existing rules are also still in effect.

PR #2557, opened 2022-12-07
Use active_stage_id to display active on detail page

DanielRyanSmith:
The active stage will now be determined by a matching intent stage, as well as the active_stage_id field on the feature detail page.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -319,7 +319,8 @@ class ChromedashFeatureDetail extends LitElement {
     const processStage = this.findProcessStage(feStage);
     if (processStage === null) return nothing;
     const fields = DISPLAY_FIELDS_IN_STAGES[processStage.outgoing_stage];
-    const isActive = (this.feature.intent_stage_int == processStage.outgoing_stage);
+    const isActive = (this.feature.active_stage_id === feStage.stage_id &&
+        this.feature.intent_stage_int == processStage.outgoing_stage);
     if (fields === undefined || fields.length == 0) {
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-feature-detail.js
```diff
@@ -319,7 +319,8 @@ class ChromedashFeatureDetail extends LitElement {
     const processStage = this.findProcessStage(feStage);
     if (processStage === null) return nothing;
     const fields = DISPLAY_FIELDS_IN_STAGES[processStage.outgoing_stage];
-    const isActive = (this.feature.intent_stage_int == processStage.outgoing_stage);
+    const isActive = (this.feature.active_stage_id === feStage.stage_id &&
```
jrobbins, line 322: Should this be logical-OR?  Are you using both conditions so that the old condition can work before running the migration script?  If that's the case, I think it should be `||` for OR.

jrobbins (changes requested):

client-src/elements/chromedash-feature-detail.js
DanielRyanSmith: Actually, I think we can just match the stage ID here instead of using the intent stage as well. This was a holdover from when there was a chance that a single stage entity could have multiple process stages associated with it (e.g. "prepare to ship" and "ship"). I'll make the change so that the stage ID is the only necessary match.

DanielRyanSmith: This PR is ready for another look, but we can also talk about this implementation if it needs more conversation

jrobbins (approved):

PR #2559, opened 2022-12-08
Show early stable release dates on roadmap.

jrobbins:
As will be discussed at our team meeting tomorrow, the chrome release schedule is changing slightly with M110.

We need to keep in sync with this change made to chromiumdash:
https://chrome-internal.googlesource.com/infra/infra_internal/+/c0bc402375948a00185b3ca3b649641a87149af8

In this PR:
* Starting with M110, we should display the JSON field `final_beta` instead of `stable_date`.
* Add an (i) icon with a popup that gives a tiny bit of information and a link to find more.


KyleJu (approved):

KyleJu:
> This is what it looks like: image

NIt: I think the "See detailed dates" does not have to be on a new line. wdyt?

PR #2560, opened 2022-12-08
Update start param in URL

### LINKED ISSUE #2343: Implement pagination in feature search API and newfeatures UI

KyleJu:
Fix #2343. update the start= query string parameter in the URL.


KyleJu:
This PR needs #2550

PR #2561, opened 2022-12-08
Use specific stage data for viewing and editing feature data

DanielRyanSmith:
This change updates pieces of the UI to display and edit existing feature stages rather than display the same single stages for each feature type. This will allow for adding UI functionality to create multiple stages of the same type.

Major changes by file:
- chromedash-app.js: Change the stage editing path to include the stage ID.
- chromedash-feature-detail.js: Generate stages shown based on stages associated with the feature.
- chromedash-guide-stage-page.js: Get stage data based on stage ID in path. Check if the new field `FeatureEntry.active_stage_id` matches the given stage ID. If so, denote this stage as active in the UI.
- chromedash-process-overview.js: Display process stages based on existing feature stages.
- cs-client.js: Add new GET endpoint for stage data based on stage ID.
- stage_helpers.py: Add new function `get_feature_stage_ids_list()` that creates a list of dictionaries describing basic information about stages associated with the feature.
- guide.py: Update `process_post_data()` to check if a `stage_id` param is used. If so, this denotes that all changes are to a single stage, and a new method, `update_single_stage()`, can be used to quickly update this stage.


client-src/elements/chromedash-guide-stage-page.js
```diff
@@ -292,7 +295,7 @@ export class ChromedashGuideStagePage extends LitElement {
     const formattedFeature = formatFeatureForEdit(this.feature);
     return html`
       <form name="feature_form" method="POST"
-        action="/guide/stage/${this.featureId}/${this.stageId}">
+        action="/guide/stage/${this.featureId}/${this.stageId}/${this.intentStage}">
```
jrobbins, line 298: Is the server-side of this safe for users who already have a tab open on our app?  I can't put my finger on the right line of server-side code, but I want to be sure that someone who had a tab open can still submit.  It used to have the intentStage (called stagId) directly after the feature ID, and I'm not seeing how we are still backwards compatible with that.

DanielRyanSmith: Sorry, I did not know that was a requirement, but it is possible.

I originally planned to only pass the stage ID in the URL, but realized we also need the intent stage with this new implementation, so I re-added it to the end of the URL, which is why we have it ordered this way. I've made the change in [this commit](https://github.com/GoogleChrome/chromium-dashboard/pull/2561/commits/649d53b54dbfd661191318181ff5cd61d096faff) to the URL address to be specified as `/guide/stage/{featureId}/{intentStage}/{stageId}`, and then to handle the backwards-compatible case where a user submits without a stage ID.

PR #2562, opened 2022-12-08
View and edit stage-specific data on edit_all page

DanielRyanSmith:
This change updates the edit-all page to be populated with and edit the fields of stages associated with the feature. This change will dynamically change the fields on the edit all page to match the feature type and stages that the user has selected for their feature, as well as make changes specifically to the stages in which they are editing. Since there is currently no way for users to create multiple stages, this change should be mostly invisible for users at the moment.


client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -157,24 +137,76 @@ export class ChromedashGuideEditallPage extends LitElement {
     `;
   }
 
+  findProcessStage(feStage) {
```
jrobbins, line 140: This is in utils.js now.

client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -157,24 +137,76 @@ export class ChromedashGuideEditallPage extends LitElement {
...
+      formsToRender.push(this.renderStageFormFields(
+        formattedFeature, processStage, feStage, formFields));
+      allFormFields = [...allFormFields, ...formFields];
+    }
+
+    return [allFormFields, formsToRender];
+  }
+
   renderForm() {
+    const formattedFeature = formatFeatureForEdit(this.feature);
```
jrobbins, line 199: Could we use the instance variable `this.featureForEdit` here?   It just seems like two ways to get the same result, so we should pick one.

jrobbins (approved):

PR #2569, opened 2022-12-09
remove "SHIPPED" stage types

DanielRyanSmith:
The migration script for calculating the active stage is written under the assumption that the “shipped” stage type will exist under many feature types. This PR removes those stage types that have been agreed upon as being okay to remove.


internals/processes.py
```diff
@@ -274,7 +274,7 @@ def process_to_dict(process):
       [],
       [],
       core_enums.INTENT_SHIP, core_enums.INTENT_SHIPPED,
-      stage_type=core_enums.STAGE_BLINK_SHIPPED),
+      stage_type=core_enums.STAGE_BLINK_SHIPPING),
```
jrobbins, line 277: Should this whole `ProcessStage` be deleted?  (Same for the other 2)

DanielRyanSmith: The problem here is that if we outright remove this, features that have set this stage as active will not see "Revisit" options to edit their feature, because it the previous stages will not register as "prior" stages. We could work to make this change, but I also added a fix in [this commit](https://github.com/GoogleChrome/chromium-dashboard/pull/2569/commits/c9a6e55304f1cc42a312dee6a8e32621c99bbaeb) which only renders process stages that match actual stage entities. Maybe this is a better quick fix for now?

DanielRyanSmith:
@jrobbins Optionally here, we can just ignore this change and save it for next week, as long as we don't run the script to create the active_stage_id field.

jrobbins (approved):
Hold on to this until after today's release goes live.

PR #2570, opened 2022-12-09
Fix rendering gate chips on details page.

jrobbins:
Update the field name of the stage's ID.
I think this changed in the API after I did my initial testing, so basically a bad logical merge.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -320,7 +320,7 @@ class ChromedashFeatureDetail extends LitElement {
   }
 
   renderGateChips(feStage) {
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-feature-detail.js
```diff
@@ -320,7 +320,7 @@ class ChromedashFeatureDetail extends LitElement {
   }
 
   renderGateChips(feStage) {
-    const gatesForStage = this.gates.filter(g => g.stage_id == feStage.id);
+    const gatesForStage = this.gates.filter(g => g.stage_id == feStage.stage_id);
```
DanielRyanSmith, line 323: This is a redundant property name, which is my fault. We can look into changing this at some point moving forward.

DanielRyanSmith (approved):

PR #2571, opened 2022-12-10
Fix pagination indexes

KyleJu:
Fix pagination range issue.

- Parse query params as number, rather than string
- When there is no feature, it should display {start} - {start} of 0

Screen Shot 2022-12-09 at 5 38 46 PM

### DIFF
--- a/client-src/elements/chromedash-all-features-page.js
+++ b/client-src/elements/chromedash-all-features-page.js
@@ -50,12 +50,12 @@ export class ChromedashAllFeaturesPage extends LitElement {
 
     if (this.rawQuery.hasOwnProperty('start') &&
       !Number.isNaN(parseInt(this.rawQuery['start']))) {
-      this.start = +this.rawQuery['start'];
+      this.start = parseInt(this.rawQuery['start']);
     }
 
     if (this.rawQuery.hasOwnProperty('num') &&
       !Number.isNaN(parseInt(this.rawQuery['num']))) {
-      this.num = +this.rawQuery['num'];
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-all-features-page.js
```diff
@@ -48,12 +48,12 @@ export class ChromedashAllFeaturesPage extends LitElement {
       this.query = this.rawQuery['q'];
     }
 
-    if (this.rawQuery.hasOwnProperty('start')) {
-      this.start = this.rawQuery['start'];
+    if (this.rawQuery.hasOwnProperty('start') && !Number.isNaN(+this.rawQuery['start'])) {
+      this.start = +this.rawQuery['start'];
```
jrobbins, line 52: Can we use `parseInt()` here rather than `+`?
Some future maintainer might not understand why there is a `+` there.

client-src/elements/chromedash-feature-table.js
```diff
@@ -181,7 +181,7 @@ class ChromedashFeatureTable extends LitElement {
   renderPagination() {
     // Indexes of first and last items shown in one-based counting.
     const firstShown = this.start + 1;
-    const lastShown = this.start + this.features.length;
+    const lastShown = this.features.length ? this.start + this.features : firstShown;
```
jrobbins, line 184: For the case where there are zero results, it might be better to not show the pagination at all.  That could be added as a new condition just before line 205:
`if (this.features.length === 0) { return nothing; }`

jrobbins (changes requested):

PR #2580, opened 2022-12-13
Add /newfeatures page rollout banner

### LINKED ISSUE #2518: Implement staged rollout of new feature list page
Before making the new feature list page the default, we should give users the option to try it.
And after making the new feature list page the default, we should offer users the ability to access the old page if they feel they need to.

* [ ] Add a banner or something on /features to allow users to try out /newfeatures
* [ ] Move the new feature list to /features and the old feature list page to /oldfeatures
* [ ] Add a banner or something on the new feature list page to allow users to navigate to /oldfeatures

**Additional context**
I don't think that we need to store a user preference for opting in or out, instead users will just navigate to the alternative URL during that visit.

KyleJu:
Fix #2518.

Screen Shot 2022-12-12 at 11 42 41 PM

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -70,7 +70,6 @@ class ChromedashApp extends LitElement {
       bannerMessage: {type: String},
       bannerTime: {type: Number},
       pageComponent: {attribute: false},
-      featureHref: {type: String},
       contextLink: {type: String}, // used for the back button in the feature page
     };
   }
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -18,6 +18,14 @@ class ChromedashApp extends LitElement {
           width: 100%;
         }
 
+        #rollout {
+          width: 100%;
+          text-align: center;
+          padding: 1em;
+          color: black;
+          background: lightgrey;
+        }
+
         #app-content-container {
           display: flex;
           flex-direction: column;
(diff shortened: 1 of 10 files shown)


client-src/elements/chromedash-app.js
```diff
@@ -62,6 +70,7 @@ class ChromedashApp extends LitElement {
       bannerMessage: {type: String},
       bannerTime: {type: Number},
       pageComponent: {attribute: false},
+      featureHref: {type: String},
```
jrobbins, line 73: Where is this used?  I see a function getFeatureHref() that is used instead.

client-src/elements/chromedash-app.js
```diff
@@ -253,7 +274,8 @@ class ChromedashApp extends LitElement {
                 .user=${this.user}
                 .appTitle=${this.appTitle}
                 .googleSignInClientId=${this.googleSignInClientId}
-                .currentPage=${this.currentPage}>
+                .currentPage=${this.currentPage}
+                ?isNewFeatures=${true}>
```
jrobbins, line 278: Is this supposed to always be true?
I thought that maybe you were trying to offer the user two modes with a boolean to determine which mode was active.  But, there needs to be some way for the user to affect that boolean's value.

KyleJu: I actually had a bit trouble setting the correct boolean values because for /features, the chromedash-header is rendered in _base.html, while it is here for /newfeatures. We have two ways to render the same tabs. Perhaps we can talk more in our meeting

KyleJu: Per our offline discussions, we will stick to static tab for now.

jrobbins (approved):

PR #2582, opened 2022-12-13
Introduce permission to comment. More perms checks in approvals dialog.

jrobbins:
Now that gate chips allow more users to open the approvals dialog, be more careful about which widgets they see in that dialog.  Also, limit commenting to only registered users so that we don't need to deal with spam.


dlaliberte (approved):
LGTM

Issue #2519, opened 2022-11-29
Display a short stage name in front of gate chips


jrobbins:
**Is your feature request related to a problem? Please describe.**
We currently show gate chips on the myfeatures page in the pending-my-review box.  These gate chips clearly show the gate type (e.g., "API owners").  However, the stage is not shown until after the user clicks to see that information in the approvals dialog.  The stage name could be useful context, e.g., the API owners might want to focus their efforts on features that are seeking approval to ship before looking at features that are seeking approval for origin trials.  Also, it is possible for a feature owner to be seeking approval on multiple stages at the same time, so showing the stage name will disambiguate among multiple rows of gate chips (one row for each stage).  

**Describe the solution you'd like**
Basically, implement the TODO comment in chromedash-feature-row.js function renderActiveStageAndGates().

The layout is already done, so I think we just need a value for the variable `stageName`.  Also, some of the names of stages can be pretty long, so we should probably use one-word or two-word labels for the state names to keep the display compact.

**Additional context**
The chromedash-feature-table element fetches a gates for each feature, and then it passes that information to chromedash-feature-row.  However, in that data the stages are only represented by the stage_id numeric field of each gate.  We don't yet make a call to get the list of stages for each feature.

DanielRyanSmith:
Addressed in #2634

PR #2583, opened 2022-12-14
Implement the layout and some details for the "gate column"

jrobbins:
This new element will eventually replace the approvals dialog.   The "gate column" is sort of like a tall, narrow version of the approvals dialog except that it shows exactly one gate at a time (whereas the dialog mixed together four gates).

This PR is not a complete implementation, in fact, it makes no user-visible change yet, but I think that it is enough to review and merge so that we can build on it.

In this PR:
* chromestatus-app.js: Refactor the way that the app content is laid out so that some pages shift the content to the left, leaving room for a sidebar on the right.  Define a method to show the gate column.   The CSS for the sidebar is a little tricky in that it positions the sidebar relative to the blank space reserved for it, but it also allows the sidebar to overlap the main page content if the window is narrow, and it keeps the sidebar in view as the user scrolls.
* chromedash-feature-detail.js and util.js: Refactor a method to a util function so that it can be reused.
* chromedash-gate-column.js: This lays out the pieces of information that will be displayed in the gate column, but it doesn't get real data for everything yet.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -43,6 +43,32 @@ class ChromedashApp extends LitElement {
           width: 100%;
         }
 
+        #content-sidebar-space {
+          position: sticky;
+          flex-shrink: 100;
+          width: 360px;
+        }
+
+        #sidebar {
+          position: absolute;
+          top: 0;
+          right: 0;
+          width: var(--sidebar-width);
+          bottom: 0;
+        }
+        #sidebar[hidden] {
+          display: none;
+        }
+        #sidebar-content {
+          position: sticky;
+          top: 10px;
+          height: 85vh;
+          border: var(--sidebar-border);
+          border-radius: var(--sidebar-radius);
+          background: var(--sidebar-bg);
+          padding: var(--content-padding);
+        }
+
         @media only screen and (max-width: 700px) {
           #content {
             margin-left: 0;
@@ -63,6 +89,7 @@ class ChromedashApp extends LitElement {
       bannerTime: {type: Number},
       pageComponent: {attribute: false},
       contextLink: {type: String}, // used for the back button in the feature page
+      sidebarHidden: {type: Boolean},
     };
   }
 
@@ -77,6 +104,7 @@ class ChromedashApp extends LitElement {
     this.bannerTime = null;
     this.pageComponent = null;
     this.contextLink = '/features';
+    this.sidebarHidden = true;
   }
 
   connectedCallback() {
@@ -197,6 +225,14 @@ class ChromedashApp extends LitElement {
     this.updateURLParams('q', e.detail.query);
   }
 
+  showSidebar() {
+    this.sidebarHidden = false;
+  }
+
+  hideSidebar() {
+    this.sidebarHidden = true;
+  }
+
   /**
  * Update window.locaton with new query params.
  * @param {string} key is the key of the query param.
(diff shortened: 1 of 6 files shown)


jrobbins:
There's nothing in any part of the existing UI that allows the user to activate the gate column yet, however if you open the console and enter: 
`document.querySelector('chromedash-app').showGateColumn({id:3510}, {stage_type: 160}, {id: 3520, feature_id: 3510, team_name: 'API owners'})`  
then you will see:
image

jrobbins:
Off-topic regarding issue #2519:  The utils.js function findProcessStage() is what I was talking about in our meeting today, I forgot that that part already existed.  However, it just has the fill stage name whereas I think we need some kind of 4 to 10 letter short stage name.

client-src/elements/chromedash-app.js
```diff
@@ -240,6 +284,33 @@ class ChromedashApp extends LitElement {
     return url;
   }
 
+  renderContentAndSidebar() {
+    const wide = (this.pageComponent &&
+                  this.pageComponent.tagName == 'CHROMEDASH-ROADMAP-PAGE');
+    if (wide) {
+      return html`
+        <div id="content-component-wrapper" wide>
```
KyleJu, line 292: What is the `wide` here?

client-src/elements/chromedash-gate-column.js
```diff
@@ -0,0 +1,255 @@
...
+       }
+
+       #votes-area th {
+         font-weight: bold;
+       }
+
+       #review-status-area {
+         margin: var(--content-padding-half) 0;
+       }
+
```
KyleJu, line 39:
```suggestion
```

client-src/elements/chromedash-gate-column.js
```diff
@@ -0,0 +1,255 @@
...
+  constructor() {
+    super();
+    this.user = {};
+    this.feature = {};
+    this.stage = {};
+    this.gate = {};
+    this.process = {};
+    this.votes = [];
+    this.comments = [];
+    this.loading = true;
```
KyleJu, line 66:
```suggestion
    this.loading = false;
```
Maybe false by default?

client-src/elements/chromedash-gate-column.js
```diff
@@ -0,0 +1,255 @@
...
+    ]).then(([process, approvalRes, commentRes]) => {
+      this.process = process;
+      this.votes = approvalRes.approvals.filter((v) =>
+        v.gate_id == this.gate.id);
+      this.comments = commentRes.comments;
+      this.needsSave = false;
+      this.loading = false;
+    }).catch(() => {
+      showToastMessage('Some errors occurred. Please refresh the page or try again later.');
+      this.handleCancel();
```
KyleJu, line 89: where is handleCancel defined?

client-src/elements/chromedash-gate-column.js
```diff
@@ -0,0 +1,255 @@
...
+
+  renderVoteReadOnly(vote) {
+    // TODO(jrobbins): background colors
+    return this.findStateName(vote.state);
+  }
+
+  renderVoteMenu(state) {
+    // hoist is needed when <sl-select> is in overflow:hidden context.
+    return html`
+      <sl-select name="${this.gate.id}"
```
KyleJu, line 161: Nit: fix the indentation here

client-src/elements/chromedash-app.js
jrobbins: It is an attribute used to trigger a CSS rule to make the main content of the page wide on certain pages, like the roadmap page.  Those pages won't have a sidebar.   I suppose that it could be done with a CSS class instead of a separate attribute, but this is existing code so I kept it as-is in this PR.

client-src/elements/chromedash-gate-column.js
jrobbins: It was not defined yet.  I have added it now.

client-src/elements/chromedash-gate-column.js
jrobbins: Done

KyleJu (approved):

PR #2598, opened 2022-12-21
Stage field organization

DanielRyanSmith:
This change adjusts the approach of how we define UI form fields, and reorganizes them to allow for easier manipulation, partitioning, and reading. It is a step toward easier display of feature stages and the possibility of multiple stages of the same types being displayed.

In `form_definition.js`, the field arrays have been changed to objects which describe how the form should be displayed in JSON. Each stage has JSON that describes the fields for each stage, and also partitions them into sections, which can be adjusted to better organize the data displayed to users in the future.

Additionally, stage forms definitions can be changed based on the stage type, which allows for more customizable forms for each feature type.


client-src/elements/form-definition.js
```diff
@@ -319,41 +287,36 @@ const IMPLSTATUS_ORIGINTRIAL = [
...
 const EXISTING_ORIGINTRIAL = [
   'experiment_goals', 'experiment_risks',
   'experiment_extension_reason', 'ongoing_constraints',
   'intent_to_experiment_url', 'intent_to_extend_experiment_url',
-  'i2e_lgtms', 'origin_trial_feedback_url', 'comments',
+  'i2e_lgtms', 'origin_trial_feedback_url',
 ];
 
-const PSA_IMPLEMENT = ['motivation', 'spec_link', 'standard_maturity', 'comments'];
+const PSA_IMPLEMENT = ['spec_link', 'standard_maturity'];
```
DanielRyanSmith, line 308: @jrobbins Do we still want to have different fields for the same general stage type? For instance, this definition excludes the `intent_to_implement_url`, `explainer_links`, and `spec_link` from the developer-facing code changes feature type.

DanielRyanSmith: Regardless, this file likely needs a greater cleanup and organization to make these definitions easier to ingest and edit

jrobbins: > @jrobbins Do we still want to have different fields for the same general stage type? For instance, this definition excludes the `intent_to_implement_url`, `explainer_links`, and `spec_link` from the developer-facing code changes feature type.

Yes, I expect that the fields offered in each stage will still be determined by the feature type.  That could happen indirectly because stage types are determined by feature type.

client-src/elements/form-definition.js
```diff
@@ -85,378 +95,456 @@ export function formatFeatureForEdit(feature) {
...
+
+  [STAGE_PSA_IMPLEMENT_FIELDS]: PSA_IMPLEMENT_FIELDS,
+  [STAGE_PSA_DEV_TRIAL]: FLAT_DEV_TRIAL_FIELDS,
+  [STAGE_PSA_SHIPPING]: PSA_PREPARE_TO_SHIP_FIELDS,
+
+  [STAGE_DEP_DEV_TRIAL]: FLAT_DEV_TRIAL_FIELDS,
+  [STAGE_DEP_DEPRECATION_TRIAL]: DEPRECATION_ORIGIN_TRIAL_FIELDS,
+  [STAGE_DEP_SHIPPING]: DEPRECATION_PREPARE_TO_SHIP_FIELDS,
+  [STAGE_ENT_ROLLOUT]: FLAT_ENTERPRISE_PREPARE_TO_SHIP_FIELDS,
+  [STAGE_ENT_SHIPPED]: FLAT_PREPARE_TO_SHIP_FIELDS,
```
DanielRyanSmith, line 549: This is not an ideal solution to mirror the enums server side, but maybe we can go with this for now and plan a better approach later on.

DanielRyanSmith:
@jrobbins I went a little wild and changed the way we define fields entirely, but the display overall should be mostly the same. The major difference is that fields are no longer repeated in different stages.

client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -135,62 +132,64 @@ export class ChromedashGuideEditallPage extends LitElement {
...
+          stageId=${stageId}
           value=${formattedFeature[field]}>
         </chromedash-form-field>
       `)}
     </section>
-    `];
+    `;
+  }
+
+  getForms(formattedFeature, feStages) {
```
jrobbins, line 161: This function is kind of hard to understand.  That may be an existing problem, but looking at it again now, it needs at least a docstring.

client-src/elements/form-definition.js
```diff
@@ -85,378 +95,457 @@ export function formatFeatureForEdit(feature) {
...
+// Note STAGE_* enum values 500-999 are reseverd for future WP processes.
+
+// Define enterprise feature processes.
+// Note: This stage can ge added to any feature that is following any process.
+const STAGE_ENT_ROLLOUT = 1061;
+const STAGE_ENT_SHIPPED = 1070;
+
+// TODO(danielrsmith): Differentiate origin trial and trial extension fields.
+export const FORMS_BY_STAGE_TYPE = {
+  [STAGE_BLINK_INCUBATE]: FLAT_INCUBATE_FIELDS,
```
jrobbins, line 529: TIL, you can use square brackets to make JS use var values as keys rather than using the names.

jrobbins (approved):

client-src/elements/chromedash-guide-editall-page.js
DanielRyanSmith: Agreed, and a more thorough explanation has been added.

PR #2600, opened 2022-12-21
fix CC emails property name

DanielRyanSmith:
The property name for the CC emails field had been changed to its new name, but the UI is still reliant on the old names, which is causing the field to not be populated. This is a simple change to fix that.

### DIFF
--- a/api/converters.py
+++ b/api/converters.py
@@ -413,7 +413,7 @@ def feature_entry_to_json_verbose(fe: FeatureEntry) -> dict[str, Any]:
   }
   d['tags'] = d.pop('search_tags', None)
   d['editors'] =  d.pop('editor_emails', [])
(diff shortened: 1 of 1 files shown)


api/converters.py
```diff
@@ -413,7 +413,7 @@ def feature_entry_to_json_verbose(fe: FeatureEntry) -> dict[str, Any]:
   }
   d['tags'] = d.pop('search_tags', None)
   d['editors'] =  d.pop('editor_emails', [])
-  d['cc_emails'] = d.pop('cc_emails', [])
+  d['cc_recipients'] = d.pop('cc_emails', [])
```
jrobbins, line 416: We want the UI to use `cc_emails` eventually, right?  So, it would be good to hav a TODO comment saying that.

jrobbins (approved):

Issue #2334, opened 2022-10-13
Replace the approvals dialog with a new approvals column


jrobbins:
**Describe the solution you'd like**
We will replace the current approvals dialog because it is different than the UI used by an internal tool, and because it is modal and covers up other page content.

The new UX will consist of a left-hand column that occupies about 1/3 of the main content width when in use.  When not in use, that area will be blank or used for information that we are OK with covering up.  It will be like a dialog box in that it has a title area and an X icon to close it, however it will normally stay open and be non-modal.

The contents of the approvals column will include: 
* name of the stage and gate
* status indicator for the overall gate
* existing vote values and widget for a user with permission to enter or change their vote
* future: read-only instructions configured by the approvers
* list of review requirements
* list of previous comments
* text area to submit a new comment
* checkbox to post that comment to the appropriate blink-dev thread, if any
* submit button for comments

**Additional context**

This UI element will be opened when the user clicks on a Gate chip.

The user will be asked to confirm losing data if the user has unsaved changes in the approval column, and then they close the column, close the browser tab, or navigate to a different Gate.

jrobbins:
Remaining steps now:
* Something for SLO
* Make comments actually send email notifications
* colors for vote rows

jrobbins:
This was done long ago.

PR #2601, opened 2022-12-21
Display comments in gate column.

jrobbins:
This is progress on issue #2334.  This PR replaces a TODO for rendering comments with code to actually render comments.  

Unlike comments in the approval dialog, the gate column is a more narrow space where some details are better laid out vertically.  And, the comments in the gate column are intended to be in reverse chronological order with the text field above the newest comment.

The comment posting behavior of the gate column is different than the approvals dialog: the dialog closes after posting, whereas the gate column updates to show the new comment list (including the user's newly posted comment).

In this PR:
* chromedash-activity-log: Add two new properties `narrow` and `reverse` to configure the display as needed for the gate column context.
* chromedash-app: add an auto scrollbar if the sidebar content gets too tall to fit.
* chromedash-gate-column: Add the activity log component and form controls to post.  When the user posts a comment, reset the form and reload the comments.
* Also, I added some `?` operators to handle the case where the user is signed out so that `this.user` is null.
* And, I am now using `font-weight: 500` instead of `font-weight: bold` because 500 looks better in safari.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -29,21 +29,31 @@ class ChromedashGateColumn extends LitElement {
          right: var(--content-padding-quarter);
        }
 
-       #votes-area {
-         margin: var(--content-padding) 0;
-       }
-
-       #votes-area table {
-         border-spacing: var(--content-padding-half) var(--content-padding);
-       }
-
-       #votes-area th {
-         font-weight: bold;
-       }
-
        #review-status-area {
          margin: var(--content-padding-half) 0;
        }
+
+       #votes-area {
+         margin: var(--content-padding) 0;
+       }
+       #votes-area table {
+         border-spacing: var(--content-padding-half) var(--content-padding);
+       }
+       #votes-area th {
+         font-weight: 500;
+       }
+
+        #controls {
+          padding: var(--content-padding);
+          text-align: right;
+          display: flex;
+          justify-content: space-between;
+          align-items: center;
+        }
+        #controls * + * {
+          padding-left: var(--content-padding);
+        }
+
     `];
   }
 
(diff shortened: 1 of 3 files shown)


jrobbins:
Here's how it looks:
image

client-src/elements/chromedash-gate-column.js
```diff
@@ -235,10 +282,40 @@ class ChromedashGateColumn extends LitElement {
     `;
   }
 
+  renderControls() {
+    if (!this.user || !this.user.can_comment) return nothing;
+    // TODO(jrobbins): checkbox to also post to intent thread.
+
+    return html`
+    <sl-textarea id="comment_area" rows=2 cols=40
```
KyleJu, line 290: Nit: indentation

client-src/elements/chromedash-gate-column.js
```diff
@@ -235,10 +282,40 @@ class ChromedashGateColumn extends LitElement {
...
+    `;
+  }
+
+
   renderComments() {
     return html`
       <h2>Comments &amp; Activity</h2>
-      TODO(jrobbins): Comments go here
+      ${this.renderControls()}
+      <div id="comment-area">
```
KyleJu, line 310: comment-area is a different id? how is it used?

client-src/elements/chromedash-gate-column.js
jrobbins: Fixed.

client-src/elements/chromedash-gate-column.js
jrobbins: Removed.  I had added that `<div>` because I thought I might need it do add some padding, but it turned out that I only needed the padding on the `<p>` elements.

KyleJu (approved):

PR #2603, opened 2022-12-22
Implement voting in gate column

jrobbins:
This is progress on #2334.  The gate column will now correctly display existing votes and allow a user who has permission to vote submit a new vote.

In this PR:
* converters.py: Add the gate ID to the gate JSON representation.  And, clarify that requested_on is a datetime, whereas next_action is just a date.  In the unit tests, the gate must be saved to NDB so that it has an ID generated.
* permissions_api.py: Instead of just a boolean that says that the current user can approve anything, pass a list of the specific gate types that the user is allowed to vote on.
* chromedash-gate-column.js:
    * Use lit refs instead of querySelector() for cleaner code.
    * Change the meaning of `this.needsSave` to be about saving votes.  Use `this.needsPost` for the "Post" button.
    * Add a `this.showSaved` property to control when a "Saved" text is shown after the user has saved their vote.
    * When the user selects a value from the vote menu, show a "Save" button which actually saves the new value to the server.
    * Save horizontal space by only showing the username@ rather than the full email address for votes.  The full email address is shown on hover.
    * Where has been no voting yet, and the current user cannot vote, show a "No review activity yet" message rather than an empty table of votes.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -1,4 +1,5 @@
 import {LitElement, css, html, nothing} from 'lit';
+import {ref, createRef} from 'lit/directives/ref.js';
 import './chromedash-activity-log';
 import {showToastMessage, findProcessStage} from './utils.js';
 
@@ -17,7 +18,10 @@ export const STATE_NAMES = [
 ];
 
 
-class ChromedashGateColumn extends LitElement {
+export class ChromedashGateColumn extends LitElement {
+  voteSelectRef = createRef();
+  commentAreaRef = createRef();
+
   static get styles() {
     return [
       ...SHARED_STYLES,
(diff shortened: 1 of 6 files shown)


jrobbins:
Here's what it looks like when the user has selected their vote value but not saved it yet:
image

client-src/elements/chromedash-gate-column.js
```diff
@@ -153,6 +162,21 @@ class ChromedashGateColumn extends LitElement {
...
 
+  handleSelectChanged() {
+    this.needsSave = true;
+  }
+
+  handleSave() {
+    // TODO(jrobbins): This should specify gate ID rather than gate type.
+    window.csClient.setApproval(
+      this.feature.id, this.gate.gate_type,
+      this.voteSelectRef.value.value)
```
KyleJu, line 173: this.voteSelectRef.value.value or this.voteSelectRef.value?

KyleJu (approved):
LGTM with a comment

client-src/elements/chromedash-gate-column.js
jrobbins: It is supposed to be `value.value`, even though it looks odd.  The first `value` is getting the value of the ref, which gives the textarea element.  The second `value` gets the value of that element, which is the text that the user typed.

Issue #2337, opened 2022-10-13
Implement a button to request a review


jrobbins:
**Describe the solution you'd like**
Implement a button for a feature owner to request a review on a particular gate.
When the feature owner presses this button, a Gate entity will be created in the database, and a notification email will be sent to the appropriate approvers (#2340).

All review requirements fields must be filled in with some valid value (even if it is a string like "TBD"), otherwise a preflight check dialog will be shown.

PR #2604, opened 2022-12-23
Display gate status or "Request review" button.

jrobbins:
This is progress toward issues #2334 and #2337.  Basically, an area near the top of the gate column is supposed to show a "Request review" button, and then later display the status of the overall review gate, until it ultimately becomes a checkmark icon and the word "Approved" or an anti icon and the word "Denied".

Also, when the user casts their vote, the app will now update everything on the page to reflect the new state of that gate.

Code in this PR to update page content after a vote:
* chromedash-app: Listen for event `refetch-needed` and then call the `refetch()` method on both the main page content element and the gate column.   Iff that method is implemented.
* chromedash-feature-page: Implement `refetch()` to reload all gate data so that the content in the middle of the page has the up-to-date icons and activity.
* chromedash-feature-page, chromedash-myfeatures-page, and chromedash-feature-table:  Implement `refetch()` to refresh the feature list.

Main code changes to chromedash-gate-column: 
* Represent the vote options the same way that we represent other enums.
* Implement refetch()
* Fix a bug where it was possible for the user to see both the "Save" button and the "Saved" notification at the same time.
* Fire the `refetch-needed` event when the user requests a review or saves a vote.
* Rendering logic and permission checking to display the gate state or "Request review" button.
* The vote menu no longer includes "Review requested" because that will be done via a button, however we still display "Review requested" read-only value for the user who requested the review.  In the case where a reviewer requested the review, we preselect "No response" for their vote.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -6,15 +6,23 @@ import {showToastMessage, findProcessStage} from './utils.js';
 import {SHARED_STYLES} from '../sass/shared-css.js';
 
 
-export const STATE_NAMES = [
-  [7, 'No response'],
-  [1, 'N/a or Ack'],
-  [2, 'Review requested'],
-  [3, 'Review started'],
-  [4, 'Needs work'],
-  [8, 'Internal review'],
-  [5, 'Approved'],
-  [6, 'Denied'],
+export const PREPARING = 0;
+export const REVIEW_REQUESTED = 2;
+export const STATE_NAMES = {
+  NO_RESPONSE: [7, 'No response'],
+  NA: [1, 'N/a or Ack'],
+  REVIEW_STARTED: [3, 'Review started'],
+  NEEDS_WORK: [4, 'Needs work'],
+  INTERNAL_REVIEW: [8, 'Internal review'],
+  APPROVED: [5, 'Approved'],
+  DENIED: [6, 'Denied'],
+};
+
+export const ACTIVE_REVIEW_STATES = [
+  REVIEW_REQUESTED,
+  STATE_NAMES.REVIEW_STARTED[0],
+  STATE_NAMES.NEEDS_WORK[0],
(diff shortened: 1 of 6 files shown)


jrobbins:
Here's how it looks to a feature owner who is about to request a review:
image

Here's how it looks to an anon visitor before the review is started:
image

Here's after the review was requested:
image

Here's after the gate is approved:
image

client-src/elements/chromedash-gate-column.js
```diff
@@ -6,15 +6,23 @@ import {showToastMessage, findProcessStage} from './utils.js';
...
+export const REVIEW_REQUESTED = 2;
+export const STATE_NAMES = {
+  NO_RESPONSE: [7, 'No response'],
+  NA: [1, 'N/a or Ack'],
+  REVIEW_STARTED: [3, 'Review started'],
+  NEEDS_WORK: [4, 'Needs work'],
+  INTERNAL_REVIEW: [8, 'Internal review'],
+  APPROVED: [5, 'Approved'],
+  DENIED: [6, 'Denied'],
+};
```
DanielRyanSmith, line 19: A possible suggestion here: it might be more readable when referencing objects like this to have a nested object with two properties, like `name` and `value`,
e.g.
```js
export const STATE_NAMES = {
  NO_RESPONSE: {
    name: 'No response',
    value: 7,
  },
  NA: {
    name: 'N/a or Ack',
    value: 1,
  },
  ...
};
```

This would allow references to read like `STATE_NAMES.NO_RESPONSE.value` rather than referencing an index.

DanielRyanSmith (approved):
Looks great - big steps forward. 🙂

client-src/elements/chromedash-gate-column.js
jrobbins: Good point.  Let's keep that in mind and maybe redo all our enums sometime.

PR #2611, opened 2022-12-28
create new stages on the feature detail page

DanielRyanSmith:
This change adds a new "Add stage" button to the feature detail page if the user has edit access to the feature, allowing the user to create an additional stage for some specific stage types. Those stage type options are based on the feature type. The button displays a dialog box which allows the user to select the stage type from a dropdown. Once the user has selected a stage type option and selects "Submit", the new stage will be created and the page will be reloaded.

Screenshot 2022-12-28 at 1 05 46 PM
Screenshot 2022-12-28 at 1 07 09 PM
Screenshot 2022-12-28 at 1 06 23 PM

**Note**: This change is part of a broader set of changes that will be required for the work to be complete, namely:
- Enterprise feature rollout stages should show on the feature detail page for non-enterprise feature types.
- The stage fields on the feature detail page should match the values that exist within the stage entity.
- Additional stages that have been created should be able to be archived or soft-deleted.


client-src/elements/chromedash-app.js
```diff
@@ -169,7 +169,7 @@ class ChromedashApp extends LitElement {
       this.pageComponent.contextLink = this.contextLink;
       this.currentPage = ctx.path;
       this.pageComponent.appTitle = this.appTitle;
-      if (this.pageComponent.featureId != this.gateColumnRef.value.feature?.id) {
+      if (this.pageComponent.featureId != this.gateColumnRef.value?.feature?.id) {
```
DanielRyanSmith, line 172: I was occasionally getting some errors when reloading the page until making this change.

DanielRyanSmith:
@jrobbins maybe there needs to be some explanation text in the dialog box about what exactly is being done here

client-src/elements/chromedash-add-stage-dialog.js
```diff
@@ -0,0 +1,123 @@
...
+        style="width:16rem"
+      >
+        <sl-menu-item value="0" disabled>Select a stage to create</sl-menu-item>
+        ${this.renderSelectMenuItems()}
+      </sl-select>
+      <sl-button variant="primary"
+        @click=${this.handleStageCreate}
+        ?disabled=${!this.canSubmit}
+        size="small"
+      >Create Stage</sl-button>
```
jrobbins, line 108: Let's use sentence case: Create Stage -> Create stage.

client-src/elements/chromedash-app.js
jrobbins: Sorry about that.  I have the same fix in my current PR.

client-src/elements/chromedash-add-stage-dialog.js
```diff
@@ -0,0 +1,123 @@
...
+import {CREATEABLE_STAGES, FORMS_BY_STAGE_TYPE} from './form-definition.js';
+
+
+let AddStageDialogEl;
+
+
+export async function openAddStageDialog(featureId, featureType) {
+  if (!AddStageDialogEl) {
+    AddStageDialogEl = document.createElement('chromedash-add-stage-dialog');
+    AddStageDialogEl.featureId = featureId;
```
jrobbins, line 12: This only sets the featureId when the chromedash-add-stage-dialog is first created.   If the user clicks away to close the dialog, then reopens it on a different feature, it would add a stage to the original feature.

client-src/elements/chromedash-feature-detail.js
```diff
@@ -388,6 +390,18 @@ class ChromedashFeatureDetail extends LitElement {
     return this.renderSection(summary, content);
   }
 
+  renderAddStageButton() {
+    if (!this.canEdit) {
+      return nothing;
+    }
+
+    return html`
+    <sl-button variant="primary" size="small" @click="${
```
jrobbins, line 399: Let's not make it primary because adding a stage is an infrequent operation.

jrobbins (changes requested):
Overall this looks very good and straightforward.  The dialog re-initialization is the main hold-up.

client-src/elements/chromedash-feature-detail.js
```diff
@@ -29,6 +30,7 @@ class ChromedashFeatureDetail extends LitElement {
     this.canEdit = false;
     this.feature = {};
     this.gates = [];
+    this.isAddingStage = false;
```
jrobbins, line 33: I don't see where this is used.

jrobbins (changes requested):

client-src/elements/chromedash-add-stage-dialog.js
DanielRyanSmith: Good catch. I added a check that saves the `featureId` and hits this block of code if a new feature ID is passed in.

client-src/elements/chromedash-add-stage-dialog.js
DanielRyanSmith: Fixed!

client-src/elements/chromedash-feature-detail.js
DanielRyanSmith: Changed 👍

client-src/elements/chromedash-feature-detail.js
DanielRyanSmith: This is a ghost from a previous implementation attempt that I missed removing. Thanks for catching!

jrobbins (approved):

PR #2612, opened 2022-12-29
Display and edit stage-specific data for multiple stages of the same type

DanielRyanSmith:
This is a follow-up PR for #2611. This can be more easily reviewed once that PR is reviewed/merged.

This change make it so that stages on the feature detail page, as well as the edit-all and stage edit page, populate data accurately for all stages, even multiple stages of the same stage type.

description of changes by file:
- api/converters.py: Update the `stage_to_json_dict` function to more accurately give all information contained in a stage entity, as well as change the `stages` property on `feature_entry_to_json_verbose` to reference a full list of those stage info objects, sorted by stage type.
- api/stages_api.py: Remove duplicated `stage_to_json_dict` function and use the function in converters.py.
- client-src/elements/chromedash-feature-detail.js: Add logic to display the field values from the stage info objects rather than the feature object if the fields are stage-specific.
- client-src/elements/chromedash-guide-editall-page.js: Add logic to display the field values from the stage info objects rather than the feature object if the fields are stage-specific. Additionally, add a name modifier to differentiate multiple stages of the same stage type.
- client-src/elements/chromedash-process-overview.js: Add a name modifier to differentiate multiple stages of the same stage type. Additionally, refactor button display logic to have a default "Edit" selection, which will display if an active stage is of the same stage type as the stage to be rendered.
- internals/core_models.py: Add lists of renamed milestone fields to MilestoneSet, separated by stage types. (These values can likely be used in other parts of the code in order to remove similar references, namely guide.py).
- main.py: Add a new endpoint for creating a stage.

Screenshot 2022-12-28 at 4 34 13 PM
Two of the same stage type can exist and are displayed with separate data based on the entity.
Screenshot 2022-12-28 at 4 34 49 PM
All stages are visible including those of the same stage type, with a number differentiation. A stage of the same type can also be set as the active stage without causing UI errors or marking the same stage types as also active.


api/converters.py
```diff
@@ -236,7 +236,9 @@ def _prep_stage_gate_info(
   ship_type = STAGE_TYPES_SHIPPING[fe.feature_type]
   rollout_type = STAGE_TYPES_ROLLOUT[fe.feature_type]
 
-  stages = Stage.query(Stage.feature_id == d['id'])
+  # Get all stages associated with the feature, sorted by stage type.
+  stages = Stage.query(Stage.feature_id == d['id']).order(Stage.stage_type)
```
jrobbins, line 240: This may require a new index.

api/converters.py
```diff
@@ -267,42 +270,63 @@ def _prep_stage_gate_info(
...
-  'ios_last',
-  'webview_first',
-  'webview_last'
-]
-
-def stage_to_dict(stage: Stage) -> dict[str, Any]:
+
+INTENT_THREAD_FIELD_NAMES = {
+  'ot': 'intent_to_experiment_url',
+  'ship': 'intent_to_shi'
```
jrobbins, line 276: Typo: intent_to_shi -> intent_to_ship_url.

api/stages_api.py
```diff
@@ -69,31 +71,6 @@ class StagesAPI(basehandlers.APIHandler):
       'rollout_platforms',
       'enterprise_policies']
 
-  def _stage_to_json_dict(self, stage: Stage) -> dict:
```
jrobbins, line 72: Thanks for doing this.  I recently noticed that there was redundant code in these functions.

client-src/elements/chromedash-add-stage-dialog.js
```diff
@@ -0,0 +1,124 @@
+import {LitElement, css, html} from 'lit';
```
jrobbins, line 1: This looks like a change from your other PR.

jrobbins (changes requested):
This is good progress.  I have a few initial comments, and then I will review it again after you rebase and update this PR.

api/converters.py
DanielRyanSmith: What is the best way to determining this? To push it to staging and see if it complains about indexes? I'm not certain the protocol 😅

jrobbins: Yeah, running on staging is probably the best way.   Based on the query added here, I could guess the needed index, but it is better to have NDB tell us.

jrobbins (approved):

PR #2613, opened 2022-12-29
Offer API Owners a checkbox to post to the intent thread.

jrobbins:
This is progress on issue #2334.  The old approvals dialog had a sl-select menu to allow the API owners to choose which intent thread to post their comment to, or to just post in chromestatus.   For the new gate column, exactly one gate is ever shown at a time, so we just need a checkbox.  It is shown for the "API Owners" gates, and it is enabled iff the thread URL has an email message ID that our server will be able to reply to (so that the message appears in the correct thread).

In this PR:
* Fix a bug in chrome status-app that caused an exception on first SPA load.
* setContext() now takes a stageId rather than a full Stage JS object because the stage parts of the feature don't have all the fields of the Stage model.  Now the gate column fetches the full Stage using the stages API.
* Logic to include the gate_type when posting a comment with the checkbox checked.
* Logic to render the checkbox.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -29,6 +29,7 @@ export const ACTIVE_REVIEW_STATES = [
 export class ChromedashGateColumn extends LitElement {
   voteSelectRef = createRef();
   commentAreaRef = createRef();
+  postToThreadRef = createRef();
 
   static get styles() {
     return [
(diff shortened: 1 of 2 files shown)


jrobbins:
Here's what it looks like for a API Owners gate that has a valid intent thread specified:
image

(I took that screenshot before adding the hyperlink for "intent thread").

DanielRyanSmith:
> setContext() now takes a stageId rather than a full Stage JS object because the stage parts of the feature don't have all the fields of the Stage model. Now the gate column fetches the full Stage using the stages API.

This is something I'm changing in the other PRs that are open. I can refactor this to accommodate depending on what we end up merging first.

client-src/elements/chromedash-gate-column.js
```diff
@@ -476,9 +478,46 @@ export class ChromedashGateColumn extends LitElement {
     `;
   }
 
+  gateHasIntentThread() {
+    return this.gate.team_name == 'API Owners';
```
DanielRyanSmith, line 482:
```suggestion
    return this.gate.team_name === 'API Owners';
```

DanielRyanSmith (approved):

PR #2619, opened 2023-01-03
Edit metadata from the feature detail page

DanielRyanSmith:
This change adds a new SPA route to edit the metadata fields associated with a feature, as well as adding a button link to the new route on the feature detail page. The metadata edit page works identically to the single stage edit page for the most part, but uses a new route due to not being associated with a specific stage.


client-src/elements/chromedash-guide-metadata-page.js
```diff
@@ -0,0 +1,213 @@
+import {LitElement, css, html} from 'lit';
```
DanielRyanSmith, line 1: This file is very close to `chromedash-guide-stage-page.js`, but does not query for stage data and attempt to use Stage entity information. I had considered trying to make that page work with both metadata and stage information, but it seemed that the logic would be unnecessarily complex when trying to accommodate both use cases, and this felt more easy to come back and refactor/edit.

client-src/components.js
```diff
@@ -31,10 +31,10 @@ setBasePath('/static/shoelace');
 // Configure shoelace to also find material design 24pt outline icons
 // like: <sl-icon-button library="material" name="unfold-more">
 // See developer-documentation.md for instructions adding icons.
-import { registerIconLibrary } from '@shoelace-style/shoelace/dist/utilities/icon-library.js';
+import {registerIconLibrary} from '@shoelace-style/shoelace/dist/utilities/icon-library.js';
 registerIconLibrary('material', {
-  resolver: name =>  `/static/shoelace/assets/material-icons/${name}.svg`,
-  mutator: svg => svg.setAttribute('fill', 'currentColor')
+  resolver: name => `/static/shoelace/assets/material-icons/${name}.svg`,
+  mutator: svg => svg.setAttribute('fill', 'currentColor'),
```
DanielRyanSmith, line 37: small formatter changes

client-src/elements/chromedash-guide-metadata-page.js
jrobbins: That's fine, we can look for ways to refactor commonalities later.

jrobbins (approved):

PR #2626, opened 2023-01-05
Deprecate LGTM fields

DanielRyanSmith:
This change disables the existing lgtm fields and makes them read-only.

### DIFF
--- a/client-src/elements/form-field-specs.js
+++ b/client-src/elements/form-field-specs.js
@@ -897,7 +897,7 @@ export const ALL_FIELDS = {
 
   'i2e_lgtms': {
     type: 'input',
-    attrs: MULTI_EMAIL_FIELD_ATTRS,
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
@jrobbins this is a possibility to handle these fields, but I know you have expressed the prospect of going an alternate path. I am receptive to any approach, but currently these fields exist in the UI but cannot be edited or displayed due to the field not existing on FeatureEntry entities. Let me know if you have opinions on this.

jrobbins:
I think this is a good change, but too soon.  We could probably merge this near the end of Q1.

jrobbins:
Looking at this again, I think I was too conservative when I said that we should keep these fields for now.
The launching-features page does not instruct users to fill in these fields, and the API Owner votes are now visible on (a dialog box of) the feature details page.

So, I think that we could go ahead and land this.

jrobbins (approved):

PR #2627, opened 2023-01-06
Make read-only field display driven by feStage and fix missing fields

jrobbins:
This fixes a couple of different problems with the display of fields on the feature detail page:
* Some fields were being listed missing when those fields were not relevant to the feature's feature type.
* Some fields were never displaying a value, even if one was filled in.
* Some fields were labeled with a variation of the source code name of the field rather than an end-user readable name.

The old approach of using DISPLAY_FIELDS_IN_STAGES was based on the assumption that we would skip fields that had no value.  So, DISPLAY_FIELDS_IN_STAGES listed all fields for a given process stage, regardless of feature type.  This approach cannot work with the new UX where we display all fields, even those that lack values.  Instead, we will now use FORMS_BY_STAGE_TYPE, which is keyed by stage type, and the set of stages in a feature are relevant to its feature type.

In this PR:
* Delete `DISPLAY_FIELDS_IN_STAGES`, and use `FORMS_BY_STAGE_TYPE` in a way very similar to the way that the edit-all page decides which fields to show.
* Add / fix some items in `fieldNameMapping`.
* add `displayLabel` to the field specs that use `''` for their editing label.  On the editing forms, these fields don't have labels and are positioned closed to other fields that have useful labels that provide context to the user.  For the read-only display, for now I decided to use a more consistent pattern where each field is independent and can have its own validation checkmark.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -1,12 +1,16 @@
 import {LitElement, css, html, nothing} from 'lit';
 import {openAddStageDialog} from './chromedash-add-stage-dialog';
-import {DISPLAY_FIELDS_IN_STAGES} from './form-field-specs';
+import {makeDisplaySpec} from './form-field-specs';
+import {
+  FLAT_METADATA_FIELDS,
+  FORMS_BY_STAGE_TYPE} from './form-definition';
+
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Thanks for taking care of this - It seems like a better approach 🙂

PR #2628, opened 2023-01-06
Fix missing gate chips.

jrobbins:
A recent PR must have changed the JSON representation of stages from using `stage_id` to just plain `id`.  This PR just updates places where the old `stage_id` was referenced.


DanielRyanSmith (approved):
Ah, sorry about this. I guess I didn't manage to update all instances where this is referenced. Definitely my fault!

Issue #2639, opened 2023-01-10
Record and display comments on a per-gate basis


jrobbins:
**Is your feature request related to a problem? Please describe.**
Our old approvals dialog showed all approval gates and all review comments in one dialog.  But, our new gate column is supposed to show only one gate at a time, along with the comments that relate specifically to that one gate.

**Describe the solution you'd like**
Two parts:
* [x] When a reviewer posts a comment using the gate column, it calls chromedash-gate-column.js `handlePost()`, which calls `postComment()` with null for the gate ID.   Instead, it should pass the actual gate ID.  On the server-side, comments_api.py `do_post()` should use the passed in `gate_id` URL param instead of the existing code which gets the `gate_id` by querying using `gate_type`.   I don't believe that `gate_type` was ever actually passed, so some of the existing code is unreachable and can be removed.
* [x] When displaying comments in chromedash-gate-column.js, we should display comments that have the gate_id set to the current gate, plus legacy comments that have no gate_id set, plus activities that have the current gate_id set.  This should mainly be done on the server-side.  chromedash-gate-column.js should call `getComments()` specifying the gate_id, and the server should decide which Activities to include in the response.  Sorting should still be chronological.

**Additional context**
After both this issue and the issue about creating an activity for a vote are done, we should see each vote generate an activity that appears in the gate column only for that gate.

jrobbins:
>  I don't believe that gate_type was ever actually passed, so some of the existing code is unreachable and can be removed.

Correction: `gate_type` was passed and is still needed.   If you see `field_id` in old code, that translates into `gate_type` now.

PR #2644, opened 2023-01-10
Display correct comments for gate

KyleJu:
#2639. Display comments on a per-gate basis. Remove incorrect use of gate_type on the server side.

- Change field_id to gate_id in the comment APIs
- Retrieve comments by querying gate_id on the server side.
- Display comments in activities that have the gate_id set to the current gate, plus legacy comments that have no gate_id set.

### Follow-up PRs
Fix do_post() in Comment APIs
Test on staging after fixing Post

### DIFF
--- a/internals/review_models.py
+++ b/internals/review_models.py
@@ -334,7 +334,8 @@ class Activity(ndb.Model):  # copy from Comment
     query = Activity.query().order(Activity.created)
     query = query.filter(Activity.feature_id == feature_id)
     if gate_id:
-      query = query.filter(Activity.gate_id == gate_id)
+      # Fetch activities with a gate_id or None value.
+      query = query.filter(Activity.gate_id.IN([gate_id, None]))
     acts = query.fetch(None)
     if comments_only:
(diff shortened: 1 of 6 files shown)


internals/review_models.py
```diff
@@ -334,7 +334,8 @@ def get_activities(cls, feature_id: int, gate_id: Optional[int]=None,
     query = Activity.query().order(Activity.created)
     query = query.filter(Activity.feature_id == feature_id)
     if gate_id:
-      query = query.filter(Activity.gate_id == gate_id)
+      # Fetch activities with a gate_id or None value.
+      query = query.filter(Activity.gate_id.IN([gate_id, None]))
```
jrobbins, line 338: Cool.  I didn't know that we could use `None` with the `IN` operator.

jrobbins (approved):

internals/review_models.py
KyleJu: I didn't know either. I think ndb set the non-existing field value to None so I tried the None value. It appears that it worked based on the tests

Issue #1915, opened 2022-06-06
Features list loads slowly and blocks rendering


jpmedley:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
The features.json file loads slowly, sometimes taking as long as 30 seconds or more.

**To Reproduce**
1. Open Chrome Status.
2. Reload the page.

If it happens to no load slowly:
1. Open Chrome Status.
2. Open DevTools.
3. Click "Empty Cache and Hard Reload."

**Expected behavior**
I expect its load time to be comparable to resources on the same page, with the same size.

**Additional context**
<img width="963" alt="Screen Shot 2022-06-06 at 11 10 40 AM" src="https://user-images.githubusercontent.com/8204171/172190429-0776394c-1945-48c6-96b7-618018a1d73a.png">

jrobbins:
The existing feature list has several problems, including slow performance because of requesting features.json which is a huge file.   The long-term replacement is the new feature page which does server-side searches and returns paginated results.  That page is still a work in progress.

jyasskin:
The new page is currently slower than the old page, with a 7s delay on one server-side query:
![Chrome devtools network timing output, showing 7s of "Waiting for server response" on a request for features?q=first&num=100](https://user-images.githubusercontent.com/83420/211384896-f5e71f16-2503-4d1a-a59e-c34305ab70c7.png)

jrobbins:
Yeah, something got slower.   I have an idea of what the slow part might be.

PR #2651, opened 2023-01-10
Speed up search results that don't need stages.

jrobbins:
This should address the speed concerns about the new feature list page expressed in some comments on #1915.

In this PR:
* Use `converters.feature_entry_to_json_basic()` because the new feature list does not need stage data.
* Cache using a key that includes the word 'basic' so that we don't overwrite any cache entry that does include stage data.
* Load and store to redis using batch operations rather than individual operations in a loop.
* Make `get_by_ids()` keyword parameter `update_cache`  default to `True`, and only set cache values if that value is true.

Testing with a tainted version on prod I see a pretty good speed improvement:
* Old code: 7s on cache miss, 2s on cache hit
* New code: 2s on cache miss, 1.5s on cache hit
* I tested by just doing the default query on /newfeatures and then paginating forward and back

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -538,7 +538,7 @@ def get_all(limit: Optional[int]=None,
 
 
 def get_by_ids(feature_ids: list[int],
-    update_cache: bool=False) -> list[dict[str, Any]]:
+               update_cache: bool=True) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Thanks for taking care of this 🙂  I forgot to take a look personally, but it looks the issue was related partly to what we expected.

PR #2653, opened 2023-01-11
Clean up do_post in Comment APIs

KyleJu:
Fix https://github.com/GoogleChrome/chromium-dashboard/issues/2639.
- Remove the set_approval and set_vote in comment POST API
- Remove new_state from the API
- Rename field_id to gate_type
- Rename gate_type to gate_id in the API params
- Pass in gate_id when post comments

### TODO 
Staging tests

### DIFF
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -82,7 +82,7 @@ class CommentsAPI(basehandlers.APIHandler):
     feature = self.get_specified_feature(feature_id=feature_id)
     user = self.get_current_user(required=True)
     gate_type = self.get_param(
-        'gate_type', required=False)
+        'gateType', required=False)
 
     comment_content = self.get_param('comment', required=False)
     if comment_content:
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -81,8 +81,8 @@ class CommentsAPI(basehandlers.APIHandler):
     gate_id = kwargs.get('gate_id', None)
     feature = self.get_specified_feature(feature_id=feature_id)
     user = self.get_current_user(required=True)
-    gate_type = self.get_param(
-        'gateType', required=False)
+    post_to_thread_type = self.get_param(
+        'postToThreadType', required=False)
 
     comment_content = self.get_param('comment', required=False)
     if comment_content:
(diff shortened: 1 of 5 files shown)


jrobbins:
I apologize for giving incorrect info in the issue.  Looking at the code in approvals_defs and review_models, it seems that gate_type is still needed.  If you see `field_id`, that is the same as `gate_type`, not a database ID.

Also, parameter `state` is never non-None, so some lines of code could be deleted from doPost().

jrobbins (changes requested):
See comment in PR discussion.

api/comments_api.py
```diff
@@ -78,51 +78,24 @@ def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
...
-      gate_id = None
-      if len(gates) > 0:
-        gate_id = gates[0].key.integer_id()
-      # Schema migration double-write.
       comment_activity = Activity(feature_id=feature_id, gate_id=gate_id,
           author=user.email(), content=comment_content)
       comment_activity.put()
 
-    if post_to_approval_field_id:
+    if gate_type:
```
jrobbins, line 96: It is true that this is a gate_type, but there is also a boolean mixed in there: This parameter is passed iff the user wants to post to the email thread.  So, maybe name it `post_to_thread_type` for now.  

Later, I think we could reduce this to be just a boolean `post_to_thread` that would know which thread based on `gate_id`, but we can't break the approvals dialog yet.

client-src/elements/chromedash-approvals-dialog.js
```diff
@@ -521,12 +521,12 @@ class ChromedashApprovalsDialog extends LitElement {
     const commentText = commentArea.value.trim();
     const postToSelect = this.shadowRoot.querySelector(
       '#post_to_approval_field');
-    const postToApprovalFieldId = postToSelect && postToSelect.value || 0;
+    const gateType = postToSelect && postToSelect.value || 0;
```
jrobbins, line 524: Let's leave something about "posting to a thread" in this variable name.   E.g., `postToThreadType`.

client-src/js-src/cs-client.js
```diff
@@ -250,15 +250,15 @@ class ChromeStatusClient {
     return this.doGet(url);
   }
 
-  postComment(featureId, gateId, state, comment, postToApprovalFieldId) {
+  postComment(featureId, gateId, comment, gateType) {
```
jrobbins, line 253: Same here, keep it something like `postToThreadType`.

jrobbins (changes requested):
Looks good, I just think we need one name change in a few places.

jrobbins (approved):

PR #2662, opened 2023-01-12
Vote API drive-by fix

KyleJu:
A follow-up for #2653. Call vote API after posting a new comment.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -206,12 +206,12 @@ export class ChromedashGateColumn extends LitElement {
   handlePost() {
     const commentArea = this.commentAreaRef.value;
     const commentText = commentArea.value.trim();
-    const postToApprovalFieldId = (
+    const postToThreadType = (
         this.postToThreadRef.value?.checked ? this.gate.gate_type : 0);
     if (commentText != '') {
       window.csClient.postComment(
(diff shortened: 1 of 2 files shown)


KyleJu:
Need to git main merge after https://github.com/GoogleChrome/chromium-dashboard/pull/2653 is merged

client-src/elements/chromedash-gate-column.js
```diff
@@ -209,10 +209,13 @@ export class ChromedashGateColumn extends LitElement {
...
         this.postToThreadRef.value?.checked ? this.gate.gate_type : 0);
     if (commentText != '') {
-      window.csClient.postComment(
-        this.feature.id, this.gate.id, commentText,
-        Number(postToThreadType))
-        .then(() => this.reloadComments());
+      Promise.all([
+        window.csClient.postComment(
+          this.feature.id, this.gate.id, commentText, Number(postToThreadType)),
+        window.csClient.setVote(featureId, this.gate.id, null),
```
jrobbins, line 215: There is actually a separate "Save" button that appears right next to the vote select widget.  So, we don't want the "Post" button to save the vote.

jrobbins (changes requested):
The changes to reviews_api.py look good, but I don't think that we need the gate-column changes.

jrobbins (approved):

Issue #2637, opened 2023-01-09
"Edit" link from guide links below form label.


jyasskin:
**Describe the bug**
Clicking an "edit" link in the guide should scroll to the target field's label, not just the `<input>` element.

**To Reproduce**
1. Go to https://chromestatus.com/guide/edit/5745285984681984
2. Hover over "Request signals" (or most other items in the "Progress" column)
3. Click on the "Edit" link that appears.
4. Note that you don't see the title of the target field:

![the Edit Feature page, scrolled to an unlabeled select box](https://user-images.githubusercontent.com/83420/211425971-2743e83e-131a-4a46-a231-bf965ff4159d.png)

**Expected behavior**

![the Edit Feature page, scrolled to a labeled "Safari views" select box](https://user-images.githubusercontent.com/83420/211426021-50c0c40f-17cb-4222-b979-f5bbb2abad2f.png)

PR #2671, opened 2023-01-13
Fixes "Edit" link from guide links below form label.

Spectre-ak:
Fixes "Edit" link from guide links below form label #2637 

Updated `scrollToPosition` of `chromedash-guide-stage-page`.

Currently `scrollIntoView` is being called on `sl-select` fetched using id and its scrolling to select element
image

Proposed changes
To call `scrollIntoView`  on the heading element (`<b>Safari views:</b>`)
image

cc: @jrobbins

### DIFF
--- a/client-src/elements/chromedash-guide-stage-page.js
+++ b/client-src/elements/chromedash-guide-stage-page.js
@@ -162,7 +162,9 @@ export class ChromedashGuideStagePage extends LitElement {
       const hash = decodeURIComponent(location.hash);
       if (hash) {
         const el = this.shadowRoot.querySelector(hash);
-        el.scrollIntoView(true, {behavior: 'smooth'});
+        if (el) {
+          this.shadowRoot.querySelector(`chromedash-form-field[name="${el.name}"] tr th b`).scrollIntoView(true, {behavior: 'smooth'});
+        }
       }
     }
   }


client-src/elements/chromedash-guide-stage-page.js
```diff
@@ -162,7 +162,9 @@ export class ChromedashGuideStagePage extends LitElement {
       const hash = decodeURIComponent(location.hash);
       if (hash) {
         const el = this.shadowRoot.querySelector(hash);
-        el.scrollIntoView(true, {behavior: 'smooth'});
+        if (el) {
+          this.shadowRoot.querySelector(`chromedash-form-field[name="${el.name}"] tr th b`).scrollIntoView(true, {behavior: 'smooth'});
```
jrobbins, line 166: Do we need the `tr th b`?  What if it just scrolled into view the custom chromedash-form-field element that has name="..."?

Spectre-ak: ya i tried that, `this.shadowRoot.querySelector(chromedash-form-field[name="${el.name}"]).scrollIntoView(true, {behavior: 'smooth'})`, but a gap is coming on top of the label
Sample images

image
image

jrobbins (approved):
Thanks for the explanation.  This fix is certainly helpful to users.

PR #2676, opened 2023-01-18
Highlight selected gates

### LINKED ISSUE #2640: Highlight the currently selected gate chip
When the user (shift-)clicks on a gate chip, that opens the gate column to display information about that gate.  However, if the user then gets up from their desk or switches to a different browser tab for a while, when they come back to look again it is not visually clear which gate is being shown in the gate column.  The gate column textually shows the stage and gate name, but not the feature name.  That is enough textual information to disambiguate which gate chip was selected on the feature detail page, but it is not enough textual information for the feature list pages.  So, it would be good to visually highlight the gate chip that was used.

* In chromestatus-app, add a property for the most recently selected gate and set its value in `showGateColumn()`.
* In chromestatus-app, whenever the most recently selected gate is changed, or when the user navigates to a new path, pass the gate property to `this.pageComponent`.  We are not using lit's normal property passing here because we are programmatically constructing the element.
* In chromedash-gate-chip, define a property to hold the most recently selected gate passed in from the app.
* In chromedash-gate-chip, if `this.gate` == the one passed in from the app, then add a border that is 2px solid var(--dark-spot-color)

KyleJu:
Fix #2640. Highlight the previously selected gate chip from both feature detail page and my features page. The previously selected chip will have a border highlight when user navigates to a different page.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -102,7 +102,7 @@ class ChromedashApp extends LitElement {
       pageComponent: {attribute: false},
       contextLink: {type: String}, // used for the back button in the feature page
       sidebarHidden: {type: Boolean},
-      selectedGate: {type: Number},
+      selectedGateId: {type: Number},
     };
   }
 
@@ -118,7 +118,7 @@ class ChromedashApp extends LitElement {
     this.pageComponent = null;
     this.contextLink = '/features';
     this.sidebarHidden = true;
-    this.selectedGate = 0;
+    this.selectedGateId = 0;
   }
 
   connectedCallback() {
@@ -150,7 +150,7 @@ class ChromedashApp extends LitElement {
     page('/myfeatures', (ctx) => {
       this.pageComponent = document.createElement('chromedash-myfeatures-page');
       this.pageComponent.user = this.user;
-      this.pageComponent.selectedGate = this.selectedGate;
+      this.pageComponent.selectedGateId = this.selectedGateId;
       this.contextLink = ctx.path;
       this.currentPage = ctx.path;
       this.hideSidebar();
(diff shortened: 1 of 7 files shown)


client-src/elements/chromedash-gate-chip.js
```diff
@@ -144,9 +146,9 @@ class ChromedashGateChip extends LitElement {
     const stateName = GATE_STATE_TO_NAME[this.gate.state];
     const className = stateName.toLowerCase().replaceAll(' ', '_');
     const iconName = GATE_STATE_TO_ICON[this.gate.state];
-
     return html`
       <sl-button pill size="small" class=${className}
+        style="border:${this.gate.id == this.selectedGate ? '2px solid var(--dark-spot-color)' : nothing}"
```
jrobbins, line 151: For non-selected gates, I think this would produce `style="border:"`, which is invalid CSS.  It probably gets harmlessly ignored, but it looks like an error.  I would prefer if you used a CSS class or Lit's styleMap or classMap or make the entire style attribute value be in the true part of the condition so that yields `style=""` or avoid generating the style attribute at all in that case.

client-src/elements/chromedash-app.js
```diff
@@ -102,6 +102,7 @@ class ChromedashApp extends LitElement {
       pageComponent: {attribute: false},
       contextLink: {type: String}, // used for the back button in the feature page
       sidebarHidden: {type: Boolean},
+      selectedGate: {type: Number},
```
jrobbins, line 105: Since it is the ID rather than a gate JSON object, I think this name would be more clear as `selectedGateId`.

jrobbins (changes requested):

client-src/elements/chromedash-gate-chip.js
```diff
@@ -144,9 +148,13 @@ class ChromedashGateChip extends LitElement {
     const stateName = GATE_STATE_TO_NAME[this.gate.state];
     const className = stateName.toLowerCase().replaceAll(' ', '_');
     const iconName = GATE_STATE_TO_ICON[this.gate.state];
-
+    let selectedBorder;
```
jrobbins, line 151: To avoid the lint error: maybe you could initialize it to an empty string, then you will not need ifDefined().

jrobbins (approved):
This change looks good to me, after you get past the lint error.

Issue #2336, opened 2022-10-13
Hard-coded gate questionnaires


jrobbins:
**Describe the solution you'd like**
Each `Gate` should have block of read-only "questionnaire" displayed on it.
At first, the value of that field can be populated by a constant in approval_defs.py based on the gate_type.

**Additional context**
Later, this will be made data-driven and we will implement an admin page to edit questionnaire text to be used when new reviews are requested.

PR #2677, opened 2023-01-18
Add simple gate questionnaires and stage action buttons

jrobbins:
This is progress on #2336.

Approving teams have various criteria for how they grant approvals.  E.g., most teams will require specific fields to be filled in.  Whereas, other teams might have additional requirements such as sending an email or sending a PR.  We handle that by displaying a short block of text authored by the approving team, and by allowing process stages to define action URLs that are shown as buttons instead of the "Request review" button.

In this PR:
* chromedash-gate-column.js: 
    * Add a render method for action URLs that shows them as link buttons instead of the "Request review" button.
    * Add render methods to display the "questionnaire" text, if there is any defined for the current gate type.  It is styled and followed by a brief instruction to respond via comments.
* Also in this PR: 
    * Add text to explain who can view comments.
    * Improve the JS constants for gate types


client-src/elements/chromedash-gate-column.js
```diff
@@ -302,18 +314,37 @@ export class ChromedashGateColumn extends LitElement {
...
-    } else {
+    if (!this.userCanRequestReview()) {
       return html`
         Review has not been requested yet.
       `;
     }
+
+    const processStage = findProcessStage(this.stage, this.process);
+    if (processStage?.actions?.length > 0) {
+      return processStage.actions.map(act =>
```
KyleJu, line 339: Enclose this part with a <div> </div>? html` <div> ${processStage.actions.map(...)}</div>`

KyleJu: Also specify the layout here when there are multiple processStage.action

client-src/elements/chromedash-gate-column.js
```diff
@@ -462,12 +493,24 @@ export class ChromedashGateColumn extends LitElement {
...
+    `;
   }
 
   renderQuestionnaire() {
-    // TODO(jrobbins): Implement questionnaires later.
-    return nothing;
+    const questionnaireText = GATE_QUESTIONNAIRES[this.gate.gate_type];
+    const instructions = (
+      questionnaireText ?
+        html`Please post responses in the comments below.` :
```
KyleJu, line 507:
```suggestion
        Please post responses in the comments below. :
```
Does it need html tags?

client-src/elements/chromedash-gate-column.js
```diff
@@ -462,12 +493,24 @@ export class ChromedashGateColumn extends LitElement {
...
+        html`&nbsp`);
```
KyleJu, line 508:
```suggestion
        "&nbsp";
```
Same here?

client-src/elements/chromedash-gate-column.js
```diff
@@ -525,6 +568,10 @@ export class ChromedashGateColumn extends LitElement {
         @keypress=${this.checkNeedsPost}
         placeholder="Add a comment"
         ></sl-textarea>
+       <div class="instructions">
+         Comments will be visible publicly.
+         Only reviewers will be notified.
```
KyleJu, line 573:
```suggestion
         Only reviewers will be notified when a comment is posted.
```
Maybe clarify?

client-src/elements/form-field-enums.js
```diff
@@ -133,11 +133,18 @@ export const DEPRECATED_FIELDS = new Set([
   'standardization',
 ]);
 
+export const GATE_TYPES = {
```
KyleJu, line 136: Are these searchable fields for the chromedash-feature-filter? if so, should we add corresponding fields to the backend as we..?

jrobbins: No, I don't think we need to search on gate type.

client-src/elements/chromedash-gate-column.js
jrobbins: So far we only ever have zero or one action.  And, I think that regular flow layout works fine for these buttons.

client-src/elements/chromedash-gate-column.js
jrobbins: I think that the `&nbsp;` needs it, so I used it for the other one to make the two conditional branches consistent.

client-src/elements/chromedash-gate-column.js
jrobbins: I think it needs it, otherwise the `&` will get escaped.

jrobbins: But, you do make me realize that it needs a semicolon.

KyleJu (approved):

PR #2678, opened 2023-01-18
Fix sidebar width on mobile.

jrobbins:
On mobile in vertical orientation, the gate column width could be wider than the device screen, which is very awkward.   So, I set a max-width based on the the view width.


jrobbins:
Here's how it looks before this change on a simulated iPhone SE (which is only 375 CSS pixels wide)
image

And here is how it looks with this change:
image

KyleJu (approved):

PR #2680, opened 2023-01-18
Show stage short name in front of row of gate chips.

jrobbins:
This is a tweak to your previous change to show the gate names for issue #2519.    I think I got mixed up between stages and gates when I was explaining it to you verbally.  The gate chips are grouped by stage, we just need to display the stage name as a label for each row of gate chips.   Gates also have a name, but we are currently showing the team name for each gate.

Visually, the app looks the same before and after this PR.  The difference will come when we start defining more gates for other approving teams.  Those gate chips will be compactly displayed on a single row for each stage.

In this PR:
* Redo the short name constants because we need stage names
* Remove a layer of <div>s because we want multiple gate chips to display on a single row
* Add logic to consider OT extension stages, however there is no way for a user to request an extension review via the UI yet, so this logic does not render anything yet.


DanielRyanSmith (approved):
Looks great!

PR #2681, opened 2023-01-19
Phase out ApprovalsAPI Get

KyleJu:
A follow-up for #2586, phasing out the ApprovalsAPI GET method on the frontend.

- **gate_type** in do_get of ApprovalsAPI is always None because it is not required in the API (see deleted **getApprovals(featureId)**). Make this **gate_type** None in the VotesAPI
- Add an optional parameter **gate_id** for VotesAPI
- Phase out all ApprovalsAPI GET methods

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -34,11 +34,7 @@ class VotesAPI(basehandlers.APIHandler):
     feature_id = kwargs['feature_id']
     gate_id = kwargs.get('gate_id', None)
     # Note: We assume that anyone may view approvals.
-    if gate_id is not None:
-      votes = Vote.get_votes(feature_id=feature_id, gate_id=gate_id)
-    else:
-      votes = Vote.get_votes(feature_id=feature_id)
-
+    votes = Vote.get_votes(feature_id=feature_id, gate_id=gate_id)
     dicts = [converters.vote_value_to_json_dict(v) for v in votes]
     return {'votes': dicts}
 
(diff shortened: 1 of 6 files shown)


api/reviews_api.py
```diff
@@ -32,9 +32,13 @@ class VotesAPI(basehandlers.APIHandler):
   def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
     """Return a list of all vote values for a given feature."""
     feature_id = kwargs['feature_id']
-    gate_id = kwargs['gate_id']
+    gate_id = kwargs.get('gate_id', None)
     # Note: We assume that anyone may view approvals.
-    votes = Vote.get_votes(feature_id=feature_id, gate_id=gate_id)
+    if gate_id is not None:
```
jrobbins, line 37: I don't think that you need this if-statement because if gate_id is None, that is the same as not passing it.

client-src/js-src/cs-client.js
```diff
@@ -200,20 +200,18 @@ class ChromeStatusClient {
...
-  }
-
   setApproval(featureId, gateType, state) {
     return this.doPost(
         `/features/${featureId}/approvals`,
         {gateType: Number(gateType), state: Number(state)});
   }
 
   getVotes(featureId, gateId) {
+  if (gateId) {
```
jrobbins, line 210: It looks like you need to indent this if-statement so that it is inside the function.

jrobbins (approved):

PR #2682, opened 2023-01-20
Fix selected gate display.

jrobbins:
Set the selectedGateId on the selected page component because it is not passed via a `${value}` in a render method.

Also, do the border via the CSS part rather than the overall sl-button element.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -279,11 +279,14 @@ class ChromedashApp extends LitElement {
 
   hideSidebar() {
     this.sidebarHidden = true;
+    this.selectedGateId = 0;
+    this.pageComponent.selectedGateId = 0;
   }
 
   showGateColumn(feature, stageId, gate) {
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-app.js
```diff
@@ -279,11 +279,14 @@ class ChromedashApp extends LitElement {
 
   hideSidebar() {
     this.sidebarHidden = true;
+    this.selectedGateId = 0;
```
KyleJu, line 282: If we set **this.selectedGateId = 0;** we would lose the previous selected gate when we navigate between pages

jrobbins: I think it is OK to lose that on navigations that close the sidebar.   The navigations were we want to maintain the selectedGateId state are the navigations that do not close the sidebar: right now that is only when navigating into the feature detail page iff a certain condition is met.   It would be nice to have it when navigating to myfeatures as well, but there's no way to check the condition that would be needed until after data loads.

KyleJu (approved):

PR #2685, opened 2023-01-20
Update shoelace to 2.0.0-beta.88.  Fix chromedash-textarea.js incompatibilities

dlaliberte:
This shoelace update resulted in an infinite loop in our subclass of the SlTextarea, since form field validation is done differently.  

There were also some issues with setting up components in tests properly.  Subdividing tests so that after each new component, we `await component.updateComplete` is a good idea, if not always essential.


package.json
```diff
@@ -76,7 +76,7 @@
     "gulp-uglify-es": "^3.0.0",
     "ini": ">=3.0.1",
     "kind-of": ">=6.0.3",
-    "lit-analyzer": "^1.2.1",
+    "lit-analyzer": "^0.0.6",
```
dlaliberte, line 79: Is this the right thing to do?  I ran the suggested command to remove 3 vulnerabilities, which might have resulted in this change of the lit-analyzer version.  But version 0.0.6 is older than 1.2.1, right?

dlaliberte:
One possible unintended side-effect is that, while validation is being done correctly for textarea fields, it now pops up an "Invalid" notice after every keypress, as long as the entered text is still invalid.  You don't have to close it or anything - but it is a flashy unnecessary reminder that you haven't finished entering your url or email address.

package.json
jrobbins: @jcscottiii I recall you saying that there was something more complex about upgrading lit-analyzer to a new version that was still in beta.  Dan made this change and it seems to pass tests.  Is this change OK (even if we need to do more later)?

jcscottiii: The change involves moving from this 3rd party lit analyzer to a [1st party lit analyzer](https://www.npmjs.com/package/@lit-labs/analyzer). But before adopting that one, we should wait until it is stable (at least 1.0.0).

It's safe to make the change. But I'm more afraid that we may be missing out on more lint rules by downgrading. Here's the [CHANGELOG](https://github.com/runem/lit-analyzer/blob/master/CHANGELOG.md) over time for the current lit analyzer. Since those vulnerable packages are only for lit analyzer for linting, I think it safe to keep it at 1.2.1.

client-src/elements/chromedash-textarea.js
```diff
@@ -26,20 +26,22 @@ export class ChromedashTextarea extends SlTextarea {
...
    * @param {string} value
-   * @return {boolean}
+   * @return {boolean} Return true if value is valid.
    */
   customCheckValidity(value) {
     if (this.multiple) {
       if (this.chromedash_split_pattern &&
           this.chromedash_single_pattern) {
         const items = value.split(new RegExp(this.chromedash_split_pattern));
+        console.info('items', items);
```
jrobbins, line 36: Left over debugging line.

client-src/elements/chromedash-textarea.js
```diff
@@ -26,20 +26,22 @@ export class ChromedashTextarea extends SlTextarea {
...
         const singleItemRegex =
             new RegExp('^' + this.chromedash_single_pattern + '$', '');
         const valid = items.every((item) => {
           if (!item) {
             // ignore empty items
             return true;
           }
+          console.info(`item: "${item}"`);
```
jrobbins, line 44: Same.

package.json
jrobbins: Thanks for the reminder.  Let's not downgrade. Let's keep it at 1.2.1.

jrobbins (changes requested):

client-src/elements/chromedash-textarea.js
dlaliberte: done

package.json
dlaliberte: Restored to 1.2.1.

jcscottiii (approved):

jrobbins (approved):

PR #2690, opened 2023-01-25
Define team gates for privacy and security

jrobbins:
This is more progress on #2336.

In this PR:
* Define approval gates for the privacy and security teams for the OT and Ship stages.
* Sort the gates to display the API Owners gate last.
* For gates that don't have questionnaire text, don't show that section in the gate column
* Make WriteMissingGates create the new gates on existing stages.  And, make that handler more scalable.


jrobbins:
FYI, I already ran the script on staging and deployed a version of the app there for a demo that was scheduled for yesterday but that has been rescheduled.  Here is what it looks like:
image

internals/schema_migration.py
```diff
@@ -558,8 +593,8 @@ def get_template_data(self, **kwargs):
           fe.active_stage_id = stage.key.integer_id()
         active_stages_set += 1
       fe.put()
-      
-    
+
```
KyleJu, line 596:
```suggestion
```

KyleJu (approved):

PR #2696, opened 2023-01-27
Cleanup form field validation behavior and style

dlaliberte:
This PR stops calling reportValidity for textarea after every user keystroke, which caused the "invalid" popup to show. 

We also add a data-user-invalid style for chromedash-form-field  sl-input or chromedash-textarea.  The style simply shows a thin red border, barely visible while editing because the focus border partly obscures it, but when the field is blurred, only the red border remains so it stands out.  This provides good feedback both instantly and when blurring.  

image

### DIFF
--- a/client-src/sass/forms-css.js
+++ b/client-src/sass/forms-css.js
@@ -18,8 +18,8 @@ export const FORM_STYLES = [
       vertical-align: top;
     }
 
-    table td:first-of-type { 
-      width: 60% 
+    table td:first-of-type {
+      width: 60%
     }
 
     table .helptext {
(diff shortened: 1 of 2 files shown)


client-src/sass/forms-css.js
```diff
@@ -18,8 +18,8 @@ export const FORM_STYLES = [
       vertical-align: top;
     }
 
-    table td:first-of-type { 
-      width: 60% 
+    table td:first-of-type {
+      width: 60%
```
jrobbins, line 22: Style: I would prefer a semicolon at the end of the line so that future edits can add or move lines with a cleaner diff and less chance of a mysterious syntax error.

jrobbins (approved):
LGTM after considering two style comments.
Thanks for getting this done.  It avoids a headache for users.

client-src/sass/forms-css.js
dlaliberte: This change happened by some auto-formatting feature of vscode, which just removed trailing spaces.  I'll add the semicolons.

PR #2706, opened 2023-01-30
Phase out Set Approval APIs on frontend

KyleJu:
Migrate SetApprovals to SetVotes. 

Gate ID is used in the APIs instead of Field ID, assuming that there is a one-to-one relationship between Gate ID and Gate Type (Field ID) per feature. It won't be true once we introduce multiple stage entities for the same type of stage, e.g. multiple shipping gates.


jrobbins (approved):
I think it is a fair assumption for now that there will be only copy of each stage.   But, it won't hold true forever, luckily we will probably phase out the approvals dialog before people start using multiple stages.

Issue #1812, opened 2022-03-30
The old logo must be updated


ghost:
**Describe the bug**
Starting at version 100 the Chrome browser had the logo changed to a new version with better colors. But the website "Chrome Platform Status" still displays the old logo.

**URL of the logo that should be updated to the new version:**
https://chromestatus.com/static/img/chrome_logo.svg

kneerightful:
I just forked it, updated it to the new logo, the SVG file size has been greatly reduced as well.

**New logo url:**
https://github.com/kneerightful/chromium-dashboard/blob/main/static/img/chrome_logo.svg

kneerightful:
Hey, @jpmedley

I updated it to the new logo, the SVG file size has also been greatly reduced.
Just replace the old logo with the new one in the /static/img/ folder.

**New logo url:**
https://raw.githubusercontent.com/kneerightful/files/main/chrome_logo.svg

Kunal0220:
Hi @kneerightful ,I guess there is a small issue with running the project on the local system , Once it gets sorted , I will change the Old Logo with the New one and fix this issue.
Thank You :)

jrobbins:
@Kunal0220, still interested in making this change?

Kunal0220:
Would love to , can I go ahead?

jrobbins:
Sure, this is still needed.
We should be using the logo that currently shows up on chrome.com:
https://www.google.com/chrome/static/images/chrome-logo-m100.svg

KUNDAN1334:
i am intrested in this and want to change the  logo but  can you guide me in which file i have to make changes

Prasun60:
Hi, I see this issue isn't closed yet. I am interested to solve this issue, can you tell me which file should I replace ( location of the old SVG), as I checked that the old svg (/statict/img/chrome_logo.svg) and the one you have provided here are the same.

awanishyadav967:
> Sure, this is still needed. We should be using the logo that currently shows up on chrome.com: https://www.google.com/chrome/static/images/chrome-logo-m100.svg

Review my PR ,i added one

Rishabh-mikku:
Hi, I have understood this issue completely and I am interested in this.

Rishabh-mikku:
Can you please guide me in which file I have to make changes.

zelfroster:
Hey @jrobbins, sorry to ping you.
I wanted to ask if this is issue is still open, and Can I work on it?

[...]

PR #2710, opened 2023-02-01
#1812 issue : Updated old logo

awanishyadav967:
I have changed all old logo to new ones. Please review it once.


jrobbins:
Thanks for sending this PR.  But, ultimately, we went with a CSS-based solution that has just one image file.

Issue #2340, opened 2022-10-14
Generate email notifications to approvers when a new review is requested


jrobbins:
**Is your feature request related to a problem? Please describe.**
In an earlier issue #2337 we implement a button to request a review, but the reviewers need to know that they are being asked to take action.

**Describe the solution you'd like**
When a review is requested, send an email to the users who have permission to approve that Gate.

The email should consist of:
* Descriptive text explaining that they need to take action
* A link to the feature entry detail page with additional path info, a query string, or an anchor will select the appropriate gate

PR #2720, opened 2023-02-02
Reland Create /tasks/email-reviewers in notifier

KyleJu:
#2340. Create a /tasks/email-reviewers in notifier. This task receives gate_type as a param and retrieves approvers as the email recipients.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -146,15 +146,11 @@ def apply_subscription_rules(
   return results
 
 
-def make_email_tasks(fe: FeatureEntry, is_update: bool=False,
-    changes: Optional[list]=None):
+def make_email_tasks(fe: FeatureEntry, target_emails: list[str],
+    is_update: bool = False, changes: Optional[list] = None):
   """Return a list of task dicts to notify users of feature changes."""
   if changes is None:
     changes = []
-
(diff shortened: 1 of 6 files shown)


internals/notifier.py
```diff
@@ -146,16 +146,12 @@ def apply_subscription_rules(
   return results
 
 
-def make_email_tasks(fe: FeatureEntry, is_update: bool=False,
-    changes: Optional[list]=None):
+def make_email_tasks(fe: FeatureEntry, target_emails: list[str],
```
jrobbins, line 149: I think you need to keep this function as-is for the existing email change notifications, and then create a new, separate function to make email tasks for the reviewers.  The reason is that for the reviews, we don't want to notify blink component owners, feature starrers, or anyone else other than the reviewers.

jrobbins (changes requested):
Your approach looks right, but because we don't want to notify feature starrers, etc., we should have more separation between the existing change notification logic and this new reviewer notification logic.

internals/notifier.py
KyleJu: Done

jrobbins (approved):
This looks better, thanks for the change.

Issue #2723, opened 2023-02-03
The checkmark on the UseCounter sorting menu is missing


jrobbins:
**Describe the bug**
We use a checkmark to indicate the current sort order.  However, that seems to have broken.  Shoelace provides helpful error message on the console:

`The checked attribute can only be used on menu items with type="checkbox" `

jrobbins:
The fix is live.

PR #2724, opened 2023-02-03
Add type="checkbox" because shoelace got more strict.

jrobbins:
This should resolve issue #2723.

This updates our code to cope with a "BREAKING" shoelace change in a previous version.  We didn't notice it first because the breakage was not very bad.
https://shoelace.style/resources/changelog?id=_200-beta88


jrobbins:
Before (broken):
image

After (fixed):
image

KyleJu (approved):

PR #2728, opened 2023-02-08
Allow feature editors to request reviews.

jrobbins:
This fixes a problem that we chatted about last week: requesting a review did not work because only a reviewer could set the REVIEW_REQUESTED vote.

Now, a user who can edit a feature should be able to request a review.

In this PR:
* Add logic to VotesAPI to allow a user who can edit a feature to set a REVIEW_REQUESTED vote.
* Add an additional check to make sure that the specified gate is actually associated with the specified feature.
* Add a new unit test file for this file.  Copy and update the unit tests that we had for the ApprovalsAPI.  Add new tests for new permission check.

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -43,23 +43,41 @@ class VotesAPI(basehandlers.APIHandler):
     feature_id = kwargs['feature_id']
     gate_id = kwargs['gate_id']
     feature = self.get_specified_feature(feature_id=feature_id)
+    new_state = self.get_int_param('state', validator=Vote.is_valid_state)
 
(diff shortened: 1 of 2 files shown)


KyleJu:
> This fixes a problem that we chatted about last week: requesting a review did not work because only a reviewer could set the REVIEW_REQUESTED vote.
> 
> Now, a user who can edit a feature should not be able to request a review.
> 

You meant "should be able to"?

KyleJu (approved):

jrobbins:
> You meant "should be able to"?

Fixed in the PR description.

PR #2732, opened 2023-02-11
Refactor the preflight dialog into its own element.

jrobbins:
I'd like to reuse the preflight dialog on the new gate column UI when the user requests to generate an intent email.
There should be no user-visible change in this PR.

In this PR:
* Refactor the dialog into a separate custom element
* Make it work more like our other dialogs:
   * Instead of each action having its own non-visible dialog, create only one dialog instance, and populate it when opening it.
* Rather than precompute item_stage_map, the code now creates specific entries as needed.  That can be more CPU cycles, but the number of iterations on these loops is well under 100.

### DIFF
--- a/client-src/elements/chromedash-preflight-dialog.js
+++ b/client-src/elements/chromedash-preflight-dialog.js
@@ -125,9 +125,7 @@ export class ChromedashPreflightDialog extends LitElement {
   }
 
   renderDialogContent() {
-    console.log(this);
     if (this.feature == null) {
-      console.log('feature is null');
       return nothing;
     }
     const prereqItems = [];
--- a/client-src/elements/chromedash-preflight-dialog.js
+++ b/client-src/elements/chromedash-preflight-dialog.js
@@ -1,0 +1,175 @@
+import {LitElement, css, html, nothing} from 'lit';
+import './chromedash-callout';
+import {findFirstFeatureStage} from './utils';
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-preflight-dialog.js
```diff
@@ -0,0 +1,175 @@
...
+        if (itemName == pi.name) {
+          return {...pi, stage: s};
+        }
+      }
+    }
+    throw new Error('prerequiste is not a defined progress item');
+  }
+
+  renderDialogContent() {
+    console.log(this);
```
dlaliberte, line 128: Remove debugging.

client-src/elements/chromedash-preflight-dialog.js
```diff
@@ -0,0 +1,175 @@
...
+    if (this.feature == null) {
+      console.log('feature is null');
```
dlaliberte, line 130: Should this be a console.warn, and stay in the code?

dlaliberte (approved):
It's much cleaner this way, as a separate component. Great move.

client-src/elements/chromedash-preflight-dialog.js
jrobbins: Removed

client-src/elements/chromedash-preflight-dialog.js
jrobbins: Removed

PR #2734, opened 2023-02-13
Add initial openapi tools and setup

jcscottiii:
# Why
OpenAPI allows us to have better coupling with our data model between frontend and backend for free. A developer describes the OpenAPI specification and can generate frontend and backend code. 

List of [generators](https://openapi-generator.tech/docs/generators/) supported by OpenAPI

# Debates
## Debate 1: Commit generated code vs do not commit generated code.
Background: Choose whether to commit or not commit the generated code
### Option 1: Committing generated code
  - Pros: No need to worry about what generated code went into a build
  - Cons: PRs can be bloated. 
### Option 2: Do not commit generated code
  - Pros: PRs will not be bloated.
  - Cons: Have to make sure everything is pinned and still have to worry if machine A is generating different code from machine B. Have to install `openapi-generator-cli` globally else you would run into a chicken and egg problem with npm packages during install (I need the generated package at install time but the CLI is not installed yet. 
### Decision
While I personally do not like bloating PRs and I also experimented with not committing the generated code, I did not like the cons enough. In the end, I decided to **go with Option 1** and commit the generated code. We commit generated mock code for wpt.fyi for example ([search for "mock" in wpt.fyi](https://github.com/web-platform-tests/wpt.fyi/find/main)). I will place the generated code in a high level folder called "gen". This will help the reviewer avoid spending time reviewing generated code.

# Other Manual Changes
- Add the middleware for the generated middleware client for csrf token and XSSI.
- Add XSSI protection to do_put. The rest of the methods were already done. Just not PUT.
- Add the use of lit-labs/context for the generated client  on the frontend - More about it in the documentation
  - Add context for the client
- use correct package for rollupBabel in gulpfile.babel.js
- add commonjs to rollup . needed for generated code.


jrobbins:
In the PR description, did you mean to say that you decided to go with option 1?

jcscottiii:
Good catch. Yep!

jcscottiii:
A lot of this is generated the first time around. Most of the reviews would be on the client context changes I added to the frontend and the two documentation pages.

jcscottiii:
This PR has been split up into three PRs to make it somewhat more readable to see the initial scaffolding and manual changes I made:

- #2769 
- #2770 
- #2771 (depends on 2770)

Closing this one

Issue #2730, opened 2023-02-08
Need preflight checks on new gate column prior to intent generation


jrobbins:
The old process-overview editing page had a preflight check dialog box that opened prior to navigating to the intent preview page.
We need that for the new gate column as well.

PR #2738, opened 2023-02-15
Add preflight dialog when generating intent via gate column.

jrobbins:
This should resolve issue #2730.

In this PR:
* Load the feature progress when opening the gate column
* For the stage actions, which are the links to generate the intent: instead of directly linking to the intent generation page, check the prerequisites and open the preflight dialog if needed.
* Make a little bit of the logic for the preflight "Edit" links simpler and more robust.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -1,6 +1,10 @@
 import {LitElement, css, html, nothing} from 'lit';
 import {ref, createRef} from 'lit/directives/ref.js';
 import './chromedash-activity-log';
+import {
+  openPreflightDialog,
+  somePendingPrereqs,
+} from './chromedash-preflight-dialog';
 import {autolink, showToastMessage, findProcessStage} from './utils.js';
 import {GATE_QUESTIONNAIRES} from './form-definition.js';
 
@@ -115,6 +119,7 @@ export class ChromedashGateColumn extends LitElement {
       feature: {type: Object},
       stage: {type: Object},
       gate: {type: Object},
+      progress: {type: Object},
       process: {type: Object},
       votes: {type: Array},
       comments: {type: Array},
@@ -131,6 +136,7 @@ export class ChromedashGateColumn extends LitElement {
     this.feature = {};
     this.stage = {};
     this.gate = {};
+    this.progress = {};
     this.process = {};
     this.votes = [];
     this.comments = [];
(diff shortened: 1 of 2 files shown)


jrobbins:
Here is what it looks like when the preflight dialog opens:
image

client-src/elements/chromedash-gate-column.js
```diff
@@ -146,12 +152,14 @@ export class ChromedashGateColumn extends LitElement {
     this.gate = gate;
     const featureId = this.feature.id;
     Promise.all([
+      window.csClient.getFeatureProgress(featureId),
```
dlaliberte, line 155: Is this redundant with the following line?  Oh, subtle, "Progress" vs "Process".  Hmm, I have confused those before. I wonder if one of them could be changed to something more different.   Since "process" is also a verb, maybe there is a similar word that is a noun only.  Just something to think about for now.

client-src/elements/chromedash-preflight-dialog.js
```diff
@@ -99,18 +99,16 @@ export class ChromedashPreflightDialog extends LitElement {
...
   renderEditLink(stage, feStage, pi) {
-    const featureId = this.feature.id;
-    let editEl = nothing;
-    if (pi.field) {
-      editEl = html`
+    if (pi.field && stage && feStage) {
+      return html`
         <a class="edit-progress-item"
-           href="/guide/stage/${featureId}/${stage.outgoing_stage}/${feStage.id}#id_${pi.field}"
+           href="/guide/stage/${this.feature.id}/${stage.outgoing_stage}/${feStage.id}#id_${pi.field}"
```
dlaliberte, line 105: Does this actually go to the right location?  Do we have an id for a light-dom element as opposed to the form field which is in shadow-dom?

dlaliberte (approved):
A couple comments, maybe for future work.

client-src/elements/chromedash-gate-column.js
jrobbins: Yeah, it is a bit of a tongue twister.  I'll make a note to myself to revisit this naming.  Also, once we get the fields to be organized by gate, then these items will become less important compared to the progress through the set of needed gates.  Once we get to that point, I might have more insight into a name that differentiates the two.

client-src/elements/chromedash-preflight-dialog.js
jrobbins: It seems to scroll the the correct form field.

PR #2742, opened 2023-02-16
Generate new routes for admin/blink.html [4/5]

jcscottiii:
GET /componentsusers - The getter function that list all components and possible users and whether they are owners or just subscribed
PUT /components/{componentId}/users/{userId} - Add a user to a component
DELETE /components/{componentId}/users/{userId} - Remove a user from a component


api.yaml
```diff
@@ -15,10 +15,111 @@ info:
...
+            required:
+              - id
+              - name
+
+    ComponentUsersRequest:
+      description: "Traits about the user in relation to the component"
+      properties:
+        owner:
+          type: boolean
+          description: Impacts this user's ownership. For PUT, add ownership. For DELETE, remove ownership.
```
jrobbins, line 125: I'm concerned that this yaml file will be really big and hard to maintain.  The YAML syntax uses a ton of vertical space and there is no include directive.  Also, I wonder if indentation errors could be hard to track down.

I don't like the idea of layering on another code generation step, but what do you think about creating a simple python script (that would allow us to use python dicts or dataclasses, and python imports) that generates the api.yaml file (or a pretty-printed JSON file)?

api.yaml
```diff
@@ -15,10 +15,111 @@ info:
...
+                $ref: '#/components/schemas/ComponentsUsers'
+  /components/{componentId}/users/{userId}:
+    put:
+      summary: Add a user to a component
+      operationId: addUserToComponent
+      parameters:
+        - in: path
+          name: componentId
+          schema:
+            type: integer
```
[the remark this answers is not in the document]
jcscottiii: Good catch!

api.yaml
jcscottiii: I figured out how to split the YAML into multiple files. More details here: https://github.com/GoogleChrome/chromium-dashboard/pull/2769#discussion_r1126956635

This should prevent the openapi file from growing too large. Also, with that we don't need to write our own code generation to build the final YAML.

openapi/componentsusers.schemas.yaml
```diff
@@ -0,0 +1,37 @@
+ComponentsUsers:
+  properties:
+    users:
+      type: array
+      items:
```
jrobbins, line 5: Even though module yaml is a big improvement over having one huge yaml file, the yaml syntax is making me think that we should look at GRPC and pRPC again.

jcscottiii: Is there something that is missing in YAML syntax that the proto files from GRPC and pRPC files provide?

# Alternative 1: protobuf file - > Generated pRPC server and manually written frontend

Pros:
- pRPC servers works on App Engine
- Familiarity with proto files

Cons:
- pRPC does not generate frontend clients ([example](https://osscs.corp.google.com/chromium/infra/infra/+/main:appengine/monorail/static_src/shared/typedef.js;l=10-11?q=prpc%20-f:test&ss=chromium%2Finfra%2Finfra:appengine%2Fmonorail%2Fstatic_src%2F) Monorail has to manually write the typedefs for frontend. This is one thing that OpenAPI handles for us) 

# Alternative 2: protobuf file -> Generated GRPC server and GRPC frontend client

Pros:
- With [GRPC-Web](https://github.com/grpc/grpc-web) or [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway), can get a generated frontend
- Familiarity with proto files

Cons:
- GRPC is not supported by App Engine. Could migrate to Cloud Run but that's a big shift in infrastructure.
- GRPC-Gateway requires additional infrastructure to have a proxy service


If I were to pick between the two options, I would pick option 2. That way we don't have to keep manually updating the client whenever we generate the backend in option 1. 

# My recommendation
Stay with this OpenAPI PR. This will allow developers to get better type checking automatically for frontend and backend. As a bonus, we can generate [protobuf](https://github.com/OpenAPITools/openapi-generator/blob/master/docs/generators/protobuf-schema.md) files from OpenAPI as a quick migration path if want to move to alternative 2.

jrobbins: > Is there something that is missing in YAML syntax that the proto files from GRPC and pRPC files provide?

The proto syntax is just more compact and easier to edit.  I suppose that I will get used to this syntax with time.

jrobbins (approved):

PR #2743, opened 2023-02-16
Migrate admin/blink.html to SPA [5/5]

jcscottiii:
Add implementations of the three new API routes
Remove old code for templated admin/blink.html

| Mode      | Old | New |
| ----------- | ----------- | ----------- |
| Regular      | image | image |
| Edit   | image  | image |

Note: The shift is due to the sidebar [panel](https://github.com/GoogleChrome/chromium-dashboard/blob/f0ed8038fa1ed3c8a2ebde04d7bdd4ae51714391/client-src/elements/chromedash-app.js#L352-L382). I didn't think it is big deal. If it is, I can add a conditional to make the `wide` variable evaluate to true for this page.

## Fixes 
Also the current admin/blink page is broken. If an admin currently navigates to https://cr-status-staging.appspot.com/admin/blink , clicks edit mode and clicks the drop down for the users, it contains a list but no names are displayed.


api/components_users.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from chromestatus_openapi.models import (
+  ComponentsUsers,
+  ComponentsUsersComponentsInner,
```
jrobbins, line 18: It's not your fault, but I don't like these generated model object names.  I just think that they will be hard to work with.  Is there an alternative style of using these where we create the overall response object early and then call type-safe methods to mutate it to build up the final response without all the mid-level objects being explicit in our code?  (That's one way to use protobufs.)

jcscottiii: I 100% agree that the current names are bad. But they can be changed. (Naming is always hard haha). Do you have any recommendations?

I thought about changing `ComponentsUsers` -> `ComponentsUsersResponse`, `ComponentsUsersComponentsInner` -> `OwnersAndSubscribersOfComponent` and `ComponentsUsersUsersInner` to `ComponentsUser`.


Could you give an example of what you mean for the style.

This is something that is currently possible but not sure if it meets what you want.

```py
resp = ComponentsUsers()

# build up users
users = []
# do stuff
resp.users(users)


# build up the components
# do stuff
resp.components(components[1:])


return resp.to_dict()
```

jrobbins: Yeah, I think that might be a little more clear.  Also, I always prefer to use the assignment operator syntax if possible rather than the setter function syntax.

In your `# do stuff` section, you would still need to construct the inner objects and reference their classes by name.  Those inner class names are kind of long and we need all those imports.  

Protocol buffers can avoid that a bit having repeated fields offer an `.add()` method.    Here's an example from https://protobuf.dev/getting-started/pythontutorial/ that does not need to reference the PhoneNumber class by name.
```
import addressbook_pb2
person = addressbook_pb2.Person()
person.id = 1234
person.name = "John Doe"
person.email = "jdoe@example.com"
phone = person.phones.add()
phone.number = "555-4321"
phone.type = addressbook_pb2.Person.HOME
```

It can even forward the add() keyword parameters to the child object constructor, like:
`person.phones.add(number="555-4321", type=addresbook_pb2.Person.HOME)`
See https://protobuf.dev/reference/python/python-generated/

This allows for child objects to be added without mentioning their class names, which makes the application code more succinct.


In terms of naming the model classes, I like the idea of using the word Response and your other suggested changes.

jcscottiii: I have changed the class names in https://github.com/GoogleChrome/chromium-dashboard/pull/2742 and rebased this PR with those changes.

The `add` method is a nice feature. Thanks for the example and explanation! I can see the benefit of the one line `.add(all the kwargs)`.

client-src/elements/chromedash-admin-blink-component-listing.js
```diff
@@ -0,0 +1,252 @@
...
+      VARS,
+      LAYOUT_CSS,
+      css`
+      :host {
+        display: flex;
+      }
+
+      :host([editing]) .owners_list_add_remove {
+          opacity: 1;
+          pointer-events: all
```
jrobbins, line 19: Style: Include trailing semicolons to make the file easier to edit and make future diffs more clear.
https://google.github.io/styleguide/htmlcssguide.html#Declaration_Stops

client-src/elements/chromedash-admin-blink-page.js
```diff
@@ -0,0 +1,212 @@
...
+    };
+  }
+
+  /** @type {import('chromestatus-openapi').DefaultApiInterface} */
+  _client;
+
+  /** @type {Array<import('chromestatus-openapi').OwnersAndSubscribersOfComponent>} */
+  components;
+
+  /** @type {Map<int, import('chromestatus-openapi').ComponentsUser} */
```
jrobbins, line 80: Does this need a closing angle bracket `>`?

framework/basehandlers.py
```diff
@@ -573,6 +573,14 @@ def get_template_data(self, **defaults):
           self, feature_id)
       if redirect_resp:
         return redirect_resp
+    # Validate the user has admin permissions and redirect if needed.
+    if defaults.get('require_admin_site'):
+      user = self.get_current_user()
+      # Should have already done the require_signin check.
+      # If for reason, we don't let's treat it as the main 403 case.
+      if not user \
```
jrobbins, line 581: Style: Instead of using `\` for line continuation, wrap the expression in parens.
https://google.github.io/styleguide/pyguide.html#32-line-length

client-src/elements/chromedash-admin-blink-component-listing.js
```diff
@@ -0,0 +1,252 @@
...
+      id: {type: Number},
+      name: {type: String},
+      subscriberIds: {type: Array},
+      ownerIds: {type: Array},
+    };
+  }
+
+  constructor() {
+    super();
+    this._client = window.csOpenApiClient;
```
jrobbins, line 101: I assume that this global is initialized in another PR.

jrobbins (approved):
Thanks for making these changes.   This all looks good now, except for some minor style nits to consider.

client-src/elements/chromedash-admin-blink-component-listing.js
jrobbins: Yep, I see it in SPA.html.

client-src/elements/chromedash-admin-blink-page.js
jcscottiii: Good catch. Somehow auto-complete still worked without it but definitely want to add it for correctness.

image

client-src/elements/chromedash-admin-blink-component-listing.js
jcscottiii: Will do!

framework/basehandlers.py
jcscottiii: Good catch!

PR #2746, opened 2023-02-16
Use object params in notifier_helpers

KyleJu:
Use object params in notifier_helpers. Notifier expects an object param


jrobbins (approved):
Whoops, I should have noticed that in the first review.

KyleJu:
> Whoops, I should have noticed that in the first review.

nah that is my bad

PR #2761, opened 2023-02-21
Get the correct gate state before change

KyleJu:
Get the correct gate state before change. The old_state and  new_state were the same [before this change](https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/GAjgQUaEJJc)


jrobbins (approved):
D'oh, I should have spotted this defect in an earlier review.

PR #2767, opened 2023-02-22
Notify devrel of feature changes

### LINKED ISSUE #2764: Treat devrel_emails field like we treat the cc_emails field
The developer relations team would like more notifications about new and updated features.  There is a devrel_emails field in each feature entry, but we don't actually send notifications to it.

We should treat the devrel_emails field like we treat the cc_emails field.  There should be another call to accumulate_reasons() for it around line 178 of notifier.py.

KyleJu:
Fix #2764. Notify devrel of feature changes

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -180,8 +180,13 @@ def make_feature_changes_email(fe: FeatureEntry, is_update: bool=False,
     'You are CC\'d on this feature'
   )
   accumulate_reasons(
-      addr_reasons, watcher_emails,
-      'You are watching all feature changes')
+    addr_reasons, fe.devrel_emails,
+    'You are CC\'d on this feature'
+  )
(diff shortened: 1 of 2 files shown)


internals/notifier.py
```diff
@@ -180,8 +180,13 @@ def make_feature_changes_email(fe: FeatureEntry, is_update: bool=False,
     'You are CC\'d on this feature'
   )
   accumulate_reasons(
-      addr_reasons, watcher_emails,
-      'You are watching all feature changes')
+    addr_reasons, fe.devrel_emails,
+    'You are CC\'d on this feature'
```
jrobbins, line 184: Let's change the reason text to:
`You are a devrel contact for this feature.`

jrobbins (approved):
LGTM after making one wording change.

PR #2769, opened 2023-02-23
Add OpenAPI documentation [1/5]

jcscottiii:
Changes:

- README.md - Update mention that java is needed both the Datastore emulator (existing dependency) and now openapi-generator-cli
- developer-documentation.md - Add section about adding a new API with example
- openapi-documentation.md - Add documentation about openapi. Maintenance expectations. Explanation of each additional property.


DanielRyanSmith (approved):
Thanks for the documentation addition - did you write all of this from scratch? Seems quite thorough 🙂

jcscottiii:
@DanielRyanSmith thanks! And I did! Took some time 😅

developer-documentation.md
```diff
@@ -79,3 +79,332 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+</details>
+
+### Step 5: Incorporate Into Backend
+
+Currently, the repository is configured to use the generated Python data models for the backend. *Once all routes are generated by OpenAPI, it would be wise to revisit using the controllers as well*
+
+- Open `main.py`
+- Locate the `api_routes` variable.
+- Add a route.
+  - In this example, it would be `Route(f'{API_BASE}/features/<int:feature_id>', features_api.FeaturesAPI)`.
```
jrobbins, line 381: So, how is this different than how we write our python code on the server side before introducing OpenAPI?

What are the generated model classes for?

jrobbins: > So, how is this different than how we write our python code on the server side before introducing OpenAPI?
> 
> What are the generated model classes for?

Ah,  I think I see it come together in your PR [5/5].

developer-documentation.md
```diff
@@ -79,3 +79,332 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+          '200':
+            description: An updated feature
+            content:
+              application/json:
+                schema:
+                  $ref: '#/components/schemas/Feature'
+
+  components:
+    schemas:
+      Feature:
```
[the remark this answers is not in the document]
jcscottiii: There is an arbitrary [object](https://swagger.io/docs/specification/data-models/data-types/#object) type (Check the "Free-Form Object" section) which could help. The only difference is that the key has to be of type string.  The value can be any type (int, array, another object)

developer-documentation.md
```diff
@@ -79,3 +79,332 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+        description: >-
+          Get a feature by ID. More details about this here.
+          Also, can do more comments
+        operationId: getFeatureById
+        parameters:
+          - name: feature_id
+            in: path
+            description: Feature ID
+            required: true
+            schema:
```
jrobbins, line 152: What do you think about using the YAML "flow" syntax to compress some of the most common lines?
https://yaml.org/spec/1.2.2/#74-flow-collection-styles
E.g., `schema: {"type": "integer", "format": "int64"}`

developer-documentation.md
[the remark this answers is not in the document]
jrobbins: OK I am thinking that we will use that for our FeatureEntry responses for /api/v0.   Then use more specific types when we design v1.

jrobbins (approved):

developer-documentation.md
```diff
@@ -79,3 +79,332 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+
+
+#### Step 1b: Add Operations
+
+Operations = HTTP verbs. (e.g. GET, POST, PUT, etc)
+
+- Add the operation(s) under the path.
+- Ensure each operation has a `summary`, `description` and `operationId`
+- If your path has path parameters, describe the parameters now too.
+  - Mark these are required as well.
```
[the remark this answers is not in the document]
jcscottiii: Will do!

developer-documentation.md
```diff
@@ -79,3 +79,332 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+  <summary>Example (click to expand)</summary>
+  
+  #### api.yaml
+  ```yaml
+  paths:
+    /features/{feature_id}:
+  ...
+      post:
+        summary: Update a feature by ID.
+        description: >-
```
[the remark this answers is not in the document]
jcscottiii: Great question!
It depends on the text. This example would be better suited with `|` as you mentioned. (I will change it here). Other times, you might not want it ([Example](https://github.com/GoogleChrome/chromium-dashboard/pull/2742/files#diff-f1cb3b3ecb32bb4cd8b5cb7a4f7cc5217afb2b62468af3db51b4cc9be379d671R25))

developer-documentation.md
jcscottiii: I would be hesitant to do this syntax for a few reasons:
- I would want to be consistent with every object.
- Current tooling doesn't work with this. (Lose auto-completed suggestions)

In the meantime, I remember there was a question about splitting this into multiple files. I wasn't sure at the time but I did some digging and found out that it can indeed be [done](https://swagger.io/docs/specification/using-ref/). (see `$ref Syntax`) If length becomes a problem, we can explore this.

jcscottiii: I figured out how to split the files out. I have updated the instructions to do it. The other PRs have been updated to the new modular system.

Another good resource about is https://blog.pchudzik.com/202004/open-api-and-external-ref/

jcscottiii:
@jrobbins I figured out the modular openapi file design and updated the docs

developer-documentation.md
```diff
@@ -79,3 +79,264 @@ address of any Google account that you own, such as an `@gmail.com` account.
...
+  _clientConsumer;
+
+  constructor() {
+    super();
+    this._clientConsumer = new ContextConsumer(this, chromestatusOpenApiContext, undefined, true);
+  }
+
+  fetchData() {
+    // Important to call .value to get the client from the context.
+    this.clientConsumer.value.getFeature();
```
jrobbins, line 338: This should have an underscore, right?

jrobbins (approved):

PR #2770, opened 2023-02-23
Add initial api.yaml and generated code [2/5]

jcscottiii:
This PR is a refactor of #2734 to contain the minimal list of manual changes. Other manual changes from that PR will separated to another PR to allow for easier review.
Manual changes:
- api.yaml:  All the necessary fields to have a valid api.yaml.
- package.json: add `openapi`, `openapi-validate`, `openapi-frontend` and `openapi-backend` scripts. Add @openapitools/openapi-generator-cli as a dependency to do the generation of the code. add gen/ folder to mypy exclusion
- .github/workflows/ci.yml: add gen/ folder to mypy exclusion

Generated changes:
- openapitools.json: Made by openapi-generator-cli on initial generate so that subsequent generation use the same pinned version of the JAR.
- gen/*: The generated code
- package-lock.json: Updated because I added the new devDependency


gen/py/chromestatus_openapi/chromestatus_openapi/controllers/default_controller.py
```diff
@@ -0,0 +1,18 @@
...
+
+def componentsubscribers_get():  # noqa: E501
+    """componentsubscribers_get
+
+     # noqa: E501
+
+
+    :rtype: Union[None, Tuple[None, int], Tuple[None, int, Dict[str, str]]
+    """
+    return 'do some magic!'
```
jrobbins, line 18: Where does this come from? 
In other words, where do we write our API handlers?

jrobbins: Ah.  I'm thinking that this controller is not used.  The 'do some magic' string is just a placeholder in the generated code and we do not intend to fill it in because we have our own way of setting up python request handlers in main.py.

jrobbins (approved):

jrobbins (approved):

PR #2771, opened 2023-02-23
Manual changes to support using openapi [3/5]

jcscottiii:
This PR is a refactor of  #2734 to contain the rest of the manual changes to start using the changes in #2770

.devcontainer/Dockerfile
- Install java in devcontainer

.devcontainer/devcontainer.json
- Make devcontainer automatically install VSCode OpenAPI plugin mentioned in #2769
- Make devcontainer automatically install VSCode plugin for Lit

client-src/elements/chromedash-app.js
- Use the generated client in the frontend.
- Leverage lit-labs/context to provide the client.
  - Docs about context and use-cases: https://lit.dev/docs/data/context/

client-src/elements/chromedash-app_test.js
- Add tests for the changes in client-src/elements/chromedash-app.js

client-src/contexts/openapi-context.js
- Declare a context for the openapi client.

framework/basehandlers.py
- add do_put function. previously, all other HTTP verbs had a `do_<verb>` version except PUT.
- Add put function that uses defensive_jsonify. This enables XSSI for PUT requests.
  - This needed change is safe to do because the only route using PUT is the admin/blink page which I am migrating into the SPA.

gulpfile.babel.js
- Fix usage of rollupBabel. Previously, @rollup/plugin-node-resolve's default export was named rollupBabel. Which is the wrong package. Fixed to use @rollup/plugin-babel
- Using the correct parameters to rollupBabel from https://www.npmjs.com/package/@rollup/plugin-babel
- Add client-src/contexts/*.js to lint and watch targets
- Add custom onwarn for rollup that comes from using the generated frontend code. Check comment for explanation

package.json
- Add lit-labs/context. Needed for changes in client-src/contexts/openapi-context.js
- Tell NPM to use the generated javascript changes via the `workspaces` key.
  - Instead of adding as a dependency like [I initially did here](https://github.com/GoogleChrome/chromium-dashboard/pull/2734/commits/ccd5c8e71b14219299808575ec22e60ababa8cb6#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R101), rather keep the "dependencies" only for ones pulled from npm remotely.
  - workspaces are meant more for packages that comes from the filesystem
  - Docs: https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true

requirements.txt
- Tell pip to install the python package.

package-lock.json
- Generated changes from adding context and workspaces.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -1,8 +1,12 @@
 import {LitElement, css, html, nothing} from 'lit';
+import {ContextProvider} from '@lit-labs/context';
+import {chromestatusOpenApiContext} from '../contexts/openapi-context';
 import {ref, createRef} from 'lit/directives/ref.js';
 import {showToastMessage} from './utils';
 import page from 'page';
 import {SHARED_STYLES} from '../sass/shared-css.js';
+
+import {DefaultApi as Api, Configuration} from 'chromestatus-openapi';
 
 
 class ChromedashApp extends LitElement {
@@ -91,6 +95,48 @@ class ChromedashApp extends LitElement {
     `];
   }
 
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * Refresh the XSRF Token, if needed. Add to headers.
+   *
+   * @param {import('chromestatus-openapi').RequestContext} req
+   * @return {Promise<import('chromestatus-openapi').FetchParams>}
+   */
+  async xsrfMiddleware(req) {
+    return window.csClient.ensureTokenIsValid().then(() => {
+      const headers = req.init.headers || {};
+      headers['X-Xsrf-Token'] = [window.csClient.token];
+      req.init.headers = headers;
+      return req;
+    });
+  }
+
+
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * Server sends XSSI prefix. Remove it and allow the client to parse.
+   *
+   * @param {import('chromestatus-openapi').ResponseContext} context
+   * @return {Promise<Response>}
+   */
+  async xssiMiddleware(context) {
+    const response = context.response;
+    return response.text().then((rawResponseText) => {
+      const XSSIPrefix = ')]}\'\n';
+      if (!rawResponseText.startsWith(XSSIPrefix)) {
+        throw new Error(
+          `Response does not start with XSSI prefix: ${XSSIPrefix}`);
+      }
+      return new Response(
+        rawResponseText.substring(XSSIPrefix.length),
+        {
+          status: response.status,
+          statusText: response.statusText,
+          headers: response.headers,
+        });
+    });
+  }
+
   static get properties() {
     return {
       user: {type: Object},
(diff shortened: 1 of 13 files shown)


jrobbins:
It will take me another day or two to review this.

requirements.txt
```diff
@@ -23,10 +23,13 @@ flask-cors==3.0.10
 funcsigs==1.0.2
 Jinja2==3.1.2
 MarkupSafe==2.1.1
-Werkzeug==2.1.2
+Werkzeug==2.2.3
```
jrobbins, line 26: Do you need this change in this PR?  We previously had this and then I reverted it because I was getting HTTP 405 responses when posting feature editing forms.

jrobbins (commented):
This looks good, but I have one concern about the Werkzeug version change.
And, I'd like to hold this change until after our release this week.

requirements.txt
jcscottiii: This is not needed. I think I must've created my changes with that bad version. Will revert.

client-src/elements/chromedash-app.js
```diff
@@ -120,6 +167,16 @@ class ChromedashApp extends LitElement {
     this.contextLink = '/features';
     this.sidebarHidden = true;
     this.selectedGateId = 0;
+    this._clientProvider = new ContextProvider(
+      this,
+      chromestatusOpenApiContext,
+      new Api(
```
jrobbins, line 173: How would you feel about making this a global variable (like csClient) rather than using lit contexts?

jcscottiii: I am 50/50 on this. 
I was reading around that this context [pattern](https://lit.dev/docs/data/context/#services) might be the preferred way to pass such a service in lit.

Some things I read:

- [yaqs/6510228263272448](http://yaqs/6510228263272448)
- https://www.patterns.dev/posts/provider-pattern
- [dont-use-window-for-declaring-global-functionsvariables](http://go/js-practices/externs#dont-use-window-for-declaring-global-functionsvariables)
- [discouraged-export-options](https://goto.google.com/typescript-practices/exports#discouraged-export-options)

With all that said, given that it is in lit labs (aka pre stable), I wouldn't be opposed to using the window global now and have an item in the backlog for once it becomes stable.

Let me know your thoughts.

jrobbins: I prefer the global variable.  This client is used throughout the app, so it is effectively global.  Removing the context stuff would be one less concept to understand.   If we start using contexts for passing app state around, then we could use it for this too, but I don't see the need for it yet.

jcscottiii: I have implemented the global variable. Turns out since we are importing it, it cannot follow the same gulp process as the existing cs client. I added notes about it in this commit https://github.com/GoogleChrome/chromium-dashboard/pull/2771/commits/ae39bcd1465ea17165a38210661797fae2ecb50a

Now there's a new gulp task `rollup-cjs`  to convert it from the ES module to CJS. Then it can be used like the existing cs client.

Issue #2693, opened 2023-01-26
Add the field for intent subject line to the new schema


jrobbins:
**Describe the bug**
Intent email threads can a several different variations on the subject line.  We can detect many of these variations and recognize the thread as an intent thread.   However, unless we store the subject line, we cannot post replies that will appear as part of the thread.

The old schema Feature class has fields for those subject lines:
https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/core_models.py#L190

We should:
* [x] Add a single field for intent subject line to the Stage class
* [ ] Update the detect_intent code to set that field
* [x] Migrate old subject line values to the new field

KyleJu:
@DanielRyanSmith I am happy to take on this issue if you haven't gotten to it!

DanielRyanSmith:
@KyleJu I've actually gotten a good portion of this done. I'll try and submit a PR in the next 24 hours for it. I'll let you know when it's ready!

KyleJu:
> @KyleJu I've actually gotten a good portion of this done. I'll try and submit a PR in the next 24 hours for it. I'll let you know when it's ready!

Perfect!

PR #2772, opened 2023-02-23
Migrate subject line fields to new schema

DanielRyanSmith:
Part of #2693

This change adds a script that checks the old Feature entities for subject line fields and adds them to the corresponding new Stage field `intent_subject_line`. Adding back the subject line matching functionality for posing comments to intent threads will come in a later PR.


jrobbins (approved):
I checked the number of features that have subject lines via pantheon.   It looks like only 90 to 160 will require writes, so this migration script should run within the time limit.

PR #2773, opened 2023-02-24
Add default mailing list for devrel

### LINKED ISSUE #2765: Set up a mailing list as the default value for devrel_emails
The developer relations team would like to be notified of all new features and feature updates even if there is no assigned devrel contact specified on the feature entry.

1. Set up an internal mailing list with a name like devrel-chromestatus-all and configure it to allow posting by our app.
2. Set that address as a default value when a new FeatureEntry is created via the UI.  (I don't think that we need to update existing entries.)

KyleJu:
Fix #2765. Created a mailing list devrel-chromestatus-all@google.com for devrel


jcscottiii (changes requested):
This change would probably fit better in the `settings` package since this is not an enumerated list

KyleJu:
> This change would probably fit better in the `settings` package since this is not an enumerated list

Settings has most the constant environment variables but you are also correct! I moved them to the top of the page instead!

KyleJu:
@jrobbins FYI for the mailing list, in case you need to communicate it to the devrel team

jcscottiii (approved):

Issue #2700, opened 2023-01-28
Roadmap i info modal not getting dismissed


Spectre-ak:
**Describe the bug**
Popup modal on the roadmap page(`chromedash-roadmap-milestone-card`) is only getting dismissed on clicking the same button(`i`) used to launch it. We're able to open the info (popup) modal for each milestone.

**To Reproduce**
1. Go to [https://chromestatus.com/roadmap](https://chromestatus.com/roadmap)
2. Click on `i` information button on STABLE, NEXT UP and DEV
3. Scroll milestones by clicking on Next or (Previous) 3 times
4. Repeat step 2,3

You'll see something like this
![image](https://user-images.githubusercontent.com/62694340/215259189-8206bf11-2ac4-48f3-98e3-425f1d56ef89.png)

**Expected behavior**
Popup modal should get dismissed whenever we click outside the modal. So that at a time no more than one pop-up modal is visible to the user.

cc: @jrobbins

PR #2774, opened 2023-02-24
Added hide popup modal on roadmap milestone

Spectre-ak:
This PR fixes Roadmap `i` info modal not getting dismissed #2700.

Added hide popup to trigger on focus out of `i` button(`sl-icon-button`), which will set `infoPopupRef` to null.
This also hides the popup when the `detailed dates` link is opened.

cc: @jrobbins


jrobbins (approved):
Thanks for doing this.

I tested this on Chrome and it seems to work with keyboard and mouse.   In Firefox, mouse clicks work, but keyboard navigation does not go in the right order.  In Safari and found that it only works with keyboard navigation.  Overall, this is an improvement for the majority of our users and I don't see any downside.

PR #2777, opened 2023-02-28
Lgtm fields migration

### LINKED ISSUE #2665: LGTM fields do not work on the page
**Describe the bug**
The feature fields to denote LGTM approvals for intents are displayed as empty and cannot be added to. The field can be filled in and submitted, but no changes will be saved and the field remains empty. This is because the LGTM fields were not moved over to the FeatureEntry field when handling the schema migration. A number of things are reliant on the fields, including the process overview field which is tracking the fields to determine if aspects of the process have been completed.

**Expected behavior**
These fields need to either be deprecated, or a script needs to be created to migrate them back from Feature entities to FeatureEntry entities. Additionally, another metric or approach will need to be created for determining process completeness.

KyleJu:
Fix #2665. Migrate LGTM fields from feature entities to Vote entities. This PR caps the number of migration per run at 100, in case of App Engine timeout.

This endpoint needs to be hit repeatedly until it returns `0 of 0 lgtms fields migrated.`

### DIFF
--- a/internals/schema_migration.py
+++ b/internals/schema_migration.py
@@ -701,3 +701,63 @@ class MigrateSubjectLineField(FlaskHandler):
         count += len(stages_to_update)
 
     return f'{count} subject line fields migrated.'
+
+
+class MigrateLGTMFields(FlaskHandler):
+
+  def get_template_data(self, **kwargs) -> str:
+    """Migrates old Feature subject lgtms to their respective votes."""
+    self.require_cron_header()
+
+    count = 0
+    votes_to_create = []
+    for f in Feature.query():
+      if (f.i2e_lgtms and not self.has_existing_vote(f.i2e_lgtms, GATE_API_ORIGIN_TRIA, f)):
+        # i2e_lgtms (Intent to Experiment)'s gate_type is GATE_API_ORIGIN_TRIAL.
+        votes_to_create.append(self.create_new_vote(
+            f.i2e_lgtms, GATE_API_ORIGIN_TRIAL, f))
+        count += 1
+
+      if (f.i2s_lgtms and not self.has_existing_vote(f.i2s_lgtms, GATE_API_SHIP, f)):
+        # i2s_lgtms (Intent to Ship)'s gate_type is GATE_API_SHIP.
+        votes_to_create.append(self.create_new_vote(
+            f.i2s_lgtms, GATE_API_SHIP, f))
+        count += 1
+
+    ndb.put_multi(votes_to_create)
+    return f'{count} lgtms fields migrated.'
+
+  def has_existing_vote(self, email, gate_type, f):
+    f_id = f.key.integer_id()
+    # Find the correct gate for a feature_id and gate_type.
+    gate = approval_defs.get_gate_by_type(f_id, gate_type)
+    if not gate:
+      logging.info('Gate entity not found for the given feature. '
+                   'Cannot find vote.')
+      return False
+
+    gate_id = gate.key.integer_id()
+    # Find the vote with.
+    existing_vote = Vote.query(Vote.set_by == email,
+                               Vote.gate_id == gate_id).get()
+
+    # If a vote exists, we do not overwrite existing vote state.
+    if existing_vote:
+      return True
+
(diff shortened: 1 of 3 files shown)


internals/schema_migration.py
```diff
@@ -701,3 +701,63 @@ def get_template_data(self, **kwargs) -> str:
...
+      if (f.i2s_lgtms and not self.has_existing_vote(f.i2s_lgtms, GATE_API_SHIP, f)):
+        # i2s_lgtms (Intent to Ship)'s gate_type is GATE_API_SHIP.
+        votes_to_create.append(self.create_new_vote(
+            f.i2s_lgtms, GATE_API_SHIP, f))
+        count += 1
+
+    ndb.put_multi(votes_to_create)
+    return f'{count} lgtms fields migrated.'
+
+  def has_existing_vote(self, email, gate_type, f):
```
jrobbins, line 730: I think that calling this in a loop in prod will timeout.  We have over 2000 Feature entities on prod.  I think that making 2000 * 2 * 2 = 8000 queries could exceed the time limit for a request.
It would be better to prefetch all existing Vote entities and put them into a dictionary keyed by feature_id with each value being a list of Votes. 

I don't think that you need to fetch the Gates because Vote has feature_id and gate_type.

internals/schema_migration.py
```diff
@@ -701,3 +701,63 @@ def get_template_data(self, **kwargs) -> str:
...
+            f.i2e_lgtms, GATE_API_ORIGIN_TRIAL, f))
+        count += 1
+
+      if (f.i2s_lgtms and not self.has_existing_vote(f.i2s_lgtms, GATE_API_SHIP, f)):
+        # i2s_lgtms (Intent to Ship)'s gate_type is GATE_API_SHIP.
+        votes_to_create.append(self.create_new_vote(
+            f.i2s_lgtms, GATE_API_SHIP, f))
+        count += 1
+
+    ndb.put_multi(votes_to_create)
```
jrobbins, line 727: Creating the Votes in a batch is good because it is fast.  However, when we set votes using approval_defs.set_vote() that also updates the state of the Gate.

I imagine three ways to address it:

1. Do a loop here to call set_vote() so that the Gate state is updated.  That will be much slower, but it could make partial progress on each run of the migration script so long as you don't try to write Votes for features that already have Votes for that gate_type.
2. Change set_vote() to call a new function set_votes_batch() that could set N Votes and update all the relevant Gate's at one time using put_multi() calls.
3. Allow the Votes to be written without updating the Gate states, but then run /admin/schema_migration_gate_status to update all the Gate states afterwards.  I had previously run this successfully in prod, but after you run your migration script, there will be more Votes, so this might timeout.
4. Or, we could change our approach to migration scripts so that each request to run a migration script has `start` and `num` query string parameters.  That would allow us to loop over a subset of Feature entities on each script run.  E.g., try updating 500 features at a time.

jrobbins (changes requested):
The approach seems right, but you are not updating the Gate state.  And, I am concerned that it will exceed the 10 minute GAE timeout on prod.

internals/schema_migration.py
KyleJu: Addressed

internals/schema_migration.py
KyleJu: I tweaked your idea a little bit and cap the number of votes created at 100 at a time. We could repeated hit the end point until the new vote created is 0. WDYT?

DanielRyanSmith:
In order to do this efficiently, it might be worth creating the new votes using `ndb.put_multi()` and then separately invoking the old script I wrote to recalibrate all Gate entity states back when they weren't set (/admin/schema_migration_gate_status). You should not have to worry about the timeout if these are handled independently

jrobbins (commented):
This approach looks like it will scale fine.
Go ahead with the unit tests and I will approve.

KyleJu:
> This approach looks like it will scale fine. Go ahead with the unit tests and I will approve.

Make a bit more changes and added tests

jrobbins (approved):

Issue #2780, opened 2023-03-01
Changing multiple votes in the same dialog can cause timeouts


DanielRyanSmith:
**Describe the bug**
When selecting multiple new vote values in the approvals dialog box, and optionally adding a comment as well, the modal has a tendency to time out or not finish each request to change the vote. The "Save" button is not disabled while the requests are in progress, so it's easy for an impatient or confused user to try and click "Save" again, resulting in duplicates of the requests being sent for each click.

Also, some values do not change regardless of how many votes are simultaneously changed. I have noticed that I cannot change my "Intent to Ship" vote for the feature "Horseless Carriages" in any form on [staging](https://cr-status-staging.appspot.com) 

**To Reproduce**
1. Open up the approvals dialog box for a feature.
2. Change 3-4 vote values to new values. Optionally, add a comment.
3. Click "Save".

The page generally does not change at this point and the modal does not close. Clicking "Save" again submits each request again.

This is a screenshot after changing each value to "Approved" and reloading the page after ~30 seconds of waiting for the modal to disappear. No vote values were changed but comments were added describing my changing of the values.
![Screenshot 2023-03-01 at 2 38 05 PM](https://user-images.githubusercontent.com/56164590/222281565-1ee506b8-816e-4ef0-93f4-093e2a8348e5.png)

**Expected behavior**
"Save" button while requests are being processed. Ideally, some loading icon would also display.
Timeouts and vote selections being unchanged need further investigation.

DanielRyanSmith:
I am able to reproduce the timeout issues locally, and they seem to be related to emailing subscribers of vote changes. The server console hangs at this point:
```
INFO:root:Enqueueing /tasks/email-subscribers task to projects/cr-status-staging/locations/us-central1/queues/default
INFO:root:Making request to 'http://localhost:8080/tasks/email-subscribers'
INFO:root:Enqueueing /tasks/email-subscribers task to projects/cr-status-staging/locations/us-central1/queues/default
INFO:root:Making request to 'http://localhost:8080/tasks/email-subscribers'
INFO:root:Enqueueing /tasks/email-subscribers task to projects/cr-status-staging/locations/us-central1/queues/default
INFO:root:Making request to 'http://localhost:8080/tasks/email-subscribers'
INFO:root:Enqueueing /tasks/email-subscribers task to projects/cr-status-staging/locations/us-central1/queues/default
INFO:root:Making request to 'http://localhost:8080/tasks/email-subscribers'
```

[...]

PR #2782, opened 2023-03-02
Display existing gates for approvals dialog

### LINKED ISSUE #2693: Add the field for intent subject line to the new schema

DanielRyanSmith:
This PR includes work necessary to partially fix #2693 and #2780.

This change updates the approvals dialog section to populate approvals based on actual gates that are associated with the feature rather than a predefined set of gates. This means that the approvals dialog is now functional with the new schema and can support multiples of the same gate, as well as some newly added gates (debuggability, etc.).

This change is also important for updating votes and posting comments to mailing lists in a consistent way. We were currently trying to match gate types with new user approvals, which does not have consistent functionality when trying to update approvals in the db and use specific intent thread URLs to post to mailing lists.

- api/comments_api.py: Add a check to ensure we have a valid gate ID before trying to post a comment to a mailing list. If we have a gate ID (not just a gate type), we can derive the intent thread that we are intended to post to with 100% accuracy.
- api/features_api.py: This is a small optimization change to call `FeatureEntry.get_by_id()` instead of `feature_helpers.get_by_ids()` in the get_one_feature method, since we are in fact fetching 1 feature.
- client-src/elements/chromedash-approvals-dialog.js: Most changes made here. In most places that referenced and used gate type to match votes, we instead use the gate IDs of the gates we already request for. This makes it much easier to display all gates, update votes, and post comments to mailing lists accurately.
- framework/basehandlers.py: Small change that has been affecting mypy. [Mypy does not recognize that an exception will be thrown for flask.abort()](https://stackoverflow.com/a/68749627). This can be fixed separately with the `types-Werkzeug` package.
- internals/detect_intent.py: Adds back the intent and subject line detection with the ability to save these to the Stage fields. A TODO has been added for finding a way to handle this for multiples of the same stage type, as we don't currently have the ability to detect differences of the same intent type (e.g. two "Intent to Extend Experiment" threads associated with the same feature).
- internals/notifier.py: Posting to mailing list now uses the gate ID that is passed in to find the associated Stage entity, and to use that entity to derive the intent thread URL.
- internals/notifier_helpers.py: Small change to fix a ValueError that occurs when an existing Gate entity has an old value that is no longer valid (e.g. 0, "preparing").
- internals/notifier_test.py: Change testing of get_thread_id to match new implementation.

### DIFF
--- a/client-src/elements/chromedash-approvals-dialog.js
+++ b/client-src/elements/chromedash-approvals-dialog.js
@@ -18,32 +18,6 @@ export const STATE_NAMES = [
 ];
 const PENDING_STATES = [2, 3, 4, 8];
 
-const APPROVAL_DEFS = [
-  {name: 'Intent to Prototype',
-    id: 1,
-    threadField: 'intent_to_implement_url',
-  },
-  {name: 'Intent to Experiment',
-    id: 2,
-    threadField: 'intent_to_experiment_url',
-  },
-  {name: 'Intent to Extend Experiment',
-    id: 3,
-    threadField: 'intent_to_extend_experiment_url',
-  },
-  {name: 'Intent to Ship',
-    id: 4,
-    threadField: 'intent_to_ship_url',
-  },
-];
-
-const GATES_BY_FEATURE_TYPE = {
-  0: new Set([1, 2, 3, 4]),
-  1: new Set([1, 2, 3, 4]),
-  2: new Set([4]),
-  3: new Set([2, 3, 4]),
-};
-
 let approvalDialogEl;
 
 export async function openApprovalsDialog(user, feature) {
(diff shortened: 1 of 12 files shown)


client-src/elements/chromedash-approvals-dialog.js
```diff
@@ -509,40 +497,43 @@ class ChromedashApprovalsDialog extends LitElement {
...
+      if (voteValue !== -1) {
+        const changedGate = this.gates.find(g => g.id === Number(gateId));
+        if (!changedGate) {
           continue;
         }
         promises.push(
           window.csClient.setVote(
-            this.feature.id, gatesByField[0].id,
-            this.changedApprovalsByField.get(fieldId)));
+            this.feature.id, changedGate.id,
```
jrobbins, line 509: Changing the protocol between the client and server can be a problem for users who have the SPA open in a browser tab at the time that we deploy the new version.  However, this dialog was already broken so I have already told the API Owners to use the gate column.  So, I think that we can go ahead with deploying this PR in a single step.

jrobbins (approved):

PR #2791, opened 2023-03-05
Update Chrome logos

Aryan20:
Fixes: #1812 
Update the chrome logos to the one launched along with v100


jrobbins:
Thanks for sending this PR.  But, ultimately, we went with a CSS-based solution that has just one image file.

PR #2802, opened 2023-03-08
Add Enterprise feature flow UI/UX changes

yanndago:
Remove the concept of active stage for enterprise policies by hiding it from the UI

Rename the "Start" stage button to edit for enterprise features

Add an id to the "edit all" section headers to serve as an anchor for navigations

Navigate to the "edit all" page instead of the "edit" page after creating a new enterprise feature and automatically scroll to the first rollout section

Remove the edit icon from the feature page that navigates to the edit page

Adding a new stage on enterprise features only allows for the rollout stage to be adde


jrobbins:
In the PR description:
> Adding a new stage on enterprise features only allows for the rollout stage to be adde

you are missing the end of the last line.

client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -102,7 +104,9 @@ export class ChromedashGuideEditallPage extends LitElement {
       if (hash) {
         const el = this.shadowRoot.querySelector(hash);
         if (el) {
-          this.shadowRoot.querySelector(`chromedash-form-field[name="${el.name}"] tr th b`).scrollIntoView(true, {behavior: 'smooth'});
+          const formField = this.shadowRoot
```
jrobbins, line 107: Dan is rewriting this functionality in 
https://github.com/GoogleChrome/chromium-dashboard/pull/2804

I'll ask him to look at what you are doing here and incorporate the same case logic.

jrobbins (approved):

client-src/elements/chromedash-guide-editall-page.js
dlaliberte: It looks like the change here is mostly just reformatting, except for the (formField || el).  Is the "formField", which is really only the header associated with the form field, sometimes missing? 

I suppose we have a conflict between my change and this change, both unsubmitted at the moment.  Is it OK to leave this line unchanged for now, or does it cause an error if the "formField" element (really the header) is missing?

dlaliberte (approved):

client-src/elements/chromedash-add-stage-dialog.js
```diff
@@ -90,17 +91,22 @@ class ChromedashAddStageDialog extends LitElement {
   }
 
   renderStageSelect() {
+    const offerChoice = CREATEABLE_STAGES[this.featureType].length > 1;
+    const initialValue = offerChoice ? 0 : CREATEABLE_STAGES[this.featureType][0];
+    this.canSubmit = !offerChoice;
```
DanielRyanSmith, line 96: Will the user be able to submit the stage creation if there is only 1 stage choice?

pages/guide.py
```diff
@@ -190,7 +190,7 @@ def process_post_data(self, **kwargs):
 
     # TODO(jrobbins): Make this be /feature/ID after ability to edit
     # from the feature detail page is complete.
-    redirect_url = '/guide/edit/' + str(key.integer_id())
+    redirect_url = '/guide/editall/' + str(key.integer_id()) + '#rollout1'
```
DanielRyanSmith, line 193: Does this mean that the user will be redirected to the editall page when creating a stage on the feature detail page?

DanielRyanSmith (approved):
Some clarifying questions, but otherwise LGTM

client-src/elements/chromedash-add-stage-dialog.js
yanndago: Yes they will be that is why I set the "this.canSubmit"

pages/guide.py
yanndago: No adding a new stage keeps you on the same page, it is only when you create a new enterprise feature that it brings you from the new feature page to the edit all page.

PR #2803, opened 2023-03-08
Convert Adoption gate questionnaire into fields.

jrobbins:
Last week we learned that the person who would have been an Adoption gate approver actually prefers to just have fields added for the API owners to review.

In this PR:
* Remove the adoption gate, including its questionnaire.
* Add three fields to Feature and FeatureEntry:
    * field spec, form definition, and guide.py


internals/core_models.py
```diff
@@ -66,11 +66,14 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
...
 
-  # Gate: Adoption
+  # Topic: Adoption (reviewed by API Owners.  Auto-approved gate later?)
   motivation = ndb.TextProperty()
   devtrial_instructions = ndb.TextProperty()
   activation_risks = ndb.TextProperty()
   measurement = ndb.TextProperty()
+  availability_expectation = ndb.TextProperty()
+  adoption_expectation = ndb.TextProperty()
+  adoption_plan = ndb.TextProperty()
```
DanielRyanSmith, line 76: Is there ever a scenario where multiple adoption gates would exist for the same feature? If so, it might be useful for these fields to live on the Stage or Gate kinds.

DanielRyanSmith: Also, I was going to suggest that maybe we add comments here to explain the usage of each of these fields, but I think that would be a useful for all of our fields. I'll write up an issue as a possible enhancement.

client-src/elements/form-definition.js
```diff
@@ -389,6 +389,9 @@ const FLAT_PREPARE_TO_SHIP_FIELDS = {
         'webview_risks',
         'anticipated_spec_changes',
         'i2s_lgtms',
+        'availability_expectation',
+        'adoption_expectation',
+        'adoption_plan',
```
DanielRyanSmith, line 394: Are these fields intended to be available and visible for every feature?

DanielRyanSmith (approved):
Asking some clarifying questions, but otherwise LGTM.

internals/core_models.py
jrobbins: > Is there ever a scenario where multiple adoption gates would exist for the same feature? If so, it might be useful for these fields to live on the Stage or Gate kinds.

Good question, but in this case I think that the feature owner would revise these fields rather than have separate copies of them.  Even a feature with multiple launches is still one feature that gets adopted by web developers.

client-src/elements/form-definition.js
jrobbins: I think just for every feature with feature_type of "new feature incubation".

PR #2806, opened 2023-03-09
Enable all gates and add reviewers.

jrobbins:
This will cause newly created features to have gates for security, privacy, enterprise, debuggability, testing, and API Owners.
I have also set up the approvers that we know so far.
This should get us ready to test on staging with at least one of the new approvers.


internals/core_enums.py
```diff
@@ -269,45 +269,45 @@
...
-         [#GATE_DEBUGGABILITY_ORIGIN_TRIAL,
+         [GATE_DEBUGGABILITY_ORIGIN_TRIAL,
           GATE_API_ORIGIN_TRIAL]),
         (STAGE_BLINK_EXTEND_ORIGIN_TRIAL, [GATE_API_EXTEND_ORIGIN_TRIAL]),
         (STAGE_BLINK_SHIPPING,
          [
-          #GATE_PRIVACY_SHIP, GATE_SECURITY_SHIP, GATE_ENTERPRISE_SHIP,
-          #GATE_DEBUGGABILITY_SHIP, GATE_TESTING_SHIP,
+          GATE_PRIVACY_SHIP, GATE_SECURITY_SHIP, GATE_ENTERPRISE_SHIP,
+          GATE_DEBUGGABILITY_SHIP, GATE_TESTING_SHIP,
```
DanielRyanSmith, line 278: It looks like all of these gates are created for every feature type. Is it confirmed that we want to create each of these gates for every feature upon creation?

DanielRyanSmith (approved):
This looks fine, but I implemented the schema under the assumption that there would be 1 gate per stage. I can't think of any place where this assumption change might lead to incorrect behavior, but I wouldn't be surprised if it was the case. It is something that we should look out for, and ideally add more tests for the future.

internals/core_enums.py
jrobbins: Good point.  I'll remove the gates on the FEATURE_TYPE_CODE_CHANGE_ID case.

KyleJu (approved):
LGTM after addressing Daniel's comment

PR #2810, opened 2023-03-09
Fix missing ctx var in a couple of places.

jrobbins:
This makes the feature creation pages work.


dlaliberte:
Looks fine to me.  I don't think I changed those lines that were missing the ctx parameter, unless it was accidental.

dlaliberte (approved):

PR #2811, opened 2023-03-10
Implement beforeunload for SPA

dlaliberte:
The motivation is to avoid leaving the current "page" without having saved any form field changes.


client-src/elements/chromedash-app.js
```diff
@@ -138,99 +138,177 @@ class ChromedashApp extends LitElement {
...
+            ctx.handled = false;
+            return false;
+          }
+        }
       }
+
+      // Loading new page.
+      this.pageComponent = document.createElement(componentName);
+      changesMade = null;
+      resetBeforeUnloadHandler();
```
jrobbins, line 185: It seems overly complicated to register and unregister the same event handler on every page navigation.  Could we just register it once for the whole app and then store a boolean named `dirty` that is checked inside that event handler?

client-src/elements/chromedash-app.js
```diff
@@ -138,99 +138,177 @@ class ChromedashApp extends LitElement {
...
+        };
+        window.addEventListener('beforeunload', beforeUnloadHandler);
+      });
+
+      window.setTimeout(() => {
+        // Allow submit button to proceed, if the form is valid.
+        // The submit button is not necessarily loaded yet, hence the timeout.
+        // Note: in general, there are several more ways to submit a form.
+        const submitButton =
+          this.pageComponent.shadowRoot.querySelector('input[type="submit"');
```
jrobbins, line 207: What about submit buttons that are in the sidebar?
Do we have any submit buttons that are nested inside multiple levels of web components?
What about the Cancel button?

client-src/elements/chromedash-app.js
dlaliberte: Indeed it does seem overly complicated, but I'm not sure whether it is possible to register it once for the whole app.  There are a few complications: 

1. We only want the beforeunload event to show the dialog if there is an unsaved change to a form field.  We need to handle changes anyway, so setting changesMade (aka dirty) is consistent.  If we simply return from the beforeunload handler when there are not changesMade, that could work.
2. The Submit button (or 'submit' event?) needs to ignore the changes while the submit is happening.  I'm not sure I need to remove the beforeunload event handler to  make that happen; that's just what I saw others doing.  
3. Restoring the beforeunload event handler after a failure to successfully submit (for whatever reason) is the last complication.  There is no way, in general, to detect whether a failure might occur.  Invalid form values is one thing, but other failures are possible.
4. Once the beforeunload event handler is set, it seems form field changes are captured automatically, even those inside shadow DOM.  It's a bit mysterious exactly how that might be happening.  So when a "page" changes, the previous changes need to be forgotten, which also seems to require removing and restoring the beforeunload event handler. 

I'll do some more experimenting to see if I can simplify things a bit.  First priority was just getting it working.  After cleaning up some more, I'll look into whether we can test page changes including simulating the user responding to the beforeunload dialog.

client-src/elements/chromedash-app.js
dlaliberte: Are the sidebar submit buttons for the same form, or a different form?  I'm looking in the pageComponent only, so far.

There are, in general, other ways to submit the form, which are not handled here.  But maybe handling the 'submit' event for the form will deal with all of them.  Not sure.  

I though about the Cancel button, but actually, it seems like a good idea to warn the user that they are losing changes, in case they did not intend to click the cancel button.  Otherwise, it is way too easy to lose everything by clicking on the cancel button which is right next to the submit button.  There's a meme about that :)

client-src/elements/chromedash-app.js
dlaliberte: Seems to work fine to just return from the beforeunload handler if there are no changes (or during submit, if we are pretending there are no changes).  The overall logic is about the same, but we only have to set up the beforeunload handler one time.

dlaliberte:
Tests are needed for inter-page event transitions, but that is going to be a non-trivial change that I can address separately. Just posted: #2841.

jrobbins (approved):

client-src/elements/chromedash-app.js
dlaliberte: BTW, I changed from listening to 'click' on the Submit button to listening to 'submit' on the form.  This catches all ways the form might be submitted.

client-src/elements/chromedash-app.js
dlaliberte: Perhaps I over-simplified the beforeunload handling.  

Regarding [Page Lifecycle API - Chrome Developers](https://developer.chrome.com/blog/page-lifecycle-api/#the-beforeunload-event)  Taymon Beal says:

> Per the above page, though, [permanent use of beforeunload] still breaks the bfcache, so don't install a permanent beforeunload handler. Instead, register a handler whenever the user makes a change, and then unregister it whenever you persist the current state, so that the handler only exists during those moments when there is unsaved work to warn the user about.

It's easy enough for me to switch back to doing that, but it can wait until I figure out how to test the inter-page event  transitions.

PR #2814, opened 2023-03-10
Fix display of feature updated-on and refactor date rendering.

jrobbins:
This fixes a problem that I noticed while testing on staging today.  The "Last updated on" line on the feature detail page was not showing any date.  This is because the JSON for the feature is a little different than it used to be: we are doing less UI-centric string formatting on the backend because it should be done on the frontend.

In this PR:
* Create some new util.js functions to render dates.
* Update existing spots where we render dates to call the new functions.

### DIFF
--- a/client-src/elements/chromedash-activity-log.js
+++ b/client-src/elements/chromedash-activity-log.js
@@ -1,5 +1,6 @@
 import {LitElement, css, html, nothing} from 'lit';
-import {autolink} from './utils.js';
+import {autolink, renderAbsoluteDate, renderRelativeDate,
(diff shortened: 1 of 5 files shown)


jrobbins:
Old (broken):
image

New (fixed):
image

Note that there is some whitespace inside the parens.  That is not desirable, but it is an existing problem with our use or sl-relative-date.

DanielRyanSmith (approved):
Looks great! The extra whitespace bothered me a lot with my implementation of sl-relative-date also, but I wasn't able to find a solid solution for it.

Issue #2797, opened 2023-03-06
Schema migration code cleanup


DanielRyanSmith:
The schema migration has been completed, although the legacy entities are still being written to. It is now time to clean up some of the leftover pieces of the old schema in order to move fully away and into the new schema.

- [x] Move old kind models to a legacy models file.
- [x] Disable writing to legacy models.
- [x] Remove all remaining uses of the legacy models.

Edit: Additional step needed:
- [x] Update email templates to use the new schema.

DanielRyanSmith:
After looking further, it seems that we cannot completely phase out Feature double-writes until we update the email templates to use the new FeatureEntry fields, which will take some work.

PR #2819, opened 2023-03-13
Move legacy Feature helpers and converters to separate file

DanielRyanSmith:
Part of #2797 

Moves the legacy helper and converter methods for the Feature kind into separate files.


jrobbins (approved):
I'm glad to see that only one caller is affected.

Issue #2820, opened 2023-03-13
Escalate notifications for features that have not been verified for accuracy


DanielRyanSmith:
In order to maintain the accuracy of feature data, we have included notifications to have feature owners verify the veracity of important feature fields just before major milestones. This has yielded results, but we would like to expand this in order to ensure that these prompts to verify accuracy are followed and to be warned of instances where the verification process is now followed. In order to do this, it would be useful to "escalate" the accuracy notification emails. Specifically, a notification will be escalated on the 3rd email sent without feature owner action.

In this context, escalation means:
- changing the phrasing of accuracy verification emails, by adding "Escalation" in subject line and an additional sentence in beginning of body explaining the escalation.
- cc'ing the notification email to the WebStatus team.
- optionally notifying spec mentors, editors, or other users with feature edit access.

Steps to implement:
- [x] Feature accuracy notification and page health check
- [x] Add `outstanding_notifications` field to FeatureEntry to track the number of notifications sent without response.
- [x] Add section in email script to cc the WebStatus team (and possibly other users with feature edit access).

DanielRyanSmith:
The code for this is now in production. We can close this until something goes awry. 😅

PR #2827, opened 2023-03-14
Feature accuracy verification handles multiple stages of the same type

DanielRyanSmith:
Part of #2820

This change updates the feature accuracy verification page to handle and reference data from multiples of the same stage type. Additionally, it adds a fix to `guide.py` that marked some fields as changed despite not being edited.

Explanation of changed by file:
- client-src/elements/chromedash-guide-verify-accuracy-page.js: Changes the page to function similarly to the editall page, which displays a sections of fields from stages that exist as part of the feature.
- client-src/elements/form-definition.js: Write a new set of form definitions that are only used on the accuracy verification page. These are the most important pieces of stage data for the feature owner to verify (mostly milestone fields).
- pages/guide.py: Fixes a bug in which the editall stage updates were not using the `touched()` method check to see if the fields exist on the page. This has not been an issue on the editall page yet because every editable field that exists in the stage entity is displayed on the editall page. With the new feature accuracy changes, this bug appeared and would delete values on stage entities that were not displayed on the page. This is now fixed.
Additionally, the `touched()` method has been slightly refactored to have the `form_fields` list passed into it as an argument, which avoids creating the list from the form fields string in multiple invocations of the method.

### DIFF
--- a/client-src/elements/chromedash-guide-verify-accuracy-page.js
+++ b/client-src/elements/chromedash-guide-verify-accuracy-page.js
@@ -38,6 +38,8 @@ export class ChromedashGuideVerifyAccuracyPage extends LitElement {
     this.feature = {};
     this.loading = true;
     this.appTitle = '';
+    this.previousStageTypeRendered = 0;
+    this.sameTypeRendered = 0;
   }
 
   connectedCallback() {
--- a/client-src/elements/chromedash-guide-verify-accuracy-page.js
+++ b/client-src/elements/chromedash-guide-verify-accuracy-page.js
@@ -1,9 +1,15 @@
 import {LitElement, css, html} from 'lit';
 import {ref} from 'lit/directives/ref.js';
-import {showToastMessage} from './utils.js';
+import {flattenSections, showToastMessage} from './utils.js';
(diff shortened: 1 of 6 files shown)


client-src/elements/chromedash-guide-verify-accuracy-page.js
```diff
@@ -115,28 +121,135 @@ export class ChromedashGuideVerifyAccuracyPage extends LitElement {
...
+    if (!stageFields) return nothing;
+
+    // Add a number differentiation if this stage type is the same as another stage.
+    let numberDifferentiation = '';
+    if (this.previousStageTypeRendered && this.previousStageTypeRendered === feStage.stage_type) {
+      this.sameTypeRendered += 1;
+      numberDifferentiation = ` (${this.sameTypeRendered})`;
+    } else {
+      this.previousStageTypeRendered = feStage.stage_type;
+      this.sameTypeRendered = 1;
```
jrobbins, line 138: I would be good to initialize these in the constructor so that they are not appearing first here.

jrobbins (approved):

Issue #2786, opened 2023-03-02
Increase test coverage to 100% for api directory


jrobbins:
**Describe the bug**
You can run a test coverage report by doing `npm run coverage` then `npm run view-coverage` and browsing localhost:8080.

Our unit test coverage for the /api directory is pretty good, but there are a few gaps:
* [ ] channels_api.py
* [ ] converters.py
* [ ] stages_api.py
* [ ] A few untested lines in the other files

Ideally untested code would become tested code by adding new unit tests.   However, if needed we can mark specific lines as excluded from the coverage metrics.  https://coverage.readthedocs.io/en/stable/excluding.html

KyleJu:
<img width="728" alt="Screenshot 2023-03-14 at 11 13 16 AM" src="https://user-images.githubusercontent.com/8611520/225100030-1cf0aa16-fd19-4d55-b71a-fa2415897f22.png">

Converters.py has been improved to 98%

KyleJu:
With #2836, all API files have achieved at least 90% test coverage

PR #2829, opened 2023-03-14
Increase stages_api coverage to 100%

KyleJu:
#2786. Increase stages_api coverage to 100%, plus a drive-by fix in using the correct `field`.

### DIFF
--- a/api/stages_api.py
+++ b/api/stages_api.py
@@ -105,12 +105,12 @@ class StagesAPI(basehandlers.APIHandler):
       if field in body:
         if stage.milestones is None:
           stage.milestones = MilestoneSet()
-        setattr(stage.milestones, field, body['field'])
+        setattr(stage.milestones, field, body[field])
 
     # Keep the gate type that might need to be created for the stage type.
     gate_type: int | None = None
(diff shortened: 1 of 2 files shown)


api/stages_api.py
```diff
@@ -105,12 +105,12 @@ def _update_stage_vals(self, stage: Stage, feature_type: int,
       if field in body:
         if stage.milestones is None:
           stage.milestones = MilestoneSet()
-        setattr(stage.milestones, field, body['field'])
+        setattr(stage.milestones, field, body[field])
```
jrobbins, line 108: This seems like a pretty significant bug fix.   Was the old code ever getting run?

DanielRyanSmith: Yeah, this is a large bug. 😅 
Luckily, this code is only being run at the moment when creating a new blank stage. PATCH requests are not wired up yet, which is why we haven't seen any issue yet. Thanks for catching this @KyleJu!

KyleJu: Caught it along the way when I wrote the tests. Anytime!

jrobbins (approved):

PR #2832, opened 2023-03-14
Escalate accuracy verification notifications after 3 weeks

DanielRyanSmith:
Part of #2820

This change allows reminder notifications to be escalated, and specifically adds escalation functionality to accuracy verification emails. The next Accuracy verification notification will be escalated after two previous notifications with no action taken.

Escalated emails have the following differences:
- Owners, editors, spec mentors, the creator, and the WebStatus group email are all notified rather than just owners.
- The subject line is prefixed with "ESCALATED: ".
- The email template has an explanation at the top of the body about the meaning of an escalated notification.

- internals/core_models.py: Added a new `outstanding_notifications` field which tracks the number of accuracy verification notifications that have been sent with no acknowledgement.
- internals/reminders.py: Most changes made here. The `AbstractReminderHandler` class will now allow for two new class variables, `ALLOW_ESCALATIONS` and `ESCALATION_CHECK`. The are a boolean value to enable escalations and a Callable to determine which notifications to escalate.
Additionally, an optional `cleanup()` method can now be implemented on child classes to be invoked after notifications have be sent. In the case of feature accuracy, this is used to increment the outstanding notifications of each feature that was notified. Some amount of type hinting was also added to functions/methods in this file.
- pages/guide.py: A new line is added to set `outstanding_notifications` to 0 when the `accurate_as_of` date is being reset, which means the user has responded to one of our notifications.
- templates/accuracy_notice_email.html: Updated template to have conditional changes to the email body based on whether the notification has been escalated or not.

### DIFF
--- a/internals/reminders.py
+++ b/internals/reminders.py
@@ -63,7 +63,7 @@ def choose_email_recipients(
 
 def build_email_tasks(
     features_to_notify, subject_format, body_template_path,
-    current_milestone_info, allow_escalation, escalation_check):
+    current_milestone_info, escalation_check):
   email_tasks = []
   beta_date = datetime.fromisoformat(current_milestone_info['earliest_beta'])
   beta_date_str = beta_date.strftime('%Y-%m-%d')
@@ -73,14 +73,14 @@ def build_email_tasks(
     # to use Stage entity fields before removing this Feature use.
     feature = Feature.get_by_id(fe.key.integer_id())
     # Check if this notification should be escalated.
-    is_escalated = allow_escalation and escalation_check(fe)
+    is_escalated = escalation_check(fe)
     body_data = {
       'id': fe.key.integer_id(),
       'feature': feature,
       'site_url': settings.SITE_URL,
       'milestone': mstone,
       'beta_date_str': beta_date_str,
-      'is_escalated': is_escalated
+      'is_escalated': is_escalated,
     }
     html = render_template(body_template_path, **body_data)
     subject = subject_format % fe.name
@@ -105,10 +105,6 @@ class AbstractReminderHandler(basehandlers.FlaskHandler):
   FUTURE_MILESTONES_TO_CONSIDER = 0
   MILESTONE_FIELDS: list[str] = list()  # Subclasses must override
 
-  ALLOW_ESCALATION = False  # By default, reminders do not escalate.
-  # Function to check if a feature notification should be escalated.
-  ESCALATION_CHECK: Callable | None = None  # Subclasses must override
-
   def get_template_data(self, **kwargs):
     """Sends notifications to users requesting feature updates for accuracy."""
     self.require_cron_header()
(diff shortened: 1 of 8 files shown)


internals/reminders.py
```diff
@@ -148,10 +188,13 @@ class FeatureAccuracyHandler(AbstractReminderHandler):
   """Periodically remind owners to verify the accuracy of their entries."""
 
   ACCURACY_GRACE_PERIOD = timedelta(weeks=4)
+  ALLOW_ESCALATION = True 
+  # Notifications after the 2nd with no acknowledgement will be escalated.
+  ESCALATION_CHECK = lambda self, f: f.outstanding_notifications >= 2
```
jrobbins, line 193: Why not define a method in the base class that always returns False and then override that method here?  Then you would not need ALLOW_ESCALATION.

internals/reminders.py
```diff
@@ -89,18 +117,30 @@ def get_template_data(self, **kwargs):
...
     email_tasks = build_email_tasks(
         features_to_notify, self.SUBJECT_FORMAT, self.EMAIL_TEMPLATE_PATH,
-        current_milestone_info)
+        current_milestone_info, self.ALLOW_ESCALATION, self.ESCALATION_CHECK)
     notifier.send_emails(email_tasks)
 
     message =  f'{len(email_tasks)} email(s) sent or logged.'
     logging.info(message)
+
+    if hasattr(self, 'cleanup'):
```
jrobbins, line 126: Rather than checking if there is a cleanup method defined, why not define one here that simply has a `pass` statement, and then override that method in subclasses when needed?

Also, the name `cleanup` is not ideal because it is not cleaning up resources that are left in some messy state.  Maybe try something like `post_process` or let it be more specific to the one way that it is actually used now with a name like `update_counts`.

internals/reminders.py
```diff
@@ -54,18 +72,24 @@ def build_email_tasks(
...
     feature = Feature.get_by_id(fe.key.integer_id())
+    # Check if this notification should be escalated.
+    is_escalated = allow_escalation and escalation_check(fe)
     body_data = {
       'id': fe.key.integer_id(),
       'feature': feature,
       'site_url': settings.SITE_URL,
       'milestone': mstone,
       'beta_date_str': beta_date_str,
+      'is_escalated': is_escalated
```
jrobbins, line 83: Add trailing comma.

internals/reminders.py
DanielRyanSmith: My thought process was that if we might have different implementations of the method that take different arguments, but even then, my implementation is not capable of passing in different arguments here, so your suggestion would be better overall! I've made the change. 🙂

internals/reminders.py
DanielRyanSmith: This is better, and I've implemented it this way. 🙂

internals/reminders.py
```diff
@@ -15,9 +15,10 @@
 from datetime import datetime, timedelta
 import json
 import logging
-from typing import Optional
+from typing import Callable
```
jrobbins, line 18: This is no longer needed.

internals/reminders.py
```diff
@@ -54,18 +72,24 @@ def build_email_tasks(
     # on old Feature entity milestone fields and will need to be refactored
     # to use Stage entity fields before removing this Feature use.
     feature = Feature.get_by_id(fe.key.integer_id())
+    # Check if this notification should be escalated.
+    is_escalated = escalation_check(fe)
```
jrobbins, line 76: Now that you have a method defined in the base class and overridden in the subclass, you don't need to pass `esclation_check`, instead you can just call `self.should_escalate_notification`.

jrobbins (approved):
LGTM after considering 2 comments.

internals/reminders.py
DanielRyanSmith: Removed!

internals/reminders.py
DanielRyanSmith: This is happening in `build_email_tasks()`, which is a separate function and not a class method - so we would either have to pass the check function in or pass in the handler object (but I might be missing something)

jrobbins: Oops.  Nevermind that part.

PR #2837, opened 2023-03-16
Update email templates to use new schema

DanielRyanSmith:
Part of the schema migration and tangentially related to #2820 

This PR updates the email templates that were previously using fields from the old schema to use the new schema and support multiple stages and fields of the same type.

- internals/stage_helpers.py: create a new function `get_stage_info_for_templates()` which aggregates Stage entities associated with the feature that are needed to render information on the email templates. The function determines whether it is needed to display the milestone table in an email, which is represented by the `should_render_mstone_table` attribute in the returned dictionary.
- internals/notifier.py: Updates the `format_email_body()` function to use the new `get_stage_info_for_templates()` helper function. Also adds a helper function`_determine_milestone_string()` that moves existing logic into a separate function for clarity.
- internals/reminders.py: Updates `build_email_tasks()` function to to use the new `get_stage_info_for_templates()` helper function.
- pages/intentpreview.py:  Updates `get_page_data()` function to to use the new `get_stage_info_for_templates()` helper function.
- templates/blink/intent_to_implement.html: Updates the email template to loop through stages to print all relevant data for each field that was previously represented by a single field on the Feature entity.
- templates/estimated-milestones-table.html: Updates the email template to loop through stages to print all relevant data for each field that was previously represented by a single field on the Feature entity.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -57,6 +57,7 @@ def _determine_milestone_string(ship_stages: list[Stage]) -> str:
                 if mstone is None else min(ms.android_first, mstone))
       has_android_mstone = True
   
+  # Add the android suffix if there were only android milestones.
   milestone_str = str(mstone)
   if not has_desktop_mstone and has_android_mstone:
     milestone_str = f'{milestone_str} (android)'
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -43,25 +43,22 @@ from internals.user_models import (
 
 def _determine_milestone_string(ship_stages: list[Stage]) -> str:
   """Determine the shipping milestone string to display in the template."""
-  has_desktop_mstone = False
-  has_android_mstone = False
-  mstone: int | None = None
-  for stage in ship_stages:
-    ms: MilestoneSet = stage.milestones or MilestoneSet()
-    if ms.desktop_first:
-      mstone = (ms.desktop_first
-                if mstone is None else min(ms.desktop_first, mstone))
-      has_desktop_mstone = True
-    elif not has_desktop_mstone and ms.android_first:
-      mstone = (ms.android_first
-                if mstone is None else min(ms.android_first, mstone))
-      has_android_mstone = True
-  
-  # Add the android suffix if there were only android milestones.
-  milestone_str = str(mstone)
-  if not has_desktop_mstone and has_android_mstone:
-    milestone_str = f'{milestone_str} (android)'
-  
+  # Get the earliest desktop and android milestones.
+  first_desktop = min(
+      (stage.milestones.desktop_first for stage in ship_stages
+        if stage.milestones and stage.milestones.desktop_first),
+    default=None)
+  first_android = min(
+      (stage.milestones.android_first for stage in ship_stages
+        if stage.milestones and stage.milestones.android_first),
+    default=None)
+
+  # Use the desktop milestone by default if it's available.
+  milestone_str = str(first_desktop)
+  # Use the android milestone with the android suffix if there are no
+  # desktop milestones.
+  if not first_desktop and first_android:
+    milestone_str = f'{first_android} (android)'
   return milestone_str
 
 
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -41,26 +41,35 @@ from internals.user_models import (
     AppUser, BlinkComponent, FeatureOwner, UserPref)
 
 
+def _determine_milestone_string(ship_stages: list[Stage]) -> str:
+  """Determine the shipping milestone string to display in the template."""
+  has_desktop_mstone = False
+  has_android_mstone = False
+  mstone: int | None = None
+  for stage in ship_stages:
+    ms: MilestoneSet = stage.milestones or MilestoneSet()
+    if ms.desktop_first:
+      mstone = (ms.desktop_first
+                if mstone is None else min(ms.desktop_first, mstone))
+      has_desktop_mstone = True
+    elif not has_desktop_mstone and ms.android_first:
+      mstone = (ms.android_first
+                if mstone is None else min(ms.android_first, mstone))
+      has_android_mstone = True
+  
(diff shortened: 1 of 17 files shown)


internals/stage_helpers.py
```diff
@@ -58,7 +76,91 @@ def get_feature_stage_ids_list(feature_id: int) -> list[dict[str, int]]:
...
+  """Gather the information needed to display the estimated milestones table."""
+  # Only milestones from DevTrial, OT, or shipping stages are displayed.
+  id = fe.key.integer_id()
+  f_type = fe.feature_type or 0
+  proto_stage_type = STAGE_TYPES_PROTOTYPE[f_type]
+  dt_stage_type = STAGE_TYPES_DEV_TRIAL[f_type]
+  ot_stage_type = STAGE_TYPES_ORIGIN_TRIAL[f_type]
+  extension_stage_type = STAGE_TYPES_EXTEND_ORIGIN_TRIAL[f_type]
+  ship_stage_type = STAGE_TYPES_SHIPPING[f_type]
+
```
jrobbins, line 97: Nit: Use at most single blank lines inside function bodies.

internals/stage_helpers.py
```diff
@@ -14,11 +14,29 @@
...
+    STAGE_TYPES_PROTOTYPE,
+    STAGE_TYPES_DEV_TRIAL,
+    STAGE_TYPES_ORIGIN_TRIAL,
+    STAGE_TYPES_EXTEND_ORIGIN_TRIAL,
+    STAGE_TYPES_SHIPPING)
+from internals.core_models import FeatureEntry, MilestoneSet, Stage
+
+
+# Type return value of get_stage_info_for_templates()
+class StageTemplateInfo(TypedDict):
```
jrobbins, line 32: I like this use of types.

internals/stage_helpers.py
```diff
@@ -58,7 +76,91 @@ def get_feature_stage_ids_list(feature_id: int) -> list[dict[str, int]]:
...
+    # This is used to check if rendering the milestone table is needed.
+    'should_render_mstone_table': False,
+    # Note if any intent URLs are seen while organizing.
+    # This is used to check if rendering the table is needed.
+    'should_render_intents': False
+  }
+
+  for s in Stage.query(Stage.feature_id == id):
+    # Stage info is not needed if it's not the correct stage type
+    # or has no milestones specified.
```
jrobbins, line 115: But, you do include stages that have no milestones specified.

internals/stage_helpers.py
```diff
@@ -58,7 +76,91 @@ def get_feature_stage_ids_list(feature_id: int) -> list[dict[str, int]]:
...
+    if (s.stage_type != proto_stage_type and
+        s.stage_type != dt_stage_type and
+        s.stage_type != ot_stage_type and
+        s.stage_type != extension_stage_type and
+        s.stage_type != ship_stage_type):
+      continue
+
+    # If an intent thread is present in any stage,
+    # we should render the intents template.
+    stage_info['should_render_intents'] = (
```
jrobbins, line 125: You could use `stage_info['should_render_intents'] |= s.intent_thread_url is not None`
but it seems more clear as:
```
if s.intent_thread_url is not None:
  stage_info['should_render_intents'] = True
```

internals/stage_helpers.py
```diff
@@ -58,7 +76,91 @@ def get_feature_stage_ids_list(feature_id: int) -> list[dict[str, int]]:
...
+    'dt_stages': [],
+    'ot_stages': [],
+    'extension_stages': [],
+    'ship_stages': [],
+    # Note if any milestones that can be displayed are seen while organizing.
+    # This is used to check if rendering the milestone table is needed.
+    'should_render_mstone_table': False,
+    # Note if any intent URLs are seen while organizing.
+    # This is used to check if rendering the table is needed.
+    'should_render_intents': False
```
jrobbins, line 110: nit: add trailing comma

internals/stage_helpers.py
```diff
@@ -58,7 +76,91 @@ def get_feature_stage_ids_list(feature_id: int) -> list[dict[str, int]]:
...
+      stage_info['proto_stages'].append(s)
+
+    # Make sure a MilestoneSet entity is referenced to avoid errors. 
+    if s.milestones is None:
+      s.milestones = MilestoneSet()
+
+    m: MilestoneSet = s.milestones
+    if s.stage_type == dt_stage_type:
+      # Dev trial's announcement URL is rendered in templates like an intent.
+      stage_info['should_render_intents'] = (
```
jrobbins, line 140: Same.

jrobbins (approved):
LGTM after considering some coding suggestions and nits.

internals/stage_helpers.py
DanielRyanSmith: Yes, much clearer. 🙂

internals/stage_helpers.py
DanielRyanSmith: RIght! I'll edit this comment. This is from an older implementation that did used this helper function to ONLY gather data for rendering the estimated milestones table. I updated it be used for any email template that needs stage data.

internals/stage_helpers.py
DanielRyanSmith: Fixed

internals/notifier.py
```diff
@@ -41,26 +41,36 @@
     AppUser, BlinkComponent, FeatureOwner, UserPref)
 
 
+def _determine_milestone_string(ship_stages: list[Stage]) -> str:
+  """Determine the shipping milestone string to display in the template."""
+  has_desktop_mstone = False
```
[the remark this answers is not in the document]
DanielRyanSmith: Much clearer to me, and not only that, but my version has the corner case of choosing an earlier android milestone (because it is the min), but not marking it with the `(android)` suffix if there is also a later desktop milestone that exists. So not just better, but more correct. 🙂

PR #2844, opened 2023-03-17
Add type definitions for Stage and FeatureEntry JSON representation

DanielRyanSmith:
This PR adds a two major Python type definitions that describe the structure of the JSON representation of the Stage entity and the verbose JSON representation of the FeatureEntry entity. The types have been defined in their own file, data_types.py. Additionally, the respective converters functions have been updated to strictly adhere to to the return type definition.

Previously, the converter functions would translate any defined attribute on the entities to a dictionary property, which made it easier to add newly created properties to the JSON representation (allows us to lazily slap any defined attribute on the JSON), but added a structure to our APIs that meant it is almost impossible to know exactly what the data structure of the return values look like. The hope is that these definition give us a better understanding of exactly what data we are serving from our APIs.

This change should also shed some light on redundant and unneeded information we are serving to the client, and will allow us to possibly create a better structure for our v1 API.

- api/converters.py: Most changes exist here. The bulk of the changes involve removing the use of the `{entity}.to_dict()` functionality, which would return a dictionary with an unknown set of attributes. Instead, each attribute is explicitly added to the dictionary to ensure it is exactly what data we want to be a part of the representation. Also, the `del_none()` function use has been removed, which completely removed any attributes that were marked as None on the JSON. These attributes should be defined on the JSON but marked as null rather than removed entirely in order to maintain a consistent data structure.
- api/features_api.py: Change some methods to use the new data type.
- internals/core_models.py: Small unrelated change to remove some old->new Stage property definitions that didn't exist (there are no iOS origin trials).
- internals/data_types.py: TypedDicts are defined here, as well as inner TypedDicts that exist in the JSON representation of a FeatureEntry entity.


DanielRyanSmith:
@jrobbins We can adjust/scrap this if we find out this overlaps with the OpenAPI implementation, but hopefully it's worth landing beforehand regardless.

PR #2845, opened 2023-03-17
List unfinished gates as prerequisites in the preflight dialog

jrobbins:
This should finish #2730.

The preflight dialog for generating an intent checks that certain feature fields are filled in.  It will now also check that other gates are finished (approved, denied, or NA).  The unfinished gates are listed with links that cause the gate column to focus on that gate.

In this PR:
* Add a new function somePendingGates() to allow the gate column to check if the preflight dialog should open.
* Load gate data just before checking the status of gates.  
    * Since this is now async, the case where we immediately open the new tab gets blocked by Safari's popup blocker, so work around it with setTimeout().
* Refactor and rename several gate- and vote-related constants into form-field-enums.js.
* In the preflight dialog, render `<li>` items for each unfinished gate


jrobbins:
Old screenshot:
image

New screenshot:
image

KyleJu (approved):

PR #2853, opened 2023-03-21
Fix to the beforeunload handling especially for the metadata form

dlaliberte:
This PR is mostly a refactoring of the chromedash-app methods to enable the chromedash-guide-metadata to call the form "submit" handler.  This special handling is required since that form does not have its own "page" but instead it is part of the guide feature overview page.  

Note that the Cancel button does not warn about unsaved changes, and furthermore, the unsaved changes will trigger a confirm dialog on the overview page when leaving it.  Rather than fixing this, we will refactor the metadata form page to  be a proper page.


jrobbins (approved):
This refactoring looks much better to me.

PR #2855, opened 2023-03-22
Use "N/A" and "INT" instead of icons.

jrobbins:
Finish adding support for the internal_review vote value.  Also, there is no intuitive icon for the concept of a "N/A" or "Internal review" vote, so instead of trying to use icons, use short textual abbreviations. 

In this PR:
* approval_defs.py: compute the gate state when there is a vote with value INTERNAL_REVIEW.
* chromedash-gate-chip.js:  allow either an icon or a textual abbreviation that is shown in very bold letters

### DIFF
--- a/client-src/elements/chromedash-gate-chip.js
+++ b/client-src/elements/chromedash-gate-chip.js
@@ -68,7 +68,7 @@ class ChromedashGateChip extends LitElement {
        border: 2px solid var(--dark-spot-color);
      }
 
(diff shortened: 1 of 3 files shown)


jrobbins:
Screenshot before:
image
Note that Enterprise looks exactly like no votes, even though it has an "internal review" vote. 

Screenshot after:
image

KyleJu (approved):

PR #2856, opened 2023-03-22
Use CSS outline rather than border to highlight selected gate chip.

jrobbins:
While working on adding support for "INT" and "N/A" text in gate chips, I noticed that the text baseline was getting messed up whenever I selected a gate chip: The text for the selected chip moves down about 3 pixels.   Apparently, shoelace calculates the line-height using a formula that includes the border width.   To avoid that, let's highlight the selected gate chip using the `outline` property rather than `border`.

### DIFF
--- a/client-src/elements/chromedash-gate-chip.js
+++ b/client-src/elements/chromedash-gate-chip.js
@@ -58,7 +58,7 @@ class ChromedashGateChip extends LitElement {
      }
 
      sl-button.selected::part(base) {
-       border: 2px solid var(--dark-spot-color);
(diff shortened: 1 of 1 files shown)


jrobbins:
Screenshot before:
image
Note that "Security" is pushed downward a little and does not appear vertically centered relative to the chip or icon.

Screenshot after:
image

jrobbins:
I switched to box-shadow to work-around the fact that current safari does not round the outline.  It looks exactly the same.

KyleJu (approved):

PR #2869, opened 2023-03-27
Remove Feature double-writes from guide.py

DanielRyanSmith:
Part of #2797.

Removes the double-writes to Feature entities in guide.py, as well as references in the test data page.

Edit: Also includes the removal of some unused Feature imports from other Python files.

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -27,7 +27,6 @@ from framework import permissions
 from internals import core_enums, notifier_helpers
 from internals import stage_helpers
 from internals.core_models import FeatureEntry, MilestoneSet, Stage
-from internals.legacy_models import Feature
 from internals.review_models import Gate
 from internals import processes
 from internals import search_fulltext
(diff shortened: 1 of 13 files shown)


pages/guide.py
```diff
@@ -98,15 +78,14 @@ def process_post_data(self, **kwargs):
         breaking_change=self.form.get('breaking_change') == 'on',
         blink_components=blink_components,
         tag_review_status=processes.initial_tag_review_status(feature_type))
-    feature_entry.put()
+    key: ndb.Key = feature_entry.put()
```
jrobbins, line 81: Why introduce a local variable for this result if it is not used?

jrobbins (approved):
This is great: simpler code and it will speed up form processing requests.
LGTM after considering one question.

pages/guide.py
DanielRyanSmith: This was how the key was referenced when creating the old Feature entity to reference in other places. This variable is used a bit further down, but it is not used in all the places it could be. I've refactored the code to use this variable anywhere it should be referenced.

PR #2871, opened 2023-03-28
Add a display name to trial and ship stages

### LINKED ISSUE #2646: Feature stages can have label names to differentiate multiples
In the current implementation of the feature detail page, a feature's stages will display with numbers next to the stage's name if there are multiples of the same stage type (e.g "Origin trial (2)" ). Without specific differentiation, this can become confusing. A new field can be added that allows users to give their own label to each stage in the case of multiples of the same stage existing.

Additional requirements:
- This field will only display to the user for stages that can appear in multiples (origin trial, shipping, etc.).
- This label will only appear on the feature detail page if there are multiples of the same stage type.
- The field description should convey to the user that this label is optional and will only display if multiples of the same stage type exist for the feature.

DanielRyanSmith:
Fixes #2646 

This change adds a new field to the Stage entity, `display_name`. This field will represent the name that should be displayed on headers/titles on the feature detail page and any stage edit pages. This field is mostly used to discern stages of the same type from each other, and so it is currently only editable for origin trial and shipping stages.

---
A new field will display for OT and shipping stages to add an optional display name to the feature detail page.
Screenshot 2023-03-27 at 7 35 41 PM

When a stage has a display name, it is displayed on the feature detail page, along with the process name in parentheses.
Screenshot 2023-03-27 at 7 34 37 PM

On the editall and verify accuracy pages, these display names will also be visible. Trial extension stages will be prefaced with a trial stage display name as well if it exists.
Screenshot 2023-03-27 at 7 43 09 PM


jrobbins:
This PR looks good, but I'd like a somewhat different UX:
* Let's always show the hard-coded stage type name first, followed by a colon and the display name (if any)
    * e.g., `Origin Trial: for revised API`
* To reduce user-visible complexity in the most common cases, let's only offer the field to edit the display name if there are multiple stages of that type on the feature

What do you think?

DanielRyanSmith:
@jrobbins This has now been updated with the changes you suggested above.
- The display name is now shown with the process name first, then the display name provided by the user, e.g. "Origin Trial: Updated trial based on feedback"
- The display name field is only available on stages where multiples of the same stage type exist on the feature.

jrobbins (approved):

PR #2876, opened 2023-03-28
Fix dedup check in set_vote.

### LINKED ISSUE #2873: "Intent to Ship: JSON.parse source text access" got 2 LGTMs, but counting as 3
Not sure how this happened, but when I went to approve in the tool (after emailing an LGTM2), I saw that Yoav had approved twice apparently - and it appears to accept this as 3 LGTMs.

Screenshot 2023-03-28 at 9 25 29 AM

jrobbins:
This should resolve #2873.

As discussed on chat today, our LGTM detection on blink-dev messages was not detecting existing votes.


DanielRyanSmith (approved):
Thanks for the investigation and fix 🙂

PR #2880, opened 2023-03-30
Remove old/duplicated fields from FeatureEntry and Stage entities

### LINKED ISSUE #2645: Update client to use the new feature field names
Currently, many of the client-side field variables are referenced with old field names that have since been renamed. The current implementation has these properties retroactively renamed back to their old names for client-side reference. It would be useful for code health and maintenance to address this and add continuity to our field names.

DanielRyanSmith:
Fixes #2645

This change removes the old field names from the FeatureEntry and Stage JSON representations and references them using their new field names.

**NOTE:** Stage fields that have different form descriptions depending on the stage type will continue to use a unique name for the field spec, but are referenced from the JSON using the new field name.
For example, `intent_thread_url` has a different form name and description depending on stage type (intent to implement, intent to experiment, etc.). The form field will still be represented with the old unique name, but the stage JSON will contain only the field `intent_thread_url`.

---
#### Old milestone representations
-  dt_milestone_desktop_start
-  dt_milestone_android_start
-  dt_milestone_ios_start
-  dt_milestone_webview_start
-  ot_milestone_desktop_start
-  ot_milestone_android_start
-  ot_milestone_webview_start
-  ot_milestone_desktop_end
-  ot_milestone_android_end
-  ot_milestone_webview_end:

#### Fields moved to Stage kind
- intent_to_implement_url
- intent_to_experiment_url
- intent_to_extend_experiment_url
- intent_to_ship_url
- ready_for_trial_url
- origin_trial_feeback_url
- experiment_goals
- experiment_risks
- announcement_url
- experiment_extension_reason
- rollout_details
- rollout_impact
- rollout_milestone
-  intent_to_implement_url
-  intent_to_experiment_url:
-  intent_to_extend_experiment_url
-  intent_to_ship_url
-  ready_for_trial_url
- finch_url

---
#### Fields with their original Feature names
-  intent_to_implement_url:
-  intent_to_experiment_url: str  | None
-  intent_to_extend_experiment_url: str | None
-  intent_to_ship_url: str | None
 - dt_milestone_desktop_start
 - dt_milestone_android_start
 - dt_milestone_ios_start
 - dt_milestone_webview_start
 - ot_milestone_desktop_start
 - ot_milestone_android_start
 - ot_milestone_webview_start
 - ot_milestone_desktop_end
 - ot_milestone_android_end
 - ot_milestone_webview_end
 - shipped_milestone
 - shipped_android_milestone
 - shipped_ios_milestone
 - shipped_webview_milestone

---
Changes by file:
- api/converters.py: Remove the above fields from the FeatureEntry and Stage JSON representation
- client-src/elements/utils.js: Add a new helper function that obtains a form field value from the correct JSON field of a stage (e.g. 'ot_milestone_desktop_start' form field is populated with 'intent_stage_url' stage field).
- internals/data_types.py: Remove references to old fields from the JSON data types.
- client-src/elements/form-field-specs.js: Add a name value for intent fields to ensure the intent value is pulled correctly from the stage entity.
- client-src/elements/form-field-enums.js: Add a mapping of the unique form field names to their generic field name that exists on the stage entity.


DanielRyanSmith:
@jrobbins, I've tried to find all cases where we might use these old fields and it seems like all of the client-side uses are taken care of (unless I missed something). There is some risk here if some other app is pulling this feature data from out API, but this change will only affect the API calls that are querying for a specific feature, not a list of features (which uses the converter function `feature_entry_to_json_basic()`.

DanielRyanSmith:
> What about existing versions of the SPA that are already loaded into a user's browser tab?

That should not be a risk because we are not adding any new fields to the entities - only removing existing ones and changing how we reference them on the client side. A lot of the fields were duplicated to allow for referencing them the old way and the new way - this change mostly removes all of that duplication as if we're already on step 2 that you mentioned.

jrobbins (approved):

PR #2881, opened 2023-03-30
Update enterprise help texts

yanndago:
The also allows us to have two different help texts , one for enterprise features and one for regular features.


jrobbins (approved):

dlaliberte:
Wish I had time to work on the conditional help text mechanism before this.  Rather than adding more code tied to specific variations of the app, the idea is that we could instead declare that certain conditions, such as being an "enterprise feature" to corresponding help text.  But we can refactor this code later, once we have such a capability.

Issue #2872, opened 2023-03-28
Phase out guide pages and use Feature/Stage API for updates


DanielRyanSmith:
Feature and Stage edits are processed using the `guide.py` logic rather than sending changes to one of our APIs. This file is getting overly complex and hard to manage, and is not the way we handle most user requests.

- [ ] Build out the Features API and Stages API to handle all use cases for creating/updating these entities.
- [ ] Phase out the guide.py implementation and make all changes through the new API implementation.

PR #2882, opened 2023-03-30
Implement PATCH request for Features API

DanielRyanSmith:
Part of #2872

This change adds PATCH request support to the Features API, which allows specific fields to be edited on FeatureEntry entities. This will not currently be used by any client aspects, but will be a necessary change for phasing out or old approach to feature edits (guide.py).
This request will only make changes if:
- the user has edit permission for the feature.
- the feature exists.
- the fields provided can be mutated by normal users.

List of changes by file:
- api/features_api.py: Add a new PATCH request function. This request will require providing a list of fields that should be changed on the FeatureEntry entity, as well as their values.
-  internals/core_models.py: Adds a new list of fields that normal users should not have access to mutate.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, TypedDict
+from datetime import datetime
 
 from api import converters
 from framework import basehandlers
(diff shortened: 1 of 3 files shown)


internals/core_models.py
```diff
@@ -29,6 +29,17 @@
...
+  # Fields that should not be mutated by user edit requests.
+  FIELDS_IMMUTABLE_BY_USER = frozenset([
+    'id',
+    'created',
+    'creator_email',
+    'outstanding_notifications',
+    'deleted',
+    'star_count',
+    'feature_type',
+  ])
```
DanielRyanSmith, line 41: It might be worth double-checking that these are the only fields we would like to protect from users with traditional edit access (non-admins).

jrobbins: accurate_as_of?  We might want the API-centric version of the accurate_as_of confirmation page to be able to set it, but it should be the server that determines the value rather than something passed in.

updated and updater_email.  These will be overwritten by your code, but it would still be better to block them in case of future code changes and because blocking them will give an error message to clients that mistakenly try to set these fields.

api/features_api.py
```diff
@@ -84,6 +84,36 @@ def do_get(self, **kwargs):
   # TODO(jrobbins): do_post
 
   # TODO(jrobbins): do_patch
```
jrobbins, line 86: You can delete this old TODO comment.

api/features_api.py
```diff
@@ -84,6 +84,36 @@ def do_get(self, **kwargs):
...
+    redirect_resp = permissions.validate_feature_edit_permission(
+        self, feature_id)
+    if redirect_resp:
+      return redirect_resp
+
+    # Update each field specified in the field mask.
+    for field in fields_to_update:
+      # Each field specified should be a valid field that exists on the entity.
+      if not hasattr(feature, field):
+        self.abort(500, msg=f'FeatureEntry has no attribute {field}')
```
jrobbins, line 108: 400

api/features_api.py
```diff
@@ -84,6 +84,36 @@ def do_get(self, **kwargs):
...
+      if field in FeatureEntry.FIELDS_IMMUTABLE_BY_USER:
+        self.abort(500, f'FeatureEntry field {field} is immutable.')
```
jrobbins, line 110: 400

jrobbins (approved):
LGTM after considering some simple change suggestions.

PR #2893, opened 2023-03-31
Bug fix: enums display as their expanded versions on feature detail page

DanielRyanSmith:
This change updates the logic that previously would not expand enum values for the feature detail page if the int value could be referenced successfully from the FeatureEntry JSON.

~~Additionally, this contains a fix to normalize the `web_views` field text to display as plural, which correlates with the text for `ff_views` and `safari_views`.~~ Edit: Not making this change, as it might be intended.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -337,7 +337,7 @@ class ChromedashFeatureDetail extends LitElement {
       web_dev_views_notes: 'browsers.webdev.view.notes',
       other_views_notes: 'browsers.other.view.notes',
     };
-    if (!value && fieldNameMapping[fieldName]) {
+    if (fieldNameMapping[fieldName]) {
       value = this.feature;
       for (const step of fieldNameMapping[fieldName].split('.')) {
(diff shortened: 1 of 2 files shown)


KyleJu:
@DanielRyanSmith python_tests is failing

DanielRyanSmith:
Removed the small text change as it's not really part of this problem. Tests should pass now (verified running the presubmit).

KyleJu:
This change certainly doesn't hurt, but does it mean that `this.feature[fieldName];` isn't empty when `fieldName` is an enum field?

DanielRyanSmith:
> This change certainly doesn't hurt, but does it mean that this.feature[fieldName]; isn't empty when fieldName is an enum field?

Yes, we have both the enum int value representation and the text value in separate properties of the JSON representation for FeatureEntry. The current implementation grabs the int value from the JSON and then does not grab the text value if the int value existed as a property. This change makes it so that if any fields are explicitly defined by the field mapping above, then it will pull the field from there. (I hope I explained that in a way that makes sense 😅)

KyleJu (approved):

PR #2910, opened 2023-04-11
Remove the rest of old schema references

DanielRyanSmith:
This change removes any existing references in the codebase to legacy models, as well as the unused `/admin/subscribers` page, which was the last component to still use legacy models.


internals/schema_migration.py
```diff
@@ -1,175 +0,0 @@
...
-
-from framework.basehandlers import FlaskHandler
-from internals import approval_defs
-from internals.core_models import FeatureEntry, Stage
-from internals.legacy_models import Feature
-from internals.review_models import Gate, Vote
-from internals.core_enums import *
-
-
-class EvaluateGateStatus(FlaskHandler):
```
jrobbins, line 26: We might still need this if we ever detect that the gate status has gotten out of sync with the votes on that gate.  Maybe keep it as-is or rename the file to something like "batch maintenance".

internals/schema_migration.py
```diff
@@ -1,175 +0,0 @@
...
-        count += 1
-        if len(batch) > BATCH_SIZE:
-          ndb.put_multi(batch)
-          batch = []
-
-    ndb.put_multi(batch)
-    return f'{count} Gate entities updated.'
-
-
-class WriteMissingGates(FlaskHandler):
```
jrobbins, line 48: We will need this when we add more gate types to backfill data on existing feature entries.

main.py
```diff
@@ -233,13 +230,6 @@ class Route:
   Route('/tasks/detect-intent', detect_intent.IntentEmailHandler),
   Route('/tasks/email-reviewers', notifier.FeatureReviewHandler),
   Route('/tasks/email-comments', notifier.FeatureCommentHandler),
-
-  Route('/admin/schema_migration_gate_status',
```
jrobbins, line 237: We'll need the gate-related routes.

jrobbins (changes requested):

internals/schema_migration.py
DanielRyanSmith: I agree, and I definitely think we'll need to write scripts like this in the future. I've renamed the file to `maintenance_scripts.py` and changed the paths to reflect this in `main.py`. If you would prefer a different naming convention, I can update them!

jrobbins (approved):

PR #2912, opened 2023-04-12
Align the Firefox views field with Mozilla's possible views.

### LINKED ISSUE #2909: Align Mozilla signal options with actual Mozilla signals
Chrome Status's options for Firefox signals include all of:

* Shipped/Shipping
* In development
* Positive
* No signal
* Negative
* Neutral
* N/A
* Under consideration
* Important
* Worth prototyping
* Non-harmful
* Defer
* Harmful

[Mozilla's standards positions repository only lists](https://github.com/mozilla/standards-positions#possible-specification-positions):

* positive
* neutral
* negative
* defer
* under consideration 

In https://github.com/mozilla/standards-positions/pull/680, @martinthomson mapped "important" and "worth prototyping" to "positive", "non-harmful" to "neutral", and "harmful" to "negative".

We should align with those changes, and stop offering developers so many options to pick from that they can miss the right option.

I propose that we put the remaining options in the following order:

* No signal
* N/A
* Under consideration
* Defer
* Positive
* Negative
* Neutral
* In development
* Shipped/Shipping

I'm not certain we should include "In development"; it probably depends on whether the API owners consider that a replacement for asking explicitly.

The code for this is at https://github.com/GoogleChrome/chromium-dashboard/blob/4dda7c5c37f13f44e3d1eb1c99d484fc4231d66a/client-src/elements/form-field-enums.js#L269-L287

jyasskin:
This should fix #2909. Someone with write access to the database should bulk-update the deprecated values to their replacements before you merge this.

### DIFF
--- a/client-src/elements/form-field-enums.js
+++ b/client-src/elements/form-field-enums.js
@@ -260,11 +260,11 @@ export const REVIEW_STATUS_CHOICES = {
 // NEUTRAL = 8
 // SIGNALS_NA = 9
 // GECKO_UNDER_CONSIDERATION = 10
-// GECKO_IMPORTANT = 11
-// GECKO_WORTH_PROTOTYPING = 12
-// GECKO_NONHARMFUL = 13
+// GECKO_IMPORTANT = 11 # Deprecated; use PUBLIC_SUPPORT
+// GECKO_WORTH_PROTOTYPING = 12 # Deprecated; use PUBLIC_SUPPORT
+// GECKO_NONHARMFUL = 13 # Deprecated; use NEUTRAL
 // GECKO_DEFER = 14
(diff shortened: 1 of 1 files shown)


jyasskin:
I started a [blink-api-owners-discuss thread](https://groups.google.com/a/chromium.org/g/blink-api-owners-discuss/c/Hh9h7oHWXIA/m/4qxhhVGZCAAJ) to make sure the API owners don't have any problems with this change.

jrobbins:
The intent of this PR looks fine.  I'll write a separate PR to bulk-update existing data.

I'll need to check if we need more steps to roll this out given that users might have an existing browser tab open running our current SPA version.

jrobbins (approved):

PR #2947, opened 2023-04-27
Set detected intent_thread_url even if it was an empty string.

jrobbins:
Fix a problem that Yoav spotted: recent features with requested reviews don't have the intent_thread_url filled in.

I believe that the cause was that when users edit the fields for that stage, the intent_thread_url is set to an empty string when they leave that field blank.  That is fine.  But, in detect_intent the code would only set intent_thread_url if it was previously None.

In this PR:
* Change the condition to look for a value that is falsey, allowing either `None` or `''`.
* Add a unit test
* Add a bunch of logging as an example of doing more logging.
* Split the originally condition so that we can log two cases separately.


DanielRyanSmith (approved):
Thanks for the quick fix!

PR #2952, opened 2023-05-01
Make the new comments page also show activity.

jrobbins:
This keeps the list of overall activities on a feature entry visible.


jrobbins:
Screenshot:
image

PR #2967, opened 2023-05-02
Validate search parameter types.

jrobbins:
This improves our handling of bad requests, such as an attempted SQL-injection attack.

In this PR:
* Check if the query value would be accepted as a field value for that field, if not, raise a ValueError
* Improve the text of the 400 response when such a request is rejected
* Add unit tests

### DIFF
--- a/internals/search_queries.py
+++ b/internals/search_queries.py
@@ -62,6 +62,14 @@ def single_field_query_async(
     return []
 
   # TODO(jrobbins): Handle ":" operator as substrings for text fields.
+
+  try:
+    # If the field can be set to val, then it can be compared to val.
+    field._validate(val)
+  except ndb.exceptions.BadValueError:
+    logging.info('Wrong type of value for %r: %r' % (field, val))
+    raise ValueError('Query value does not match field type')
+
   if (operator == '='):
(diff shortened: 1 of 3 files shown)


internals/search_queries.py
```diff
@@ -62,6 +62,14 @@ def single_field_query_async(
     return []
 
   # TODO(jrobbins): Handle ":" operator as substrings for text fields.
+
+  try:
+    # If the field can be set to val, then it can be compared to val.
+    field._validate(val)
```
KyleJu, line 68: I believe the validation uses **isinstance** to validate. It would not work if val is a list of integer, e.g. '5,6,7,8' where field is **Integer** and val is **String**. It will run **query.filter(field.IN(val_list))**

jrobbins: True, but I don't think that was previously supported either.  We only supported the quick-OR syntax with string fields.
I have added a TODO to add support for quick-OR on integer fields.

KyleJu: Yea I realized that it was the case. Thanks for adding a comment there!

KyleJu (approved):

PR #2972, opened 2023-05-03
Revert "npm: bump @babel/preset-env from 7.21.4 to 7.21.5"

jrobbins:
Reverts GoogleChrome/chromium-dashboard#2954
This may have caused #2957 to fail.


jrobbins:
I'll test more locally before reverting and re-landing so many PRs.

PR #2973, opened 2023-05-03
Reorganize feature page to clearify stage sections.

### LINKED ISSUE #2969: The Development Stages header shouldn't be above Metadata
The Development stages header is currently right above the Metadata section which is not a stage itself. We should ensure that the Development stages header comes right before the actual stages sections.

image

jrobbins:
This should resolve #2969.

In this PR:
* Define a set of CSS styles that is shared between chromedash-feature-page and chromedash-feature-detail.
* Move the metadata section just above the heading "Development stages".
* Add a "stage" CSS class and change the CSS selector used find the stage accordion sections to look for that class.  This makes the expand-all / collapse-all link work for just the stage sections.
* Create a new "History" heading in the overview box:
    * add an entry created date
    * move the link to the comments & activity page here

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -28,6 +28,29 @@ import './chromedash-gate-chip';
 import {autolink, findProcessStage, flattenSections} from './utils.js';
 import {SHARED_STYLES} from '../sass/shared-css.js';
 
+export const DETAILS_STYLES = [css`
+      sl-details {
+        border: var(--card-border);
+        box-shadow: var(--card-box-shadow);
+        margin: var(--content-padding-half);
+        border-radius: 4px;
+        background: var(--card-background);
+      }
+      sl-details::part(base),
+      sl-details::part(header) {
+        background: transparent;
+      }
+      sl-details::part(header) {
+        padding-bottom: 8px;
+      }
+
+      .card {
+        background: var(--card-background);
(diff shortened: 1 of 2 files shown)


jrobbins:
Screenshot showing changes
image

client-src/elements/chromedash-feature-detail.js
```diff
@@ -466,7 +467,7 @@ class ChromedashFeatureDetail extends LitElement {
         @sl-after-show=${this.updateCollapsed}
         @sl-after-hide=${this.updateCollapsed}
         ?open=${isActive || defaultOpen}
-        class=${isActive ? 'active' : ''}
+        class="${isActive ? 'active' : ''} ${isStage ? 'stage' : ''}"
```
KyleJu, line 470: How would it look like if it is both active and stage? class="activestage"?

KyleJu:
> Screenshot showing changes image

Could perhaps add a bit vertical space between "all comments & activity" and  "last updated..."

client-src/elements/chromedash-feature-detail.js
jrobbins: It has whitespace between the expressions, so it becomes `class="active stage"` which works.  That also means that that space is always there, but it does hurt anything.

jrobbins:
I added some margin between the paragraphs.

KyleJu (approved):

PR #2983, opened 2023-05-05
Update enterprise release notes UI

yanndago:
* Remove list style from feature items
* Add the owners and last updated date to features, making sure to mark the as items to remove


jrobbins (approved):
Just so I understand, the `<To remove ... >` text is stuff that is supposed to be edited out by the person who is copying and pasting release notes from chromestatus into a Google doc, right?

PR #2984, opened 2023-05-05
Show "Pending my review" box to users who can approve any gate.

jrobbins:
This should resolve a problem reported yesterday by bheenan, as well as taking care of a pending TODO comment.

Previously the only users doing approval were the API Owners, and the permission check to display the "Features pending my review" box on the My Features page checked specifically for ability to approve the API Owners gate.  That was wrong because it leaves out approvers on other teams.  I had written a TODO comment to check for specific approval permissions, but not resolved that TODO until now.

In this PR:
* Change chromedash-myfeatures-page to check `this.user.approvable_gate_types`.
    * Also change the approvals dialog, even though it will soon be deleted.
* Remove all other client-side references to `can_approve`
    * Most of them were passing info down to feature-row so that the stamp icon could be offered, but that icon was previously removed, so the canApprove property was no longer used.

### DIFF
--- a/api/permissions_api.py
+++ b/api/permissions_api.py
@@ -36,7 +36,7 @@ class PermissionsAPI(basehandlers.APIHandler):
       approvers = approval_defs.get_approvers(field_id)
       user_data = {
         'can_create_feature': permissions.can_create_feature(user),
-        # TODO(jrobbins): phase out can_approve and use approvable_gate_types
+        # TODO(jrobbins): delete unused can_approve in an upcoming release.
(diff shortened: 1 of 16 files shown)


KyleJu (approved):

jstenback:
Would a re-wording to "My pending reviews" be better than the current "Pending my review" make sense while we're at it? To my brain that's easier to parse, but that might just be me...

jstenback:
All good, and that full wording makes a lot more sense. Thanks!

Issue #2384, opened 2022-10-24
Phase out sass


jcscottiii:
[From](https://github.com/GoogleChrome/chromium-dashboard/issues/2304#issuecomment-1287248445) @jrobbins 

> I'd also like to phase out our use of SASS, because we don't use much of it. We've already converted almost all of it to lit-css. I'll leave it up to you if you want to upgrade SASS as part of this work or phase it out to avoid the need to upgrade.

Left over work from #2304

jrobbins:
This was done in #3970.

Issue #900, opened 2020-07-23
Re-write remaining SASS code as plain CSS


jrobbins:
We currently have about 12 SASS files that style individual pages.  We don't make use of SASS functionality other than things that could be easily done in modern CSS.  So, we will rewrite these in plain CSS and phase out SASS.

alokkumarak:
hello sir,
please assign this to me.
I can convert all your SASS file to CSS.

PR #2998, opened 2023-05-10
Stop using scss compiler, and use css files directly.

dlaliberte:
This PR only disables use of the "clean" and "styles" tasks of gulp, to avoid using the scss compiler, and use the css files directly.

Related to #2384 and #900

The remaining static/css files still being used, mostly in tests, are:
base.css, main.css, forms.css, features/features.css, and features/launch.css.


jrobbins:
> I'm not sure why `client-src/elements/chromedash-enterprise-release-notes-page_test.js` needs to be updated.

That code was originally merged without semicolons at the ends of those lines.   When we run the server, it does a JS lint command first which automatically adds them.   I already merged in a PR that also added them, so if you rebase you should find that you get the semicolons without that diff in your PR.


> Also, client-src/elements/chromedash-activity-page_test.js is failing with:

Yeah, that is an existing problem.

jrobbins (approved):

PR #3001, opened 2023-05-10
Restrict access to deleted feature entries.

jrobbins:
Today I was asked to delete an unneeded feature entry, which I did.  But then I realized that we still serve the feature detail page and guide page for deleted features with not indication that they are marked deleted.    I do want to continue to serve these pages to users who can edit the feature, so that it could be undeleted in the future.

In this PR:
* The GET handler for the features_api returns 404 when the feature has been marked deleted and the user does not have permission to edit it.  This will cause the client-side to show "Feature not found".
* For users who still see these pages for deleted features, add a warning banner to top.  And, move the existing unlisted warning to the top.
* Add a shared style for warnings.

Note that the code for renderWarnings() appears in two places, but we expect to delete the guide metadata page fairly soon.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -117,7 +117,7 @@ class FeaturesAPI(basehandlers.APIHandler):
     # Only fields with defined data types are allowed.
     if field not in self.FIELD_DATA_TYPES_CREATE:
       self.abort(400, msg=f'Field "{field}" data format not found.')
-    
+
     field_type = self.FIELD_DATA_TYPES_CREATE[field]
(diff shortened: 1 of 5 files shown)


jrobbins:
Screenshots

Unlisted feature detail page
image

Unlisted feature guide edit page
image

Deleted feature detail page
image

Deleted feature guide edit page
image

Non-deleted, non-unlisted feature pages don't display any warning
image

KyleJu (approved):

PR #3013, opened 2023-05-16
Update PATCH request handler to make stage entity changes

DanielRyanSmith:
This change has NO effect on the current UX behavior and functionality. Its purpose is to lay the framework for handling API calls more effectively and removing the current handling in guide.py.

This change updates the existing (unused) PATCH request handler to accept both feature and stage entity changes in a single request. The request format as assumed to be the format defined in the `formatFeatureChanges` helper function in `utils.js`. Some small renaming changes have been made to this file.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -23,84 +23,120 @@ from framework import permissions
 from framework import rediscache
 from framework import users
 from internals.core_enums import *
-from internals.core_models import FeatureEntry, Stage
+from internals.core_models import FeatureEntry, MilestoneSet, Stage
 from internals.review_models import Gate
 from internals.data_types import VerboseFeatureDict
 from internals import feature_helpers
 from internals import search
 import settings
 
+
+# Data type for lists defining field data type information.
+FIELD_INFO_DATA_TYPE = list[tuple[str, str]]
+
+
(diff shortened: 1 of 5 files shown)


api/features_api.py
```diff
@@ -23,84 +23,120 @@
...
+
+# Data type for lists defining field data type information.
+FIELD_INFO_DATA_TYPE = list[tuple[str, str]]
+
+
 class FeaturesAPI(basehandlers.APIHandler):
   """Features are the the main records that we track."""
 
+  ############################
+  ###### Mutable Fields ######
```
jrobbins, line 42: I love ASCII art, but if these kind of banners are seem needed, then maybe we should be doing something else to organize and clarify the code instead.   E.g., maybe this big constant should go into a server-side python version of form-field-specs.js.  That might also help with #3016.

jrobbins (changes requested):

api/features_api.py
DanielRyanSmith: I agree that it would make sense to move the constants over to a specs file, and made the change accordingly. My hope was to separate the relevant methods used for each request with the big comments, but it's overall unnecessary, so I removed them. 🙂

jrobbins (approved):

PR #3029, opened 2023-05-21
Add hide obsolete support

### LINKED ISSUE #2565: Hide OBSOLETE_* UseCounters by default, but offer a checkbox to show them
Our metrics pages show a bunch of UseCounters that have names starting with the word "OBSOLETE_".  Having these shown looks like an error because users would not normally want to be looking at obsolete data. However the technical meaning of that word in the UseCounter name is just that no new data is being collected.

Hide OBSOLETE_ UseCounters from the stack rank page by default, but offer the ability to show them, e.g., via a checkbox that says "Include obsolete".

**Additional context**
This was discussed in our team meeting today.

pingren:
Close #2565

CleanShot 2023-05-21 at 12 15 31@2x

- Hide property entries using `display: none`. Rather than mutating `viewList` or `tempList` with sorting logic. CSS is easier to implement and performance wise.
- Update total counts accordingly.

### DIFF
--- a/client-src/elements/chromedash-stack-rank.js
+++ b/client-src/elements/chromedash-stack-rank.js
@@ -14,6 +14,7 @@ class ChromedashStackRank extends LitElement {
       sortType: {type: String},
       sortReverse: {type: Boolean},
       tempList: {attribute: false},
+      shouldHideObsolete: {state: true, attribute: false},
     };
   }
 
@@ -25,6 +26,8 @@ class ChromedashStackRank extends LitElement {
     this.maxPercentage = 100;
     this.sortType = 'percentage';
     this.sortReverse = false;
+    this.shouldHideObsolete = true;
+    this.obsoleteCount = 0;
   }
 
 
@@ -80,6 +83,10 @@ class ChromedashStackRank extends LitElement {
 
       .stack-rank-item {
         border-top: var(--table-divider)
+      }
+
+      .stack-rank-item-hidden {
+        display: none;
       }
 
       .stack-rank-item-name {
@@ -151,6 +158,8 @@ class ChromedashStackRank extends LitElement {
     setTimeout(() => {
       this.scrollToPosition();
     }, 300);
+
+    this.obsoleteCount = this.viewList.filter((item) => item.obsolete).length;
   }
 
   scrollToPosition(e) {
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-stack-rank.js
```diff
@@ -178,10 +187,19 @@ class ChromedashStackRank extends LitElement {
...
+  }
+
   renderSubHeader() {
     return html`
       <div id="subheader">
-        <p class="title-text">Showing <span>${this.viewList.length}</span> properties</p>
+        <p class="title-text">Showing <span>${this.viewList.length - (
+          this.shouldHideObsolete ? this.obsoleteCount : 0
+        )}</span> properties</p>
+        <sl-checkbox ?checked=${this.shouldHideObsolete} @input=${this.handleChangeHideObsolete} >
```
jrobbins, line 200: Why `@input` rather than `@sl-change`?

client-src/elements/chromedash-stack-rank.js
```diff
@@ -219,7 +237,8 @@ class ChromedashStackRank extends LitElement {
   renderStackRank(displayedList) {
     return html`
       ${displayedList.map((item) => html`
-        <li class="stack-rank-item" id="${item.property_name}">
+        <li class="stack-rank-item${(this.shouldHideObsolete && item.obsolete) ? ' stack-rank-item-hidden' : ''}"
```
jrobbins, line 240: I think it would be OK for the class attribute to have a trailing space when the item is not hidden.  That would allow you to write the code more cleanly with a space before the `$` and no leading space inside the string constant `'stack-rank-item-hidden'`.

jrobbins (commented):
This PR looks good.  After we talk later today I will approve it.  I just have two bits of feedback.  Also, you will need to work through the steps for the CLA (contributor licensing agreement).

client-src/elements/chromedash-stack-rank.js
pingren: I found `@input` here:

https://github.com/GoogleChrome/chromium-dashboard/blob/ae58989d1e1e16e42e18501df7776466dc1c6db8/client-src/elements/chromedash-settings-page.js#L102

I was wondering why `@input` works. I checked
https://github.com/shoelace-style/shoelace/blob/next/src/components/checkbox/checkbox.ts#L106 
only emits events starts with `sl-`.

`@sl-change` is documented but for now I leave it as it is.

jrobbins (approved):

PR #3044, opened 2023-05-26
Add link helper

pingren:
In this PR: 

- `link_helper.Link` fetches info on links found in the feature entry
  - Currently supports only [Chromium issue tracker url](https://bugs.chromium.org/p/chromium/issues/detail?id=1000000)
- `FeatureLink` indexes all links in features in database
  - When a user changes fields of a feature, then indexing should update accordingly
    - If a link is new, its information is parsed and updated immediately
    - If a link already exists, only other fields are updated.
    - If a link is not used in any fields of a feature, the corresponding feature should be removed from `FeatrueLink.feature_ids`.
    - If a link is not used by any features (`FeatrueLink.feature_ids` is empty), then the feature link should be removed.


internals/link_helpers.py
```diff
@@ -0,0 +1,86 @@
...
+    """Parse the information from the chromium bug tracker."""
+
+    endpoint = 'https://bugs.chromium.org/prpc/monorail.Issues/GetIssue'
+
+    issue_id = self.url.split('id=')[-1]
+
+    # csrf token is required, its expiration is about 2 hours according to the tokenExpiresSec field
+    # technically, we could cache the csrf token and reuse it for 2 hours
+
+    csrf_token = re.findall(
```
jrobbins, line 44: I think this would only work for a request from a user that is currently signed in to bugs.chromium.org.   When our python server makes such a request, there is no signed in user, so instead it will generate a redirect to a login page. 

 You could try the same trick with a request to `"https://bugs.chromium.org/"`.

It would be better for us to register as a monorail API client using oauth.  But we can leave that as a TODO comment for now.

internals/link.py
```diff
@@ -0,0 +1,93 @@
...
+        type=link.type,
+        url=link.url,
+        information=link.information
+    )
+    logging.info(f'Created link {link.url}')
+  logging.info(f'Indexed link {link.url} for feature {feature_id}')
+  feature_link.put()
+
+
+def _remove_link(link: Link, fe: 'FeatureEntry') -> None:
```
jrobbins, line 67: I don't think that you need the single quotes around `'FeatureEntry'` because that class has been imported into this file.

internals/link.py
```diff
@@ -0,0 +1,93 @@
+import logging
+from typing import Any
+from internals.core_models import FeatureEntry
+from internals.link_helpers import Link
+
+from google.cloud import ndb  # type: ignore
+
+
+class FeatureLinks(ndb.Model):
```
jrobbins, line 9: It would be good to add a comment here to explain how and when the `FeatureLinks` entities are used.   E.g., "This information is used to display a preview of information from the linked page, which often saves the user from needing to click.".

internals/link_helpers.py
pingren: @jrobbins 

It will get valid `csrf_token` even it is not signed in, `requests.get` will not follow redirection by default.

Prove:
- CleanShot 2023-05-29 at 12 16 29

- If token is not valid, `test_parse_chromium_tracker` won't pass.

I notices some issues requires permission to view.

We can leave a TODO to use an API client.

internals/link_helpers.py
```diff
@@ -0,0 +1,84 @@
+import re
+import requests
+import json
+import logging
+LINK_TYPE_CHROMIUM_BUG = 'chromium_bug'
+LINK_TYPE_UNKNOWN = 'unknown'
+LINK_TYPES_REGEX = {
+    # https://bugs.chromium.org/p/chromium/issues/detail?id=
+    LINK_TYPE_CHROMIUM_BUG: re.compile(r'https://bugs.chromium.org/p/chromium/issues/detail?.*'),
```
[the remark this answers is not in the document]
pingren: Fixed

pingren:
@jrobbins Thanks for reviewing. 
I updated the PR based on your comments. Let me know if we could merge it. This PR won't affect production database since I didn't commit changes in `guide.py`.

CleanShot 2023-05-29 at 13 22 55

internals/link_helpers.py
```diff
@@ -0,0 +1,87 @@
...
+LINK_TYPE_UNKNOWN = 'unknown'
+LINK_TYPES_REGEX = {
+    # https://bugs.chromium.org/p/chromium/issues/detail?id=
+    LINK_TYPE_CHROMIUM_BUG: re.compile(r'https://bugs\.chromium\.org/p/chromium/issues/detail\?.*'),
+    # any other links
+    LINK_TYPE_UNKNOWN: re.compile(r'https?://.*'),
+}
+
+
+class Link():
```
jrobbins, line 17: It's potentially confusing to have `class Link` defined in a file that is not `link.py`.
Maybe you should rename `link.py` to `feature_links.py`.

Also, I am not sure that you need class `Link`.    I think you could make the methods of `Link` into top-level functions in this file and have them return tuples rather than constructing a `Link` object.  That might be easier to understand than using the current names `FeatureLinks` and `Link`, which are logically similar to each other.

internals/link.py
```diff
@@ -0,0 +1,95 @@
+import logging
```
jrobbins, line 1: Each of these files should have the copyright header comments.
You can copy them from any of the other python files.
Just make sure to edit the year number to be 2023 in these new files.

jrobbins (changes requested):

internals/link.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _remove_link(link: Link, fe: FeatureEntry) -> None:
+  """Un-index the given link."""
+  feature_id = fe.key.integer_id()
+  feature_links = FeatureLinks.query(FeatureLinks.url == link.url).fetch(None)
+  feature_link: FeatureLinks = feature_links[0] if feature_links else None
+  if feature_id in feature_link.feature_ids:
+    feature_link.feature_ids.remove(feature_id)
+    feature_link.put()
```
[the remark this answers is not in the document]
jrobbins: It would look like:
```
  if feature_id in feature_link.feature_ids:
     feature_link.feature_ids.remove(feature_id)
     if not feature_link.feature_ids:
       # delete the link if it is not used by any feature
       feature_link.key.delete()
       logging.info(f'Delete indexed link {link.url}')
    else:
       feature_link.put()
       logging.info(f'Updated indexed link {link.url}')
```

Or even better, reverse the two clauses:
```
  if feature_id in feature_link.feature_ids:
     feature_link.feature_ids.remove(feature_id)
     if feature_link.feature_ids:
       feature_link.put()
       logging.info(f'Updated indexed link {link.url}')
    else:
       # delete the link if it is not used by any feature
       feature_link.key.delete()
       logging.info(f'Delete indexed link {link.url}')
```

pingren:
@jrobbins PR updated and all comments are addressed, ready to merge 💯

jrobbins (approved):
You project is off to a good start!

jrobbins:
I went ahead and merged this change for you, but for future PRs you are welcome to merge the change after it has been approved.  You should squash the PR so that it creates only a single change on the main branch, and then delete the branch after doing that (it can always be restored).

Issue #3022, opened 2023-05-19
Implement server-side changes for representing SLOs


jrobbins:
Implement:
* [x] New fields in the Gate entity
* [x] Convert those fields to JSON in converters.py.
* [x] Set those fields in response to votes.
* [x] Calculate the number of business days that a review has been pending.
* [x] Define optional specific SLOs for each gate type.
* [ ] Backfill `responded_on` for previously reviewed gates.

jrobbins:
Most of this has been done.  The last step is the backfill.

jrobbins:
This was done last year.

PR #3062, opened 2023-06-08
Implement server-side changes for review SLOs

jrobbins:
This should resolve issue #3022.  This is just server-side logic to calculate SLOs and start storing response dates.  There is no user-visible effect yet.

In this PR:
* Add gate.responded_on date field and add logic to set it.
* Implement functions to count week days and compare to SLO limits
* Include SLO info in the Gate API response JSON.


internals/slo.py
```diff
@@ -0,0 +1,99 @@
...
+
+def weekdays_between(start: datetime.datetime, end: datetime.datetime) -> int:
+  """Return the number of Pacific timezone weekdays between two UTC dates."""
+  d_ptz = start.astimezone(PACIFIC_TZ)
+  # The day of the request does not count.
+  d_ptz = d_ptz.replace(hour=23, minute=59, second=59)
+  end_ptz = end.astimezone(tz=PACIFIC_TZ)
+  weekday_counter = 0
+  while d_ptz < end_ptz and weekday_counter < MAX_DAYS:
+    d_ptz = d_ptz + datetime.timedelta(days=1)
```
KyleJu, line 41: if d_ptz is a Friday and end_ptz is saturday, I suspect this function will return  0. Should be move `d_ptz = d_ptz + datetime.timedelta(days=1)`  after the `is_weekday(d_ptz)` check?

internals/slo_test.py
```diff
@@ -0,0 +1,230 @@
...
+  def test_weekdays_between__same_day(self):
+    """There are zero days between different times on the same day."""
+    start = datetime.datetime(2023, 6, 7, 12, 30, 0)
+    end = datetime.datetime(2023, 6, 7, 14, 15, 10)
+    actual = slo.weekdays_between(start, end)
+    self.assertEqual(0, actual)
+
+  def test_weekdays_between__normal(self):
+    """We can count the weekdays between dates."""
+    start = datetime.datetime(2023, 6, 7, 12, 30, 0)
```
KyleJu, line 48: Could you add a  Friday `start` test case?

internals/slo.py
```diff
@@ -0,0 +1,99 @@
...
+      gate.responded_on = recent_vote_time
+      return True
+
+  return False
+
+
+def record_comment(
+    feature: FeatureEntry, gate: Gate, user: User,
+    approvers: list[str]) -> bool:
+  """Record Gate SLO response time if needed.  Return True if changed."""
```
KyleJu, line 87:
```suggestion
  """Record Gate SLO response time if needed. Return True if changed."""
```

KyleJu (approved):
LGMT with two comments to address

internals/slo.py
jrobbins: I think returning 0 is correct: the day of the request does not count, and Saturday is a weekend, so neither one of them should increase the count.

Issue #3024, opened 2023-05-19
Display SLO info at the top of the gate column


jrobbins:
On the gate column UI, if there is a review pending, then display an icon and message about the SLO status.

E.g., if a review was requested on a Monday, then the user would see
* Monday: `Review SLO status:` `gray clock icon` `2 days to respond` 
* Tuesday: `Review SLO status:` `gray clock icon` `1 day to respond`
* Wednesday: `Review SLO status:` `gray clock icon` `Response due today`
* Thursday: `Review SLO status:` `red clock icon` `Response overdue by 1 day`

jrobbins:
This was done last year.

PR #3066, opened 2023-06-09
Display SLO status near the top of the gate column.

jrobbins:
This should resolve issue #3024.  It shows one line of SLO info near the top of the gate column.

The SLO info takes several forms depending on the state of the gate as calculated by server-side logic:
* If the review has not started, then show "N days for initial response (i)".
* If the review is in progress, but it is not due yet, then show "N days remaining".
* If the review is in progress and it is due today, then show "initial response is due today".
* If the review is in progress and is overdue, then show "N days overdue" in red.
* If the review is done, then show "took N days for initial response"
* Use "1 day" if the count is 1, otherwise use "days"
* Offer an info icon that explains a little more if the use hovers over it.


KyleJu (approved):

jrobbins:
Thanks for the review.   I'm going to hang on to this until next week because I also need to write a PR to backfill responded_on values for old reviews.

PR #3070, opened 2023-06-09
Use chromedash-login-required-page for all components that require signin

yanndago:
All pages that require sign in will now stay on the same url but display a message that asks the user to sign in. After signing in, the actual page content will be shown.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -305,7 +305,6 @@ class ChromedashApp extends LitElement {
       this.hideSidebar();
     });
     page('/guide/new', (ctx) => {
-      debugger;
       if (!this.setupNewPage(ctx, 'chromedash-guide-new-page')) return;
       if (ctx.querystring.search('loginStatus=False') == -1) {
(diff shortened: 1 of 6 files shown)


client-src/elements/chromedash-app.js
```diff
@@ -305,16 +305,18 @@ class ChromedashApp extends LitElement {
       this.hideSidebar();
     });
     page('/guide/new', (ctx) => {
+      debugger;
```
jrobbins, line 308: This should be removed.

settings.py
```diff
@@ -64,7 +64,7 @@ def get_flask_template_path() -> str:
 
 # This is where the an anon user is redirected if they try to access a
 # page that requires being signed in.
-LOGIN_PAGE_URL = '/features?loginStatus=False'
+LOGIN_PAGE_URL = '?loginStatus=False'
```
jrobbins, line 67: Neat.  I didn't know you could redirect to the same page by starting with "?".

jrobbins (approved):
LGTM after removing the debugger statement.

PR #3082, opened 2023-06-13
Allow users to add new stages without having to reload the edit all page and losing their uunsaved data

yanndago:
- Modify the "Add stage dialog" to receive a custom submit handler that replaces the default behavior of creating a new stage.
- Modify the edit all page to add a stage on the client only via a custom submit handler that it passes to the "Add stage dialog"
- Modify the form field names of stages added only on the client so that they can be created on submit
- Modify the server so that it creates new stages for the stages that were only created on the client

### DIFF
--- a/api/stages_api.py
+++ b/api/stages_api.py
@@ -109,22 +109,17 @@ class StagesAPI(basehandlers.APIHandler):
         setattr(stage.milestones, field, body[field])
 
     # Keep the gate type that might need to be created for the stage type.
-    gate_type: int | None = None
+    gate_type: int | None = stage_helpers.get_gate_for_stage(feature_type, s_type)
     # Update type-specific fields.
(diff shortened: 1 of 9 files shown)


jrobbins:
I'll defer to Daniel's review on this PR because he has been working more closely in this area of the code.

DanielRyanSmith (commented):
I don't know if someone has requested this to land soon, but I'm wondering if we can hold off on this change for a while until the changes to phase out the code in `guide.py` have been landed, which should be very soon. This adds a lot of complexity to the old submission process which is being overhauled, and this work will need to be translated to fit that. Sorry for the collision here; I was not aware this was something you were working on.

DanielRyanSmith (approved):
I tried this out locally and it seems to function just as intended, even when adding multiple stages in the same edit. 🙂

Issue #3027, opened 2023-05-19
Generate notification email to a suggested reviewer


jrobbins:
If a feature editor suggests a specific reviewer, chromestatus should generate an email notification to that reviewer prompting them to take on the review.

PR #3086, opened 2023-06-14
Implement a simple JSON report about overdue review gates.

jrobbins:
This is a first step toward resolving #3027.  But, before we start sending SLO reminder emails to users, I want to get a feel for how many gates are overdue.  By looking at this report, we might decide to adjust the SLO limits or make some other change.

In this PR:
* Implement a new handler, but this one just returns JSON rather than displaying an HTML page.
* Generate a list of overdue review gates in a way that does not require any new ndb index.
* Reorganize that list into a dictionary keyed by recipient.

### DIFF
--- a/internals/slo.py
+++ b/internals/slo.py
@@ -58,11 +58,6 @@ def remaining_days(requested_on: datetime.datetime, slo_limit: int) -> int:
   return slo_limit - weekdays_between(requested_on, now_utc())
 
 
-def is_overdue(requested_on: datetime.datetime, slo_limit: int) -> bool:
-  """Return True if a review is overdue."""
-  return remaining_days(requested_on, slo_limit) < 0
-
-
 def record_vote(gate: Gate, votes: list[Vote]) -> bool:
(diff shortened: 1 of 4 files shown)


internals/slo.py
```diff
@@ -97,3 +92,21 @@ def record_comment(
       return True
 
   return False
+
+
+def is_gate_overdue(gate, appr_fields, default_slo_limit) -> bool:
+  """Return True if a gate's review is overdue."""
+  if gate.requested_on is None or gate.responded_on is not None:
```
KyleJu, line 99:
```suggestion
  if gate.requested_on is None or gate.responded_on is None:
```
Checking None?

KyleJu: Actually it means  `is not Nonne` that the gate has been responded?

KyleJu (approved):

internals/slo.py
jrobbins: Yep, that's right.  If it is not none, then a reviewer already responded.

PR #3088, opened 2023-06-16
Avoid cluttering test output with expected error logging.

jrobbins:
This PR cleans up the unit test output.  It should not change the tests themselves.

When testing a case that is expected to log an error, we should mock `logging.error` so that any error logging output does not clutter the output of `npm test`.  I always like to see the output look like `................................... OK`.


pingren (approved):
LGTM. Thanks!

PR #3093, opened 2023-06-16
Ensure that newly created stages have an id that keeps increasing in the database

yanndago:
It seems that having the same parameter for every place we create a new stage ensures that the ID keeps increasing while having multiple parameters seems to give random ID. The ID is used to sort the stages when we  fetch them, so they should be sorted in the order they were created.

### DIFF
--- a/internals/core_models.py
+++ b/internals/core_models.py
@@ -17,6 +17,7 @@
 # https://stackoverflow.com/a/33533514
 from __future__ import annotations
 
+from datetime import datetime
 from typing import Any
 
 from google.cloud import ndb  # type: ignore
@@ -287,3 +288,4 @@ class Stage(ndb.Model):
   enterprise_policies = ndb.StringProperty(repeated=True)
 
   archived = ndb.BooleanProperty(default=False)
+  created = ndb.DateTimeProperty(auto_now_add=True, default=datetime.min, required=True)
--- a/api/converters.py
+++ b/api/converters.py
@@ -211,8 +211,6 @@ def stage_to_json_dict(
     'webview_last': milestones.webview_last,
   }
 
-  print(d['created'])
-
   return d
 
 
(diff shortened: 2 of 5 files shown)


internals/stage_helpers.py
```diff
@@ -45,7 +45,8 @@ class StageTemplateInfo(TypedDict):
 
 def create_feature_stage(feature_id: int, feature_type: int, stage_type: int) -> Stage:
   # Create the stage.
-  stage = Stage(feature_id=feature_id, stage_type=stage_type)
+  stage = Stage(feature_id=feature_id)
```
jrobbins, line 48: If it needs to be created in this specific way, then we should put a comment here otherwise a future developer could revert this change.

I am not sure that this will work because I seem to recall seeing some documentation about how datastore clients can reserve a pool of IDs and then allocate specific IDs out of that block.  But, of course, each GAE instance is a separate datastore client with its own reserved pool of IDs.  E.g., a user adds a stage using GAE instance #1 and then the same user adds another stage a few minutes later, but that POST gets processed by GAE instance #2, the IDs used could be out of order.

Instead, it might be better to add a `created` date time field to each stage.  That could also have clock skew between GAE instances but that should be less than the time between two requests from the same user.

jrobbins (commented):
I'm not sure about this approach.

internals/core_models.py
```diff
@@ -287,3 +288,4 @@ class Stage(ndb.Model):
   enterprise_policies = ndb.StringProperty(repeated=True)
 
   archived = ndb.BooleanProperty(default=False)
+  created = ndb.DateTimeProperty(default=datetime.min, required=True)
```
jrobbins, line 291: You could use `created = ndb.DateTimeProperty(auto_now_add=True)` so that it is automatically set.

yanndago: For some reason, doing that updates the created time of existing stages and messes up the ordering more.

yanndago: I updated the change to fix this

internals/stage_helpers.py
yanndago: Changed the approach to adding a created field that is auto added, and updated when editing the stages

PR #3098, opened 2023-06-18
Update stale feature links

pingren:
In this PR:

## Feature links updates on the fly
- `feature_link` update time older than 30 minutes ago will be considered stale.
- server will update the link if `update_stale_links = True`
- if any links are stale, client will send a second request to get the latest information.

Possible improvements:

- `_index_feature_links` could get many links to update, we could improve its performance by updating/requests concurrently.
- if two or more users visit a feature at the same time, the updates could be triggered more than once (which is acceptable since any of them will still get the latest information). We could use a global cache (such as redis) to prevent the same feature's links from being updated multiple times.

## Parsing GitHub issues and markdown file

- Use https://github.com/fastai/ghapi as client
  - parse GitHub issues and get its information
  - parse markdown file and get its title on the first line
- improve url components parsing with `urllib.parse`

Possible improvements:
- Add an access_token to increase GitHub API rate limit, since
   

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-personal-accounts

### DIFF
--- a/api/feature_links_api.py
+++ b/api/feature_links_api.py
@@ -22,7 +22,7 @@ from internals import feature_links
 class FeatureLinksAPI(basehandlers.APIHandler):
   """FeatureLinksAPI will return the links and its information to the client."""
 
-  def get_feature_links(self, feature_id: int, update_stale_links: bool = False):
+  def get_feature_links(self, feature_id: int, update_stale_links: bool):
     feature = FeatureEntry.get_by_id(feature_id)
     if not feature:
       self.abort(404, msg='Feature not found')
--- a/api/feature_links_api.py
+++ b/api/feature_links_api.py
@@ -22,16 +22,22 @@ from internals import feature_links
 class FeatureLinksAPI(basehandlers.APIHandler):
   """FeatureLinksAPI will return the links and its information to the client."""
 
-  def get_feature_links(self, feature_id: int):
+  def get_feature_links(self, feature_id: int, update_stale_links: bool = False):
     feature = FeatureEntry.get_by_id(feature_id)
     if not feature:
       self.abort(404, msg='Feature not found')
-    return feature_links.get_by_feature_id(feature_id)
+    return feature_links.get_by_feature_id(feature_id, update_stale_links)
 
   def do_get(self, **kwargs):
 
     feature_id = self.get_int_arg('feature_id', None)
+    update_stale_links = self.get_bool_arg('update_stale_links', True)
     if feature_id:
-      return self.get_feature_links(feature_id)
+      data, has_stale_links = self.get_feature_links(
+          feature_id, update_stale_links)
+      return {
+          "data": data,
+          "has_stale_links": has_stale_links
+      }
     else:
       self.abort(400, msg='Missing feature_id')
--- a/client-src/elements/chromedash-feature-page.js
+++ b/client-src/elements/chromedash-feature-page.js
@@ -175,9 +175,18 @@ export class ChromedashFeaturePage extends LitElement {
 
     window.csClient.getFeatureLinks(this.featureId).then(
       (featureLinks) => {
-        this.featureLinks = featureLinks;
+        this.featureLinks = featureLinks?.data || [];
+        if (featureLinks?.has_stale_links) {
+          // delay 2 seconds to fetch server to get latest link information
+          setTimeout(this.refetchFeatureLinks.bind(this), 2000);
+        }
       },
     );
+  }
+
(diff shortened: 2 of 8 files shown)


internals/feature_links.py
```diff
@@ -14,12 +14,19 @@
...
+import datetime
+from threading import Semaphore
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 from internals.core_models import FeatureEntry
 from internals.link_helpers import Link
 
 from google.cloud import ndb  # type: ignore
 
+executor = ThreadPoolExecutor()
```
jrobbins, line 26: I was thinking that you would use Google Cloud Tasks to enqueue a request that would run a separate handler.  There are examples of doing this in our code.   E.g.,
https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/notifier.py#L534

The advantage of using Google Cloud Tasks is that each task execution is a separate request that can be found more easily in the logs, if the task handler fails it is automatically retried, and there is an admin page in the Google Cloud Console that shows the pending task queue.

client-src/elements/chromedash-feature-page.js
```diff
@@ -175,11 +175,20 @@ export class ChromedashFeaturePage extends LitElement {
...
+        if (featureLinks?.has_stale_links) {
+          // delay 2 seconds to fetch server to get latest link information
+          setTimeout(this.refetchFeatureLinks.bind(this), 2000);
+        }
       },
     );
   }
 
+  async refetchFeatureLinks() {
+    featureLinks = await window.csClient.getFeatureLinks(this.featureId, false);
```
jrobbins, line 188: When passing a keyword argument, please use the name of the keyword.

pingren, line 188: @jrobbins is below changes what you prefer?

```suggestion
    const updateStaleLinks = false
    featureLinks = await window.csClient.getFeatureLinks(this.featureId, updateStaleLinks);
```

jrobbins: My mistake.  I was thinking of python code.   What you had originally is fine for javascript.

internals/feature_links.py
jrobbins: I believe that Cloud Tasks should run the task in under one second. 

However, the indexing process will need to wait for the requests for each URL to finish, which could take several seconds.

So, maybe the time between the initial API call and the retry should be more like 10 seconds instead of 2 seconds.

api/feature_links_api.py
```diff
@@ -22,16 +22,22 @@
 class FeatureLinksAPI(basehandlers.APIHandler):
   """FeatureLinksAPI will return the links and its information to the client."""
 
-  def get_feature_links(self, feature_id: int):
+  def get_feature_links(self, feature_id: int, update_stale_links: bool = False):
     feature = FeatureEntry.get_by_id(feature_id)
     if not feature:
       self.abort(404, msg='Feature not found')
-    return feature_links.get_by_feature_id(feature_id)
+    return feature_links.get_by_feature_id(feature_id, update_stale_links)
```
jrobbins, line 29: This is the place where you call a function that has a keyword parameter, so it should be called like:
`feature_links.get_by_feature_id(feature_id, update_stale_links=update_stale_links)`

pingren:
PR updated to use cloud task.

To setup GitHub access token. Please visit https://github.com/settings/personal-access-tokens/new

Reference: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token

internals/link_helpers.py
```diff
@@ -46,17 +55,62 @@ def __init__(self, url: str):
...
+  def _parse_github_issue(self) -> dict[str, object]:
+    """Parse the information from the github issue tracker."""
+
+    parsed_url = urlparse(self.url)
+    path = parsed_url.path
+    owner = path.split('/')[1]
+    repo = path.split('/')[2]
+    issue_id = path.split('/')[4]
+
+    information = github_api_client.issues.get(
```
jrobbins, line 97: What errors do we need to handle?  What happens if this issue does not exist or the repo is not public?

jrobbins (approved):
LGTM after considering one question.

jrobbins (approved):
Go ahead and merge this PR.

PR #3104, opened 2023-06-20
Count posts to blink-dev as SLO initial responses.

jrobbins:
The API Owners often respond on the blink-dev mailing list before they do anything in chromestatus.  That should count as an initial response for the SLO.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -24,7 +24,8 @@ from framework import users
 from internals import core_enums
 from internals import approval_defs
 from internals.core_models import FeatureEntry, Stage
-from internals.review_models import Vote
+from internals.review_models import Vote, Gate
+from internals import slo
 
 
 FIELDS_REQUIRING_LGTMS = [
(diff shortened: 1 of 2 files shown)


internals/detect_intent.py
```diff
@@ -284,3 +286,21 @@ def create_approvals(self, feature: FeatureEntry,
         # REVIEW_REQUESTED votes.
         approval_defs.set_vote(feature_id, approval_field.field_id,
             Vote.REVIEW_REQUESTED, from_addr)
+
+  def record_slo(self, feature, approval_field, from_addr) -> None:
```
KyleJu, line 290: I remember that we had a script to backfill all the slo by filtering through earliest activities. Should this consider as a part of activities as well?

jrobbins: It wouldn't be so easy because it would need to access the Google Groups message archives.  And, I don't think that it would be needed because we don't really need the SLO dates to be accurate on old gates that have been approved already.

KyleJu (approved):

Issue #2669, opened 2023-01-13
Stages of the same type aren't displayed in a consistent order


DanielRyanSmith:
If a feature has multiple stages of the same type, the order that those stages are display on the feature detail page is not consistent. Ideally, these stages would be displayed in order of first to last based on creation date, but the Stage entity currently does not have a "created" field.

PR #3107, opened 2023-06-21
Add creation date field to stages and backfill them to the feature creation date

yanndago:
We need a creation date on the stages so that they are properly sorted.
Existing stages need to get backfilled with a value which will be the feature creation date.
All new stages will automatically have the creation date set when they are created.


api/converters.py
```diff
@@ -159,7 +156,9 @@ def _prep_stage_info(
     elif s.stage_type == rollout_type:
       stage_info['rollout'] = s
     stage_info['all_stages'].append(stage_dict)
+    print(stage_dict['created'])
```
DanielRyanSmith, line 159: Remove print statement

internals/core_models.py
```diff
@@ -17,6 +17,7 @@
 # https://stackoverflow.com/a/33533514
 from __future__ import annotations
 
+from datetime import datetime
```
DanielRyanSmith, line 20: Remove unused import

internals/maintenance_scripts.py
```diff
@@ -193,3 +193,26 @@ def get_template_data(self, **kwargs):
...
+      stage.created = feature_entry.created
+      batch.append(stage)
+      count += 1
+      if len(batch) > BATCH_SIZE:
+        ndb.put_multi(batch)
+        logging.info(f'Finished a batch of {BATCH_SIZE}')
+        batch = []
+
+    ndb.put_multi(batch)
+    return f'{count} Stages entities updated of {stages.count()} available stages.'
```
DanielRyanSmith, line 218: nit: New line here

DanielRyanSmith (approved):
Thanks for addressing this! It should fix #2669
LGTM after some cleanup

Issue #2928, opened 2023-04-22
Require feature developers to provide a reasoning when omitting a feature flag


past:
[Recent feature launch guidance](https://groups.google.com/a/chromium.org/g/blink-dev/c/jhJLN9drXy4) requires featurue flags for most feature changes. Chromestatus already has a field for specifying the flag name, but it would be better to ensure that when that field is empty/missing the developers reasoning is included.

past:
After #1239 we have a "Flag Name" field, but feature authors have been using it for both the flag name as it appears in chrome://flags (the intent in #1209) and as the base::Feature flag name that is controlled by Finch. We should disambiguate them by adding a separate "Finch flag" field and an additional field when the Finch flag is unset to provide justification. There should also be a warning indicator when the Finch flag is unset and the justification field is also empty (maybe a blocking error even?).

past:
@jrobbins can you elaborate on what remains to be done to close this issue?

jrobbins:
> There should also be a warning indicator when the Finch flag is unset and the justification field is also empty (maybe a blocking error even?).

I need to add this part to the preflight dialog.  I'll work on that today.

jyasskin:
@jrobbins Should this be closed?

PR #3108, opened 2023-06-22
Add fields for finch flag name and non-finch justification.

jrobbins:
This implements most of issue #2928.

In this PR:
* Add FeatureEntry fields for finch_name and non_finch_justification.
* Add field specs for those in various places in the code, and add them to the dev trial forms.  Also, add more detail to the existing flag_name field-level docs.
* Display those values in the intent emails.  The non_finch_justification is shown if it has a value, otherwise "None" is shown if there is no finch_name.
* Make the finch_name searchable.  The non_finch_justification could be searchable in the future.

### DIFF
--- a/api/api_specs.py
+++ b/api/api_specs.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Data type for lists defining field data type information.
-FIELD_INFO_DATA_TYPE = list[tuple[str, str]]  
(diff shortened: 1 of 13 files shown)


DanielRyanSmith (approved):
LGTM! This looks like a great example of how to add a new FeatureEntry field in the future. 🙂

Issue #3014, opened 2023-05-17
Change main site navigation to a hamburger menu


jrobbins:
Replace the 4 top tabs with a hamburger menu consisting of 4 items.
Move the "Stats" submenu items to indented items in the hamburger menu.

On desktop, keep the existing location for the site name, "Create feature" button, and sign-in button / account menu.
But on mobile, move the buttons and account menu into the top of the hamburger menu.
This would probably be accomplished by having those items in both places in the DOM, but using CSS media queries to have one or the other instances of those elements be hidden.

You would probably use sl-icon-button and sl-drawer
https://shoelace.style/components/drawer?id=contained-to-an-element
Or, you could build it using standard HTML elements, e.g., `<aside>` or `<nav>` for the hamburger menu area, a checkbox that looks like the button to open it, and a CSS rule that conditionally displays the `<aside>` based on the checkbox state.

Ideally, on desktop the hamburger menu would default to being initially open, whereas it would default to being closed on mobile.  Also, while the drawer would overlap content on mobile, on desktop the content of the page should add left margin to avoid overlap in the situation where the hamburger menu drawer is open.

We should look up accessibility guidelines for a hamburger menu.  E.g., does the screen reader read it even when it is closed?

Wireframes:
https://docs.google.com/drawings/d/1szNbci26b2zV960a5WT3F8eDFNFwjWbqaZ7f_vHycHE/edit

jrobbins:
A quick search for "accessible hamburger menu" turned up several plain HTML / CSS / JS implementations.  But the key points seem to be:
* Add `role="navigation" aria-label="Main menu button"` on the button
* Add `role="navigation" aria-label="Main menu"` on the menu itself
* Test that the whole thing works using keyboard navigation via tab and space keys
* Check the tab order: e.g., it should go in sequence from the button into the menu iff the menu is open
* Check the whole thing using an actual screen reader

KyleJu:
After https://github.com/GoogleChrome/chromium-dashboard/pull/3134 and https://github.com/GoogleChrome/chromium-dashboard/pull/3135, the outstanding issues are 
- log-in does not work properly in `chromedash-featurelist`, on mobile
- content shifting does not work properly in `chromedash-featurelist`, on mobile

PR #3112, opened 2023-06-26
Drawer menu on desktop

KyleJu:
#3014.


client-src/elements/chromedash-drawer.js
```diff
@@ -0,0 +1,165 @@
...
+          pointer-events: none;
+        }
+        nav [active] {
+          color: var(--nav-link-active-color);
+          border-bottom: var(--nav-link-active-border);
+        }
+        nav [active] a {
+          color: var(--nav-link-active-color);
+        }
+        nav .nav-dropdown-container {
```
jrobbins, line 59: Maybe we could just make the Stats item always open rather than collapsable.  We have enough space and it would simplify the code and the UX if it was just always open.

client-src/elements/chromedash-drawer.js
```diff
@@ -0,0 +1,165 @@
...
+    if (drawer.open) {
+      drawer.hide();
+    } else {
+      drawer.show();
+    }
+  }
+
+  renderTabs() {
+    return html`
+      <sl-drawer label="Menu" placement="start" open=true class="drawer-placement-start" contained>
```
jrobbins, line 137: Can we hide the "X" icon and the whole row that says "Menu" by specifying attribute "no header"?  The user can already toggle the menu closed by clicking the hamburger icon again.

KyleJu: done

client-src/elements/chromedash-drawer.js
KyleJu: Done

jrobbins (changes requested):
This looks very good so far. 

Remember to update the TODOs listed in the PR description.

jrobbins:
Also, PR title says "on desktop" but you did incorporate support for mobile into the same PR, so the title should be updated.

KyleJu:
> Also, PR title says "on desktop" but you did incorporate support for mobile into the same PR, so the title should be updated.

Addressed all comments and TODOs. It is easier if this is reviewed commit by commit.

The only caveat is that the old `chromedash-featurelist` does not shift like other pages. I tried a few things programmatically but felt like I spent enough time making it work . Maybe it is time to use the new feature page?

jrobbins (approved):
Non-shifting old all-features page is OK for now.

KyleJu:
There are a few more issues with chromedash-featurelist
- my feature is shown when user isn't log=inn
- shifting isn't working
- sign-in icon does not render properly

Will send out a follow-up PR for them

PR #3119, opened 2023-06-28
Show review request at top of column.

jrobbins:
Dharani got feedback from a feature owner saying that they had "become a reviewer" after requesting a review.  The confusion came from (1) we were treating the list of votes like a log that started with the review request vote and displaying all of them under a header that said "reviewer", and (2) for a while we did have a bug where a vote widget was shown in some cases where the vote should have been read-only.  So now, we will show the name of the user who requested the review outside of the table of votes.  This makes the "Reviewer" header an accurate description of the users listed in the vote table.

In this PR:
* Show the name of the user who requested the review as part of the line that says when the review was requested.
* In the vote table, show nothing for the review request vote, unless it was requested by a user who can do a review (e.g., a site admin or one of the reviewers who is also a feature owner)
* Display "No activity yet" if there are zero votes, other than the review request vote.
* Display the review requester name whenever a review request vote exists, even if the review has concluded already.  This is needed because otherwise it would not be clear who requested the review, unless the user scrolls to the earliest comment.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -394,12 +394,18 @@ export class ChromedashGateColumn extends LitElement {
     `;
   }
 
-  renderReviewStatusActive() {
-    return html`
-      Review requested on
-      ${renderAbsoluteDate(this.gate.requested_on)}
-      ${renderRelativeDate(this.gate.requested_on)}
-    `;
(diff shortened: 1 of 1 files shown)


jrobbins:
Screenshot showing review request at the top and not included in the votes table.
image

KyleJu (approved):

Issue #3106, opened 2023-06-21
Error causing problem in pattern validation for HTML inputs since version update


g-battaglia-h-farm:
Error causing problem in pattern validation for HTML inputs, at least since version:
 114.0.5735.133 

In console is visible this errore:
```
Pattern attribute value ^[\w!#$%&'*+/=?`{|}~^-]+(?:\.[\w!#$%&'*+/=?`{|}~^-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,6}$ is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /^[\w!#$%&'*+/=?`{|}~^-]+(?:\.[\w!#$%&'*+/=?`{|}~^-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,6}$/v: Invalid character in character class
```

I tested with old version of both Chrome and Chromium and the issue was not present in versione:
113.0.5672.92

The issue is also not present in Firefox and Firefox based browsers.

dlaliberte:
Which field are you entering a value for, if any, such that this error results?   Part of this pattern looks like an email address pattern, but I don't recognize the beginning as anything we are generating explicitly.

jrobbins:
I can repro it in Chrome at this page:
https://cr-status-staging.appspot.com/guide/stage/5766413109690368/5/5709334269394944

GuidoPX:
I have the same issue with a much simpler pattern: [-a-z0-9_]
The dash/hyphen causes the error and it doesn't matter if it is placed at start or at end of the expression.
I don't know since which chrome/webkit version this is the case. I am using 114.0.5735.134. Same issue in Edge 114.0.1823.51 but not in Firefox.

Kovalchik8:
I have the same problem in Chrome with input patterns. Seems like a /v regex flag is using now by default and has a different syntax than before. 

https://groups.google.com/a/chromium.org/g/blink-dev/c/gIyvMw0n2qw

gustavoazo:
A regular expression is a powerfull tool for fast text pattern matching.
But they are also intractable, in my view. Or maybe I should just admit that despite my modest experience I never had the guts to dive deeper into regular expressions to become an expert. I'm not. And so I don't envy those people who actually develop these engines.
That being said, things get hairy when a error like the one reported in the OP is also hapenning with a regexp I'm using for url validation in a HTML form, using required attribute. 
[...]

PR #3122, opened 2023-06-28
Resolve JS regex syntax warning

jrobbins:
This should resolved issue #3106.

After a lot of trial and error, I finally gave up and read the docs.  After finding nothing in the docs, I finally searched the web and found: https://stackoverflow.com/questions/76285652/valid-with-the-regexp-u-flag-but-not-with-the-v-flag.

I don't know how the `\v` flag gets added to the regex, but it is there in the warning message.

Basically the solution is:
* Use backslashes before `/`, `{`, `}`, and `-`.
* Replace the characterset `[^\s]` with the built-in opposite one `\S`.


dlaliberte (approved):
Great fix.  Surprising and difficult to track down, I can imagine.

PR #3125, opened 2023-07-02
Feature links rendering

pingren:
In this PR:

- Render all feature links similar to Google Docs style by default
- Support rendering `LINK_TYPE_GITHUB_ISSUE` and `LINK_TYPE_GITHUB_MARKDOWN`
- Fallback if feature link is invalid

截屏2023-07-02 21 11 27
截屏2023-07-02 23 53 07

### DIFF
--- a/client-src/elements/feature-link.js
+++ b/client-src/elements/feature-link.js
@@ -20,6 +20,11 @@ function enhanceChromeStatusLink(featureLink, text) {
   // const ccRefs = information.ccRefs;
   const openedTimestamp = information.openedTimestamp;
   const closedTimestamp = information.closedTimestamp;
+
+  if (!text) {
+    text = summary;
+  }
+
   function renderTooltipContent() {
     return html`<div class="feature-link-tooltip">
     ${summary && html`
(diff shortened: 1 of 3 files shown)


client-src/elements/feature-link.js
```diff
@@ -54,32 +60,155 @@ function enhanceChromeStatusLink(featureLink, text) {
...
+        <div slot="content">${renderTooltipContent()}</div>
+        <sl-badge class="tag">
+          <sl-tag size="small" class="badge" variant="${statusRef.meansOpen ? 'success' : 'neutral'}">${statusRef.status}</sl-tag>
+          <img src="https://bugs.chromium.org/static/images/monorail.ico" alt="icon" class="icon" />
+          ${text}
+        </sl-badge>
+    </sl-tooltip>
+  </a>`;
+}
+function enhanceGithubIssueLink(featureLink, text) {
```
jrobbins, line 74: Style: Use two blank lines between top-level declarations.

client-src/elements/feature-link.js
```diff
@@ -54,32 +60,155 @@ function enhanceChromeStatusLink(featureLink, text) {
...
+  return html`<a class="feature-link" href="${featureLink.url}" target="_blank" rel="noopener noreferrer">
+    <sl-tooltip style="--sl-tooltip-arrow-size: 0;--max-width: 50vw;">
+        <div slot="content">${renderTooltipContent()}</div>
+        <sl-badge class="tag">
+          <img src="https://docs.github.com/assets/cb-600/images/site/favicon.png" alt="icon" class="icon" />
+          ${'Markdown: ' + text}
+        </sl-badge>
+    </sl-tooltip>
+  </a>`;
+}
```
jrobbins, line 183: Style: 2 blank lines needed here also.

client-src/elements/feature-link.js
```diff
@@ -20,6 +21,11 @@ function enhanceChromeStatusLink(featureLink, text) {
   // const ccRefs = information.ccRefs;
   const openedTimestamp = information.openedTimestamp;
   const closedTimestamp = information.closedTimestamp;
+
+  if (!text) {
+    text = summary;
```
jrobbins, line 26: We should limit the number of characters displayed.  How about displaying the full title if it is less than 60 characters.   If it is more than 60, then instead display only the first 45 plus "..." plus the last 10 characters.

jrobbins (approved):
LGTM after resolving some review comments.

client-src/elements/feature-link.js
jrobbins: Let's make it shorter:
How about displaying the full title on the page if it is less than 50 characters? If it is more than 50, then instead display only the first 35 plus "..." plus the last 10 characters.

The full summary line should still be displayed in the tooltip.

pingren: 截屏2023-07-03 21 29 02

PR #3132, opened 2023-07-04
Use a github access token stored in ndb.

jrobbins:
The GitHub API limits un-authenticated requests to 60 per hour, whereas authenticated requests have a quota of 5000 per hour.    Using one GitHub token might be enough for our app, but using multiple tokens could be helpful in the future.

In this PR:
* Add a new ndb entity for ApiCredential.  It stores the API tokens.
* Add code to query existing ApiCredential entities and choose one of the tokens based on recent failure timestamp.
* Call that code when setting up the GitHub API client.
* When a failure is detected, record the timestamp and choose a new token (which would be different if there is more than one)


pingren (approved):
Suggestions: Instead of randomly choosing a token, we could use [`x-ratelimit-reset` headers from Github API](https://docs.github.com/en/rest/guides/best-practices-for-integrators?apiVersion=2022-11-28#dealing-with-rate-limits) and update NDB entry with a timestamp and switch to valid ones.

If 5000/hour is enough for current traffic, we can leave a TODO.

According to [Github Docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-personal-accounts)
> User access token requests are limited to 5,000 requests per hour and per authenticated user.

jrobbins:
Please re-review.  I changed it to order API credentials by the time of the most recent failure when using each credential.  So, when one GhAPI token hits a rate limit and starts returning errors, our code will switch to using a different token in ndb, if one exists.

KyleJu (approved):

pingren (approved):

Issue #3137, opened 2023-07-07
JavaScript error with dropdowns


lukewarlow:
**Guide UX Feedback**
When editing a feature most of the dropdowns are broken.

![image](https://github.com/GoogleChrome/chromium-dashboard/assets/32498324/f2659140-877b-4fe4-b03f-82bd8a91e0d1)

```
Uncaught (in promise) ReferenceError: process is not defined
    at computePosition$1 (floating-ui.core.mjs:88:7)
```

![image](https://github.com/GoogleChrome/chromium-dashboard/assets/32498324/9cbd5f54-72ff-4488-841b-d2c9c629a62d)

lukewarlow:
Opening dev tools and running the below code "fixes" the issue.

```js
process = {
    env: {
        NODE_ENV: "PRODUCTION"
    }
}
```

jrobbins:
I'm sorry for letting this bug through, especially since it blocks users from doing work on the site.

It looks like it got through our release candidate testing because we allowed in some late library changes and tested a slightly different version on our staging server than what was deployed to our production server.  

Thanks for the bug report and the work-around.  

I have deployed it to our production server and it seems to be working now.

lukewarlow:
Would sure be nice to just be able to use popover attribute and CSS anchor positioning. Some day soon maybe! 😄 

No worries. Glad the workaround helped!

jrobbins:
I'll keep this issue open until a better fix is merged and deployed.

jcscottiii:
Related issues about floating-ui:

- https://github.com/floating-ui/floating-ui/issues/1506
- https://github.com/floating-ui/floating-ui/issues/1575

The first one has some linked implementations to resolve it. However, I'm not sure what changed as to why it's needed now.

jrobbins:
Upgrading all of our npm modules, including floating-ui, seems to have resolved the problem.
The work-around has been removed.
These changes were deployed to chromestatus.com yesterday.

PR #3139, opened 2023-07-08
Work around sl-select bug.

jrobbins:
This works-around issue #3137 by using a simplified version of the code suggested in the bug report.

The underlying defect seems to be in the floating-ui package.  It seems to have been fixed floating-ui/core-1.2.5.
https://github.com/floating-ui/floating-ui/pull/2251

But we are on floating-ui/core/1.2.2  because we have not been running `npm upgrade`.  Running that seems to resolve the problem without needing this work-around, but I made this PR as an option so that we have time to carefully test the impact of upgrading all our JS libs.


jrobbins:
This is already live in production as a -tainted version.

dlaliberte (approved):

PR #3141, opened 2023-07-08
Fix Cancel button on stage editing page.

jrobbins:
I was testing editing widgets and hit cancel.  I was surprised to see that my changes were in fact saved.

This looks like it might have been an autocomplete error in 
https://github.com/GoogleChrome/chromium-dashboard/pull/2981


dlaliberte (approved):

DanielRyanSmith (approved):
Thanks for catching this - this was a change from me that slipped in somehow.

PR #3143, opened 2023-07-09
Validate feature links http status code

pingren:
In this PR:
- add a helper function to validate feature_link
- add `is_error` and `http_error_code` in feature_link schema
   - `is_error` stands for any error happened when parsing links. For example:
       1. user input wrong url, url points to 403/500. **`http_error_code` will store http code for this case.**
       2. APIs to fetch information is down or changed.
       3. all tokens hit Github rate limiting.
       4. ...any exceptions raised.
 - When update existing indexed feature link, update information if there is no error. In this way, old information will not overwritten by `None`
 - Display http error code to user  

截屏2023-07-10 23 52 24

### DIFF
--- a/client-src/elements/feature-link.js
+++ b/client-src/elements/feature-link.js
@@ -165,6 +165,12 @@ function enhanceGithubMarkdownLink(featureLink, text) {
 
   function renderTooltipContent() {
     return html`<div class="feature-link-tooltip">
+    ${title && html`
+    <div>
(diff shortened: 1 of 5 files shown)


jrobbins (approved):

jrobbins (approved):
Don't add too much to this PR.  Merge it and start a new one if you will go beyond the work described in the PR description.

PR #3150, opened 2023-07-11
Add finch name or justification to preflight dialog.

jrobbins:
This should finish #2928.

In this PR:
* Define a new progress item for specifying finch info.
* Add that progress item to the dev trials stage in each process.
* Make that progress item a prerequisite for each the intent-to-ship actions, which makes it listed in the preflight dialog.
* Implement a progress detector function to decide when this item has been satisfied.
* Also, fix a typo in the finch_name help text


jrobbins:
Screenshot of preflight dialog when the user tries to generate an intent to ship email without filling in the finch feature name:
image

Progress item checkmark after finch feature name is filled in:
image

DanielRyanSmith (approved):

PR #3155, opened 2023-07-13
Set defaultOpen for drawer

KyleJu:
Close Drawer initially in _base.html


jrobbins (approved):

jrobbins:
I looked at the failing test and it was just that we hit a rate limit on the GitHub API for the GSoC student's work.   I'll have him use more mocks rather than real requests.

I re-ran the tests for you and think that they are likely to pass this time.

PR #3160, opened 2023-07-13
Change form fields to use updated names

DanielRyanSmith:

Changes in this PR:
- Add updated names in form field specs to allow these fields to save correctly in PATCH requests for the Features API, and no longer required rename mapping in `guide.py`.


pages/guide.py
```diff
@@ -223,18 +223,12 @@ def get_template_data(self, **defaults):
...
       ('tag_review_status', 'int'),
       ('webview_risks', 'str'),
-      ('comments', 'str'),
+      ('feature_notes', 'str'),
       ('breaking_change', 'bool'),
       ('ongoing_constraints', 'str')]
 
   # Old field name, new field name
   RENAMED_FIELD_MAPPING: dict[str, str] = {
-      'owner': 'owner_emails',
```
jrobbins, line 232: What about users who have a tab open on editing page at the time that we deploy this change?  Won't they be submitting a POST form that has the old names that still need to be mapped?  We could hold of on deleting these mappings and deploy that part a week later.

jrobbins (changes requested):

pages/guide.py
DanielRyanSmith: Good idea - I've added back the old mapping, so this scenario should not cause any problems for users

jrobbins (approved):

PR #3161, opened 2023-07-13
Enable stage edit page to use Features API PATCH request

DanielRyanSmith:
!! This change is built on top of all PRs #3156-#3160 and should be reviewed after ALL of those changes have been reviewed and landed.

This PR updates the stage edit page to use the endpoint for the Features API upon form submission and removes the use `guide.py` to save field changes.


api/features_api_test.py
```diff
@@ -465,7 +465,10 @@ def test_patch__stage_changes(self):
       'stages': [
         {
           'id': self.ship_stage_1_id,
-          'intent_thread_url': new_intent_url,
+          'intent_thread_url': {
+            'field_display_name': 'shipped_milestone',
```
jrobbins, line 469: Normally I think of a "display name" as something that is shown to users, but `shipped_milestone` is in snake-case so it looks more like a "field name" that still needs to be converted to a display name at some point.

client-src/elements/chromedash-guide-stage-page.js
```diff
@@ -125,13 +126,18 @@ export class ChromedashGuideStagePage extends LitElement {
...
-    event.preventDefault();
+  handleFormSubmit(e, hiddenTokenField) {
+    e.preventDefault();
+    const submitBody = formatFeatureChanges(this.fieldValues, this.featureId);
 
     // get the XSRF token and update it if it's expired before submission
     window.csClient.ensureTokenIsValid().then(() => {
       hiddenTokenField.value = window.csClient.token;
-      event.target.submit();
+      return csClient.updateFeature(submitBody);
```
jrobbins, line 136: Maybe in a follow-up CL, the 2 lines above this might be possible to eliminate.  IIRC, the hiddenTokenField was only needed because we were doing HTML form submission.  If we make an API call instead, csClient should handle the token for that without needing any token-management code here.

jrobbins (approved):

api/features_api_test.py
DanielRyanSmith: That's true. This property is misnamed, and is meant to represent the form field name on the client side (e.g. `shipped_milestone`) rather than the field in the database (`desktop_first`). It is used for writing activities and notification emails. I've changed the name to `form_field_name`. Hopefully this is better, but I am willing to change it if it's still unclear. 🙂

PR #3162, opened 2023-07-13
Remove listener to copy over OT milestone fields

DanielRyanSmith:

This change removes event listeners on the stage edit page to keep track of fields that have been interacted with, as this is already handled separately.
Additionally, the listener to mirror over OT milestone field values into other input fields has been removed, as it is not functional with the new implementation of form submission and tracking "touched" fields. If this functionality would like to be kept, it will need to be implemented in a separate way, but it's likely not needed with the upcoming changes to Origin Trials integration, or can be addressed at that time.

### DIFF
--- a/client-src/elements/chromedash-guide-stage-page.js
+++ b/client-src/elements/chromedash-guide-stage-page.js
@@ -121,7 +121,7 @@ export class ChromedashGuideStagePage extends LitElement {
       this.handleFormSubmit(event, hiddenTokenField);
     });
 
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
Let's track an issue to reimplement field copying.   You might want to leave COPY_ON_EDIT in the code, or put it into the description of that new issue.

DanielRyanSmith:
@jrobbins I created a new issue and linked this PR, so I can find it for reference later 👍

PR #3174, opened 2023-07-16
Edit-all page uses Features API to make updates

DanielRyanSmith:
!! This change is built on top of PR https://github.com/GoogleChrome/chromium-dashboard/pull/3161

This PR updates the edit-all page to use the endpoint for the Features API upon form submission and removes the use guide.py to save field changes.

Also, this PR changes the existing edit-all stage creation logic to send the initial stage creation request once the stage type has been chosen, rather than waiting until the entire form has been submitted. This replaces the `AddNewStageToCreate` method with the `createNewStage` method. Two web tests have been removed that were used to test this delayed creation request.


client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -98,16 +102,34 @@ export class ChromedashGuideEditallPage extends LitElement {
...
-    event.preventDefault();
+  handleFormSubmit(e, hiddenTokenField) {
+    e.preventDefault();
+    const submitBody = formatFeatureChanges(this.fieldValues, this.featureId);
 
     // get the XSRF token and update it if it's expired before submission
     window.csClient.ensureTokenIsValid().then(() => {
       hiddenTokenField.value = window.csClient.token;
-      event.target.submit();
+      return csClient.updateFeature(submitBody);
```
jrobbins, line 112: Previous 2 lines might no longer be needed.

jrobbins (approved):

client-src/elements/chromedash-guide-editall-page.js
DanielRyanSmith: I'll open a subsequent PR and remove these calls 👍

DanielRyanSmith:
More refactoring work was needed after #3181 was merged, so it might be worth taking another pass at this. 👍

@yanndago You might be interested in reviewing changes as well. The page should still function as it does today, but we're no longer using `guide.py` to save changes.

As a side note, the conditional differences for editing enterprise features vs. other features is growing, and we might want to consider creating a separate file for making enterprise edit changes. It could do a lot for reducing the complexity of these some of the editing pages

yanndago:
@DanielRyanSmith if the functionality is the same it all looks good to me.
However I have 2 questions, with this change do unsaved changes on the page stay if a user adds a new step before saving them ?
And deleting stages archives them, maybe we could simply delete them if they only have default values.

DanielRyanSmith:
> However I have 2 questions, with this change do unsaved changes on the page stay if a user adds a new step before saving them ?

Yes - the new stage should appear as normal. The only difference is that the request to create the new stage is sent when the user selects in from the dialog, and the final form submission is handled without also creating the stage, as it's already been created.

> And deleting stages archives them, maybe we could simply delete them if they only have default values.

I think this is a good idea. @jrobbins do you have opinions on this?

PR #3181, opened 2023-07-17
Allow deletion of enterprise features and stages

yanndago:
- Features can now be deleted by their creator
- Enterprise rollout stages can be deleted
- Refactor the edit all page to use the repeat directive so that the right elements are deleted


yanndago:
The test error seems unrelated to any changes here, it is triggered locally even without changes

jrobbins (approved):

PR #3182, opened 2023-07-17
Add script to backfill feature links

pingren:
This PR add script to extract feature_links from existing features and backfill to database


jrobbins:
I think the test failures are due to a previous bug that made it past CI and got merged.  If you rebase, that test should pass now.

internals/maintenance_scripts.py
```diff
@@ -216,3 +216,11 @@ def get_template_data(self, **kwargs):
...
     ndb.put_multi(batch)
     return f'{count} Stages entities updated of {stages.count()} available stages.'
+
+class BackfillFeatureLinks(FlaskHandler):
+  def get_template_data(self, **kwargs):
+    """Backfill feature links for existing feature entries."""
+    self.require_cron_header()
+    all_feature_entries = FeatureEntry.query().fetch()
+    count = batch_index_feature_entries(all_feature_entries, True)
+    return f'{len(all_feature_entries)} FeatureEntry entities backfilled of {count} feature links.'
```
jrobbins, line 226: Style: Please add newline at end of file.

internals/maintenance_scripts.py
```diff
@@ -216,3 +216,11 @@ def get_template_data(self, **kwargs):
 
     ndb.put_multi(batch)
     return f'{count} Stages entities updated of {stages.count()} available stages.'
+
+class BackfillFeatureLinks(FlaskHandler):
+  def get_template_data(self, **kwargs):
+    """Backfill feature links for existing feature entries."""
+    self.require_cron_header()
+    all_feature_entries = FeatureEntry.query().fetch()
+    count = batch_index_feature_entries(all_feature_entries, True)
```
jrobbins, line 225: I think this might be too slow because `batch_index_feature_entries` does 2 NDB queries inside a loop that iterates over each FeatureEntry.    Our production server has about 2500 FeatureEntries, so that would be 5000 queries, plus whatever writes are needed.

I think a simple way to make sure that it could work would be to accept start= and num= query string parameters.  That way, if the first attempt times out, I can run with start=0&num=500 and then take it in steps.

jrobbins (changes requested):

internals/maintenance_scripts.py
jrobbins: After talking about this, I think that we can try without star= and num= because each run will very likely make progress.

jrobbins (approved):

PR #3184, opened 2023-07-17
Dlaliberte playwright 06 22

dlaliberte:
Add initial playwright visual tests.

 * This PR includes many changes to the configuration of the repo to support the use of docker-compose with the playwright test system.  
 * Only one set of simple tests are included at this time.    The "Add a feature" button and navigation to the form for adding a new feature are minimally tested.


jcscottiii:
# Things I tried

I tried to do this outside of docker and ran into similar issues. So I ruled out a docker networking issue.

Upon starting the database, I used dsadmin to view the database before starting.

image

I see that there are four session_secrets that my local environment created. It seems that it wasn't due to gunicorn. (Whereas if you go to the Entities in the GCP console, you will find only one entry.)

# Hypothesis

Locally, gunicorn starts [differently](https://github.com/GoogleChrome/chromium-dashboard/blob/593968b95df5e23f4c8d8943d1006cf50a40ef3f/scripts/start_server.sh#L14). It creates 4 workers all at once. (whereas GAE spins up multiple instances gradually, each with a single worker).
When the browser does a login, it will call Set-Cookie with a cookie signed with a secret. However, on the reload, it is likely that it will go to a different worker which has a different session. If you continually try to log in, eventually your browser will have the right session to worker OR it has cookies for every worker.

Due to the gradual spin up in GAE, the secret retrieval for the singleton works time wise. Whereas locally the four workers fight to create the the session secret all at the same time.


# Verified Solutions

## 1.  Start [gunicorn with one worker](https://github.com/GoogleChrome/chromium-dashboard/blob/593968b95df5e23f4c8d8943d1006cf50a40ef3f/scripts/start_server.sh#L14)

```diff
-gunicorn --bind :8080 --workers 4 main:app
+gunicorn --bind :8080 --workers 1 main:app
```

## 2. Make secret getting a transaction
```diff
@classmethod
+@ndb.transactional(retries=4)
def _get_or_make_singleton(cls):
```

Returned `last exception: Only ancestor queries are allowed inside transactions.`

Turns out that locally and in docker we are using the wrong mode compared to what is run in GCP. Upon describing the database, you will find the following information. (truncated)

```sh
$ gcloud alpha firestore databases describe  --project=cr-status-staging
...
type: DATASTORE_MODE
...

If you look at the way that we start the emulator, you will see the `consistency` flag is used for the legacy datastore ([1](https://cloud.google.com/sdk/gcloud/reference/beta/emulators/datastore/start))([2](https://cloud.google.com/datastore/docs/upgrade-to-firestore)). So to match what we currently use, I replaced the consistency flag with the ``firestore-in-datastore-mode` flag.

I only changed it for the docker instance. But all other instances **NEED TO** remove the consistency flag and use -use-firestore-in-datastore-mode to match running the emulator in the [correct mode](https://cloud.google.com/datastore/docs/tools/datastore-emulator).

I went with this option long term. To make the singleton retrieval safer.

Now locally I can see there is only one secret after using the transaction and using the correct datastore mode.

![image](https://github.com/GoogleChrome/chromium-dashboard/assets/7788930/ee819d1c-ff8d-4689-be77-5da05eba2f19)

**Comment by dlaliberte** [comment]
One of the early hacks/workarounds I did early on was to hide the google login, so it didn't get in the way of logging into the testing account.  But we probably want to allow developers to login to their own account as well, so I'll see if I can reenable the google login, now that we got everything else working.

File: packages/playwright/tests/chromedash-guide-new-page_pwtest.js [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+
+test('navigate to create feature page', async ({page}) => {
+  // console.log('navigate to create feature page');
+
+  // Expect create feature button to be present.
+  const createFeatureButton = page.locator('[data-testid=create-feature-button]');
+  await expect(createFeatureButton).toBeVisible({timeout: 30000});
+
+  // Take a screenshot of header with "Create feature" button.
+  await expect(page.locator('[data-testid=header]')).toHaveScreenshot('create-feature-button.png');
```
jrobbins, line 26: How does this play out in the scenario where a developer makes a code change and forgets to update the golden screenshot?  What kind of error message do they get and at what point do they get it?

packages/playwright/tests/test_utils.js
```diff
@@ -0,0 +1,200 @@
...
+
+  const navContainer = page.locator('[data-testid=nav-container]');
+  await expect(navContainer).toBeVisible({timeout: 20000});
+  await delay(1000);
+
+  await navContainer.hover({timeout: 5000});
+  const signOutLink = page.locator('[data-testid=sign-out-link]');
+  await expect(signOutLink).toBeVisible();
+
+  await signOutLink.hover({timeout: 5000});
```
jrobbins, line 193: Why hover over the sign-out link before clicking it?

client-src/elements/chromedash-header.js
```diff
@@ -245,18 +307,19 @@ export class ChromedashHeader extends LitElement {
...
       ${this.user ? html`
         ${this.user.can_create_feature && !this.isCurrentPage('/guide/new') ? html`
-          <sl-button href="/guide/new" variant="primary" size="small">
+          <sl-button data-testid="create-feature-button"
+            href="/guide/new" variant="primary" size="small">
             Create feature
           </sl-button>
         `: nothing }
-        <div class="nav-dropdown-container">
+        <div class="nav-dropdown-container" data-testid="nav-container">
```
jrobbins, line 315: The terms "nav" and "container" are pretty common.  I think of this as the "account indicator".

packages/playwright/tests/test_utils.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+export async function delay(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+
+/**
+ * @param {import("playwright-core").Page} page
+ */
+export function captureConsoleMessages(page) {
```
jrobbins, line 15: How is this used?

scripts/start_server.sh
```diff
@@ -4,14 +4,14 @@
...
-export PYTHONPATH=cs-env/lib/python3.10/site-packages:$PYTHONPATH
+export PYTHONPATH=cs-env/lib/python3/site-packages:$PYTHONPATH
 export GOOGLE_CLOUD_PROJECT='cr-status-staging'
 export SERVER_SOFTWARE='gunicorn'
 export GAE_ENV='localdev'
 export DATASTORE_EMULATOR_HOST='localhost:15606'
 
 
-gunicorn --bind :8080 --workers 4 main:app
+gunicorn --bind :8080 --workers 1 main:app
```
jrobbins, line 14: Is this still needed?  It's a difference running locally and how the app runs on GAE.

jrobbins (commented):
I made a first pass through and have some questions.

client-src/elements/chromedash-header.js
```diff
@@ -138,6 +138,7 @@ export class ChromedashHeader extends LitElement {
     return {
       appTitle: {type: String},
       googleSignInClientId: {type: String},
+      devMode: {type: String},
```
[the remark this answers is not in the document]
dlaliberte: We could, though we would have to change the python to not generate "True" and "False".  This was another case of using what worked and moving on.  We have a few other places where we check for 'True' and 'False' in the js code.

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
dlaliberte: If they run the pwtests locally, then they will see the failures in their console.  Failures show up in the packages/playwright/tests-results directory.  If the change that causes a failure is pushed to github, then the playwright action will fail, and an artifacts file will be created with the differences.   If some other expectation fails that is not about screenshots, then the test will fail, and they would have to look into the action log.

The `npm run pwtests-update --packages=playwright` is the command to update all the screenshots locally.  They end up in the packages/playwright/tests/__screenshots__ directory.

I think I have to correct the README about this.

packages/playwright/tests/test_utils.js
dlaliberte: That part of my experimenting to get it to work reliably.  It might not be needed after all.  I was also taking a screenshot of the hovered link just to see that clicking on it should work.  But there is no screenshot of that now since a failed screenshot during logout made the logout process even more unreliable.  Adding a TODO to check this out further.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+RUN curl https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.4/install.sh | bash \
+    && . $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default \
+    && npm config set script-shell bash
+
+
+ENV NODE_PATH $NVM_DIR/v$NODE_VERSION/lib/node_modules
```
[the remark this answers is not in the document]
jcscottiii: This indeed should use the equal sign syntax.

client-src/elements/chromedash-header.js
dlaliberte: Changed to account-indicator.

dlaliberte: Renamed.

framework/secrets.py
```diff
@@ -43,6 +43,7 @@ class Secrets(ndb.Model):
   session_secret = ndb.StringProperty()
 
   @classmethod
+  @ndb.transactional(retries=4)
```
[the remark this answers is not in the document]
jcscottiii: This is still needed. More details in https://github.com/GoogleChrome/chromium-dashboard/pull/3184#issuecomment-1697546461

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
```
jcscottiii, line 3: Shouldn't the event be the same as the other CI jobs?

https://github.com/GoogleChrome/chromium-dashboard/blob/fbcfef6e873d59a3bb91cb5172821912b6cdfa5e/.github/workflows/ci.yml#L8-L12

scripts/start_server.sh
jcscottiii: This can be reverted

.github/workflows/playwright.yml
dlaliberte: I copied the initial playwright.yml from the bikeshed version.  So, I don't know what "dependabot/**" and "daily-test-sync" are about, and you may well be right.  I'll change it and see what happes.

framework/secrets.py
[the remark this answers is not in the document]
jrobbins: Ah, so there is basically a race to create a new session secret in the case where there was none previously (which is a case that happens on every test run, even though developers only see it once when setting up their development environment).  I think this race condition exists on GAE since we migrated to python3 (and gunicorn).  We didn't encounter this problem because the session secret already existed from running in python2.

client-src/elements/chromedash-drawer.js
```diff
@@ -223,7 +268,8 @@ export class ChromedashDrawer extends LitElement {
       <a href="/settings">Settings</a>
       <a href="#" id="sign-out-link" @click=${this.handleSignOutClick}>Sign out</a>
       ${this.user.can_create_feature && !this.isCurrentPage('/guide/new') ? html`
-        <sl-button href="/guide/new" variant="primary" size="small">
+        <sl-button data-testid="create-feature-button"
```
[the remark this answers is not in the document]
jrobbins: I think you resolved this conversation without posting a comment or making a change.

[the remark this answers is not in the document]
dlaliberte: Committed change to use getByTestId instead of locator('[data-testid=xxx]').  Seems to be perfectly reliable.

jrobbins (approved):

Dockerfile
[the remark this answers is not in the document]
dlaliberte: This was updated.

jrobbins:
Please give a specific PR summary line.

jrobbins:
> Please give a specific PR summary line.

Oh well.  I did not see that you had merged it already.

PR #3194, opened 2023-07-19
Add origin trial fields in the trial creation questionnaire

DanielRyanSmith:
This PR adds the fields to collect the data needed for trial creation. These are the pieces of information collected in the trial creation questionnaire that are not currently fields in ChromeStatus.

These fields are available to users currently in origin trial stages, and their help text has been taken directly from the origin trial creation form, with some small changes.


jrobbins (approved):
This code looks good.  I have a concern about the need for one field, but if it is not needed we could drop it later.

PR #3199, opened 2023-07-20
"Add Step" button displays on edit all page for enterprise only

DanielRyanSmith:
Adding a stage is less common for non-enterprise features, so it seems better to remove it from the bottom of the edit all page for now, but keep the existing button and dialog on the feature detail page.

### DIFF
--- a/client-src/elements/chromedash-guide-editall-page.js
+++ b/client-src/elements/chromedash-guide-editall-page.js
@@ -322,15 +322,13 @@ export class ChromedashGuideEditallPage extends LitElement {
 
   // render the button to add a new stage. Displays for enterprise features only.
   renderAddStageButton() {
-    if (!this.feature.is_enterprise_feature) {
-      return nothing;
-    }
-
-    return html`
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-guide-editall-page.js
```diff
@@ -320,12 +320,16 @@ export class ChromedashGuideEditallPage extends LitElement {
     return stageIds.join(',');
   }
 
+  // render the button to add a new stage. Displays for enterprise features only.
   renderAddStageButton() {
-    const text = this.feature.is_enterprise_feature ? 'Add Step': 'Add Stage';
+    if (!this.feature.is_enterprise_feature) {
```
yanndago, line 325: Maybe we can user renderHTMLif from utils

yanndago (approved):

client-src/elements/chromedash-guide-editall-page.js
DanielRyanSmith: Good idea - changed

PR #3201, opened 2023-07-21
Make features_api also update feature_links.

jrobbins:
Feature links were not being updated after each user edit because the method to update them was not being called.  It was only being called from guide.py, and this week we are shifting to using features_api.py instead.


DanielRyanSmith (approved):
Thanks for catching!

PR #3202, opened 2023-07-21
Make gate chips more compact.

jrobbins:
We will soon have more gate chips, so I'm making them a little more compact so that they fit.  And, I'm improving the alignment between icons and text, and making the padding more symmetrical.


jrobbins:
Overall page before:
image


Overall page after:
image


Close up before, showing red clock icon appears too close to the next chip, and text is slightly too high:
image

Close up after, showing chips that are more visually balanced and distinct, and text is closer to being vertically centered:
image

DanielRyanSmith (approved):
Looks good!

PR #3203, opened 2023-07-21
Fix "set to null" implementation status bug

DanielRyanSmith:
This change fixes a bug that causes the "Set implementation stage" field to display even if there is no associated implementation status name for the stage. Some refactoring is also present to move the code to a new function, `renderSetImplField`.

### DIFF
--- a/client-src/elements/chromedash-guide-stage-page.js
+++ b/client-src/elements/chromedash-guide-stage-page.js
@@ -290,16 +290,15 @@ export class ChromedashGuideStagePage extends LitElement {
     return formSections;
   }
 
-  renderImplStatusFormSection(formattedFeature, section) {
+  // Render the checkbox to set to this implementation stage.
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-guide-stage-page.js
```diff
@@ -290,16 +290,15 @@ export class ChromedashGuideStagePage extends LitElement {
     return formSections;
   }
 
-  renderImplStatusFormSection(formattedFeature, section) {
+  // Render the checkbox to set to this implementation stage.
+  renderSetImplField(implStatusName, section) {
+    // Don't render this checkbox if there's no associated implementation stage name.
+    if (!implStatusName) {
+      return nothing;
+    }
```
DanielRyanSmith, line 298: This is the check that wasn't present that was causing the bug

PR #3207, opened 2023-07-24
Add feature-link admin page

pingren:
In this PR, I added an admin-only page that displays all feature links and statistics. This page will be useful in prioritizing and implementing more feature link types to parse more links.

### API response example
```js
{
    "total_count": 4,
    "covered_count": 1,
    "uncovered_count": 3,
    "link_types": [
        {"key": "web", "count": 3},
        {"key": "chromium_bug", "count": 1},
    ],
    "uncovered_link_domains": [
      {"key": "https://docs.google.com", "count": 2},
      {"key": "https://www.google.com", "count": 1},
    ],
},

```

### UI
image


client-src/elements/chromedash-admin-feature-links-page.js
```diff
@@ -0,0 +1,99 @@
...
+        count: v.length,
+      })).sort((a, b) => b.count - a.count);
+    };
+
+    const webLinks = this.featureLinks.filter(link => link.type === 'web');
+    const groups = {
+      types: groupBy(this.featureLinks, item => item['type']),
+      uncoveredDomains: groupBy(webLinks, item => new URL(item.url).hostname),
+    };
+    return {
```
jrobbins, line 74: Why have the `stats` layer in this nested object?

client-src/elements/chromedash-admin-feature-links-page.js
```diff
@@ -0,0 +1,99 @@
...
+      stats: {
+        totalCount: this.featureLinks.length,
+        coveredCount: this.featureLinks.length - webLinks.length,
+        uncoveredCount: webLinks.length,
+        types: countAndSortGroupBy(groups.types, 'type'),
+        uncoveredDomains: countAndSortGroupBy(groups.uncoveredDomains, 'domain'),
+      },
+      groupBy,
```
jrobbins, line 82: What is this function name doing here?

api/feature_links_api.py
```diff
@@ -41,3 +41,20 @@ def do_get(self, **kwargs):
...
+    feature_links = FeatureLinks.query().fetch(
+        projection=[
+            FeatureLinks.feature_ids,
+            FeatureLinks.url,
+            FeatureLinks.type,
+            FeatureLinks.is_error,
+            FeatureLinks.http_error_code,
+        ]
+    )
+    return [item.to_dict() for item in feature_links]
```
jrobbins, line 60: Why not implement the stats calculation in python?

Doing the stats on the server side would make it much more scalable because it would not  send every FeatureLink in the JSON response.

Also, I generally find python easier to read, test, and maintain than JS.

jrobbins (commented):
I realize that this PR is still marked DRAFT, but here are some early comments anyway.

internals/feature_links.py
```diff
@@ -211,3 +212,57 @@ def batch_index_feature_entries(fes: list[FeatureEntry], skip_existing: bool) ->
...
+          })
+      list.sort(key=lambda x: x["count"], reverse=True)
+      return list
+  
+  def get_domain_with_scheme(url):
+    scheme, host = urlparse(url).scheme, urlparse(url).netloc
+    return f"{scheme}://{host}"
+  
+  uncovered_links = [link for link in links if link['type'] == 'web']
+  link_types = group_by(links, lambda item: item['type'])
```
jrobbins, line 259: If you use `collections.Counter`, then this could be something like:
```
type_counter = Counter(item['type'] for item in links)
link_types = [
    {'key': k, 'count': c} 
    for (k, c) in type_counter.most_common().items()]

domain_counter = Counter(get_domain_with_scheme(item['url']) for item in uncovered_links)
uncommon_link_domains = [
    {'key': k, 'count': c} 
    for (k, c) in domain_couner.most_common().items()]

return { 
    ... other fields ...
   "link_types": link_types,
   "uncommon_link_domains": uncommon_link_domains,
}
```

And you would not need the 2 helper functions that you wrote.

jrobbins (approved):
The code that you have looks like it will work.
I just have a suggestion on how the code could be a bit cleaner.

PR #3213, opened 2023-07-25
Add origin trials API key variable to server settings

DanielRyanSmith:
This change adds the a new variable to the server settings, `OT_API_KEY`. The value is obtained from a local text file when in dev mode or unit test mode, and is obtained from the project's secrets manager when on staging or prod. This key will be used to query the origin trials API to obtain origin trial data.

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -1,6 +1,5 @@
 import logging
 import os
-from typing import Any, Optional
 
 
 ROOT_DIR = os.path.abspath(os.path.dirname(__file__))
@@ -12,7 +11,7 @@ def get_flask_template_path() -> str:
 
 # By default, send all email to an archive for debugging.
 # For the live cr-status server, this setting is None.
-SEND_ALL_EMAIL_TO: Optional[str] = (
+SEND_ALL_EMAIL_TO: str|None = (
     'cr-status-staging-emails+%(user)s+%(domain)s@google.com')
 
 BOUNCE_ESCALATION_ADDR = 'cr-status-bounces@google.com'
@@ -74,6 +73,8 @@ REVIEW_COMMENT_MAILING_LIST = 'jrobbins-test@googlegroups.com'
 # Truncate some log lines to stay under limits of Google Cloud Logging.
 MAX_LOG_LINE = 200 * 1000
 
+# Origin trials API URL
+OT_API_URL = 'https://staging-chromeorigintrials-pa.sandbox.googleapis.com'
 
 if UNIT_TEST_MODE:
   APP_TITLE = 'Local testing'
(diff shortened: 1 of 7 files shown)


settings.py
```diff
@@ -103,6 +105,27 @@ def get_flask_template_path() -> str:
 else:
   logging.error('Unexpected app ID %r, please configure settings.py.', APP_ID)
 
+def get_ot_api_key() -> str|None:
+  # In dev or unit test mode, pull the API key from a local file.
+  if DEV_MODE or UNIT_TEST_MODE:
+    try:
+      with open(f'{ROOT_DIR}/ot_api_key.txt', 'r') as f:
```
jcscottiii, line 112: Can you add something to the README about this file and how to generate it

settings.py
```diff
@@ -103,6 +105,27 @@ def get_flask_template_path() -> str:
...
+        return f.read().strip()
+    except:
+      return None
+  else:
+    # If in staging or prod, pull the API key from the project secrets.
+    from google.cloud.secretmanager import SecretManagerServiceClient
+    client = SecretManagerServiceClient()
+    name = client.secret_path(APP_ID, 'OT_API_KEY') + '/versions/latest'
```
jcscottiii, line 120:
```suggestion
    name = f'{client.secret_path(APP_ID, 'OT_API_KEY')}/versions/latest'
```

Use f-string

jcscottiii (changes requested):
Some minor nits

settings.py
DanielRyanSmith: I'm curious how we want to handle this for a new contributor actually. This is just a workaround for the user not having access to the GCP project secrets locally, and is a text file containing the API key. So there wouldn't really be a way to generate this file, and it would need to be obtained by someone who already has access to the key. Is there a better way to go about this? It's a problem I'm not sure the best solution for.

jcscottiii: Locally, I think it's okay that a new contributor doesn't have access to it for now.
These instructions would be for us.  You could add a new package.json command that does the [download](https://cloud.google.com/sdk/gcloud/reference/secrets/versions/access) of the staging secret from GCP. `gcloud secrets versions access --out-file=thefile.txt --project=cr-status-staging`. The instructions could say something like:

```md
# Origin Trials
To test Origin Trial functionality locally, you will need to run the following: `<insert command>`

Note: *Only developers with access to the cr-status-staging project will be able to run this command. If you need to test this and you don't have access, open a bug.*
```

Additionally, you may want to generate a separate secret for staging for the devs. One secret that will be used strictly for the server to talk to OT staging, the other one to be used locally by devs and their servers to talk to OT staging

jrobbins: I'd rather not add a lot of imperative code to settings.py.   How about specifying `OT_API_URL` here, but putting the code to access the OT API key in the existing file `framework/secrets.py` or putting it in a new file named something like`internals/otconsole.py`?

jrobbins (changes requested):

jrobbins (approved):

jcscottiii (approved):

PR #3214, opened 2023-07-25
Add new Origin Trials API

DanielRyanSmith:
This PR adds the new Origin Trials API that will be used to interact with the OT API on the server side. A new endpoint is also created to query the OT API for origin trial data.


api/origin_trials_api.py
```diff
@@ -0,0 +1,37 @@
...
+from framework import basehandlers
+import requests
+import settings
+
+
+class OriginTrialsAPI(basehandlers.APIHandler):
+
+  def do_get(self, **kwargs):
+    """Get a list of all origin trials."""
+    key = settings.OT_API_KEY
```
jcscottiii, line 26: If we are planning on adding more requests besides the GET, we could move all the OT client requests logic to a new file in `frameworks`. Then we can cleanly separate OT client code from the server handler code.

api/origin_trials_api.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 Google Inc.
```
jcscottiii, line 2:
```suggestion
# Copyright 2023 Google Inc.
```

api/origin_trials_api.py
```diff
@@ -0,0 +1,37 @@
...
+class OriginTrialsAPI(basehandlers.APIHandler):
+
+  def do_get(self, **kwargs):
+    """Get a list of all origin trials."""
+    key = settings.OT_API_KEY
+    # Return an empty list if no API key is found.
+    if key == None:
+      return []
+
+    response = requests.get(
```
jcscottiii, line 31: If the key exists, we should wrap the request and the parsing in a try-except. In the event something goes wrong, log the exception internally and send back a generic error for the frontend to display.

DanielRyanSmith: Yes, agreed - I've added a try/except block with the approach you mentioned 👍

api/origin_trials_api.py
DanielRyanSmith: I want to make sure I'm understanding correctly here - We want to separate code for handling client-side requests vs. server side requests like automated trial creation, is that right? If so, would it be better to keep the server handler code in `framework` instead, and keep the client API request handlers in this file? I'm not certain of the best organization here

api/origin_trials_api.py
```diff
@@ -0,0 +1,41 @@
...
+      return []
+
+    try:
+      response = requests.get(
+          f'{settings.OT_API_URL}/v1/trials',
+          params={'prettyPrint': 'false', 'key': key})
+      response.raise_for_status()
+      trials = response.json()['trials']
+      return trials
+    except:
```
jrobbins, line 39: Can this be `except HTTPException` or something else a little more specific?  The problem with a catch-all `except:` is that it can mask programming errors.
https://google.github.io/styleguide/pyguide.html#244-decision

api/origin_trials_api.py
```diff
@@ -0,0 +1,41 @@
...
+    if key == None:
+      return []
+
+    try:
+      response = requests.get(
+          f'{settings.OT_API_URL}/v1/trials',
+          params={'prettyPrint': 'false', 'key': key})
+      response.raise_for_status()
+      trials = response.json()['trials']
+      return trials
```
jrobbins, line 38: Unit tests would help document the expected type of `trials`.

jrobbins (changes requested):

api/origin_trials_api.py
```diff
@@ -34,8 +34,8 @@ def do_get(self, **kwargs):
           f'{settings.OT_API_URL}/v1/trials',
           params={'prettyPrint': 'false', 'key': key})
       response.raise_for_status()
-      trials = response.json()['trials']
-      return trials
-    except:
+    except requests.exceptions.HTTPError:
       self.abort(500, 'Error obtaining origin trial data from API.')
```
jcscottiii, line 38: You should print a log with an exception level as well. (So that we get an email alert of some sorts)

api/origin_trials_api.py
jcscottiii: api/origin_trials_api.py should handle the incoming request to the ChromeStatus API and returning an appropriate response.

framework/origin_trials_client.py should handle sending any requests to the OT API.

The workflow would be like this
1. incoming request to the Chromestatus API which is handled by `api/origin_trials_api.py`
2. `api/origin_trials_api.py` parses the request and validates any passed in data (Not for this current GET method but you can imagine in the POST method)
3. `api/origin_trials_api.py` calls `framework/origin_trials_client.py` to do some operation
4. `framework/origin_trials_client.py` calls framework/secrets to get the OT secret and then makes the appropriate call the OT API. (In the future, `framework/origin_trials_client.py`, could raise certain custom exceptions which could be caught by `api/origin_trials_api.py` to let it know which http error to return)
5. `api/origin_trials_api.py` gets the returned value and sends the response back

api/origin_trials_api.py
```diff
@@ -0,0 +1,41 @@
...
+
+    try:
+      response = requests.get(
+          f'{settings.OT_API_URL}/v1/trials',
+          params={'prettyPrint': 'false', 'key': key})
+      response.raise_for_status()
+    except requests.exceptions.HTTPError:
+      self.abort(500, 'Error obtaining origin trial data from API.')
+
+    trials = response.json()['trials']
```
jcscottiii, line 40: I know it's low likelihood since we work on both OT API and ChromeStatus API now, but in the event the OT API ever changes and `trials` is missing, this line will throw an uncaught exception.

Technically, we should check if there's a trials key. If not, throw a 500 (since the server did not receive expected data to send back to the end user).

jcscottiii: Also, what is the exact data that we need to return back to the frontend? We should be explicit about the shape of the data we want to send back. That will prevent accidentally oversharing in case a field is added on the OT API side. Also, it will help out on the frontend to have the expected data response schema somewhere in this code.

jcscottiii (changes requested):

api/origin_trials_api.py
DanielRyanSmith: > Technically, we should check if there's a trials key. If not, throw a 500 (since the server did not receive expected data to send back to the end user).

Gotcha - I've added a check for this.

> Also, it will help out on the frontend to have the expected data response schema somewhere in this code.

Is this where OpenAPI comes in? I know we spoke about it, but I haven't familiarized myself with it yet. 😅 If this is the appropriate time, I'm willing to do so

api/origin_trials_api.py
DanielRyanSmith: I see, that makes sense - I'll try and implement this structure

api/origin_trials_api.py
DanielRyanSmith: I went to go do this, but it seems logging is already implemented in the `self.abort` call

https://github.com/GoogleChrome/chromium-dashboard/blob/main/framework/basehandlers.py#L67-L71

api/origin_trials_api.py
jcscottiii: A dataclass will be sufficient. I think the plan is to use OpenAPI for the v1 of the API.

api/origin_trials_api.py
jcscottiii: Oh ok! Thanks for pointing that out.

internals/data_types.py
```diff
@@ -263,3 +264,31 @@ class VerboseFeatureDict(TypedDict):
...
+
+
+@dataclass
+class OriginTrialInfo():
+  def __init__(self, api_trial):
+    self.id = api_trial.get('id', '')
+    self.ot_type = api_trial.get('type', '')
+    self.display_name = api_trial.get('displayName', '')
+    self.description = api_trial.get('description', '')
+    self.origin_trial_feature_name = api_trial.get('originTrialFeatureName', '')
```
jcscottiii, line 276: Will the frontend need to know about the `originTrialFeatureName` field?

framework/origin_trials_client.py
```diff
@@ -0,0 +1,52 @@
...
+import settings
+
+
+
+# Data type for information if an error occurred. Error code + message.
+ERROR_INFO_TYPE = tuple[int, str]|None
+
+
+def get_trials_list() -> tuple[list[dict[str, Any]], ERROR_INFO_TYPE]:
+  """Get a list of all origin trials."""
```
jcscottiii, line 31: The exceptions should be documented in this method docstring  https://google.github.io/styleguide/pyguide.html#244-decision

The two exceptions will help origin_trials_api.py know which message to put in the abort.

jcscottiii (changes requested):
The frameworks/origin_trials_client.py should not have knowledge of the type of http status codes or message to return to the frontend. Instead of returning a tuple of value and error, you need to raise an error.

https://google.github.io/styleguide/pyguide.html#244-decision

Then in api/origin_trials_api.py, you can do things like:

```py
try:
  trials_list = origin_trials_client.get_trials_list()
except ExceptionType1:
  self.abort(500, 'Error obtaining origin trial data from API')
except ExceptionType2:
  self.abort(500, 'Malformed response from origin trials API')
```

Regarding creating your own Exception types vs re-raising existing ones: I'm okay re-raising the existing ones for now. We can always refactor that later on.

internals/data_types.py
```diff
@@ -263,3 +264,31 @@ class VerboseFeatureDict(TypedDict):
   is_enterprise_feature: bool
   updated_display: str | None
   new_crbug_url: str | None
+
+
+@dataclass
+class OriginTrialInfo():
+  def __init__(self, api_trial):
+    self.id = api_trial.get('id', '')
+    self.ot_type = api_trial.get('type', '')
```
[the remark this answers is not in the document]
DanielRyanSmith: Probably not needed, but I'm not certain - I've just removed it for now

internals/data_types.py
```diff
@@ -263,3 +264,31 @@ class VerboseFeatureDict(TypedDict):
...
+    self.display_name = api_trial.get('displayName', '')
+    self.description = api_trial.get('description', '')
+    self.origin_trial_feature_name = api_trial.get('originTrialFeatureName', '')
+    self.enabled = api_trial.get('enabled', None)
+    self.status = api_trial.get('status', '')
+    self.chromestatus_url = api_trial.get('chromestatusUrl', '')
```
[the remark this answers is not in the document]
DanielRyanSmith: Until trial creation is done on ChromeStatus, this is one of the only ways to identify which trials belong to which ChromeStatus features, so we'll use this on the client side, at least for a little while

framework/origin_trials_client.py
```diff
@@ -0,0 +1,52 @@
...
+  # Return an empty list if no API key is found.
+  if key == None:
+    return [], None
+
+  try:
+    response = requests.get(
+        f'{settings.OT_API_URL}/v1/trials',
+        params={'prettyPrint': 'false', 'key': key})
+    response.raise_for_status()
+  except requests.exceptions.HTTPError:
```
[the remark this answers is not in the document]
DanielRyanSmith: I want to make sure I'm doing this right - should I not be passing the error info in the return value and instead be returning an actual error? Or am I raising the error in the `get_trials_list()` function and then catching it in the `origin_trials_api.do_get()` to raise it again there?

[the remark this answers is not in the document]
jcscottiii: You'll do this part `Or am I raising the error in the get_trials_list() function and then catching it in the origin_trials_api.do_get()`. 

Then in do_get(), it will catch that specific exception and call the right abort:

```py
try:
  trials_list = origin_trials_client.get_trials_list()
except ExceptionType1: # the http one
  self.abort(500, 'Error obtaining origin trial data from API')
except ExceptionType2: # The valueerror one
  self.abort(500, 'Malformed response from origin trials API')
```

internals/data_types.py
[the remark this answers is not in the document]
jcscottiii: Makes sense! Thanks!

internals/data_types.py
jcscottiii: Wouldn't we want to use the `id` field for that purpose instead?

DanielRyanSmith: Yes, but we cannot know the OT ID since we do not have access to that until trial creation is handled on Chromestatus, or we ask feature owners to supply the OT ID in Chromestatus. We ask feature owners to supply this feature name field, and with this + the chromestatus URL, we can begin to associate origin trials to their respective feature stage in Chromestatus.

This is not an implementation we want to use permanently, but it would make it possible with what we currently have to link OTs with their feature, and then when trial creation is handled in Chromestatus, we can save the trial ID directly to make this association.

I am willing to talk more about this if you would like to talk about the efficacy of this approach 🙂 I won't deny that it's half-baked 😅

jcscottiii: No worries! Thanks for the explanation. It helped. I get what you're going for now. Let's go with what you have and we can revisit in the future if needed.

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+import settings
+
+
+
+# Data type for information if an error occurred. Error code + message.
+ERROR_INFO_TYPE = tuple[int, str]|None
+
+
+def get_trials_list() -> tuple[list[dict[str, Any]], ERROR_INFO_TYPE]:
+  """Get a list of all origin trials."""
```
jcscottiii, line 31:
```suggestion
  """Get a list of all origin trials.
  
  Raises:
    requests.exceptions.RequestException: If the request fails to connect or the HTTP
      status code is not successful.
    KeyError: If the response from the OT API is not in the expected format.
  """
```

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+  key = secrets.get_ot_api_key()
+  # Return an empty list if no API key is found.
+  if key == None:
+    return [], None
+
+  response = requests.get(
```
jcscottiii, line 37: You should still have the try-except for the request.

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+      f'{settings.OT_API_URL}/v1/trials',
+      params={'prettyPrint': 'false', 'key': key})
+  response.raise_for_status()
```
jcscottiii, line 40:
```suggestion
  try:
    response = requests.get(
        f'{settings.OT_API_URL}/v1/trials',
        params={'prettyPrint': 'false', 'key': key})
    response.raise_for_status()
  except requests.exceptions.RequestException as e:
    logging.exception("Failed to get response from OT")
    raise e
```

This will give us a friendly log statement before doing a stack trace print to the logs (which should send an email to us)

You will need to `import logging`

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+import requests
+
+from framework import secrets
+from internals.data_types import OriginTrialInfo
+import settings
+
+
+
+# Data type for information if an error occurred. Error code + message.
+ERROR_INFO_TYPE = tuple[int, str]|None
```
jcscottiii, line 27:
```suggestion
```

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_trials_list() -> tuple[list[dict[str, Any]], ERROR_INFO_TYPE]:
```
jcscottiii, line 30:
```suggestion
def get_trials_list() -> list[dict[str, Any]]:
```

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+  """Get a list of all origin trials."""
+  key = secrets.get_ot_api_key()
+  # Return an empty list if no API key is found.
+  if key == None:
+    return [], None
```
jcscottiii, line 35:
```suggestion
    return []
```

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+      f'{settings.OT_API_URL}/v1/trials',
+      params={'prettyPrint': 'false', 'key': key})
+  response.raise_for_status()
+
+  response_json = response.json()
+
+  trials_list = [asdict(OriginTrialInfo(api_trial))
+                 for api_trial in response_json['trials']
+                 if api_trial['isPublic']]
+  return trials_list, None
```
jcscottiii, line 47:
```suggestion
  return trials_list
```

framework/origin_trials_client.py
```diff
@@ -0,0 +1,47 @@
...
+  response = requests.get(
+      f'{settings.OT_API_URL}/v1/trials',
+      params={'prettyPrint': 'false', 'key': key})
+  response.raise_for_status()
+
+  response_json = response.json()
+
+  trials_list = [asdict(OriginTrialInfo(api_trial))
+                 for api_trial in response_json['trials']
+                 if api_trial['isPublic']]
```
jcscottiii, line 46:
```suggestion
  try:
    trials_list = [asdict(OriginTrialInfo(api_trial))
                   for api_trial in response_json['trials']
                   if api_trial['isPublic']]
  except KeyValue as e:
    logging.exception("Received badly formatted data from OT API")
    raise e
```

api/origin_trials_api.py
```diff
@@ -0,0 +1,38 @@
...
+class OriginTrialsAPI(basehandlers.APIHandler):
+
+  def do_get(self, **kwargs):
+    """Get a list of all origin trials.
+    Returns:
+      A list of data on all public origin trials.
+    
+    Raises:
+      RequestException: If an error occurred interacting with the OT API.
+      KeyError: If the response from the OT API is not in the expected format.
```
jcscottiii, line 29:
```suggestion
    
    Raises:
      RequestException: If an error occurred interacting with the OT API.
      KeyError: If the response from the OT API is not in the expected format.
```

This doc string should be for the get_trials_list method as the exceptions originate there. Will make a suggestion there. Remove the raises section here.

jcscottiii (changes requested):

framework/origin_trials_client.py
DanielRyanSmith: Ah, I wasn't sure if I was supposed to leave the try/except block in here along with the one in the `origin_trials_api` function

framework/origin_trials_client.py
DanielRyanSmith: Sorry, I had thought I removed these, as well as the error type hinting, but it's not showing up here... Maybe it was with an abandoned change prior. Either way, I should have checked before sending for a second review!

framework/origin_trials_client.py
DanielRyanSmith: Ah, I wasn't sure if I was supposed to leave the try/except block in here along with the one in the `origin_trials_api` function

framework/origin_trials_client.py
DanielRyanSmith: Sorry, I remember removing the code for the error returns and the type hinting, but it doesn't show up here. It might have been part of an abandoned change. Either way, I should have caught it before sending it over for re-review!

jcscottiii (approved):

PR #3215, opened 2023-07-26
Fix new_crbug_url JSON response field so that bug icon works.

jrobbins:
Carol noticed that the bug icon on the feature detail page was not doing anything.  It was always being left as None in cases where the client requested a single feature.


DanielRyanSmith (approved):
Thanks for the fix!

PR #3216, opened 2023-07-26
Defaut to opening the menu drawer on the old feature list page.

### LINKED ISSUE #3167: Navigating to All Features closes the hamburger menu
After the new UI changes switching to a different tab works as expected, leaving the left menu open to select another tab to navigate to. The only exception is https://chromestatus.com/features, which immediately closes the menu on navigation.

jrobbins:
This resolves #3167.

Basically, take the same approach that chromedash-app uses to shift page content when the menu is open and apply that to the old MPA page for the feature list.  

Having the margin shifting working allows us to make the menu open by default on desktop. This feels very natural as a user because the menu would always be open when the user using it to navigate to this page, so it should still be open on the page once the user gets there.

### DIFF
--- a/templates/features.html
+++ b/templates/features.html
@@ -56,5 +56,25 @@
       };
       loadFeatureLegendViews(VIEWS);
     })();
+
+    // Since we don't use chromedash-app on this page, re-implement a little
+    // logic for the main menu shifting the page content on desktop.
+    const isMobile = window.screen.width < 701;
+    if (!isMobile) {
+        const header = document.querySelector('chromedash-header');
+        const drawer = document.querySelector('chromedash-drawer');
+        const flexWrapper = document.querySelector('#content-flex-wrapper');
+        const shiftedStyle = 'margin-left: 210px; justify-content: start';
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):

templates/features.html
```diff
@@ -56,5 +56,25 @@ <h2>Features: <span class="num-features"></span></h2>
...
+    if (!isMobile) {
+        const header = document.querySelector('chromedash-header');
+        const drawer = document.querySelector('chromedash-drawer');
+        const flexWrapper = document.querySelector('#content-flex-wrapper');
+        const shiftedStyle = 'margin-left: 210px; justify-content: start';
+        drawer.defaultOpen = true;
+        flexWrapper.style = shiftedStyle;
+
+        function shiftPage() {
+            let slDrawer = drawer.shadowRoot.querySelector('sl-drawer');
```
DanielRyanSmith, line 72:
```suggestion
            const slDrawer = drawer.shadowRoot.querySelector('sl-drawer');
```

PR #3221, opened 2023-07-30
Handle invalid urls

pingren:
In this PR:

- Added a `valid_url` function as an extra guard in `extract_urls_from_value` to filter out invalid URLs if possible.
- On the frontend, preventing 404 errors from showing up for long text fields with `_enhanceLink`.
- Skipped indexing feature_links if parsing encountered errors. This way, invalid links won’t be indexed in the first place.

### DIFF
--- a/internals/feature_links.py
+++ b/internals/feature_links.py
@@ -61,8 +61,9 @@ def update_feature_links(fe: FeatureEntry, changed_fields: list[tuple[str, Any, 
         link = Link(url)
         if link.type:
           feature_link = _get_index_link(link, fe, should_parse_new_link=True)
-          feature_link.put()
-          logging.info(f'Indexed feature_link {feature_link.url} to {feature_link.key.integer_id()} for feature {fe.key.integer_id()}')
+          if feature_link:
+            feature_link.put()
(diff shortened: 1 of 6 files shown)


internals/link_helpers.py
```diff
@@ -47,6 +47,12 @@
 
 URL_REGEX = re.compile(r'(https?://\S+)')
 
+def valid_url(url):
+    try:
+        result = urlparse(url)
+        return all([result.scheme, result.netloc, result.path])
```
jrobbins, line 53: Does this work for `http://example.com` with no trailing slash?  I think that makes path an empty string which is considered false.

internals/feature_links.py
```diff
@@ -80,6 +81,8 @@ def _get_index_link(link: Link, fe: FeatureEntry, should_parse_new_link: bool =
   else:
     if should_parse_new_link:
       link.parse()
+      if link.is_error:
```
jrobbins, line 84: If you don't index links that have errors, then how will you generate notification emails about broken links?

pingren: During the meeting we agreed that we would only generate notifications for links that were valid before and turned broken later, not for links that were never valid.

jrobbins (approved):

PR #3231, opened 2023-08-03
If an API Owner requests a review, dont count that as a response.

### LINKED ISSUE #3228: SLO response was 0.00003 seconds
**Describe the bug**
I saw that one of our prod Gate entities had a request time and slo_initial_response time that differed by only  a tiny fraction of a second, which seems wrong because these are human responses that cannot be that fast.

The review was requested via an intent thread that was sent by a user who is both a feature owner of that feature and an API Owner on the site.  My theory is: since the intent message was sent by someone who had permissions to review it, the original request was also counted as a response.

**Expected behavior**
An email message that requests a review should never itself count as a review response.

**Additional context**
https://groups.google.com/a/chromium.org/g/blink-dev/c/pSD2HBnd720/m/YAE8xwpwAQAJ?utm_medium=email&utm_source=footer

jrobbins:
This resolves #3228.

Normally we think of reviewers as being a separate set of users from feature owners, however it is possible for a user with review permission to also be a feature owner.   In the case where an API Owner creates an intent thread, that was counting as both a review request (setting `gate.requested_on`) and a response (setting `gate.responded_on`).

In this PR:
* Keep track of whether this is the first vote on a gate.  If it is, then don't count that as a response.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -196,8 +196,9 @@ class IntentEmailHandler(basehandlers.FlaskHandler):
       return {'message': 'Feature not found'}
 
     self.set_intent_thread_url(feature, approval_field, thread_url, subject)
+    is_new_thread = detect_new_thread(feature_id, approval_field)
     self.create_approvals(feature, approval_field, from_addr, body)
-    self.record_slo(feature, approval_field, from_addr)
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Interesting bug - good catch!

PR #3234, opened 2023-08-04
Add an ndb index for /admin/feature_links.

jrobbins:
Apparently fetching data from Google Cloud NDB with a specific projection (a subset of fields) requires that there be an index with those fields.  This is needed for https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/feature_links.py#L241


jrobbins:
This is already live on staging.

Issue #3236, opened 2023-08-07
Add property `requireTouched` for form fields


DanielRyanSmith:
Currently, all form fields require the user to interact with that field or make changes in order to submit the field's value for changes in their feature. This caused a bug today, as the accuracy verification checkbox has a default "checked" (or "true") value. Most users to not interact with this field before submitting the form, which means this field was treated as not "touched" by the user, and was not part of the submitted changes in the request.

Some form fields, like the accuracy verification checkbox field, should not be required for users to interact with in order to change the value when submitting a form. Some fields can be filled with default values that the user will not change before submission, and these values should be sent over to be processed in the Features API.

Ideally, we can designate which fields do not require user interaction to be part of the request, so it would be useful to designate certain fields as not required to be touched. This `requireTouched` property could be added and referenced from the field definitions in `client-src/elements/form-definition.js`.

jrobbins:
@DanielRyanSmith All done or is there remaining work on this issue?

DanielRyanSmith:
Work on this has not been done yet. I added a special case in the code to always submit the `accurate_as_of` form field value [whether it has been touched or not](https://github.com/GoogleChrome/chromium-dashboard/blob/main/client-src/elements/utils.js#L302-L304), but it's possible that there could be more fields that should be handled the same way. In that case, the approach described above still needs to be implemented.

DanielRyanSmith:
Closing this since #3425 has landed. We have no other fields that function this way, but If new fields come up and require function similar to the `accurate_as_of` checkbox, then a `defaultTouched` or `requireTouched` property might be worthwhile in the form definition. Without that needed yet, this is likely not worth pursuing.

PR #3237, opened 2023-08-07
Accuracy verification bug fix

DanielRyanSmith:
This change fixes a bug that caused the `accurate_as_of` date to not update unless the user interacted with the checkbox to confirm accuracy.

Note: this is a quick fix, and #3236 should be implemented in order to more gracefully handle this problem.

### DIFF
--- a/client-src/elements/utils.js
+++ b/client-src/elements/utils.js
@@ -299,7 +299,7 @@ export function formatFeatureChanges(fieldValues, featureId) {
   const stages = {};
   for (const {name, touched, value, stageId, implicitValue} of fieldValues) {
     // Only submit changes for touched fields or accuracy verification updates.
-    if (!touched && (name !== 'accurate_as_of' || value !== true)) {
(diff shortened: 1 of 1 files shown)


client-src/elements/utils.js
```diff
@@ -298,8 +298,8 @@ export function formatFeatureChanges(fieldValues, featureId) {
   // Multiple stages can be mutated, so this object is a stage of stages.
   const stages = {};
   for (const {name, touched, value, stageId, implicitValue} of fieldValues) {
-    // Only submit changes for touched fields.
-    if (!touched) {
+    // Only submit changes for touched fields or accuracy verification updates.
+    if (!touched && (name !== 'accurate_as_of' || value !== true)) {
```
jrobbins, line 302: I'm not sure about the `|| value !== true part`.  Does that make it send every false checkbox every time?  Will that add noise to the activity log?

DanielRyanSmith: Good catch - updated to fix that case

jrobbins (approved):

PR #3242, opened 2023-08-08
Replace indentation with dividers in navigation drawer.

jrobbins:
Carol pointed out that the "Stats" header looks the same as clickable navigation items, which can lead to user frustration.

The material 3 guidelines show using a dividing line rather than indentation.   Their approach works best if the items have icons which effectively make all the items look indented more than the icon-less header.  We don't have icons, so I went with bold to differentiate the headers.  I also simplified the code a little bit.

https://m3.material.io/components/navigation-drawer/specs


jrobbins:
Screenshot on desktop:
image

Screenshot on mobile when signed in:
image

Screenshot on mobile when signed out:
image

KyleJu (approved):

PR #3246, opened 2023-08-10
Work around threading error with Google Cloud Logging

### LINKED ISSUE #3197: Threading exceptions from Google Cloud Logging
**Describe the bug**
When we upgraded to python3.11, we started seeing exception from Google Cloud Logging.   Whenever the instance exits, there are KeyError exceptions in the standard python threading.py file.

**To Reproduce**
1. Edit main.py to enable GCL locally (or test on staging)
2. `npm start` and wait, then hit ctrl-c.  Or, wait about 15 minutes for GAE.
3. See exceptions

**Expected behavior**
It should not raise exceptions.   

This is not a user-visible error because the exception is raised during instance shutdown rather than the processing of any user request.  However, it is an operational problem for us because each exception shows up on the Error Reporting page and causes an alert notification.   These notifications are not merged together because the exception includes a unique thread ID.

**Additional context**
A work-around for now is to disable Google Cloud Logging and let the normal python logger simply write to stdout.   This is less than ideal because log lines are not grouped under the request log entry for the request that generated the log output.  Instead, all logging for all requests is mixed together in one time-ordered stream, which can be confusing when there are concurrent requests.

jrobbins:
This resolves #3197.

For reasons still unknown, Google Cloud Logging under python 3.11 seems to cause errors when each instance shuts down.  It looks like it is somehow trying to kill its background worker thread twice, causing a KeyError in the standard python library threading.py.  I tested same versions of Google Cloud Logging that previously worked under 3.10 and it had the same error in 3.11, so the problem only arises with a combination of python 3.11 and GCL and our app.

In this PR:
* Patch Thread._delete to make it catch the KeyError.
* Remove an unneeded call to get_default_logger() 

This change is live on staging now and it seems to get us back to the point where we had things working in python 3.10.  Log lines are grouped by requests, Error Report entries have a full stack thread, but there is the ongoing low-priority problem of some log lines exceeding a maximum size.

### DIFF
--- a/main.py
+++ b/main.py
@@ -57,11 +57,11 @@ import settings
 
 # Sets up Cloud Logging client library.
 # TODO(jrobbins): Re-enable Google Cloud Logging when #3197 is resolved.
-# if not settings.UNIT_TEST_MODE and not settings.DEV_MODE:
-#   import google.cloud.logging
-#   client = google.cloud.logging.Client()
-#   client.get_default_handler()
-#   client.setup_logging()
+if True: #not settings.UNIT_TEST_MODE and not settings.DEV_MODE:
+  import google.cloud.logging
+  client = google.cloud.logging.Client()
+  client.get_default_handler()
+  client.setup_logging()
 
 
 # Note: In the URLs below, parameters like <int:feature_id> are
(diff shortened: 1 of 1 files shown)


DanielRyanSmith (approved):
Good sleuthing! It seemed like this would have been hard to find the root cause of the noisy errors. 🙂

PR #3247, opened 2023-08-11
Use GateDef for privacy and security.

jrobbins:
Instead of listing configured approver emails in our source code, we now have an option to to store these in NDB.

In this PR:
* When looking up approvers for a gate type, check for special value IN_NDB.
* Implement GateDef entity with a list of approvers and placeholders for fields to be moved there in the future.
* Implement get_gate_def() to find the needed GateDef or create a blank entity for an admin to fill in via cloud console.

### DIFF
--- a/internals/approval_defs_test.py
+++ b/internals/approval_defs_test.py
@@ -19,7 +19,7 @@ import testing_config  # Must be imported before the module under test.
 from unittest import mock
 
 from internals import approval_defs
-from internals.review_models import Gate, Vote
+from internals.review_models import Gate, GateDef, Vote
 
 
 class FetchOwnersTest(testing_config.CustomTestCase):
@@ -73,10 +73,18 @@ MOCK_APPROVALS_BY_ID = {
         'Intent to optimize',
         'You need permission to optimize',
         2, approval_defs.THREE_LGTM, 'https://example.com', 'API Owners'),
+    3: approval_defs.ApprovalFieldDef(
+        'Intent to memorize',
+        'You need permission to memorize',
+        3, approval_defs.THREE_LGTM, approval_defs.IN_NDB, 'API Owners'),
 }
 
 
 class GetApproversTest(testing_config.CustomTestCase):
+
+  def tearDown(self):
+    for gate_def in GateDef.query():
+      gate_def.key.delete()
 
   @mock.patch('internals.approval_defs.APPROVAL_FIELDS_BY_ID',
               MOCK_APPROVALS_BY_ID)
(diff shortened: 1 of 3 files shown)


internals/approval_defs_test.py
```diff
@@ -97,6 +105,34 @@ def test__url(self, mock_fetch_owner):
     mock_fetch_owner.assert_called_once_with('https://example.com')
     self.assertEqual(actual, ['owner@example.com'])
 
+  @mock.patch('internals.approval_defs.APPROVAL_FIELDS_BY_ID',
+              MOCK_APPROVALS_BY_ID)
+  @mock.patch('internals.approval_defs.fetch_owners')
+  def test__ndb_new(self, mock_fetch_owner):
+    """Some approvals will have appovers in NDB, but they are not found."""
```
DanielRyanSmith, line 112:
```suggestion
    """Some approvals will have approvers in NDB, but they are not found."""
```

internals/approval_defs_test.py
```diff
@@ -97,6 +105,34 @@ def test__url(self, mock_fetch_owner):
...
+    updated_gate_defs = GateDef.query().fetch()
+    self.assertEqual(1, len(updated_gate_defs))
+    self.assertEqual(3, updated_gate_defs[0].gate_type)
+    self.assertEqual([], updated_gate_defs[0].approvers)
+
+  @mock.patch('internals.approval_defs.APPROVAL_FIELDS_BY_ID',
+              MOCK_APPROVALS_BY_ID)
+  @mock.patch('internals.approval_defs.fetch_owners')
+  def test__ndb_existing(self, mock_fetch_owner):
+    """Some approvals will have appovers in NDB, use it if found."""
```
DanielRyanSmith, line 125:
```suggestion
    """Some approvals will have approvers in NDB, use it if found."""
```

internals/review_models.py
```diff
@@ -55,6 +54,33 @@ def get_raw_owner_file(cls, url):
...
+    query: ndb.Query = GateDef.query(GateDef.gate_type == gate_type)
+    gate_defs = query.fetch(1)
+    if gate_defs:
+      gate_def = gate_defs[0]
+    else:
+      logging.info(f'Creating empty GateDef for {gate_type}')
+      gate_def = GateDef(gate_type=gate_type)
+      gate_def.put()
+
+    return gate_def
```
DanielRyanSmith, line 81: I remember last year trying to move away from adding these class methods directly on the model and instead putting them in a `_helpers.py` file. However, I'm blanking on why exactly that was. I think it had something to do with the possibility of circular imports. Either way, I don't think this is a problem now, so it's fine here. 🙂

DanielRyanSmith (approved):
LGTM outside of small nits

PR #3249, opened 2023-08-11
Re-enable creation of privacy and security gates.

jrobbins:
These gates were previously enable but then we held off on that because we were blocked on the details of their questionnaire and reviewers.  Now those details are set, so the gates can be re-enabled.


DanielRyanSmith (approved):
These gates will be created for every new feature they are relevant to. However, they will not be retroactively created for existing features. I don't know off the top of my head if that will cause problems, but it's something we should make sure of as we land this change.

jrobbins:
> These gates will be created for every new feature they are relevant to. However, they will not be retroactively created for existing features. I don't know off the top of my head if that will cause problems, but it's something we should make sure of as we land this change.

We have a script to fill in missing gates: /admin/write_missing_gates.

PR #3235, opened 2023-08-06
Cron job to update all feature links

pingren:
In this PR:

- Add `/cron/update_all_feature_links`: a weekly cron job to update feature links
- This cron job filters feature links based on certain conditions:
  1. **stale**: If it is has not been updated for 8 days
  2. **type changed**: If its type has changed. e.g. LINK_TYPE_WEB -> new link type we support later
  3. **has error**: if last time parsing had an error
- Filter links ids are sent to cloud task `/tasks/update-feature-links`, with a maximum of 500 in batches to prevent 600 seconds timeout
- For backfill purpose, when passing query params `no_filter=true`, it will simply update all feature links without any condition check.

So to backfill all feature links with information:
1. Run `/scripts/backfill_feature_links`, this extract and backfill all feature links from FeatureEntry into Google Cloud NDB
2. Run `/cron/update_all_feature_links?no_filter=true`, this update all feature links in Google Cloud NDB with latest information

PR #3252, opened 2023-08-11
Add index for FeatureLinks cron

jrobbins:
Pingren updated an existing index on FeatureLinks to add a field needed in a new query, but ndb requires an exact match and the existing index has a field that is not in the new query.  So, ndb generated an error message that requests this additional index.

Pingren's PR: https://github.com/GoogleChrome/chromium-dashboard/pull/3235/files

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -472,3 +472,10 @@ indexes:
   - name: type
   - name: url
   - name: updated
+- kind: FeatureLinks
+  properties:
+  - name: http_error_code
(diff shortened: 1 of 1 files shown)


jrobbins:
This is live on staging.  The FeatureLinks cron job now runs successfully.

pingren (approved):

DanielRyanSmith (approved):

PR #3278, opened 2023-08-25
Permit assigned reviewer to vote on reviews.

jrobbins:
This is a step in supporting assigned reviewers, which will be used when auto-assigning a reviewer based on a rotation.

In this PR:
* Rename the existing Gate `owners` field to `assignee_emails` to avoid confusion with other uses of `owner` in the app.
* Change the permission checking function `can_approve_features` to a more accurate name `can_review_gates` and make it allow users listed in `gate.assignee_emails`.
* Update all the callers of the above.
* Remove the base data for `can_approve`.  It is no longer used because we have been using `approvable_gate_types` for a while now.


KyleJu (approved):
How do we convert the existing `owner` field to `assignee_emails`? Don't we need a script for it?

jrobbins:
> How do we convert the existing `owner` field to `assignee_emails`? Don't we need a script for it?

The owner field was never exposed in the UI so it has not been used yet.

PR #3285, opened 2023-08-28
Display "Awaiting review" when prompting a reviewer to vote.

jrobbins:
This UI tweak aligns the implementation with a detail of the UX mocks that had been missed before.  It was prompted by a question from a reviewer: "How does it know that this review is assigned to me already?"  Which is a misunderstanding because assigned reviews are still a work-in-progress with no UI yet.

The existing UX has been using the name of the signed in reviewer as a placeholder to say "this is where your review would go".  It fits in nicely with a list of other review votes from other reviewers.  That seemed like an acceptable UX before we started discussing the concept of assigned reviews with the review teams.  Now, using the signed-in user's name as a placeholder seems like a conflict with how we might display the name of an assigned reviewer.

In this PR:
* Change the loop that is used to generate the table of votes so that it only iterates over responses.  Responses are votes other than the initial review request.  This allows a simplification of `renderVoteRow()`.
* Simplify `renderVoteRow()` to always render single vote row.  It no longer needs to skip the review request, and it is no longer called for the special case of prompting for a new vote at the end of the table.
* Implement a new function `renderAddVoteRow()` specifically for the case of prompting the user to enter a new vote.  And, have it display "_Awaiting review_" rather than the user's name.

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -85,6 +85,10 @@ export class ChromedashGateColumn extends LitElement {
        #votes-area th {
          font-weight: 500;
        }
+       table .your-vote {
+         font-style: italic;
+         white-space: nowrap;
+       }
 
        #questionnaire {
          white-space: pre-wrap;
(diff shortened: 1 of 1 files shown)


jrobbins:
Old screenshot:
image

New screenshot:
image

KyleJu (approved):

PR #3287, opened 2023-08-30
Hide CSS metrics ending in "(obsolete)".

jrobbins:
A while back we started hiding UseCounters that have a name that starts with "OBSOLETE_".  Those were mainly JS/HTML use counters.  Recently I noticed that some CSS UseCounters are also marked "(obsolete)".  So, this PR detects those too.


jrobbins:
Screenshot before
image

Screenshot after
image

KyleJu (approved):

PR #3290, opened 2023-08-30
Add links to associated origin trials

DanielRyanSmith:
This PR adds direct links to the origin trial registration pages of trials that have been associated with ChromeStatus entries, opening them in a new tab when clicked. Additionally, this fixes some small spacing issues on the buttons that display together at the top-right of detail sections.

### View when user is not a feature owner/editor:
Screenshot 2023-08-30 at 1 52 36 PM

### View when user has edit access to the feature:
Screenshot 2023-08-30 at 1 51 37 PM

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -57,6 +57,7 @@ const LONG_TEXT = 60;
 class ChromedashFeatureDetail extends LitElement {
   static get properties() {
     return {
+      appTitle: {type: String},
       featureLinks: {type: Array},
       user: {type: Object},
       canEdit: {type: Boolean},
@@ -73,6 +74,7 @@ class ChromedashFeatureDetail extends LitElement {
 
   constructor() {
     super();
+    this.appTitle = '';
     this.user = {};
     this.canEdit = false;
     this.feature = {};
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-feature-detail.js
```diff
@@ -582,27 +587,44 @@ class ChromedashFeatureDetail extends LitElement {
...
+        <sl-button size="small"
+            href="/guide/stage/${this.feature.id}/${processStage.outgoing_stage}/${feStage.id}"
+            >Edit fields</sl-button>`;
+    }
+    // If we have an origin trial ID associated with the stage, add a link to the trial.
+    if (feStage.origin_trial_id) {
+      let originTrialsURL = `https://origintrials-staging.corp.google.com/origintrials/#/view_trials/${feStage.origin_trial_id}`;
+      // If this is the production host, link to the production OT site.
+      if (window.location.host === 'chromestatus.com') {
+        originTrialsURL = `https://developer.chrome.com/origintrials/#/view_trials/${feStage.origin_trial_id}`;
```
DanielRyanSmith, line 615: @jrobbins I wasn't sure of the best way to handle this differentiation of the staging and production OT URLs - maybe you have some opinions on this? Also, non-Googlers will not have access to the OT staging URL when developing, since it is a corp site. Hopefully this is still fine to reference in test environments?

jrobbins: I think checking window.location.host falls short because we want to be able to test on non-live GAE versions in the prod GAE project, which would have a host of HASH-dot-cr-status.appspot.com.  So, another way to do it would be to check the appTitle property that is passed down as far as  chromedash-feature-page but could be passed open more level to chromedash-feature-detail.

I think that it is OK to link from cr-status-staging to the OT staging site, even if non-googlers cannot follow the links.

jrobbins (approved):
LGTM after using appTitle property.

PR #3314, opened 2023-09-09
Record vote activities with amendments instead of text.

jrobbins:
We got feedback from reviewers via Dharani that the comments generated for votes look too much like human-generated comments.  The internal launch tool has separate lists for human comments and vote activities.  We can keep a unified list of activity on a gate, but the votes need to be more visually distinguished.  Also, we need an activity in the list for the initial review request.

I decided to keep the same CSS that we have, but record votes using activity amendments rather than comment content.  The amendments already have a distinct visual style because they are preceded by a dark triangle.  Also, they are much less verbose than the comment string that was being generated (which was more appropriate for the email).

In this PR:
* Start phasing out special treatment of "legacy" comments and comments_only mode.  See below.
* Fix the calculation of the old state during a vote state change.  We want that particular reviewer's vote, not the gate state.
* When notifying subscribers, create an Activity with an Amendment rather than comment content.
* Also create an Activity for the initial review request.

Legacy comments were ones generated in 2022 via the old approvals dialog box.  Those did not have a gate_id filled in, so they would have only been shown on the activity log page rather than also being shown in the specific gate where they belonged.  At the time that we launched the gate column, we wanted to show the legacy comments on the gate column, so we just displayed all textual comments that had no gate_id on every gate.   The impact of removing the legacy comments rule is that (1) old comments will only be found on the activity log page and (2) we are now able to show activities with amendments on a gate (even if they have no comment text).

### DIFF
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -61,6 +61,7 @@ class CommentsAPI(basehandlers.APIHandler):
     """Return a list of all review comments on the given feature."""
     feature_id = kwargs['feature_id']
     gate_id = kwargs.get('gate_id', None)
+    # TODO(jrobbins): Remove comments_only after next deployment.
     comments_only = self.get_bool_arg('comments_only')
     # Note: We assume that anyone may view approval comments.
     comments = Activity.get_activities(
--- a/api/comments_api.py
+++ b/api/comments_api.py
@@ -61,11 +61,8 @@ class CommentsAPI(basehandlers.APIHandler):
     """Return a list of all review comments on the given feature."""
     feature_id = kwargs['feature_id']
     gate_id = kwargs.get('gate_id', None)
-    # TODO(jrobbins): Remove comments_only after next deployment.
-    comments_only = self.get_bool_arg('comments_only')
     # Note: We assume that anyone may view approval comments.
-    comments = Activity.get_activities(
-        feature_id, gate_id, comments_only=comments_only)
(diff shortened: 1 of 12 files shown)


jrobbins:
Screenshot of activity in the gate column when the review is initially requested and one vote was cast:
image

Same activities after the user clicks on both triangles to expand details
image

For comparison, this is what the existing comment text looked liked:
image

api/comments_api.py
```diff
@@ -61,6 +61,7 @@ def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
     """Return a list of all review comments on the given feature."""
     feature_id = kwargs['feature_id']
     gate_id = kwargs.get('gate_id', None)
+    # TODO(jrobbins): Remove comments_only after next deployment.
     comments_only = self.get_bool_arg('comments_only')
```
KyleJu, line 65:
```suggestion
    comments_only = self.get_bool_arg('comments_only')
```
comments_only needs to be removed here, since it is removed on the client side. Same goes for`Activity.get_activities(
        feature_id, gate_id, comments_only=comments_only)`

jrobbins: OK, I removed this.   I had thought that it was needed by anyone who had the old SPA open in a browser tab because the old SPA would still send comments_only, but it is OK for the server to simply ignore it.

KyleJu (approved):

PR #3325, opened 2023-09-12
Add support for testing mobile (i.e. narrow window).

dlaliberte:
This PR adds support for testing in the mobile mode. 

 * Change how mobile mode is detected, to use window width rather than screen width
 * Add a floating test login button at the body level.
 * Add access to menu (chromedash-drawer) for logout and "Create feature" button.
 * Remember to await element.click() !!!


client-src/elements/chromedash-drawer.js
```diff
@@ -160,36 +163,36 @@ export class ChromedashDrawer extends LitElement {
...
 
   initializeTestingSignIn() {
-    if (this.devMode != 'True') {
-      return;
-    }
-
     // Create DEV_MODE login button for testing
     const signInTestingButton = document.createElement('button');
     signInTestingButton.innerText = 'Sign in as example@chromium.org';
+    signInTestingButton.setAttribute('type', 'button');
```
jrobbins, line 169: No need to do it in this PR, but it occurs to me that this testing button could be created using a lit-html approach like the way that regular product UI elements are created.

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
```diff
@@ -18,20 +19,30 @@ test.afterEach(async ({ page }) => {
...
+  await createFeatureButton.click();
+  if (mobile) {
+    await menuButton.click();  // To hide menu
+    delay(500);
+  }
 
   // Expect "Add a feature" header to be present.
   const addAFeatureHeader = page.getByTestId('add-a-feature');
-  await expect(addAFeatureHeader).toBeVisible();
+  await expect(addAFeatureHeader).toBeVisible({timeout:10000});
```
jrobbins, line 45: Why so many different timeout values at different places in the tests?  Would it be better to decide on a standard test step timeout?  Or maybe one for steps that require no API calls and a longer one for steps that do require API calls.

packages/playwright/tests/test_utils.js
```diff
@@ -160,11 +168,11 @@ export async function login(page) {
   // console.log('take a screenshot of header that should have "Create feature" button');
   await expect(page.getByTestId('header')).toHaveScreenshot('after-login-click.png');
 
-  // Check that we are logged in now.
-  // Expect a nav container to be present.
-  // This sometimes fails, even though the screenshot seems correct.
-  accountIndicator = page.getByTestId('account-indicator');
-  await expect(accountIndicator).toBeVisible({ timeout: loginTimeout });
+  // // Check that we are logged in now.
```
jrobbins, line 171: Double comment symbols.

jrobbins (approved):

client-src/elements/chromedash-drawer.js
dlaliberte: Yes, it could, and probably should.  It was good to start simple since there were so many unknowns.  

By the way, I will also probably eliminate the redundant screenshot checking for the login before each test, and perhaps figure out how to login just once for a set of tests.

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
dlaliberte: The varying timeouts make sense in some cases, since it takes a varying amount of time before some displays stabilize.   The default is 5000, but that often seems to not be enough.  Changing the default to, say, 10000 might be reasonable.    It seems to be required for this step, for example.

Early on, many steps were timing out, and I couldn't figure out why, and the turnaround for trying something different became very tedious.  Now that things have stabilized, I'm trying to incrementally reduce timeouts and eliminate delays.

PR #3327, opened 2023-09-13
Add 3 more fields for OT creation

DanielRyanSmith:
This change adds 3 more fields that are important for requesting the creation of an origin trial.

```
stage_description
ot_emails
ot_request_note
```

stage_description: This is written to be reused for any stage, but has a primary use of being the displayed description for the origin trial in the [OT Console](https://developer.chrome.com/origintrials/). This description often differs from the feature's description in ChromeStatus, and seemed to merit a new separate field.

ot_emails: A list of contacts that will be used when sending origin trial  related notifications and emails.

ot_request_note: The "Anything other questions or concerns?" field that will be prompted during the trial creation request form. Note that this field is not served later in Stage entity data requests, as it should be somewhat confidential compared to other stage fields.


client-src/elements/form-field-specs.js
```diff
@@ -1391,6 +1409,16 @@ export const ALL_FIELDS = {
     </ul>`,
   },
 
+  'stage_description': {
+    type: 'textarea',
+    required: true,
+    label: 'Description',
+    help_text: html`
+      <p>
+        A brief description of the feature (1-2 sentences). Displayed publicly.
```
jrobbins, line 1418: We should clarify how this relates to the main feature description.  I.e., is this the one displayed on the origin trial console?

What if we didn't have this field and just used the main feature description on the OT console?

jrobbins (approved):

client-src/elements/form-field-specs.js
DanielRyanSmith: > We should clarify how this relates to the main feature description. I.e., is this the one displayed on the origin trial console?

That's a good idea. I tried to make it more general description field in case we need to use it for other stages, but maybe it's best to make it more specific to origin trials. 🙂 

> What if we didn't have this field and just used the main feature description on the OT console?

After looking around at the active origin trials in production, it seems that this description is intended to be shorter and almost always varies from the description displayed in ChromeStatus, so I assumed we'd need a new field entirely

PR #3330, opened 2023-09-14
Hide enterprise features from /features page

yanndago:
Enterprise features were shown on /features when they should never be shown there.
This PR filters out enterprise features from the features sent to /features


jrobbins (approved):
Please add a PR description that briefly describes the problem being solved.

PR #3338, opened 2023-09-18
Fix base.css unix perms.

jrobbins:
This should avoid unexpectedly listing this file when making other PRs.


dlaliberte:
That (merging the origin/main into my branch) fixed the permissions for me.

jrobbins:
I think this worked.  I no longer see base.css listed as a modified file after running locally.

PR #3349, opened 2023-09-19
Add test of entering a blink_component name

dlaliberte:
This PR adds a single test of adding a blink_component name.  

* I discovered that (apparently) we cannot show the autocomplete datalist options programmatically.
* I also removed some login related tests, and screenshots of the "Create feature" button.
* Updated the playwright instructions in the README.


packages/playwright/tests/test_utils.js
```diff
@@ -164,7 +164,8 @@ export async function login(page) {
 
   // Take a screenshot of header that should have "Create feature" button.
   // console.log('take a screenshot of header that should have "Create feature" button');
-  await expect(page.getByTestId('header')).toHaveScreenshot('after-login-click.png');
+  // This is no longer very useful.
```
jrobbins, line 167: Is there a situation where a developer should uncomment this?  If not, can we just delete this line?

jrobbins (approved):

packages/playwright/tests/test_utils.js
dlaliberte: I guess not.  Easy to add again if we want.

PR #3350, opened 2023-09-20
Add "overdue" to the tooltip on overdue gate chips.

jrobbins:
Carol and Dharani have pointed out that the red clock icon is not obvious enough by itself.   I think users can figure it out because the same icon is used in the gate column with more text that explains it, but it would be helpful to add the word "overdue" to the tooltip on the gate chips as well.


jrobbins:
Before:
image

After:
image

Ignore the fact that a review was not even requested yet in these screenshots.  I just created a condition to show the overdue case without actually setting up realistic test data.

KyleJu (approved):

PR #3352, opened 2023-09-21
OT requests admin page

DanielRyanSmith:
This PR adds a new temporary page available only to site admins that displays any origin trial creation requests that have been submitted by users.

Screenshot 2023-09-21 at 4 26 29 PM

This is a temporary solution to allow for the existing manual process to continue. The data collected from the trial creation request is given so that it is easy to paste the data to the existing origin trials spreadsheet, with a "copy to clipboard" button available.

Requests should be cleared from this page when the origin trial is detected via the origin trial association cron job.

### DIFF
--- a/pages/ot_requests.py
+++ b/pages/ot_requests.py
@@ -1,0 +1,32 @@
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pages/ot_requests.py
```diff
@@ -0,0 +1,38 @@
...
+from internals.core_models import Stage
+
+from framework import basehandlers
+from framework import permissions
+
+
+class OriginTrialsRequests(basehandlers.FlaskHandler):
+  """Display any origin trials requests."""
+
+  TEMPLATE_PATH = 'admin/features/ot_requests.html'
```
jrobbins, line 25: Whaaa.... alright, it's temporary.

jrobbins (approved):

pages/ot_requests.py
DanielRyanSmith: lol, is this because it doesn't follow the SPA paradigm? If so, I agree with the sentiment 🙂 I wouldn't have chosen this route without knowing it was a temporary solution!

jrobbins: Yep, that was what I was thinking too.

PR #3354, opened 2023-09-22
Make certain trial creation request form fields required

DanielRyanSmith:
This PR updates some form fields on the OT creation request form to make certain fields required, as well as creating new form definitions for fields that should be required on the creation request form, but not on the field editor page. Namely:
```
ot_creation__intent_to_experiment_url
ot_creation__desktop_milestone_first
ot_creation__desktop_milestone_last
```

These fields are mostly identical to their original fields, except that they are required on the creation form page.

Additionally, some fields that really only belong on the OT creation request form have been removed from the typical fields displayed in the feature detail page.

Edit: This change has uncovered a small bug for the `r4dt_url`, and a fix for that has been added to this PR.


client-src/elements/chromedash-ot-creation-page.js
```diff
@@ -152,7 +152,7 @@ export class ChromedashOTCreationPage extends LitElement {
 
   renderFields(section) {
     const fields = section.fields.map(field => {
-      const value = getStageValue(this.stage, field);
+      let value = getStageValue(this.stage, field);
```
jrobbins, line 155: It's odd that this did not cause an error at build time.

jrobbins (approved):

client-src/elements/chromedash-ot-creation-page.js
DanielRyanSmith: I think this might also be from a git issue we spoke of before. This is causing runtime problems in the code now, but in a place that is not accessible to users. 😅

PR #3361, opened 2023-09-25
Add new prerequisites dialog for origin trial creation request form

DanielRyanSmith:
This change adds a dialog to the "Request Trial Creation" button that will display a list of prerequisite steps to complete before requesting the creation of an origin trial. The user can then select "Proceed" to move on to the creation form page.

Screenshot 2023-09-25 at 2 43 29 PM

### DIFF
--- a/client-src/elements/chromedash-feature-detail.js
+++ b/client-src/elements/chromedash-feature-detail.js
@@ -2,6 +2,7 @@ import {LitElement, css, html, nothing} from 'lit';
 import {getStageValue, renderHTMLIf} from './utils';
 import {enhanceUrl} from './feature-link';
 import {openAddStageDialog} from './chromedash-add-stage-dialog';
+import {openPrereqsDialog} from './chromedash-ot-create-prereqs-dialog';
 import {makeDisplaySpecs} from './form-field-specs';
 import {
   FLAT_ENTERPRISE_METADATA_FIELDS,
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-ot-create-prereqs-dialog.js
```diff
@@ -0,0 +1,118 @@
...
+        <br>
+        <p>
+          If you haven't already, please review the docs for
+          <a href="https://www.chromium.org/blink/origin-trials/running-an-origin-trial/">
+            running an origin trial
+          </a>.
+          If you have any further questions, contact us at origin-trials-discuss@google.com.
+        </p>
+        <br>
+        <sl-button id="continue-button" variant="primary"
```
jrobbins, line 109: Why use an @click instead of the href="" attribute?

client-src/elements/chromedash-feature-detail.js
```diff
@@ -2,6 +2,7 @@ import {LitElement, css, html, nothing} from 'lit';
 import {getStageValue, renderHTMLIf} from './utils';
 import {enhanceUrl} from './feature-link';
 import {openAddStageDialog} from './chromedash-add-stage-dialog';
+// import {openPrereqsDialog} from './chromedash-ot-create-prereqs-dialog';
```
jrobbins, line 5: Don't you need this import to not be commented out?

client-src/elements/chromedash-ot-create-prereqs-dialog.js
jrobbins: Ah, good point.  Yes, the dialog should close.  If @click does that then keep it.

client-src/elements/chromedash-feature-detail.js
jrobbins: OK, I missed the fact that the code below was in a block comment because GitHub did not display that part or take it into account for syntax coloring.

jrobbins (approved):

PR #3368, opened 2023-09-27
Correct the subject prefix for PSA shipping emails.

jrobbins:
Dharani brought to my attention a feature with type "Web-developer-facing change to existing behavior" that had an "Intent to Ship" thread and so people were looking for the privacy and security gates.  However, according to launching-features, that type of feature only requires a PSA rather than an I2S.   The email prefix was wrong.

In this PR:
* Fix the PSA action name in the process definition.
* Fix the PSA subject prefix in the intent email template

### DIFF
--- a/internals/processes.py
+++ b/internals/processes.py
@@ -432,7 +432,7 @@ PSA_ONLY_STAGES = [
        PI_UPDATED_TARGET_MILESTONE,
        PI_I2S_EMAIL,
       ],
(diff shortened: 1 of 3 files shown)


jyasskin (approved):
This matches the instructions at https://www.chromium.org/blink/launching-features/#psa-prepare-to-ship, and @chrishtr in case the API owners actually want the instructions to say something else.

KyleJu (approved):

PR #3370, opened 2023-09-27
Origin Trial creation visible for all Chromium/Google users

DanielRyanSmith:
This PR allows any users signed in with `@google.com` or `@chromium.org` accounts to request the creation of an origin trial.


api/features_api.py
```diff
@@ -348,11 +348,16 @@ def do_patch(self, **kwargs):
     if not feature:
       self.abort(400, msg=f'Feature not found for ID {feature_id}')
 
-    # Validate the user has edit permissions and redirect if needed.
-    redirect_resp = permissions.validate_feature_edit_permission(
-        self, feature_id)
-    if redirect_resp:
-      return redirect_resp
+
+    # If submitting an OT creation request, the user must have feature edit
```
jrobbins, line 352: Where's the if-statement condition that narrows this to just OT creation requests?

DanielRyanSmith: I'm going to update this to use the Stages API. It's difficult to discern when an OT creation request comes in, and the logic starts to get muddied here in the Features API. I'll update you when it's ready 🙂

DanielRyanSmith: Okay, I've updated the logic to work in the Stages API rather than the Features API. 👍

jrobbins (approved):

PR #3379, opened 2023-09-29
Reword preflight intro.  Make pre* dialogs more consistent.

### LINKED ISSUE #3378: Change "you should first do the following" to "it is strongly recommended that you do the following"
In this flow:

image

The flow allows you to send the email even if these are not complete. This is on purpose. Therefore the text in the flow should align with this design by using "strongly recommend".

jrobbins:
This should resolve #3378.

In this PR:
* Change the wording to "it is strongly recommended"
* Change closeDialog() to hide() for consistency
* Make the prevote dialog items clickable links that open the relevant other gates.
* Make the buttons at the bottom of the two dialogs consistent:
    *  First button is orange, uses the word "yet", and backs out of the dialog without doing anything.
    * Second button (in float layout) is white and uses the word "anyway" to indicate that the user is proceeding despite the prerequisites.

### DIFF
--- a/client-src/elements/chromedash-preflight-dialog.js
+++ b/client-src/elements/chromedash-preflight-dialog.js
@@ -103,6 +103,12 @@ export class ChromedashPreflightDialog extends LitElement {
       .done:hover .edit-progress-item {
         visibility: visible;
       }
+
+      sl-button {
+        float: right;
+        margin: var(--content-padding-half);
+      }
+
     `];
   }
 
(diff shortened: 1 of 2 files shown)


jrobbins:
Old preflight dialog:
image

New preflight dialog (with different data so items are not exactly the same):
image

Old prevote dialog:
image

New prevote dialog (with different data):
image

DanielRyanSmith (approved):

PR #3408, opened 2023-10-13
Prompt users to switch to a paired account that can edit or review.

jrobbins:
There have been a couple of recent email threads with feature owners where it looked like the user could not edit or request a review because they were signed into the wrong account.  Googlers who work on chrome often have both a @google.com and a @chromium.org account and can be signed into the wrong one.  So, now we prompt users to switch to the right one.

In this PR:
* In permissions_api.py: add the concept of a `paired_user` and get its permissions.
* In chromedash-feature-page.js: add logic to detect two specific cases of using the wrong account, and display all applicable warnings.

### DIFF
--- a/api/permissions_api.py
+++ b/api/permissions_api.py
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from framework import basehandlers
 from framework import permissions
+from framework.users import User
 from internals import approval_defs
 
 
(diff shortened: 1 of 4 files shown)


jrobbins:
Screenshot:
image

api/permissions_api.py
```diff
@@ -25,22 +25,37 @@ class PermissionsAPI(basehandlers.APIHandler):
...
+      'is_admin': permissions.can_admin_site(user),
+      'email': user.email(),
+      'editable_features': permissions.feature_edit_list(user)
+    }
+
+  def find_paired_user(self, user):
+    """If @google.com or @chromium.org, return the other one."""
+    username, domain = user.email().split('@', 1)
+    if domain == 'google.com':
+      return User(email=username + '@chromium.org')
```
DanielRyanSmith, line 58: Would it be feasible/useful to check if a matching user exists with a @chromium.org account if the user is using any non-chromium.org email? Ex. @gmail.com or @microsoft.com? Just curious

DanielRyanSmith (approved):
LGTM with 1 question

api/permissions_api.py
jrobbins: It's possible.  We could add that if users request it.  However, I don't think that very many chrome contributors mistakenly use their @gmail.com accounts on chromestatus.  I think that people are aware of which accounts they are using and know the difference between work and personal accounts, but the chromium and google accounts are both work accounts with similar purposes.

PR #3410, opened 2023-10-13
Upgrade playwright to 1.37.

jrobbins:
When trying to run playwright tests, one error message told me to upgrade this version.  I have not gotten it to work locally, so let's see if it works in CI.


dlaliberte (approved):
I thought we had enabled auto updating of playwright.

jrobbins:
> I thought we had enabled auto updating of playwright.

I don't really know about it.  I am just doing what I thought an error message told me to do.
How could auto-updating work when the version number is specified in a source file that we have in git?
Dependbot?

PR #3412, opened 2023-10-16
Add prereqs dialog for extension request

DanielRyanSmith:
This PR updates the `chromedash-ot-create-prereqs-dialog` component to `chromedash-ot-prereqs-dialog` to be more generic and supply prerequisite dialogs for any origin trial request. The prerequisite that is displayed is changed based on the constant that is supplied to the function that opens the dialog.

Additionally, a prereqs dialog has been added for trial extensions that will be displayed before filling out an extension request form, but is not yet made visible.

Screenshot 2023-10-16 at 1 38 34 PM


DanielRyanSmith:
It seems an unrelated error is occurring with the test `test_google_docs_url`

jrobbins (approved):
Looks good. 

You might want to rebase and then check the diff again because I think some of the off-topic changes shown now should disappear.

PR #3424, opened 2023-10-18
suppress "annotation-unchecked" mypy notes

DanielRyanSmith:
An update of mypy came with a series of notes reminding the user that untyped functions are not checked. This change suppresses those notes so that relevant mypy issues are not lost in these reminders.
Screenshot 2023-10-17 at 5 31 48 PM


jcscottiii (approved):
LGTM.

I'm getting nervous about the number of arguments we are passing to this command. If we have to make another change to this, we should move to the mypy ini [file](https://mypy.readthedocs.io/en/stable/config_file.html). That will help with readability.

PR #3427, opened 2023-10-19
Remove unneeded install command.

jrobbins:
The tests run for me without this command.  On chat, you said that it was not needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,13 @@ updates:
     reviewers:
       - "kyleju"
       - "jrobbins"
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      # Prefix all commit messages with "docker"
+      prefix: "docker"
+    reviewers:
+      - "jrobbins"
+      - "jcscottiii"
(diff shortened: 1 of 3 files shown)


dlaliberte:
I see that the Playwright CI tests are failing.  Looks like maybe this install command is relevant in that context?  If so, can we move the install command to only be used in the CI test?

jcscottiii (approved):

dlaliberte:
We used to have a comment in the "//pwtests" pseudo command like this (copied from bikeshed):
    "//pwtests": "The version of the docker playwright image used here must exactly match the versions of @playwright/test in devDependencies and the Github Actions configuration (in playwright.yml).  The version of playwright used in devDependencies should not include the ^ prefix, so it doesn't auto-upgrade and get out of sync.",

I had thought that we made changes to avoid need for this manual updating, but it seems like that must not be the case.  To avoid this problem in the future, maybe we should add this "//pwtests" commend back in.

dlaliberte:
> We used to have a comment in the "//pwtests" pseudo command like this (copied from bikeshed).

BTW, in case people missed it, in my last PR I added a "//pwtests" comment to the packages/playwright/packages.json:
```
        "//pwtests": "The version of playwright used here in devDependencies:@playwright/test 
    must be the same as the docker image version used in Dockerfile, 
    so the '^' prefix should not be included to avoid auto-upgrade of playwright 
    which would then be out of sync.",
```

Issue #2977, opened 2023-05-04
Web test logs display errors but are not failing


DanielRyanSmith:
When running `npm run webtest` on main branch, these logs appear:

```
 🚧 Browser logs:
      An error was thrown in a Promise outside a test. Did you forget to await a function or assertion?
 🚧 Browser logs on Chromium:
      TypeError: Cannot read properties of null (reading 'showMessage')
        at showToastMessage (client-src/elements/utils.js:19:11)
        at client-src/elements/chromedash-activity-page.js:79:7
 🚧 Browser logs on Firefox:
      TypeError: toastEl is null
        at showToastMessage (client-src/elements/utils.js:19:3)
        at fetchComments/< (client-src/elements/chromedash-activity-page.js:79:23)
        at promise callback*fetchComments (client-src/elements/chromedash-activity-page.js:78:8)
        at connectedCallback (client-src/elements/chromedash-activity-page.js:67:10)
        at S (node_modules/lit-html/src/lit-html.ts:1402:52)
        at T (node_modules/lit-html/src/lit-html.ts:1445:35)
        at $ (node_modules/lit-html/src/lit-html.ts:1551:11)
        at _$AI (node_modules/lit-html/src/lit-html.ts:1375:11)
        at B (node_modules/lit-html/src/lit-html.ts:2172:7)
        at litFixtureSync (node_modules/@open-wc/testing-helpers/src/litFixture.js:32:9)
        at litFixture (node_modules/@open-wc/testing-helpers/src/litFixture.js:63:28)
        at fixture (node_modules/@open-wc/testing-helpers/src/fixture-no-side-effect.js:62:12)
        at client-src/elements/chromedash-activity-page_test.js:21:29
```

The tests do not actually fail, so whatever change is causing this was merged without being flagged. The errors need to be investigated to determine the root cause and fixed.

DanielRyanSmith:
This problem continues to worsen. Running `npm run webtest` today on the main branch looks like this:

```
 🚧 Browser logs:
      An error was thrown in a Promise outside a test. Did you forget to await a function or assertion?
 🚧 Browser logs on Chromium:
      TypeError: Cannot read properties of null (reading 'showMessage')
        at showToastMessage (client-src/elements/utils.js:31:11)
        at client-src/elements/chromedash-activity-page.js:79:7
 🚧 Browser logs on Firefox:
      TypeError: toastEl is null
        at showToastMessage (client-src/elements/utils.js:31:3)
[...]

PR #3449, opened 2023-10-25
Add initial semantic check, and fix some web tests

dlaliberte:
This PR adds initial support for semantic checks for form fields.  
Currently, only a single field value can be checked.
This is to partially address #3452. 

Here is what it looks like:

image

Several miscellaneous errors that are visible after running webtest, but without causing test failures (reported in #2977), are fixed here as well.  Errors about the unknown feature_links urls remain.

Also, in the unit tests, I couldn't figure out how to make a change to the value and await the change.  Seems that `await component.updateComplete` should do it, but it didn't work.  I'll work on this in future PRs.


dlaliberte:
I've added the fieldValues array to each of the form fields so that each form field can get the values of other fields.  I still need to add a way to notify a field when another field  that it depends on is changed to rerun the checks.

client-src/elements/chromedash-form-field.js
```diff
@@ -2,52 +2,66 @@ import {LitElement, html, nothing} from 'lit';
...
     this.loading = false;
     this.forEnterprise = false;
     this.stageType = undefined;
     this.componentChoices = {};
+    this.checkMessage = '';
+  }
+
+
+  getValue() {
+    return !this.value && this.fieldProps.initial ?
```
jrobbins, line 43: Could this be `this.value || this.fieldProps.iniitial`?
Is there a logical difference?

client-src/elements/chromedash-form-field.js
```diff
@@ -81,6 +95,7 @@ export class ChromedashFormField extends LitElement {
     } else {
       fieldValue = e.target.value;
     }
+    this.value = fieldValue; // Should we do this?
```
jrobbins, line 98: Is this comment resolved now?  It seems like you are doing it.

client-src/elements/chromedash-form-field.js
```diff
@@ -91,15 +106,49 @@ export class ChromedashFormField extends LitElement {
...
+            ${
+              checkResult.message ? checkResult.message :
+              checkResult.warning ? html`<b>Warning</b>: ${checkResult.warning}` :
+              checkResult.error ? html`<b>Error</b>: ${checkResult.error}` :
+              ''
+            }
+          </span>`;
+
+        const formFieldElement = this.renderRoot.querySelector(`#id_${this.name}`);
+        if (formFieldElement && formFieldElement.setCustomValidity &&
```
jrobbins, line 134: Th first two parts of this condition could be combined to a little shorter as `formFieldElement?.setCustomValidity`

jrobbins (commented):
Some initial comments.

client-src/elements/chromedash-form-field.js
dlaliberte: I copied that code that used to be in renderWidget, to be on the safe side.  I'm not sure what it should be, but what you suggest seems reasonable.  The logic might involve whether a value is 0 or null.  We should allow 0 values, so perhaps it should be:

  this.value != null ? this.value : this.fieldProps.initial

client-src/elements/chromedash-form-field.js
dlaliberte: The comment was a reminder to myself to be careful about doing this.  We currently don't change the value property, but is there a reason why we should not do so?  It is a dynamic property, so there could potentially be an update loop, for example.  But since this is happening in the event handler for the last update, perhaps the loop is avoided?  I'm not sure.  Seems to work fine with the text fields I am checking now, but maybe other types of fields will behave differently.

client-src/elements/chromedash-form-field.js
dlaliberte: Right.  Good point.

client-src/elements/chromedash-form-field.js
jrobbins: OK, that looks like it would work.

client-src/elements/chromedash-form-field.js
jrobbins: I suppose the only way is to test each field type.

dlaliberte: Added TODO to the comment.  I'll keep an eye on this.

dlaliberte:
> I've added the fieldValues array to each of the form fields so that each form field can get the values of other fields. I still need to add a way to notify a field when another field that it depends on is changed to rerun the checks.

Doing checks with multiple fields is working now.  The origin trials desktop start and end milestones must be in order:

image

I will add a visual test of this, since it will be cumbersome to test in unit tests because, currently, the code maintains a registry of all form field components at the app level.

client-src/js-src/cs-client.js
```diff
@@ -103,24 +103,24 @@ class ChromeStatusClient {
     return JSON.parse(rawResponseText.substr(XSSIPrefix.length));
   }
 
-  doGet(resource, body) {
+  async doGet(resource, body) {
```
jrobbins, line 106: Is there a reason why `async` is needed or is this just to clarify that each of these functions returns a promise?

jrobbins (commented):
Why are there so many new and modified screenshots?  There seem to be more than the ones created by your new tests.

dlaliberte:
I can't seem to reliably delete a feature, let alone take a screenshot of the result.  So I'll avoid feature deletion for now, but resetting the database is useful for avoiding the accumulation of test features.

client-src/js-src/cs-client.js
dlaliberte: This is needed to avoid *some* of the errors after webtest runs.  More remain, but I am not sure how to fix them.

dlaliberte:
> Why are there so many new and modified screenshots? There seem to be more than the ones created by your new tests.

I replaced a '-' with '/' in the screenshot paths, so they will be more readable. I could have done this in a separate PR, but oh well.

jrobbins (approved):

PR #3451, opened 2023-10-26
Let feature owners request an N/A.

jrobbins:
Some features have a small scope to the extent that the feature owner is confident that there is no possible concern about privacy, security, etc.  For these cases, the feature owners have always been able to post a comment asking for an "N/A" response and then request a review.   However, feature owners find that unintuitive because the button specifically says "request review" whereas they are actually asking to avoid a review.

In this PR:
* Implement a second button at the top of the gate column that says "Request N/A".
* Clicking that button prompts for a rationale comment, and then posts a vote NA_REQUESTED.
* If the only vote on a gate is NA_REQUESTED, then the gate state becomes NA_REQUESTED.
* If a gate has state NA_REQUESTED, show the gate as blue for pending with prefix "N/A?".

### DIFF
--- a/client-src/elements/chromedash-gate-column.js
+++ b/client-src/elements/chromedash-gate-column.js
@@ -16,6 +16,7 @@ import {GATE_QUESTIONNAIRES} from './form-definition.js';
 import {
   GATE_PREPARING,
   GATE_REVIEW_REQUESTED,
+  GATE_NA_REQUESTED,
   VOTE_OPTIONS,
 } from './form-field-enums';
 
@@ -26,6 +27,8 @@ export class ChromedashGateColumn extends LitElement {
   voteSelectRef = createRef();
   commentAreaRef = createRef();
   postToThreadRef = createRef();
+  rationaleDialogRef = createRef();
+  rationaleRef = createRef();
 
   static get styles() {
     return [
@@ -249,11 +252,15 @@ export class ChromedashGateColumn extends LitElement {
   }
 
   handlePost() {
-    this.submittingVote = true;
     const commentArea = this.commentAreaRef.value;
     const commentText = commentArea.value.trim();
     const postToThreadType = (
-      this.postToThreadRef.value?.checked ? this.gate.gate_type : 0);
+        this.postToThreadRef.value?.checked ? this.gate.gate_type : 0);
+    this.postComment(commentText, postToThreadType);
+  }
+
+  postComment(commentText, postToThreadType=0) {
+    this.submittingVote = true;
     if (commentText != '') {
       window.csClient.postComment(
         this.feature.id, this.gate.id, commentText,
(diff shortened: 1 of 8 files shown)


jrobbins:
This feature was requested by the API Owners at this morning's meeting, and I discussed it with Dharani.

jrobbins:
Screenshot of the buttons:
image

Screenshot of a gate chip in state NA_REQUESTED:
image

client-src/elements/chromedash-gate-column.js
```diff
@@ -348,6 +355,22 @@ export class ChromedashGateColumn extends LitElement {
       });
   }
 
+  handleNARequested() {
+    this.rationaleDialogRef.value.show();
+  }
+
+  handleNARequestSubmitted() {
+    const commentText = ('An "N/A" response is requested because: ' +
```
KyleJu, line 363:
```suggestion
    const commentText = ('A "N/A" response is requested because: ' +
```

client-src/elements/chromedash-gate-column.js
```diff
@@ -411,16 +434,37 @@ export class ChromedashGateColumn extends LitElement {
         this.renderAction(processStage, act));
     }
 
+    const dialog = html`
+      <sl-dialog ${ref(this.rationaleDialogRef)} label="Request an N/A">
```
KyleJu, line 438:
```suggestion
      <sl-dialog ${ref(this.rationaleDialogRef)} label="Request a N/A">
```

internals/notifier_helpers.py
```diff
@@ -80,10 +80,11 @@ def notify_subscribers_and_save_amendments(fe: 'FeatureEntry',
 
 
 def notify_approvers_of_reviews(
-    fe: 'FeatureEntry', gate: Gate, email: str) -> None:
+    fe: 'FeatureEntry', gate: Gate, new_state: int, email: str) -> None:
   """Notify approvers of a review requested from a Gate."""
+  new_value_str = Vote.VOTE_VALUES.get(new_state, 'None')
   amendment = Amendment(field_name='review_status',
-                        old_value='None', new_value='review_requested')
+                        old_value='None', new_value=new_value_str)
```
KyleJu, line 87: line 95 needs to be updated 
`
changed_props = {
      'prop_name': 'Review status change in %s' % (gate_url),
      'old_val': 'na',
      'new_val': 'review_requested',
  }
`

KyleJu (approved):
LGTM after addressing the comments

jrobbins:
I accepted your `an` to `a` changes, but then I double-checked and I believe that `an` is correct.  So, I will change those back.
https://www.grammar.com/a-vs-an-when-to-use

KyleJu:
> I accepted your `an` to `a` changes, but then I double-checked and I believe that `an` is correct. So, I will change those back. https://www.grammar.com/a-vs-an-when-to-use

Aha that is something new to me. Thanks!

internals/notifier_helpers.py
jrobbins: Changed.

PR #3470, opened 2023-10-31
Convert intent-to-implement.html django template into lit element template

dlaliberte:
Note, this also includes the estimated-milestones-table.html.


dlaliberte:
Here is what it looks like so far:

image

jrobbins:
Daniel Smith has implemented this.

Issue #3468, opened 2023-10-30
Process seems stuck at Web Face Change PSA email


schenney-chromium:
Use this template to provide feedback on the new Blink launch process and Guide UX for editing features.

**Process Feedback**
I'm very confused by the steps for a web facing change, or maybe all changes. I drafted the Web Facing Change PSA and sent it, but there's no way to enter a link for this and now no way to craft an Intent To Ship email. I think I still need the Intent to Ship, but maybe the Web Facing Change PSA is the intent to ship in this case. Either way I'm confused.

**Guide UX Feedback**
There's a bug when you request feedback for, say Testing, then request feedback, the sidebar overlay switches back to "Privacy" with the Privacy instructions etc. This applies to any of the not-Privacy feedback overlays.

schenney-chromium:
This is the feature I'm working on. https://chromestatus.com/guide/edit/5090853643354112

It's a bit of a mess because it was not set up as a feature before being implemented.

jrobbins:
Regarding the process:
The "Web developer facing code change" type of feature entries normally only generate a PSA.  Since a review is desired in this case, I went ahead and voted "review started" to get it to show up on the reviewers' dashboard.

Sorry if feature entry type choices are unclear.  We recently changed the wording to make them more clear, but I'll ask Chris H if more changes are needed.


Regarding the UX feedback:
I haven't been able to repro that behavior.
I created a test feature entry on our staging server to try to repro it there.
https://cr-status-staging.appspot.com/feature/5667111777599488
Are you seeing the same behavior there?

schenney-chromium:
Do you want me to edit that test feature to try to repro the issue?

jrobbins:
Sure, you are welcome to edit anything on that staging server.

I figured out one way to repro the problem.  It happens when the user start by following a URL from one of the notification emails that has one gate ID in that URL.  E.g., https://cr-status-staging.appspot.com/feature/5667111777599488?gate=6241369507495936

schenney-chromium:
That would be my case. I have this url: https://chromestatus.com/feature/5090853643354112?gate=5157511535788032

I originally started with https://chromestatus.com/guide/verify_accuracy/5090853643354112 if for some reason that matters.

PR #3471, opened 2023-10-31
Fix undesired navigation to previous gate column.

jrobbins:
This should resolve the UX aspect of #3468.

The user had followed a link from a notification email that included the `gate=` query string parameter for gate A.  Then he navigated to a different gate B chip and requested a review there.  Then, surprisingly, the gate column showed gate A again.

In this PR:
* In chromedash-gate-chip.js, update the `gate=` query string parameter whenever the user clicks a gate chip.  This gives a bookmarkable URL that opens the current gate for any user that visits it rather than keeping the ID of the original gate.
* In the other files, stop passing in the rawQuery object that was parsed when a page.js navigation event occurred because it can be stale.  Instead, parse any query string parameters directly from `window.location.search`.


KyleJu (approved):
Aha that is a good fix! Thanks

PR #3474, opened 2023-11-03
Add API to update Gate assignees and notify them.

jrobbins:
This implements the server-side support for allowing users to assign a review.  A user who can edit the feature or a reviewer can set a list of assigned reviewers for a gate.  (In the UI, we will probably limit it to just one assigned reviewer at a time, but the API can support multiple.). The email address provided for each assigned reviewer must be a user who has permission to vote on that gate.  If `gate.assignee_emails` is successful set, it is stored and an email notification is triggered.

In this PR:
* reviews_api: Add the POST handler with permission checks and input validation.  Also, refactor some shared parameter parsing.
* cs-client: Implement a JS function to make the API call.  The old ApprovalConfig stuff was outdated and not used since the approvals dialog box was phased out 6 months ago.
* notifier:
   * Generalize `format_email_body` to allow for more template types, even though I am not making use of that yet.
   * Also allow `updated_email` to be passed in, because the user who edited the gate is not necessarily the updater of the feature.
   * Add handler and helper function to send the email
* main: register handlers for the new API call and the email sending task


api/reviews_api.py
```diff
@@ -16,15 +16,34 @@
...
 from framework import permissions
+from framework.users import User
 from internals import approval_defs, notifier_helpers
+from internals.core_models import FeatureEntry
 from internals.review_models import Gate, Vote
 
 
+def get_user_feature_and_gate(handler, kwargs) -> Tuple[
+    User, FeatureEntry, Gate, int, int]:
+  """Get common parameters from the the request."""
```
KyleJu, line 32:
```suggestion
  """Get common parameters from the request."""
```

api/reviews_api.py
```diff
@@ -115,3 +126,38 @@ def do_get(self, **kwargs) -> dict[str, Any]:
...
+      self.abort(403, msg='User lacks permission to assign reviews')
+
+  def validate_assignees(self, assignees, fe, gate):
+    """Each assignee must have permission to review."""
+    approvers = approval_defs.get_approvers(gate.gate_type)
+    for email in assignees:
+      reviewer = User(email=email)
+      is_approver = permissions.can_review_gate(reviewer, fe, gate, approvers)
+      if not is_approver:
+        self.abort(400, 'Assignee is not a reviewer')
```
KyleJu, line 163: nit: Maybe log the person who isn't  an assignee?

KyleJu (approved):
LGTM with nits

api/reviews_api.py
jrobbins: Fixed

api/reviews_api.py
jrobbins: All the POST data is logged in APIHandler.post() before this handler is called.

Issue #3028, opened 2023-05-19
Implement UI for suggesting a reviewer


jrobbins:
A feature editor or reviewer should be able to suggest a specific reviewer out of the pool of possible reviewers on a given gate.
As per mocks from our UX consultant, we should add a "Suggest a reviewer" (borderless) button below the votes.

It should open a small dialog box or in-line editing area with the `<select>` to choose a reviewer.

A dialog box would give more room for in-product help text that explains that suggesting a reviewer simply lists them and generates an email to them.  However, another reviewer could still respond to the review request.

PR #3475, opened 2023-11-04
Display and edit gate assigned reviewer

jrobbins:
This is progress on #3028.

This displays the assigned reviewers on a review gate.  Each reviewer is displayed as a row in the votes table, but since they haven't voted yet, it says "No response yet".  For the row where the user can enter their vote, it now displays their user name if they are an assigned reviewer, otherwise it still says "Awaiting review" (because users interpreted seeing their name as an indication that they were assigned).

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -113,18 +113,15 @@ class GatesAPI(basehandlers.APIHandler):
     if len(gates) == 0:
       return {
           'gates': [],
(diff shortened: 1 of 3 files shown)


jrobbins:
Screenshot when the current user is not one of the assigned reviewers:
image

Screenshot when the current user is one of the assigned reviewers:
image

KyleJu (approved):

PR #3476, opened 2023-11-06
Auto-assign reviewers on gates that have rotations.

jrobbins:
This is the last main piece of the assigned reviewers functionality.

If a rotations_url is configured in datastore, that URL is used when a user requests a review.  It gets the list of on-call user emails and assigns the review to them.  (Usually it would always be a list of one person.)


KyleJu (approved):

jrobbins:
Thanks for the review.  I'm going to add a little bit more logic, but the gist is the same, so probably no need to re-review.

PR #3485, opened 2023-11-09
Move debuggability reviewers into NDB.

jrobbins:
Now that I have assigned reviewers and support for rotations working, the Debuggability team also wants to have the reviewers assigned by their rotation.

In this PR:
* Move the debuggability reviewer list out of the source code and in to NDB.
* Change how failure to fetch the on-call person is handled: simply don't auto-assign instead of aborting the whole request.

### DIFF
--- a/internals/approval_defs.py
+++ b/internals/approval_defs.py
@@ -44,13 +44,7 @@ ENTERPRISE_APPROVERS = [
     'davidayad@google.com',
     'bheenan@google.com',
 ]
-DEBUGGABILITY_APPROVERS = [
-    'bmeurer@google.com',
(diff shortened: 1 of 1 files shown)


jrobbins:
Note: I already submitted CL 580271712 to enable access to the debuggability rotation on-call person's email address.  It has not been deployed yet, but regardless of that, this PR can be reviewed, merged, and deployed to chromestatus on our usual schedule.  The change to simply `return` when the rotation_url request gives a 404 means that the auto-assignments will just be a no-op until my CL gets deployed.

KyleJu (approved):

PR #3520, opened 2023-11-28
Replace omahaproxy with version history.googleapis.com

jrobbins:
The backend that chromestatus previously used was phased out and we were supposed to migrate to this other backend server instead.  The OmahaProxy backend gave info about which version of chrome was live on each channel.

In this PR:
+ Rewrite fetchchannel.py to hit the new backend, but keep the format of the returned/cached string basically the same to avoid needing to change other code

### DIFF
--- a/internals/fetchchannels.py
+++ b/internals/fetchchannels.py
@@ -24,6 +24,18 @@ OMAHA_URL_TEMPLATE = (
     'https://versionhistory.googleapis.com'
     '/v1/chrome/platforms/win/channels/%s/versions/?pageSize=1')
 
+# Expected response for the "stable" channel looks like:
+# {
+#  "versions": [
+#   {
+#    "name": "chrome/platforms/android/channels/stable/versions/119.0.6045.163",
+#    "version": "119.0.6045.163"
+#   }
+#  ],
+#  "nextPageToken": "202090764"
+# }
+# We really only need the version string.
+
 
 def get_channel_version(channel):
   """Return the version string that is live on the given channel."""
--- a/internals/fetchchannels.py
+++ b/internals/fetchchannels.py
@@ -24,27 +24,30 @@ OMAHA_URL_TEMPLATE = (
     'https://versionhistory.googleapis.com'
     '/v1/chrome/platforms/win/channels/%s/versions/?pageSize=1')
 
-def get_channel_info(channel):
+
+def get_channel_version(channel):
+  """Return the version string that is live on the given channel."""
   url = OMAHA_URL_TEMPLATE % channel
   logging.info('fetching %s' % url)
   result = requests.get(url)
   if result.status_code != 200:
     logging.info('Could not fetch channel info')
-    return None
+    return '0.0'
   channel_info = json.loads(result.content)
   logging.info('channel_info is %r', channel_info)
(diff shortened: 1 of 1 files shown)


internals/fetchchannels.py
```diff
@@ -14,17 +14,42 @@
...
+    '/v1/chrome/platforms/win/channels/%s/versions/?pageSize=1')
+
+
+def get_channel_version(channel):
+  """Return the version string that is live on the given channel."""
+  url = OMAHA_URL_TEMPLATE % channel
+  logging.info('fetching %s' % url)
+  result = requests.get(url)
+  if result.status_code != 200:
+    logging.info('Could not fetch channel info')
```
KyleJu, line 34: Print the channel here for debugging

KyleJu (approved):

internals/fetchchannels.py
```diff
@@ -14,17 +14,42 @@
...
 import json
+import logging
 import requests
 
 from framework import rediscache
 # Note: this file cannot import core_models because it would be circular.
 
+OMAHA_URL_TEMPLATE = (
+    'https://versionhistory.googleapis.com'
+    '/v1/chrome/platforms/win/channels/%s/versions/?pageSize=1')
```
jcscottiii, line 25: Just one quick question:

Is it safe to hardcode windows (`win`) for the platform?

For example: I get different results between `https://versionhistory.googleapis.com/v1/chrome/platforms/linux/channels/all/versions/?pageSize=1` and `https://versionhistory.googleapis.com/v1/chrome/platforms/win/channels/all/versions/?pageSize=1`


For windows:
```
{
  "versions": [
    {
      "name": "chrome/platforms/win/channels/extended/versions/118.0.5993.144",
      "version": "118.0.5993.144"
    }
  ],
  "nextPageToken": "2741885587"
}
```

For linux:
```
{
  "versions": [
    {
      "name": "chrome/platforms/linux/channels/stable/versions/119.0.6045.159",
      "version": "119.0.6045.159"
    }
  ],
  "nextPageToken": "69122120"
}
```

internals/fetchchannels.py
jrobbins: Done

internals/fetchchannels.py
```diff
@@ -14,17 +14,42 @@
...
+  """Return the version string that is live on the given channel."""
+  url = OMAHA_URL_TEMPLATE % channel
+  logging.info('fetching %s' % url)
+  result = requests.get(url)
+  if result.status_code != 200:
+    logging.info('Could not fetch channel info')
+    return '0.0'
+  channel_info = json.loads(result.content)
+  logging.info('channel_info is %r', channel_info)
+  version = channel_info['versions'][0]['version']
```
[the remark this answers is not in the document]
jrobbins: Done

internals/fetchchannels.py
jrobbins: I think that the old code only looked at the win platform, so I am keeping that as-is.

In your example, the win platform actually gave you back the extended channel rather than the stable channel.  In my implementation, I specify which channels we want.

jcscottiii: Makes sense!

jcscottiii (approved):

PR #3526, opened 2023-12-02
Fix for semantic  check complaining when resetting numeric field to empty string

### LINKED ISSUE #3525: Once a milestone range is set, it cannot be cleared
**Describe the bug**
I tried to remove the OT milestones for a feature entry and found that it complained that the end needed to be after the start.

**To Reproduce**
1. Go to a feature entry that has start and end OT milestones set
2. Clear out both fields
3. Submit
4. See field error on both fields, preventing submission.

**Expected behavior**
It should be allowed to have one of both fields empty, even if they were previously set.

dlaliberte:
This PR fixes #3525 by checking for empty string explicitly to return an undefined value rather than converting it to 0.

### DIFF
--- a/client-src/elements/utils.js
+++ b/client-src/elements/utils.js
@@ -394,7 +394,6 @@ export function checkMilestoneStartEnd(startEndPair, getFieldValue) {
   const {start, end} = startEndPair;
   const startMilestone = getValue(start);
   const endMilestone = getValue(end);
-  console.info(`after getValue: start: ${startMilestone} end: ${endMilestone}`);
   if (startMilestone != null && endMilestone != null) {
     if (endMilestone <= startMilestone) {
       return {error: 'Start milestone must be before end milestone'};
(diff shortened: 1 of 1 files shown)


client-src/elements/utils.js
```diff
@@ -387,12 +387,14 @@ export function checkMilestoneStartEnd(startEndPair, getFieldValue) {
...
     const value = getFieldValue(name);
     if (typeof value === 'string') {
+      if (value === '') return undefined;
       return Number(value);
     }
   };
   const {start, end} = startEndPair;
   const startMilestone = getValue(start);
   const endMilestone = getValue(end);
+  console.info(`after getValue: start: ${startMilestone} end: ${endMilestone}`);
```
jrobbins, line 397: Is this console output still needed?

jrobbins (approved):
LGTM after addressing one comment.

PR #3536, opened 2023-12-07
Add semantic check for consistency between the feature name and type when type is deprecation

dlaliberte:
This PR adds a semantic check for consistency between the feature name and type when type is deprecation.
If the name includes "deprecate" or "remove" (or variations, e.g. "Deprecation..."), then the type should only be "Feature deprecation", and vice versa.

This applies both in the "New feature" page and when editing the feature any time later (since the feature name can be changed).

image

image

### DIFF
--- a/client-src/elements/chromedash-guide-new-page.js
+++ b/client-src/elements/chromedash-guide-new-page.js
@@ -6,9 +6,10 @@ import {
   NEW_FEATURE_FORM_FIELDS,
   ENTERPRISE_NEW_FEATURE_FORM_FIELDS,
 } from './form-definition';
+import {ALL_FIELDS} from './form-field-specs';
 import {SHARED_STYLES} from '../css/shared-css.js';
 import {FORM_STYLES} from '../css/forms-css.js';
-import {setupScrollToHash} from './utils';
+import {setupScrollToHash, formatFeatureChanges} from './utils';
 
 
 export class ChromedashGuideNewPage extends LitElement {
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-guide-new-page.js
```diff
@@ -50,12 +51,15 @@ export class ChromedashGuideNewPage extends LitElement {
...
       userEmail: {type: String},
       isEnterpriseFeature: {type: Boolean},
+      fieldValues: {type: Array},
     };
   }
 
   constructor() {
     super();
     this.userEmail = '';
+    // this.isEnterpriseFeature = false; // ??
```
jrobbins, line 61: If this commented-out code is not needed, it should be deleted.

client-src/elements/chromedash-guide-new-page.js
```diff
@@ -74,14 +78,30 @@ export class ChromedashGuideNewPage extends LitElement {
 
   handleFormSubmit(event, hiddenTokenField) {
     event.preventDefault();
+    // const submitBody =
```
jrobbins, line 81: Same

client-src/elements/chromedash-guide-new-page.js
```diff
@@ -74,14 +78,30 @@ export class ChromedashGuideNewPage extends LitElement {
...
+    formatFeatureChanges(this.fieldValues, this.featureId);
 
     // get the XSRF token and update it if it's expired before submission
     window.csClient.ensureTokenIsValid().then(() => {
       hiddenTokenField.value = window.csClient.token;
       event.target.submit();
+      // return csClient.updateFeature(submitBody);
```
jrobbins, line 88: Same

client-src/elements/utils.js
```diff
@@ -400,3 +400,26 @@ export function checkMilestoneStartEnd(startEndPair, getFieldValue) {
     }
   }
 }
+
+
+export function checkFeatureNameAndType(getFieldValue) {
+  const name = (getFieldValue('name') || '').toLowerCase();
+  const featureType = Number(getFieldValue('feature_type') || '0');
+  const deprecationName =
```
jrobbins, line 408: Since these are booleans, the variable names would be more clear as `isDeprecationName` and `isDeprecationType`.

jrobbins (approved):
Thanks for doing this.  All the logic looks right.  LGTM after some style tweaks.

PR #3539, opened 2023-12-09
Implement a menu to add xfn gates.

jrobbins:
This should resolve a request from the API Owners and save me from doing manual work.  Sometimes a feature is created using the PSA feature type, but then it turns out to have a bigger impact than expected, so it needs cross-functional review and an intent-to-ship thread.

In this PR:
* Add a "..." menu to the stage accordion sections with a "Add cross-functional gates" item when appropriate.
* Add an API call to csClient and main.py for the new action.
* Add an API handler to add the needed gates.


jrobbins:
I still need to add new unit tests, but you can take a look now.

jrobbins:
OK, ready for review.

dlaliberte (approved):
I can't really review the Python code, but the JavaScript seems fine, though I wonder about the use of `confirm` and event triggering.

PR #3549, opened 2023-12-13
Fixes to milestone checks

dlaliberte:
This PR does the following:

- Don't constrain the origin trial end milestone relative to shipped milestone.
- Add 'allFields' to formFieldValues, so checks on any form can get access to any other field of the feature.
-


jrobbins (approved):
Great, effectively applying semantic checks on stage-specific forms will be a big improvement.

PR #3559, opened 2023-12-20
Check images are actual images

yanndago:
- Make form field semantic checks asynchronous
- Add a check function to the screenshot_links field to verify that all the urls point to an actual image


dlaliberte (approved):
I didn't realize you were doing this.  Looks great, and it makes me think we can do more of the URL checks from the client-side.

PR #3561, opened 2023-12-27
Switch from roboto to system-ui font.

jrobbins:
I discussed fonts with Peter, the UX designer that we are working with now.  Inspired by modernfontstacks.com, we decided to go with `system-ui` rather than `roboto`.  It makes for one less thing to load and it gives the user a font that is optimized for the browser and OS that they are using.  The visual difference is actually pretty small, see the diffs of the playwright images for details.


jrobbins:
Note, the bold text does not look as bold as it was before in the playwright screenshots, but I do see the bold text clearly on my Mac in chrome and safari.

dlaliberte (approved):

past:
This seems to have switched to a serif font in ChromeOS for most things (except the button in this screenshot):

image

jrobbins:
D'oh.  I tested chromeos with an earlier iteration of this PR and then did not test again after simplifying it.  I'll add back sans-serif.

PR #3571, opened 2024-01-05
Add back sans-serif for chromeos

jrobbins:
Apparently chromeos does not have system-ui.


jrobbins:
This is already live on staging as a tainted version, so you can check it there if you have a chromeos device.

jrobbins:
I went ahead and put this on prod because the change is very low risk and I didn't want to have an incongruous user experience.

dlaliberte (approved):

PR #3580, opened 2024-01-10
Add new first_enterprise_notification_milestone field to all features

yanndago:
This new field is used to control the first milestone where a feature will be shown in the enterprise release notes.

This field is set by default on enterprise features and can be modified only if the feature has not yet been in the release notes.
This field cannot be set to a milestone that has already been released,
For non-enterprise features that are marked as breaking change, this field is automatically added.
When a feature is no longer a breaking change, if the feature had not been in the release notes, the field will be cleared.


internals/enterprise_helpers.py
```diff
@@ -0,0 +1,148 @@
...
+  return False
+
+def can_update_first_notification_milestone(feature: FeatureEntry, new_fields: dict) -> bool:
+  """Returns whether the milestone can be used to update first_enterprise_notification_milestone.
+
+  Args:
+    existing_feature: FeatureEntry feature that needs to be updated.
+    new_fields: dict Fields that will be used to update or create the feature.
+  """
+  if ('first_enterprise_notification_milestone' not in new_fields or
```
jrobbins, line 84: These few lines of code could be simpler as:
milestone = new_fields.get('first_enterprise_notification_milestone')
if not milestone:
  return False

internals/enterprise_helpers.py
```diff
@@ -0,0 +1,148 @@
...
+                                                    new_fields['breaking_change'] == True):
+    return not has_valid_milestone_in_new_fields
+  
+  # The normal feature becomes a breaking change
+  if 'breaking_change' in new_fields and new_fields['breaking_change'] == True:
+    return not has_valid_milestone_in_new_fields
+  
+  return False
+
+def can_update_first_notification_milestone(feature: FeatureEntry, new_fields: dict) -> bool:
```
jrobbins, line 77: Other parts of our code use the word `can` for user permission checking.  This seems more like a data validation function, so maybe name it `is_valid_first_notification_milestone`.

internals/enterprise_helpers.py
```diff
@@ -0,0 +1,148 @@
...
+                                                                                     milestone)
+    if (milestone in previous_milestone_details and
+        _str_to_datetime(previous_milestone_details[milestone]['stable_date']) <= datetime.now()):
+      return False
+
+  if feature.feature_type == FEATURE_TYPE_ENTERPRISE_ID:
+    return True
+  
+  # The breaking change stays a breaking change
+  if feature.breaking_change == True and ('breaking_change' not in new_fields
```
jrobbins, line 108: This whole condition could be simpler as:
`new_fields.get('breaking_change', feature.breaking_change)`
And, with that, I don't think that you need the next if-statement.

internals/enterprise_helpers.py
```diff
@@ -0,0 +1,148 @@
...
+                                          or new_fields['breaking_change'] == True):
+    return True
+  # The Feature becomes a breaking change
+  if ('breaking_change' in new_fields and new_fields['breaking_change'] == True):
+    return True
+  
+  return False
+
+
+def get_default_first_notice_milestone_for_feature() -> datetime:
```
jrobbins, line 118: Doesn't this return an `int` for a milestone?

internals/enterprise_helpers.py
```diff
@@ -0,0 +1,148 @@
...
+    existing_feature: FeatureEntry feature that needs to be updated.
+    new_fields: dict Fields that will be used to update or create the feature.
+  """
+  if feature.first_enterprise_notification_milestone == None:
+    return False
+
+  if feature.feature_type == FEATURE_TYPE_ENTERPRISE_ID:
+    return False
+  
+  if (feature.breaking_change == True and
```
jrobbins, line 138: I think this condition could also be reduced like as in my comment above:
`if new_fields.get('breaking_change', feature.breaking_change):`

internals/feature_helpers.py
```diff
@@ -85,7 +85,10 @@ def get_features_in_release_notes(milestone: int):
   features = [dict(converters.feature_entry_to_json_verbose(f))
             for f in _get_future_results(_get_entries_by_id_async(feature_ids))]
   features = [f for f in filter_unlisted(features) 
-    if f['deleted'] == False and (f['breaking_change'] == True or f['feature_type_int'] == FEATURE_TYPE_ENTERPRISE_ID)]
+    if f['deleted'] == False and
```
jrobbins, line 88: This code could be a little cleaner if we don't compare to boolean constants.  I.e.,
`if not f['deleted'] and ... `

jrobbins (approved):
Thanks for sending a well-thought-out PR that includes tests.  

LGTM after considering several style comments.

PR #3583, opened 2024-01-12
Fix semantic checks of milestones to get field values from form or appropriate stages

dlaliberte:
Always look for a form field value first, matching the stageId if provided.  Then, if the field is not present in the form, look in the feature, using mapped field names for stage-specific names.

* For dev trials, check that the milestone is less than all shipped milestones of the corresponding type (desktop, android, etc)
* For origin trials, since there can be more than one stage of origin trials, check that the maximum start milestone is before the minimum of all shipped milestones of the corresponding type.
* I separated the one test file into more than one and added many more tests.   This resulted in more flakiness due to unintended shared sessions (which is a feature that is not easily fixable).  The shared sessions resulted in greater parallelism between tests such that one test might be logged out in the middle of testing because another test finished and logged out  So I was forced to run only one test at a time.
  * Also related, I added utils for accepting dialogs for alert, confirm, and before-unload.
* The pwtests-update command can be run on a single file by adding the `--filename=xxx` parameter.  
  * e.g. `npm run pwtests-update  --workspace=playwright --filename=chromedash-guide-stage-page_pwtest.js`
 

This PR started by reapplying all changes from #3552 except for the package-lock.json file to a new clone of the repo.


client-src/elements/chromedash-form-field.js
```diff
@@ -44,6 +45,7 @@ export class ChromedashFormField extends LitElement {
   getValue() {
     // value can be a js or python boolean value converted to a string
     // or the initial value specified in form-field-spec
+    // If value is falsy, it will be replaced by the initial value, if any.
```
jrobbins, line 48: What is the difference between this code and:
return this.value || this.fieldProps.initial;
?
Is the goal to keep this.value when they are both falsy?   I am not sure that your comment is correct in that case.

client-src/elements/chromedash-form-field.js
```diff
@@ -117,52 +129,100 @@ export class ChromedashFormField extends LitElement {
...
+    // Define function to get any other field value relative to this field.
+    // stageOrId is either a stage object or an id, or the special value
+    // 'current stage' which means use the same stage as for this field.
+    const getFieldValue = (fieldName, stageOrId) => {
+      if (stageOrId === 'current stage') {
+        stageOrId = this.stageId;
+      }
+      return getFieldValueWithStage(fieldName, stageOrId, this.fieldValues || []);
+    };
+    // Attach the feature to the getFieldValue function, in case it is needed.
```
jrobbins, line 178: Why would this be needed?  It seems pretty unusual to poke into the state of a function object.  Why not pass `feature` through `checkFunction` as a regular argument?

client-src/elements/chromedash-form-field.js
```diff
@@ -344,4 +405,49 @@ export class ChromedashFormField extends LitElement {
...
+ * @param {string} fieldName
+ * @param {number|Object|undefined} stageOrId
+ * @param {Array<{name:string, value:*}>} formFieldValues
+ * @return {*}
+ */
+function getFieldValueWithStage(fieldName, stageOrId, formFieldValues) {
+  // Iterate through formFieldValues looking for element with name==fieldName
+  // and stage == stageId, if there is a non-null stageId
+  let stageId;
+  if (typeof stageOrId === 'number') {
```
jrobbins, line 424: Couldn't we just update all callers to be sure to pass in the stage object in all cases?

client-src/elements/chromedash-guide-new-page.js
```diff
@@ -132,7 +132,7 @@ export class ChromedashGuideNewPage extends LitElement {
 
   renderForm() {
     const newFeatureInitialValues = {owner: this.userEmail};
-    this.fieldValues.allFields = newFeatureInitialValues;
+    this.fieldValues.feature = this.feature;
```
jrobbins, line 135: Thanks for doing this.  This code seems cleaner than what was there before.

packages/playwright/package.json
```diff
@@ -16,6 +16,6 @@
     },
     "author": "",
     "devDependencies": {
-        "@playwright/test": "1.41.1"
+        "@playwright/test": "1.40.1"
```
jrobbins, line 19: Is this intentional?

client-src/elements/chromedash-form-field.js
```diff
@@ -117,52 +129,100 @@ export class ChromedashFormField extends LitElement {
...
-        if (dependentField) {
-          dependentField.doSemanticCheck();
-        }
-      });
+    if (app?.pageComponent) {
+      app.pageComponent.allFormFieldComponentsList.forEach((formFieldComponent) =>
+        formFieldComponent.doSemanticCheck());
+    } else {
+      // Do the semantic check for unit testing.  Only works for isolated field.
+      this.doSemanticCheck();
```
DanielRyanSmith, line 139: Can we run this on initial page load as well? When I tested this and had separate bad inputs already submitted for my feature, I would see warnings start to populate on fields I wasn't editing. Just running this and adding the warnings on page load should fix that unusual behavior.

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
+// const OT_ALL_SHIPPED_MILESTONE_IOS_RANGE = {
+//   earlier: 'ot_milestone_ios_start',
+//   allLater: 'shipped_ios_milestone',
+//   warning: 'Origin trial should start before feature is shipped.',
+// };
+// const ALL_OT_SHIPPED_MILESTONE_IOS_RANGE = {
+//   allEarlier: 'ot_milestone_ios_start',
+//   later: 'shipped_ios_milestone',
+//   warning: 'Origin trial should start before feature is shipped.',
+// };
```
DanielRyanSmith, line 132: We can just remove this - as there should never be a check for this.

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
   earlier: 'ot_milestone_desktop_start',
+  allLater: 'shipped_milestone',
+  warning: 'Origin trial should start before feature is shipped.',
+};
+
+const ALL_OT_SHIPPED_MILESTONE_DESKTOP_RANGE = {
+  allEarlier: 'ot_milestone_desktop_start',
   later: 'shipped_milestone',
-  error: 'Origin trial must start before feature is shipped.',
+  warning: 'All origin trials should start before feature is shipped.',
```
DanielRyanSmith, line 95: Optional nit: when I encountered this warning message filling out a milestone on a shipping stage, the text felt confusing to me, since it was talking about origin trials stages, which I wasn't editing. It might be worthwhile to specifically mention milestones.

e.g. "Origin trials starting milestones should be before feature shipping milestones."

@jrobbins Maybe you have an opinion here

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
   earlier: 'dt_milestone_desktop_start',
+  allLater: 'shipped_milestone',
+  warning: 'Shipped milestone must be later than dev trial.',
+};
+
+const ALL_DT_SHIPPED_MILESTONE_DESKTOP_RANGE = {
+  allEarlier: 'dt_milestone_desktop_start',
   later: 'shipped_milestone',
-  error: 'Shipped milestone must be later than dev trial.',
+  warning: 'Shipped milestone must be later than dev trial.',
```
DanielRyanSmith, line 143: To match other wording
```suggestion
  warning: 'Shipped milestone should be later than dev trial start milestone.',
```

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
 };
 
-const DT_SHIPPED_MILESTONE_ANDROID_RANGE = {
+const DT_ALL_SHIPPED_MILESTONE_ANDROID_RANGE = {
   earlier: 'dt_milestone_android_start',
+  allLater: 'shipped_android_milestone',
+  warning: 'Shipped milestone must be later than dev trial.',
```
DanielRyanSmith, line 149:
```suggestion
  warning: 'Shipped milestone should be later than dev trial start milestone.',
```

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
+// const ALL_OT_SHIPPED_MILESTONE_IOS_RANGE = {
+//   allEarlier: 'ot_milestone_ios_start',
+//   later: 'shipped_ios_milestone',
+//   warning: 'Origin trial should start before feature is shipped.',
+// };
+
+const DT_ALL_SHIPPED_MILESTONE_DESKTOP_RANGE = {
   earlier: 'dt_milestone_desktop_start',
+  allLater: 'shipped_milestone',
+  warning: 'Shipped milestone must be later than dev trial.',
```
DanielRyanSmith, line 137:
```suggestion
  warning: 'Shipped milestone should be later than dev trial start milestone.',
```

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
   earlier: 'dt_milestone_android_start',
+  allLater: 'shipped_android_milestone',
+  warning: 'Shipped milestone must be later than dev trial.',
+};
+
+const ALL_DT_SHIPPED_MILESTONE_ANDROID_RANGE = {
+  allEarlier: 'dt_milestone_android_start',
   later: 'shipped_android_milestone',
-  error: 'Shipped milestone must be later than dev trial.',
+  warning: 'Shipped milestone must be later than dev trial.',
```
DanielRyanSmith, line 155:
```suggestion
  warning: 'Shipped milestone should be later than dev trial start milestone.',
```

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
 };
 
-const DT_SHIPPED_MILESTONE_IOS_RANGE = {
+const DT_ALL_SHIPPED_MILESTONE_IOS_RANGE = {
   earlier: 'dt_milestone_ios_start',
-  later: 'shipped_ios_milestone',
-  error: 'Shipped milestone must be later than dev trial.',
+  allLater: 'shipped_ios_milestone',
+  warning: 'Shipped milestones must be later than dev trial.',
```
DanielRyanSmith, line 161:
```suggestion
  warning: 'Shipped milestones should be later than dev trial start milestone.',
```

client-src/elements/form-field-specs.js
```diff
@@ -77,54 +83,101 @@ const OT_MILESTONE_WEBVIEW_RANGE = {
...
 };
 
-const DT_SHIPPED_MILESTONE_WEBVIEW_RANGE = {
-  earlier: 'dt_milestone_webview_start',
-  later: 'shipped_webview_milestone',
-  error: 'Shipped webview milestone must be later than dev trial.',
+const ALL_DT_SHIPPED_MILESTONE_IOS_RANGE = {
+  allEarlier: 'dt_milestone_ios_start',
+  later: 'shipped_ios_milestone',
+  warning: 'Shipped milestone must be later than dev trial.',
```
DanielRyanSmith, line 167:
```suggestion
  warning: 'Shipped milestone should be later than dev trial start milestone.',
```

DanielRyanSmith (dismissed):

client-src/elements/chromedash-form-field.js
dlaliberte: If this.value is a falsy value like 0 or '', then `!this.value` will be true.  And then, if this.fieldProps.initial is truthy, the whole condition `!this.value && this.fieldProps.initial` will be true, in which case the result of getValue will be this.fieldProps.initial.  That's what I meant by my comment, which I believe is correct.

To answer your question about `return this.value || this.fieldProps.initial`, it will have a similar but not identical behavior.  If this.value is falsy, then this.fildProps.initial will be used regardless whether the initial value is truthy.  

So, the fact that any falsy value (0, '', false) will be treated the same as null or undefined means that non-null values that are falsy will still be replaced by the initial value.  This means we can't store falsy values that differ from the initial value.  I would think we should allow 0 or false to be stored, and not be replaced by the initial value.  I was trying to make a change to check explicitly for null or undefined, with `this.value == null`, but that conflicted with something I don't recall at the moment.

client-src/elements/chromedash-form-field.js
dlaliberte: There is a call of this.doSemanticCheck() in firstUpdated.  This should result in doing the checks when the page is loaded.  But your comment reminds me that I should add a test for that.  (It worked in my manual tests, but it needs to be in the tests.)

client-src/elements/chromedash-form-field.js
dlaliberte: Adding the feature object to the getFieldValue function object was just a convenient way make it available outside of getFieldValue itself.  I needed to get the feature object to get the stages for findMinMilestone and findMaxMilestone, since they must look in the feature stages if the required fields are not in the form.  It might be reasonable to attach the feature to the formFieldValues array (which is passed as a parameter to the inner function, getFieldValueWithStage), but even if I did that, neither are available just given the getFieldValue function.  So I just punted to get the job done. 

This is all rather confusing, I admit.  It would be much simpler to have a uniform and centralized way of getting any field value, independent of whether the latest value might be in the current form or only in the feature object.  The lack of a global id for each field/feature-property makes it more confusing as well, though we do sometimes need to iterate through arrays of stages of a particular type.

client-src/elements/chromedash-form-field.js
dlaliberte: We don't actually need the stage object if the field is in the current form.  In fact, I don't think we can use the stage object in the case that the start and end OT milestones are in the current form.  But when they are not in the current form, we do need to
look through stage objects to find the max of all OT start milestones.

client-src/elements/chromedash-guide-new-page.js
dlaliberte: Right.  The previous code was a result of my multiple confusions about what was going on.  Now that I see that again, the feature object should be attached to all the fieldValues arrays.  so I wonder if we could just pass the fieldValues object to the check function, rather than the getFieldValue function, which can't do the right thing on its own.

client-src/elements/form-field-specs.js
dlaliberte: Done.  We can always add it later if needed.

client-src/elements/form-field-specs.js
dlaliberte: There are two separate checks related to OT and shipped milestones.  This one is for the shipped milestone, in which it compares all the OT start milestones with the one shipped milestone you are editing.  It doesn't matter if you are currently editing the OT start milestones, if they have a value from earlier editing.  The other check is for each of the OT start milestones, which compares each one with all the shipped milestones (for the same platform). 

Regarding the message, I was assuming that an origin trial starts with its starting milestone, but maybe that is not quite right?  So your message seems to be equivalent to mine.  I'm open to changing the language to whatever is more appropriate.

I was thinking it would be useful to include the offending milestone numbers, which should be possible with some kind of dynamic creation of tagged template literals (so the `${milestone}` expressions are not evaluated in these constant definitions, but later when the value is available).

client-src/elements/form-field-specs.js
dlaliberte: Quite right, since it is a warning rather than an error.  Also, dev trials could end later.

packages/playwright/package.json
dlaliberte: Hmm, no!  Not sure how or why that happened.  Thanks for pointing this out.

client-src/elements/form-field-specs.js
dlaliberte: done

client-src/elements/form-field-specs.js
dlaliberte: Although, dev trials don't distinguish between start and end milestones, since there is no end milestone.  So it might be confusing to mention the start milestone.

client-src/elements/chromedash-form-field.js
jrobbins: Thanks for explaining.  It's a subtle bit of logic, but I can see why it is needed.

client-src/elements/chromedash-form-field.js
jrobbins: I understand that you need the feature object, but why not pass it as an additional function argument?

jrobbins: Actually, let's go with what you have and we can look for ways to simplify later.

client-src/elements/chromedash-form-field.js
jrobbins: OK, this can be another thing to consider simplifying later.

client-src/elements/form-field-specs.js
jrobbins: As a simple improvement, let's just make the warning text include the word "milestones".  That word will help the user find the referenced field by name.

jrobbins:
This is all looking pretty good.  From all my comments, the only thing to change is to add the word "milestones" to the warning.

client-src/elements/form-field-specs.js
dlaliberte: I changed it to a more specific variation of Daniel's suggestion: 
`'All origin trials starting milestones should be before feature shipping milestone.'`
and for the OT_ALL_SHIPPED_* cases:
`'Origin trial starting milestone should be before all feature shipping milestones.'`

dlaliberte:
> This is all looking pretty good. From all my comments, the only thing to change is to add the word "milestones" to the warning.

Thanks.  There is more that I can test, and more refactoring cleanups I am tempted to dive into, but I am confident that this PR will not break anything, and does fix a couple current problems.  So it is probably better to get this in and cleanup more later.

client-src/elements/chromedash-form-field.js
dlaliberte: OK, passing yet another argument to the check function could work.  Actually, I am inclined to pass in an object as a bag of properties, which will simplify the check function properties in the form-field-specs.   But ... later.

client-src/elements/chromedash-form-field.js
dlaliberte: OK thanks.  Yes, I hope we can find some simplifications for all this.

jrobbins (approved):

client-src/elements/chromedash-form-field.js
DanielRyanSmith: I am still seeing the issue, but specifically when editing an OT stage or shipping stage in `chromedash-edit-stage-page`. I think this is because the shipping/OT milestones for the stage NOT being edited are not a part of the form.

See the below video where I begin editing a shipping stage when the the OT and shipping milestones have already been submitted with invalid values (OT desktop milestone of 123 and shipping desktop milestone of 120). The warning will pop up only after I've edited any other field on the page.

I think this can be easily fixed by doing a run of the semantic checks when the component is loaded.

https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/decc66d3-0c02-4fb3-ba12-211ad11f930b

client-src/elements/chromedash-form-field.js
```diff
@@ -71,7 +74,16 @@ export class ChromedashFormField extends LitElement {
...
+
+    // We only want to do the following one time.
+    this.setupSemanticCheck();
+
+    // We need to wait until the entire page is rendered, so later dependents
+    // are available to do the semantic check, hence firstUpdated is too soon.
+    // Do first semantic check after the document is ready.
+    document.addEventListener('DOMContentLoaded', () => (setTimeout(() => {
+      this.doSemanticCheck();
+    })));
```
DanielRyanSmith, line 86: I believe I've found the problem. When my page loaded, the `DOMContentLoaded` event had already fired before this event listener was added, which meant the initial render never occurred. I found [this section in the DOMContentLoaded docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#checking_whether_loading_is_already_complete) that describes how to handle this situation. It worked for my problem 🙂 

```suggestion
    if (document.readyState === 'loading') {
      document.addEventListener('DOMContentLoaded', () => (setTimeout(() => {
        this.doSemanticCheck();
      })));
    } else {
      this.doSemanticCheck();
    }
```

client-src/elements/chromedash-form-field.js
dlaliberte: Great fix, though it seems there ought to be a cleaner way to deal with these kinds of things.

PR #3586, opened 2024-01-12
Replace breaking_change by enterprise_impact on features

yanndago:
This changelist also removes rollout_impact on rollout stages as It becomes redundant.

A migration plan is added where features that are marked as breaking will have a medium impact and the highest rollout_impact will be translated to the enterprise_impact of the corresponding feature.

The deprecated fields are not yet removed as they are needed for the migration.

All code paths that relied on breaking_change being true will now rely on enterprise_impact being at least at low impact.

(cherry picked from commit 51e585b7cdbe8050f35f5c0b4fae2ffb43213949)


yanndago:
Screenshot 2024-01-12 at 11 45 08 AM
Screenshot 2024-01-12 at 11 45 24 AM
Screenshot 2024-01-12 at 11 45 39 AM
Screenshot 2024-01-12 at 11 57 11 AM

dlaliberte:
I'm curious whether running `npm run pwtests-update  --workspace=playwright` will be enough to resolve the playwright test errors.

api/processes_api_test.py
```diff
@@ -57,16 +57,16 @@ def test_get__default_feature_type(self):
 
   def test_get__feature_type_0(self):
     """We can get process for features with feature type 0 (New feature incubation)."""
-    self.feature_1.feature_type = 0
+    self.feature_1.feature_type = core_enums.FEATURE_TYPE_INCUBATE_ID
```
jrobbins, line 60: Thanks for improving this.

internals/maintenance_scripts.py
```diff
@@ -450,3 +450,52 @@ def get_template_data(self, **kwargs):
...
+    self.require_cron_header()
+    count = 0
+    batch = []
+    BATCH_SIZE = 100
+    feature_ids = set()
+  
+    stages: ndb.Query = Stage.query(Stage.stage_type == STAGE_ENT_ROLLOUT, Stage.archived == False)
+    # Update enterprise_impact to be the highest impact set on any of the rollout steps.
+    for stage in stages:
+      feature_entry = FeatureEntry.get_by_id(stage.feature_id)
```
jrobbins, line 466: During the overall execution of the stages loop, this line of code will fetch each feature entry about 5 times.  It would be better to fetch all the feature entries before the start of this loop and put them into a dictionary keyed by NDB IDs.  You would still need to keep a set of the IDs of the ones that were modified so that you can write them back to NDB.  You could still do the batching logic that you have already so that this code makes incremental progress (which would make it go faster on a second run after a timeout).

internals/maintenance_scripts.py
```diff
@@ -450,3 +450,52 @@ def get_template_data(self, **kwargs):
...
+      feature_ids.add(stage.feature_id)
+      if feature_entry == None:
+        continue
+      new_impact = stage.rollout_impact + 1
+      if new_impact <= feature_entry.enterprise_impact:
+        continue  
+      feature_entry.enterprise_impact = new_impact
+      batch.append(feature_entry)
```
jrobbins, line 474: Using a list of feature entries rather than a set of IDs might cause a problem in the rare case where a feature has two rollout steps that both cause enterprise_impact to be set.   That could append the same feature entry twice to the batch list, which NDB might treat as an error or it could be ambiguous as to which one is written to the datastore last.

jrobbins (commented):
Overall this looks pretty good, but I do have one suggested algorithm change for the migration script.

jrobbins (approved):

PR #3598, opened 2024-01-20
Create SECURITY.md

ScorpiusDraconis83:


jrobbins:
This doesn't seem to have a lot of info beyond what users could just get from the GitHub UI.  And, we don't release or support numbered versions of our app.  Also, we cannot accept contributions without a signed CLA.

Issue #3007, opened 2023-05-15
Create a script to populate your environment 


jcscottiii:
Taking inspiration from [wpt.fyi](https://github.com/web-platform-tests/wpt.fyi#setting-up-your-environment), it would be nice to have an optional script that will read information from staging to populate your local environment.

wpt.fyi currently has:

- bash [script](https://github.com/web-platform-tests/wpt.fyi/blob/main/util/docker-dev/dev_data.sh), which calls
- a Makefile [target](https://github.com/web-platform-tests/wpt.fyi/blob/main/Makefile#L249) that calls a go program that
- makes public API calls to read data from staging and then populate the local database.

Similarly, we could have:
- npm target that calls
- a python program that makes API calls to staging to read data and populates the local database.

jphysicsx17:
hii assign this to me

jyasskin:
A hint to whoever picks this up: there's a route handler to import some data from JSON at https://github.com/GoogleChrome/chromium-dashboard/blob/fe936c92f6aeb0d4daee4538e4cce12341d620c0/api/dev_api.py#L64-L91, and a commented-out route to call it from the dev server at https://github.com/GoogleChrome/chromium-dashboard/blob/fe936c92f6aeb0d4daee4538e4cce12341d620c0/main.py#L293.

jphysicsx17:
hey I am not getting how to start  can you give me some  more hint

PR #3606, opened 2024-01-24
Add a script to seed the dev datastore from a running server.

jyasskin:
This starts fixing #3007, but it only pulls the data available from https://github.com/GoogleChrome/chromium-dashboard/blob/fe936c92f6aeb0d4daee4538e4cce12341d620c0/api/converters.py#L470-L556, not the full set of fields.

I haven't written Python in years, so please point out modern idioms and libraries I'm missing.

### DIFF
--- a/scripts/seed_datastore.py
+++ b/scripts/seed_datastore.py
@@ -22,126 +22,86 @@ This script can also set a user as admin, to facilitate testing those parts of t
 """
 
 from datetime import datetime
-from google.cloud import ndb
 from urllib.parse import urljoin
 import argparse
 import logging
 import os
-import pprint
-import requests
 import sys
+
+import requests
+from google.cloud import ndb  # type: ignore
 
 sys.path = [os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
             ] + sys.path
-os.environ['DATASTORE_EMULATOR_HOST'] = 'localhost:15606'
-os.environ['GAE_ENV'] = 'localdev'
-os.environ['GOOGLE_CLOUD_PROJECT'] = 'cr-status-staging'
-os.environ['SERVER_SOFTWARE'] = 'gunicorn'
+os.environ.setdefault('DATASTORE_EMULATOR_HOST', 'localhost:15606')
+os.environ.setdefault('GAE_ENV', 'localdev')
+os.environ.setdefault('GOOGLE_CLOUD_PROJECT', 'cr-status-staging')
+os.environ.setdefault('SERVER_SOFTWARE', 'gunicorn')
 
+# pylint: disable=wrong-import-position
 from internals.core_enums import MISC
 from internals.core_models import FeatureEntry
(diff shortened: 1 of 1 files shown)


scripts/seed_datastore.py
```diff
@@ -0,0 +1,147 @@
...
+import argparse
+import logging
+import os
+import pprint
+import requests
+import sys
+
+sys.path = [os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+            ] + sys.path
+os.environ['DATASTORE_EMULATOR_HOST'] = 'localhost:15606'
```
jcscottiii, line 36: Use setdefault for these four environment variables in case the caller of this script wants to set their own values.

Example: `os.environ.setdefault('DATASTORE_EMULATOR_HOST', 'localhost:15606')`

scripts/seed_datastore.py
```diff
@@ -0,0 +1,147 @@
...
+    '--after',
+    type=datetime.fromisoformat,
+    help='only add features that have been updated since AFTER',
+    default='0001-01-01')
+
+args = parser.parse_args()
+client = ndb.Client()
+with client.context():
+    if args.server:
+        target = urljoin(args.server, 'features.json')
```
jcscottiii, line 62: Could you move this section into a separate function e.g. `def add_features` ?

scripts/seed_datastore.py
```diff
@@ -0,0 +1,147 @@
...
+                        'view': {
+                            'notes': fe.other_views_notes,
+                        }
+                    },
+                }
+            })
+            fe.put()
+
+    if args.admin:
+        logging.info('Making %s an admin.', args.admin)
```
jcscottiii, line 143: Could you move this into a function too? Example `def add_admin`

scripts/seed_datastore.py
jcscottiii: @DanielRyanSmith @KyleJu @jrobbins Is the features.json API also going away soon? Or only the old features page? I was just wondering because of the google cloud alerts. If so, I'd rather use an alternative API if the features.json API is slated to go away soon.

jcscottiii (commented):
Tested out the script and it works. LGTM with nits. I just have one question for the team about the future of the features.json API before approving

scripts/seed_datastore.py
jyasskin: Done.

scripts/seed_datastore.py
jyasskin: Done.

scripts/seed_datastore.py
jyasskin: Done.

jcscottiii: > Is the features.json API also going away soon?

After talking with the team, it will be supported for awhile

jcscottiii (approved):

PR #3607, opened 2024-01-24
Add helpful tools and extensions to the devcontainer.

jyasskin:
I picked yapf as the formatter extension because there was a setting for it, but it might be a better idea to adopt https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff. What are people using outside the container?

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,6 +15,8 @@
 		"vscode": {
 			// Set *default* container specific settings.json values on container create.
 			"settings": {
+				"mypy-type-checker.args": ["--ignore-missing-imports --exclude cs-env/ --exclude gen/ --exclude appengine_config.py --disable-error-code \"annotation-unchecked\""],
+				"pylint.args": ["--indent-string=\"  \""],
 				"python.defaultInterpreterPath": "/workspace/cs-env/bin/python",
 				"python.editor.tabSize": 2,
 				"python.testing.pytestEnabled": false,
@@ -51,13 +53,11 @@
 	],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
-	// pip version https://github.com/microsoft/vscode-python/issues/18802
-	// Need to install pylint inside the virutalenv because the global one does not honor it.
 	// Recently, when running `npm run clean-setup` in the container, it has
 	// not been installing the packages in the virtual environment. Instead,
 	// it has been installing in the default Python environment. We will
 	// need to call npm run setup again once the environment has been activated.
-	"postCreateCommand": "npm run clean-setup; . cs-env/bin/activate; npm run setup; python -m pip install pylint==2.13.4",
+	"postCreateCommand": "npm run clean-setup; . cs-env/bin/activate; npm run setup",
 
 	// Comment out to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
 	"remoteUser": "vscode",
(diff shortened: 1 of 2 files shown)


dlaliberte:
I don't really know what most people are using.  Bikeshed uses Black, which appears to be quite popular.  Ruff is 99.9% compatible with Black (by default), but 30x faster and 100x faster than YAPF.    Speed is not the highest priority, I would think, as long as it is not particularly noticeable.   Flexibility might be important, but then...

YAPF has lots of options, which means there are lots of choices to make.
Black has no choices.  I'm not sure I like their choice in all cases. 

I do notice lots of red squiggly lines in our python code, so I figured I must have the wrong default linter, but I don't know if we have a default linter, other than the `npm run pylint` command.

jyasskin:
I prefer formatters with no options, but all the people editing the codebase need to pick the same one or it'll reformat on every checkin. And we probably don't want one that implies lots of changes to the existing formatting, although that could be worth it if there's actually no consistent format across the existing code.

.devcontainer/devcontainer.json
```diff
@@ -17,17 +17,6 @@
...
-				"python.linting.pylintArgs": [
-					"--indent-string=\"  \""
-				],
-				"python.linting.banditEnabled": true,
-				"python.linting.mypyEnabled": true,
-				"python.linting.pylintEnabled": true,
-				"python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
-				"python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
-				"python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
-				"python.linting.pylintPath": "/workspace/cs-env/bin/pylint",
```
jcscottiii, line 30: We need to force pylint to use two spaces instead of two. The [extension](https://marketplace.visualstudio.com/items?itemName=ms-python.pylint) allows us to configure the setting.

```
"pylint.args": ["--indent-string=\"  \""]
```

jcscottiii: Similarly, can you add a config for mypy to match what is used in the package.json

```
"mypy-type-checker.args": ["--ignore-missing-imports --exclude cs-env/ --exclude gen/ --exclude appengine_config.py --disable-error-code \"annotation-unchecked\""]
```

jcscottiii (commented):
I really like that you switched to using the extensions. I had two comments to help with the config that I saw when testing out the devcontainer.

Also, in the "postCreateCommand", we can now remove installing pylint manually since the extension comes with it.

jcscottiii:
When modifying the postCreateCommand, you can remove the snippet about pylint in the comment above too 🎉

jyasskin (commented):
And I've fixed up the postCreateCommand. This'll need conflicts resolved with #3608, which I'll do once either one gets merged.

jcscottiii (approved):

dlaliberte:
> I prefer formatters with no options, but all the people editing the codebase need to pick the same one or it'll reformat on every checkin. 

Oh, I certainly don't mean that we might each choose different options.  We would have to agree, so that each change to the codebase doesn't flip-flop between each individual's preferences. 

There can also be many optional situations, such as whether to break lines in one place or another, or not at all.  I tend to prefer more lines that are shorter, and I find I frequently fight with formatters when they want to undo what I just did intentionally.

> And we probably don't want one that implies lots of changes to the existing formatting, although that could be worth it if there's actually no consistent format across the existing code.

There are many cases in the ChromeStatus code of different amounts of indentation for the same kind of situation, which Python allows, but it is superfluous, and I find it distracting.

PR #3608, opened 2024-01-25
Use `.` instead of `source` in npm scripts. Then shorten the postCreateCommand.

jyasskin:
`npm run` runs commands in `sh`, which is dash rather than bash. Dash is strictly POSIX, and `source` is an extension to POSIX.

I believe that's the bug the extra `npm run setup` in the postCreateCommand was working around, so I've removed it.


jrobbins (approved):
Thanks for doing this.

jcscottiii (approved):

jyasskin:
Thanks, and now this is merged with #3607.

PR #3612, opened 2024-01-25
Define template for assigned review notifications

jrobbins:
This is progress on modernizing our notification emails. 

In this PR:
+ Refactor the function to format a notification of an assigned review into a method of the handler class to keep the code more organized.
+ Add a new template specifically for notifications of assigned reviews
+ Change the task parameters passed for these tasks so that we have a link directly to the gate

You can see examples of these notifications here:
https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/mBYE-7nE4i0
Note that they lack CSS style so far, but we are closer to being able to add that.


KyleJu:
Won't be able to get to it. I will leave the review to @dlaliberte

internals/notifier_test.py
```diff
@@ -655,6 +655,76 @@ def test_make_review_requests_email__assigned(self, mock_f_e_b):
...
+    self.assertEqual('new@example.com', review_task_1['to'])
+
+    review_task_2 = actual_tasks[1]
+
+    # Notification to new assignee.
+    self.assertEqual(
+        'Review assigned for feature: feature one',
+        review_task_2['subject'])
+    self.assertIn('mock body html', review_task_2['html'])
+    self.assertIn('<li>The review was previously assigned to you</li>',
```
dlaliberte, line 714: The comment above sas this is notification to the new assignee, and yet the text being sent says "The review was previously assigned to you".  And it's the opposite above, for the old assignee, where the text says "The review is now assigned to you".  I think the comments should just be swapped.

dlaliberte (approved):

internals/notifier_test.py
jrobbins: Good catch.  Fixed.

jrobbins:
Here is what one of these looks like with style:
https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/DXXAHYpPx_o

In Groups:
image

And in gmail:
image

PR #3616, opened 2024-01-27
Add style attributes to emails.

jrobbins:
This adds significant visual polish to our email notifications.  Preceding PRs have updated the markup and content of our email templates, and now this one adds style.   The `<style>` element does not work on groups.google.com, so we need to use `style="..."` attributes instead.  

To avoid repeating style details many times, I have used jinja2 variables and macros.  I could have done the same thing with python code, but I felt that it was more clear to keep template content stuff in the template language rather than in python.

In this PR:
* Pass APP_TITLE because it is used in the "branding" line at the very top of each email notification body.
* Revise each email notification template:
    * Add grey "body" and white "content" divs to frame the content.  The "content" has a color-coded top border.
    * Style the `<sections>` to look like bubbles
    * Style the "context" header with a color-coded circle and large-font feature name
    * Style the "View feature details" or other main action link to look like buttons

### DIFF
--- a/templates/new-feature-email.html
+++ b/templates/new-feature-email.html
@@ -1,56 +1,70 @@
-<section id="context">
-  <div>
-    <b>New</b> feature entry:<br/>
-    <a class="feature-name" href="{{SITE_URL}}feature/{{id}}"
-          >{{feature.name}}</a>
-  </div>
-</section>
-<br/>
+{% import 'email-styles.html' as styles %}
+<div style="{{styles.body}}">
+<div style="{{styles.branding}}">{{APP_TITLE}}</div>
 
-
-<section id="details">
-  <div><b>New details entered by {{feature.creator_email}}:</div>
-
+<div style="{{styles.content}} {{styles.content_new}}">
+<section id="context" style="{{styles.section}}">
   <table>
     <tr>
-      <td>Summary:</td>
-      <td>{{feature.summary}}</td>
+      <td rowspan=2>{{styles.icon_new()}}</td>
+      <td><b>New</b> feature entry:</td>
     </tr>
     <tr>
-      <td>Unlisted:</td>
-      <td>{% if feature.unlisted %}Yes{% else %}No{% endif %}</td>
-    </tr>
-    <tr>
-      <td>Breaking change:</td>
-      <td>{% if feature.breaking_change %}Yes{% else %}No{% endif %}</td>
-    </tr>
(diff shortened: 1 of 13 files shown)


jrobbins:
Here's one example of how the "new feature notification" looks now:
New: https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/eZzD42ndXVE

Here's the way it looks to users before this change:
Old: https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/asnfXZLGzKU/m/bwMGT8G-AgAJ

Those are different features.

templates/new-feature-email.html
```diff
@@ -1,56 +1,70 @@
...
-      <td>Cc:</td>
-      <td>{{feature.cc_emails|join(", ")}}</td>
+      <td style="{{styles.td}}">Cc:</td>
+      <td style="{{styles.td}}">{{feature.cc_emails|join(", ")}}</td>
     </tr>
     <tr>
-      <td>Blink component:</td>
-      <td>{{feature.blink_components|join(", ")}}</td>
+      <td style="{{styles.td}}">Blink component:</td>
+      <td style="{{styles.td}}">{{feature.blink_components|join(", ")}}</td>
```
dlaliberte, line 48: I notice that the resulting text looks like, e.g. Blink>Compositing>Filters, which should probably have space around the ">" chars.   

Also, I am wondering what join(", ") is joining in that case.  Can there be multiple blink components, each one being a path through the hierarchy?

dlaliberte:
Sad that the <style> element doesn't work in Google Groups.  Does it work at all in gmail?  (Thinking about the intent email style)

Anyway, the resulting styled posting looks great.

dlaliberte (approved):

templates/new-feature-email.html
jrobbins: In the UI, there are no spaces around the `>` characters, so I'll leave that as-is for consistency.

Yes, the component is a repeated string field in our NDB models, but our UI only exposes the option to set a single value for now.

jrobbins:
> Sad that the <style> element doesn't work in Google Groups. Does it work at all in gmail? (Thinking about the intent email style)

It mostly worked in gmail.  Maybe I could have gotten that approach to work in gmail, but since our users also view these messages in mailing list archives, and they are free to use other email clients, using style="..." is a least-common-demonimator approach.

PR #3626, opened 2024-01-30
Make refreshing the XSRF token conditional on whether the openapi endpoint needs authentication.

jyasskin:
We can almost get rid of this middleware entirely, but the generated code is missing an `await` at https://github.com/OpenAPITools/openapi-generator/blame/59f672d9aa05005c4c3568535c39cd7a27daa4b2/modules/openapi-generator/src/main/resources/typescript-fetch/apis.mustache#L194.

### DIFF
--- a/gen/js/chromestatus-openapi/package.json
+++ b/gen/js/chromestatus-openapi/package.json
@@ -16,6 +16,6 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "typescript": "^5.3"
+    "typescript": "^4.0"
   }
 }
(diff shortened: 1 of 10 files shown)


gen/js/chromestatus-openapi/package.json
```diff
@@ -16,6 +16,6 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "typescript": "^5.3"
+    "typescript": "^4.0"
```
jcscottiii, line 19: Nothing wrong here.

But I am very curious as to how this was 5.3 beforehand haha. I see the [mustache](https://github.com/OpenAPITools/openapi-generator/blob/2129b15c8f0598126a7a259106f7018bd302eef2/modules/openapi-generator/src/main/resources/typescript-fetch/package.mustache#L33) file pins it to ^4.0

jcscottiii (approved):
LGTM

PR #3627, opened 2024-01-30
Add an admin section to the menu drawer

jrobbins:
This adds a simple section to our app's main menu with items for admins.  It is only shown to admin users, so there is no impact on regular users.

I also removed a CSS class that gave a little extra padding to the menu items seen by all users.  The change is not very noticeable to humans, but it required regeneration of the playwright screenshots.


client-src/elements/chromedash-drawer.js
```diff
@@ -235,29 +230,41 @@ export class ChromedashDrawer extends LitElement {
...
     if (IS_MOBILE && !this.loading) {
-      accountMenu = html`
-      ${this.renderAccountMenu()}
-      `;
+      accountMenu = this.renderAccountMenu();
+    }
+
+    let adminMenu = nothing;
+    if (!this.loading && this.user?.is_admin) {
+      adminMenu = this.renderAdminMenu();
```
jyasskin, line 248: I think you can replace this with just
```suggestion
    let adminMenu = this.renderAdminMenu();
```
Since `loading` seems to just guard whether `this.user` is set, and `renderAdminMenu()` checks `this.user?.is_admin` itself.

dlaliberte (approved):

PR #3628, opened 2024-01-30
Fix gate_url for staging emails.

jrobbins:
We have been generating prod URLs to the gates in the email notifications.  That's correct from the users' point of view, but it makes it a bit of a hassle to test on staging.  This PR uses settings.SITE_URL to generate the correct gate_url.  Also, it DRYs up some redundant code and adds some missing whitespace.

### DIFF
--- a/internals/notifier_helpers.py
+++ b/internals/notifier_helpers.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 from typing import TYPE_CHECKING
+import settings
 from api import converters
 from framework import cloud_tasks_helpers, users
 from internals import core_enums, approval_defs, core_models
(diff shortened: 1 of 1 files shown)


internals/notifier_helpers.py
```diff
@@ -80,6 +84,14 @@ def notify_subscribers_and_save_amendments(
     notify_feature_subscribers_of_changes(fe, amendments, is_update=is_update)
 
 
+def get_gate_url(gate: Gate) -> str:
+  """Reutrn a URL for the user to view the given gate."""
```
dlaliberte, line 88: Return

dlaliberte (approved):
One mispelig.

internals/notifier_helpers.py
jrobbins: Fixed

dlaliberte:
I see that the playwright tests timed out.  `Error: The action has timed out.`.  I previously doubled the total time from 15 minutes to 30, which I thought was more than enough for now.  I'm not sure more time would help.

jrobbins:
> I see that the playwright tests timed out. `Error: The action has timed out.`. I previously doubled the total time from 15 minutes to 30, which I thought was more than enough for now. I'm not sure more time would help.

Fourth try is the charm :)

PR #3637, opened 2024-02-01
Disable the eslint valid-jsdoc check.

jyasskin:
It's deprecated upstream
(https://eslint.org/blog/2018/11/jsdoc-end-of-life/), and it has
false-positives on some of the type expressions we need to accurately
describe our types.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
       { "outerIIFEBody": 0, "SwitchCase": 1, "ignoredNodes": ["TemplateLiteral > *"] }
     ],
     "require-jsdoc": 0,
+    "valid-jsdoc": "off",
     "no-var": 1,
     "arrow-parens": 0,
     "max-len": [2, 100, {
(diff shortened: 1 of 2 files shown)


dlaliberte (approved):
We should be looking for some path to incrementally require accurate typing.  Is there a better alternative to the eslint valid-jsdoc check?

dlaliberte:
> We should be looking for some path to incrementally require accurate typing. Is there a better alternative to the eslint valid-jsdoc check?

From the end-of-life doc page:

>  If you would like to continue checking JSDoc comments using ESLint, we suggest using the community-supported [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc) plugin. This plugin contains several rules that are more specific than the core ESLint rules being deprecated and may actually be a better fit for most projects.

jyasskin:
I think the incremental typing improvements should follow the path laid out by https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html. We're on the stage of adding `// @ts-check` to the tops of JS files that we've fixed up, and we should probably add a call to `tsc` to ... `npm run presubmit`, in another PR. I don't have a strong opinion on whether to add eslint-plugin-jsdoc or which of its rules to turn on.

PR #3638, opened 2024-02-01
Recommend the Ruff extension to everyone

jyasskin:
I removed the pylint and yapf extensions, which Ruff replaces, and I
added a pyproject.toml to hold Python tool configuration.

I also added a .vscode/extensions.json to recommend the same list of
extensions to people who don't use devcontainers.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,6 +19,7 @@
 				"pylint.args": ["--indent-string=\"  \""],
 				"python.defaultInterpreterPath": "/workspace/cs-env/bin/python",
 				"python.editor.tabSize": 2,
+				"python.languageServer": "Jedi",
 				"python.testing.pytestEnabled": false,
 				"python.testing.unittestArgs": [
 					"-v",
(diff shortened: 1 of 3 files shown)


jcscottiii (commented):
The devcontainer and vscode files look good to me.

I have a question though:

There's a pylint command in the package.json

https://github.com/GoogleChrome/chromium-dashboard/blob/1fdfcf5796d945f81c3424d7429adc0253571180/package.json#L40

And since ruff is a linter itself, it would replace pylint, right? If so, we should probably replace/remove this command.

Additionally, if ruff is replacing pylint, I will defer to @jrobbins if he has any opinions on keeping pylint in the project vs changing to ruff.

jrobbins:
It's OK with me if we add ruff and also leave the pylint target in package.json.  Nothing runs pylint on a regular basis right now, but it would be nice to still be able to consult that output... if we ever get around to cleaning up some of that output.

jrobbins (approved):

jyasskin:
Yeah, the pylint target reports a lot of violations right now. It might be worth doing a mass reformat with Ruff and then replace that npm script with a reformat run, but it's probably good to try it out for a bit with just the extension before changing the whole codebase.

PR #3639, opened 2024-02-01
Regenerate the base css.

jyasskin:
This is the change that results from `npm run start-app` on a checkout of `main`.

### DIFF
--- a/static/css/base.css
+++ b/static/css/base.css
@@ -447,9 +447,18 @@
   --sl-z-index-tooltip: 1000;
 }
 
-.sl-scroll-lock {
-  padding-right: var(--sl-scroll-lock-size) !important;
-  overflow: hidden !important;
+@supports (scrollbar-gutter: stable) {
+  .sl-scroll-lock {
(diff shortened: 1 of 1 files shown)


jrobbins (approved):

dlaliberte (approved):
I'm not seeing that change.  But I am seeing an error when it tries to connect to the server: 
```
curl: (7) Failed to connect to localhost port 15606 after 0 ms: Couldn't connect to server
Warning: Problem : connection refused. Will retry in 1 seconds. 4 retries left.
```
This occurs (for me) even though `npm start` does seem to work fine.

Anyway, this change seems fine regardless.   Maybe we are having permission issue regarding automatic update of the base.css file that isn't affecting you.  In my repo, I see:
```
(cs-env) dlaliberte@dlali16:~/chromium-dashboard$ ls -l static/css/base.css
-rw-r--r-- 1 dlaliberte primarygroup 19286 Jan 12 00:21 static/css/base.css
```
So it should be writable by me, but perhaps not by some auto-process?

PR #3640, opened 2024-02-01
Preserve casing in the OpenAPI JS bindings.

jyasskin:
It seems like it'll be confusing to have the API documentation use a different casing from the code we actually write. If we want a particular casing, we should use it in the API spec. (Do we want camelCase for these two fields?)

I left the paramNaming as camelCase because it doesn't seem to react to changes in the API casing (a bug in the generator?), and we don't want "ComponentUsersRequest" to start with a capital when used as an object member.


jcscottiii (approved):
LGTM. Thanks for this!

PR #3654, opened 2024-02-08
Add a report on which people have mentored specs recently.

jyasskin:
This currently treats any change to the feature as the mentor having done some work, which isn't correct, but it should still work to give us a rough picture of people's activity.

This report looks like:

Screenshot of the report proposed in this PR


api/spec_mentors_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+from chromestatus_openapi.models.feature_link import FeatureLink
+from chromestatus_openapi.models.spec_mentor import SpecMentor
+
+from framework import basehandlers
+from internals.core_models import FeatureEntry
+
+
+class SpecMentorsAPI(basehandlers.APIHandler):
+  """Implements the OpenAPI /specmentors path."""
```
jcscottiii, line 25:
```suggestion
  """Implements the OpenAPI /spec_mentors path."""
```

openapi/api.yaml
```diff
@@ -81,6 +81,26 @@ paths:
...
+      responses:
+        '200':
+          description: >-
+            List of all the matching spec mentors.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/SpecMentor'
```
jcscottiii, line 103: If you anticipate other consumers of this API, could you add a 400 response for the invalid `after` parameter case?

api/spec_mentors_api.py
```diff
@@ -0,0 +1,63 @@
...
+        after = datetime.fromisoformat(after_param)
+      except ValueError:
+        self.abort(400, f'invalid ?after parameter {after_param}')
+
+    # Every non-empty string is greater than '', so this will find all entries with any spec mentors
+    # set. (!= is interpreted as "less or greater".)
+    # We can't have NDB sort the results because it can only sort by the inequality condition.
+    query = FeatureEntry.query(FeatureEntry.spec_mentor_emails > '')
+    features: list[FeatureEntry] = query.fetch()
+    if after is not None:
```
jcscottiii, line 46: Could you add a few unit tests please?
1. Happy case when after is supplied
2. Happy case when after is not supplied
3. Case when the feature is unlisted but has spec_mentors to assert that they don't show up.

Case 3 can be folded into one of the other cases.

api/spec_mentors_api.py
```diff
@@ -0,0 +1,63 @@
...
+    if after_param is not None:
+      try:
+        after = datetime.fromisoformat(after_param)
+      except ValueError:
+        self.abort(400, f'invalid ?after parameter {after_param}')
+
+    # Every non-empty string is greater than '', so this will find all entries with any spec mentors
+    # set. (!= is interpreted as "less or greater".)
+    # We can't have NDB sort the results because it can only sort by the inequality condition.
+    query = FeatureEntry.query(FeatureEntry.spec_mentor_emails > '')
```
jcscottiii, line 44: Do we have any idea of how many results this should return? I know that for some queries to FeatureEntry we impose an upper limit DEFAULT_RESULTS_PER_PAGE. 

And as there are more FeatureEntries added, I'm assuming we want to encourage people to use the spec_mentor_emails field. 

The `after` parameter will protect against returning a lot of results (which the frontend client provides). But it is not a required field. So there's no upper limit on how far back in the past other clients can go.

All this said, this could be fine since this is used in a report page. So latency could be okay. But I just wanted to bring it up.

jcscottiii (commented):
Some suggestions, comments and questions.

Thanks for trying out the openapi functionality!

client-src/elements/chromedash-drawer.js
```diff
@@ -261,6 +261,9 @@ export class ChromedashDrawer extends LitElement {
         ${this.renderNavItem('/metrics/css/animated', 'CSS Animation')}
         ${this.renderNavItem('/metrics/feature/popularity', 'JS/HTML')}
 
+        <div class="section-header">Reports</div>
```
jrobbins, line 264: According to our current menu style, this should have an `<hr>` above it.  I plan to change the menu style but that hasn't happened yet.

FYI, this new "Reports" section seems like it would be interesting to a fairly narrow subset of users, so I think that when I redo the menu style using sl-tree, I will make this section initially collapsed.

jrobbins (approved):
LGTM after making one UI change.

jyasskin (commented):
I also added a playwright test to check that the report shows what it's supposed to.

api/spec_mentors_api.py
jcscottiii: I am fine with leaving it for now.

jcscottiii (approved):

PR #3666, opened 2024-02-22
OT Extension - Add endpoint to submit origin trial extension

DanielRyanSmith:
This change adds a new endpoint to the Chromestatus API in order to submit approved extension requests to be processed. Additionally, this change will redirect the user directly to the feature detail page with the extension gate open when the user submits an extension request.

More changes to actually use this endpoint in the UI will come in a subsequent PR.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -68,7 +68,7 @@ class OriginTrialsAPI(basehandlers.APIHandler):
     feature_id = int(kwargs['feature_id'])
     extension_stage_id = int(kwargs['extension_stage_id'])
     # Check that feature ID is valid.
-    if feature_id is None or feature_id == 0:
+    if not feature_id:
       self.abort(404, msg='No feature specified.')
     feature: FeatureEntry | None = FeatureEntry.get_by_id(feature_id)
     if feature is None:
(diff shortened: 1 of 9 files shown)


api/origin_trials_api.py
```diff
@@ -33,3 +37,76 @@ def do_get(self, **kwargs):
...
+    gate = Gate.query(Gate.stage_id == extension_stage.key.integer_id()).get()
+    if not gate or gate.state != Vote.APPROVED:
+      self.abort(400, 'Extension has not received the required approvals.')
+
+  def do_post(self, **kwargs):
+    """Extends an existing origin trial"""
+    feature_id = int(kwargs['feature_id'])
+    extension_stage_id = int(kwargs['extension_stage_id'])
+    # Check that feature ID is valid.
+    if feature_id is None or feature_id == 0:
```
jrobbins, line 71: `if not feature_id:` would be more pythonic.  The same change could be made again below.

jrobbins (approved):

api/origin_trials_api.py
DanielRyanSmith: Agreed. Updated!

PR #3670, opened 2024-02-24
docs: fix the version of python

kshitjj:
README.md file has the wrong Python version causing `npm run setup` to fail.

This is happening because "setup" script in `package.json` uses python3.11 while the README tells us to install `python3.10-venv`: 

https://github.com/GoogleChrome/chromium-dashboard/blob/b6d1c554ff47e39e8a893548963d9aadf339eca4/package.json#L15

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ For a one-click setup that leverages devcontainers, check out the devcontainer
 1. Install other developer tools commands
     1. node and npm.
     1. Gulp: `npm install --global gulp-cli`
-    1. Python virtual environment: `sudo apt install python3.10-venv`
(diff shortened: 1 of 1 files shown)


kshitjj:
I have signed the CLA, not sure how I can make the checks pass now.

kshitjj:
@jrobbins Can you tell how can I fix the CLA? Should I add one more commit to make the check run again? :sweat_smile:

jrobbins:
> @jrobbins Can you tell how can I fix the CLA? Should I add one more commit to make the check run again? 😅

I re-ran the test for you and it passed.

jrobbins (approved):
Thanks for catching this.

PR #3671, opened 2024-02-26
Colour code generated in the notification email

impact-maker:
Hi, 
In order to differentiate between the old and new text in the email notification changes a new function with red for deletions and green for additions has been added.


jrobbins:
That function looks potentially useful, but it should be actually called from the relevant place in the existing code.

Issue #3393, opened 2023-10-05
Email notifications incorrectly say that the feature owner made a change when a reviewer votes


jrobbins:
Example of an incorrect email notification:
```
andruud@chromium.org updated this feature:

Relaxed CSS Nesting

Components: Blink>CSS

Implementation status: Proposed

Estimated milestones:

Shipping on desktop	120
Shipping on Android	120
Shipping on WebView	120
Changes:

yoavweiss@chromium.org set review status in https://chromestatus.com/feature/5070369895743488?gate=5158552956305408:
old: no_response
new: approved
```

andruud@chromium.org did nothing to trigger this email, so that line at the top is incorrect.  yoavweiss@chromium.org is correctly cited near the bottom.

impact-maker:
@jrobbins Hi, could you guide on how to reproduce this issue.

Tabhi109:
Hi @jrobbins , Can you please provide a bit more information regarding this issue , I want to solve this and make my first open source contribution.
Looking forward for your reply.

obamwonyi:
I would also love to contribute to this .

Rishikesh63:
@jrobbins & @DanielRyanSmith I would like to work on it.
i can solve this bug using html & javascript or html & python.

jrobbins:
This is still a problem.  See screenshot below.

I started work on this and didn't finish it, so I should be the one to continue on it.  I'll try to get it done this week. 

![image](https://github.com/GoogleChrome/chromium-dashboard/assets/17365/ad6a2e4c-203f-4026-bde9-c0212fd061c8)

kshitjj:
Hey @jrobbins, 

I have made the [PR](#3674)  which I think fixes this issue, would love to get your review :D

Rishikesh63:
@jrobbins  has this solved?

PR #3674, opened 2024-02-26
fix incorrect email notification

kshitjj:
Issue: #3393

Description: Incorrect email used when creating an update feature email

The issue arises from the use of fe.updater_email which results in the owner's email being used. This happens because fe.updater_email corresponds to the owner's email, as the owner is considered the updater when they create the feature.

In the code snippet, updater_email is not passed as a parameter, causing it to default to fe.updater_email.
https://github.com/GoogleChrome/chromium-dashboard/blob/a09e8cb0a04bea37a68e635ff91e68b83fc85c37/internals/notifier.py#L94-L96

The issue was resolved by passing the correct email obtained while creating the `Changes:` part of the email as a parameter. This ensures that the appropriate email is used in the update feature email.

https://github.com/GoogleChrome/chromium-dashboard/blob/a09e8cb0a04bea37a68e635ff91e68b83fc85c37/internals/notifier_helpers.py#L141-L145

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -197,7 +197,7 @@ def add_core_receivers(fe: FeatureEntry, addr_reasons: dict[str, list[str]]):
 
 
 def make_feature_changes_email(fe: FeatureEntry, is_update: bool=False,
-    changes: Optional[list]=None):
+    changes: Optional[list]=None, local_updater_email: Optional[str]=None):
   """Return a list of task dicts to notify users of feature changes."""
   if changes is None:
     changes = []
@@ -214,7 +214,10 @@ def make_feature_changes_email(fe: FeatureEntry, is_update: bool=False,
     triggering_user_email = fe.creator_email
     template_path = 'new-feature-email.html'
 
-  email_html = format_email_body(template_path, fe, changes)
+  if local_updater_email:
+      email_html = format_email_body(template_path, fe, changes, updater_email=local_updater_email)
+  else:
+      email_html = format_email_body(template_path, fe, changes)
 
   addr_reasons: dict[str, list[str]] = collections.defaultdict(list)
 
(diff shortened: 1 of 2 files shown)


jrobbins (approved):

jrobbins (changes requested):
Actually, I think you have changed the parameters for a different task than the one that you are giving the new parameters to.

kshitjj:
I could be completely wrong here,  but I think the parameter is given to the right task `notify_subscribers_of_vote_changes()` enqueue a task to `/tasks/email-subscribers`

https://github.com/GoogleChrome/chromium-dashboard/blob/cbcd94ef0c8afa6039bece37e787d662ebd34032/internals/notifier_helpers.py#L155

which is then routed to the notifier.FeatureChangeHandler in main.py
https://github.com/GoogleChrome/chromium-dashboard/blob/cbcd94ef0c8afa6039bece37e787d662ebd34032/main.py#L264

Then FeatureChangeHandler receives the parameter
https://github.com/GoogleChrome/chromium-dashboard/blob/e0ad40211e9959cad5ccab7c9a1bc9b1804c209a/internals/notifier.py#L404-L415

kshitjj:
It seems like I pasted it in the wrong function while switching branches :sweat_smile:

jrobbins:
I still don't think that you quite got it right.  The existing code is probably more complicated than it should be, so it is challenging to make a change here.  Inspired by your change, I went ahead and worked through a fix that I think is right in #3675.  If that PR gets merged, we can close this one.  Thanks for bringing my attention to this bug.

kshitjj:
I guess I accidentally helped you. 😅

can you refer to me documentation that will help understand me the workflow?

jrobbins:
There's no documentation for that code.   You can see which functions work together by their names, and by the URI of the task generated in notification_helpers.py vs. the class that handles that task specified in main.py.

Thanks again.

PR #3676, opened 2024-02-28
Highlight Difference in notification email fix via colour code 

### LINKED ISSUE #1303: Highlight diffs in notification emails
It can be hard to see what text changed in an update notification.  The notifications contain old and new text for each field that changed, but the specific words that changed within those field values can be hard to spot.

We could use an HTML span with a pale red or green background to indicate words that changed.

In cases where the entire text has changed (or maybe over 50%) we can skip highlighting because it would not be too useful. We should try a few different thresholds to see how they work with realistic data.

This could be implemented with python's difflib.  E.g.,
>>> list(difflib.ndiff('hello there , how are you?'.split(), 'hey there , where are you?'.split()))
['- hello', '+ hey', '  there', '  ,', '- how', '+ where', '  are', '  you?']

**Additional context**
Color alone should never be used to convey important information,  so we should keep the Old and New labels.

impact-maker:
This PR fixes #1303 
Description: Highlight difference function is added which introduces colour codes to insertions and deletions in changed text of the email body


DanielRyanSmith:
As far as I can tell, the new function is not invoked anywhere, so it would not have an effect on the email output. Be sure to run `npm run test` to see possible new test failures that need to be fixed based on your change once you have updated the code to use you new function.

PR #3678, opened 2024-02-29
Update intent detection logic for multiple stages

DanielRyanSmith:
This change updates the intent detection logic to function even in the presence of multiple stages of the same type.

If multiple stages of the same type exist that correspond to a new intent thread, the logic will check if any of those stages have a set intent thread URL matching the detected thread URL. If no stages are associated with that intent thread URL and exactly one of those stages has an unset intent thread URL, it will be assumed that the newly detected intent thread is associated with that stage.

Additionally, the gate is always queried for before setting a vote, so that we definitively discern which gate the vote belongs to.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -242,24 +242,38 @@ class IntentEmailHandler(basehandlers.FlaskHandler):
       logging.info('stage_type not found for %r %r',
                    approval_field.field_id, feature.feature_type)
       return
-    # TODO(danielrsmith): A new way to approach this detection is needed
-    # now that multiple stages of the same type can exist for a feature.
-    # This will currently only detect an intent if there is only 1 stage
-    # of the specific stage type associated with the feature.
-    matching_stages: list[Stage] = Stage.query(
+
+    # Get all stages of the detected stage type that belong to the feature.
+    stages_of_type_in_feature: list[Stage] = Stage.query(
         Stage.feature_id == feature.key.integer_id(),
         Stage.stage_type == stage_type).fetch()
-    if len(matching_stages) == 0 or len(matching_stages) > 1:
-      logging.info('Ambiguous stages: %r', matching_stages)
-      return
-    if matching_stages[0].intent_thread_url:
-      logging.info('intent_thread_url was already set to %r',
-                   matching_stages[0].intent_thread_url)
+    if len(stages_of_type_in_feature) == 0:
+      logging.info('No matching stage found for feature: '
+                   f'{feature.key.integer_id()}, stage_type: {stage_type}')
       return
 
-    matching_stages[0].intent_thread_url = thread_url
-    matching_stages[0].intent_subject_line = subject
-    matching_stages[0].put()
+    # Check if this intent URL already belongs to a stage.
+    matching_stage = next((s for s in stages_of_type_in_feature
+                            if s.intent_thread_url == thread_url), None)
+    if matching_stage:
+      matching_stage.intent_subject_line = subject
+      matching_stage.put()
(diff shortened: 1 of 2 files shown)


internals/detect_intent_test.py
```diff
@@ -74,10 +74,12 @@ def test_detect_field(self):
           'Intent to Extend Experiment: Something cool',
           'Intent to Continue Experiment: Something cool',
           'Intend to Continue Experiment: Something cool',
+          'Intent to Extend Origin Trial: Something cool',
           'Intending to Continue Experiment: Something cool',
           'Request to Continue Experiment: Something cool',
           'Request to Continue Experimenting: Something cool',
           'Request to Continuing Experiment: Something cool',
+          'Request to Extend Origin Trial: Something cool',
```
DanielRyanSmith, line 82: Unrelated addition that rounds out intent detection cases for I2EEs

jrobbins:
Is this ready for review?

DanielRyanSmith:
> Is this ready for review?

Yes, thanks for the reminder

internals/detect_intent.py
```diff
@@ -228,42 +238,57 @@ def load_detected_feature(self, feature_id: Optional[int],
...
-      logging.info('Ambiguous stages: %r', matching_stages)
-      return
-    if matching_stages[0].intent_thread_url:
+    if len(stages_of_type_in_feature) == 0:
+      logging.info('No matching stage found for feature: '
+                   f'{feature.key.integer_id()}, stage_type: {stage_type}')
+      return None
+
+    # Check if this intent URL already belongs to a stage.
+    matching_stage = next((s for s in stages_of_type_in_feature
```
jrobbins, line 266: I don't think that there will ever be a match because the `thread_url` is actually contains the SMTP Message-ID value of a specific message within the thread like `/d/msgid/blink-dev/MESSAGE-ID`.  The Message-ID of the first message in a thread will not == the Message-ID of later messages in the same thread.  We use the URL of the first message in the thread to represent the whole thread, but we never store a URL for the whole thread.

I think this code would often work anyway, simply because when a feature owner creates a new origin trial, and then posts anything to a new or old "Intent to experiment" thread about that feature, they are probably talking about the new origin trial or more-or-less ready to review it.  Worst case, the review shows up on the reviewers' dashboard too soon.

To make this work precisely, I think that we would need to change the URL that is embedded in the intent email to include `?gate=GATE_ID`.  That could be a helpful change for all intents anyway.

jrobbins (approved):
I have a couple of small comments and one big one, but you could proceed even without a big change for the big one.

internals/detect_intent.py
DanielRyanSmith: I see. I agree that it's probably best for us to use the gate ID query string to make sure we're obtaining the correct gate. I can update this logic and add the query string in a subsequent PR. 🙂

internals/detect_intent.py
```diff
@@ -195,10 +195,20 @@ def process_post_data(self, **kwargs):
     if not feature:
       logging.info('Could not retrieve feature')
       return {'message': 'Feature not found'}
-
-    self.set_intent_thread_url(feature, approval_field, thread_url, subject)
+    
+    stage = self.find_matching_stage(feature, approval_field, thread_url)
+    if stage is None:
+      return {'message': (
```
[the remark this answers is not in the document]
DanielRyanSmith: Good catch - I've added a log for this.

internals/detect_intent.py
DanielRyanSmith: Adding a TODO here, but will follow up ASAP

PR #3682, opened 2024-03-02
Highlight Difference in notification email fix

impact-maker:
This PR fixes https://github.com/GoogleChrome/chromium-dashboard/issues/1303
Description: Highlight difference function is added which introduces colour codes to insertions and deletions in changed text of the email body

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -64,17 +64,18 @@ def _determine_milestone_string(ship_stages: list[Stage]) -> str:
     milestone_str = f'{first_android} (android)'
   return milestone_str
 
-def highlight_diff(old_text, new_text):
-  differ = difflib.ndiff(old_text.split(), new_text.split())
-  highlighted_text = []
-  for item in differ:
-    if item.startswith('-'):
-      highlighted_text.append(f'<span style="background-color: #FFCCCC">{item[2:]}</span>')
-    elif item.startswith('+'):
-      highlighted_text.append(f'<span style="background-color: #CCFFCC">{item[2:]}</span>')
-    else:
-      highlighted_text.append(item[2:])
-  return ' '.join(highlighted_text) 
+def highlight_diff(old_text, new_text, highlight_type):
+    differ = difflib.ndiff(old_text.split(), new_text.split())
+    highlighted_text = []
+    for item in differ:
+        text = escape(item[2:])  # Escaping the text
+        if item.startswith('-') and highlight_type == 'deletion':
+            highlighted_text.append(f'<span style="background-color: #FFCCCC">{text}</span>')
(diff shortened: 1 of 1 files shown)


impact-maker:
@jrobbins  @DanielRyanSmith  I have implemented highlight function and called it from the format email function. I ran npm run test and tested it too. Can you provide a feedback on this pull request

internals/notifier.py
```diff
@@ -79,9 +91,13 @@ def format_email_body(
     new_val = prop['new_val']
     old_val = prop['old_val']
 
-    formatted_changes += ('<li><b>%s:</b> <br/><b>old:</b> %s <br/>'
-                          '<b>new:</b> %s<br/></li><br/>' %
+    # Call highlight_diff to get the highlighted differences
+    highlighted_new_val = highlight_diff(old_val, new_val)
```
jrobbins, line 95: This produces a highlighted diff, but it is not highlighted quite the way that would be expected by people who are familiar with viewing diffs.   It puts all the highlights into the new value and nothing on the old value.  It looks like:
image

It would be better to call highlight_diff() twice, once on the old value but highlighting only deletions, and once on the new value but highlighting only additions.  Then it could look like:
image



Also, you are not escaping the text that goes into or out of the differ.  That is a problem, because the input could include markup because our users discuss HTML tags, or if a malicious user tried to do something funny.  In this example, I changed the words "list item" to "<li>":
image

jrobbins (changes requested):
Thanks for doing this.  I found a few problems that should be fixable.

impact-maker:
I have worked on the feedback provided and made the following changes
   1. Separates the highlighting into two distinct calls for old and new values, accurately reflecting the standard diff presentation.
   2. Escapes all text to prevent HTML injection, making the function safer for handling arbitrary input, including those containing   HTML tags or special characters.
  3. Uses f-strings for cleaner and more readable string formatting throughout.

internals/notifier.py
impact-maker: It now separates the highlighting into two distinct calls for old and new values
It  escapes all text to prevent HTML injection, making the function safer

internals/notifier.py
```diff
@@ -79,9 +91,13 @@ def format_email_body(
...
+
+    formatted_changes += ('<li><b>{prop_name}:</b> <br/><b>old:</b> {escape(old_val)} <br/>'
```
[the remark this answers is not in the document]
impact-maker: It now uses f-strings for cleaner and more readable string formatting

jrobbins:
I will merge your change, and then merge my enhancements to it immediately afterwards.

jrobbins (approved):

PR #3688, opened 2024-03-04
Update intent template for extensions

DanielRyanSmith:
List of changes:

- Separates the intent enums for "origin trial" and "origin trial extension", which were previously using the same intent enum.
- Adds extension process stages with "Draft Intent to Extend Experiment" action.
- Updates the intent template to contain extension milestones.


client-src/elements/form-field-enums.js
```diff
@@ -204,7 +204,7 @@ export const INTENT_STAGES = {
   INTENT_IMPLEMENT: [1, 'Start prototyping'], // Start prototyping
   INTENT_EXPERIMENT: [2, 'Dev trials'], // Dev trials
   INTENT_IMPLEMENT_SHIP: [4, 'Evaluate readiness to ship'], // Eval readiness to ship
-  INTENT_EXTEND_TRIAL: [3, 'Origin Trial'], // Origin trials
+  INTENT_ORIGIN_TRIAL: [3, 'Origin Trial'], // Origin trials
```
jrobbins, line 207: Do you also need the new enum value here?

internals/processes.py
```diff
@@ -265,9 +263,18 @@ def process_to_dict(process):
                PI_EXPLAINER.name, PI_SPEC_LINK.name,
                PI_EST_TARGET_MILESTONE.name])],
       [approval_defs.ExperimentApproval],
-      core_enums.INTENT_IMPLEMENT_SHIP, core_enums.INTENT_EXTEND_TRIAL,
+      core_enums.INTENT_IMPLEMENT_SHIP, core_enums.INTENT_ORIGIN_TRIAL,
       stage_type=core_enums.STAGE_BLINK_ORIGIN_TRIAL),
 
+  ProcessStage(
```
jrobbins, line 269: Since the Guide UX is driven by these process definitions, I think that every feature's Guide UX page will immediately show a stage to extend the origin trial.  I suppose that is OK.

DanielRyanSmith: When we send this over to the front end, the extension stages are organized a little differently (as properties of the origin trial stage), so it seems like this is not the case. 🙂 No regression there as far as I can see.

client-src/elements/form-field-enums.js
DanielRyanSmith: Yes! added.

Issue #2512, opened 2022-11-28
Toast is briefly displayed on page load


jrobbins:
On several of our pages, the chromedash-toast element briefly appears in the bottom left corner if the page is initially loaded or reloaded (not on a SPA navigation).   The fact that the toast flashed by looks like a mistake or error, even though it does not prevent the user from accomplishing any use case.  We should prevent it because it is not intended to be initially displayed.

jrobbins:
Several pages: roadmap, myfeatures, newfeatures, and probably others.

DanielRyanSmith:
https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/63db554a-dc29-4089-9871-3562f59a12b6

Here is an example of the behavior you can see on the bottom left on page refresh.

r3yc0n1c:
Hi, I'd like to work on this.

DanielRyanSmith:
Hey @r3yc0n1c, feel free to open a PR for this work, and you can pose any questions you have here as well!

r3yc0n1c:
Sure, there was some setup issues but now I got the dev-env ready, working on it... will drop a PR soon!

impact-maker:
@r3yc0n1c  Hi I have
 already made a pull request on this issue and closing it.

r3yc0n1c:
> @r3yc0n1c Hi I have already made a pull request on this issue and closing it.

Oh, I thought this is open to work, Daniel assigned it to me.

PR #3698, opened 2024-03-08
Fixed toast element is not initially displayed

impact-maker:
Issue fixed: #2512 
Description: One can prevent the toast from briefly appearing on page load, by utilising the connectedCallback lifecycle method to delay the fetching of data until after the initial rendering. With this modification, the fetchData method is only called during the connectedCallback if the dataFetched flag is false, ensuring that it runs only once during the initial page load and not during subsequent renders


impact-maker:
@jrobbins @DanielRyanSmith I have worked on this bug for quite some time now. Could you give me a feedback on this pull request

client-src/elements/chromedash-activity-prevent_toast.js
```diff
@@ -0,0 +1,59 @@
+
```
DanielRyanSmith, line 1: It doesn't seem like this component is used anywhere. I cannot tell what the intention is to avoid the initial toast being displayed for the short time window at the beginning of a new load. What do you see when you run the project locally?

DanielRyanSmith:
This is close, and fixes the problem that the toast component was being seen during page load, but causes a separate problem. The toast component now no longer shows the transition of moving up and down when it is displayed in other forms.

https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/9c6e75eb-9a39-428e-8933-09fb490dbf4f

Previously, the toast component would move up and down as it was displayed. See below for example.

https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/3df30f39-3131-4d15-bea6-ef845d8a267e

DanielRyanSmith:
A fix for this issue has since been landed.

PR #3712, opened 2024-03-15
Impact maker/highdiff branch

jrobbins:
This is a continuation of impact-maker's #3682.  By committing to this branch, I meant to amend that PR, but instead GitHub created a new one.

In my commit, I add some tests, use `re.split()`, and avoid double escaping, and use shorter CSS and lighter background colors.


jrobbins:
Screenshot:
Note that the python list syntax around the explainer_links old value is a separate problem not related to this PR.
Also, the motivation was bullet items split over a few lines, but we don't use `<pre>` in these emails, which is also a separate issue.
image

DanielRyanSmith (approved):

PR #3719, opened 2024-03-19
Fix type when editing enterprise_impact field.

jrobbins:
This is a quick change to avoid the exception that I emailed you about.  

However, there might be a better way to do this if the field types are declared somewhere and this field is missing.  I didn't trace through everything.


yanndago:
LGTM, but I think we should do the same everywhere we have a 'new_fields.get('enterprise_impact' ...), casting it to a int

yanndago (approved):

yanndago (approved):

PR #3723, opened 2024-03-19
Implement a page for feature latency metrics.

jrobbins:
This implements a new reporting page that will show features that launched in the previous calendar year and how long each one took between chromestatus entry creation and their launch milestone.

In this PR:
* Add a page to chromestatus-app.js and main.py.  And, define a simple lit component that renders a table.
* Add an openapi call that takes a date range and returns a list of `FeatureLatency` records.  This causes a lot of changes to generated code.
* Implement that API with a handler to calculate which features launched within the range.


jrobbins:
Screenshot:
image

api/feature_latency_api_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+  def setUp(self):
+    self.handler = feature_latency_api.FeatureLatencyAPI()
+
+  def test_get_date_range__unspecified(self):
+    """If query string params were not set, it returns None, None."""
+    actual = self.handler.get_date_range({})
+    self.assertEqual((None, None), actual)
+
+  def test_get_date_range__specified(self):
```
KyleJu, line 33: What about test cases for do_get?

gen/js/chromestatus-openapi/package.json
```diff
@@ -16,6 +16,6 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "typescript": "^5.4"
+    "typescript": "^4.0"
```
KyleJu, line 19: Again, should this be a separate PR?

api/feature_latency_api.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class FeatureLatencyAPI(basehandlers.APIHandler):
+  """Implements the OpenAPI /feature_latency path."""
+
+  def get_date_range(
+      self, request_args: dict[str, str]
+  ) -> tuple[datetime|None, datetime|None]:
+    """Parse start and end dates from query-string params."""
+    start_param: str | None = request_args.get('startDate', None)
```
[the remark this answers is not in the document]
jrobbins: I think of the python code using snake case, and the JS code using camel case, and the names of parameters following the convention of the code that sent the parameter.  It can look a little out of place when you have a camel case name appearing in a python file, but that tells you that it came from JS.

gen/js/chromestatus-openapi/package.json
jrobbins: I reverted these files.

jrobbins:
This is ready for review again.  It now includes the logic for the report and tests.

gen/js/chromestatus-openapi/package.json
KyleJu: Still here!

package-lock.json
```diff
@@ -75,7 +75,7 @@
     "gen/js/chromestatus-openapi": {
       "version": "1.0.0",
       "devDependencies": {
-        "typescript": "^5.4"
+        "typescript": "^4.0"
```
KyleJu, line 78: ditto. A separate PR so it can be easily reverted if something goes wrong

KyleJu (approved):
LGMT % minor comments

PR #3731, opened 2024-03-22
Reimplement input mirroring for OT milestone fields

### LINKED ISSUE #3173: Reimplement input mirroring for OT milestone fields
After #3162 has landed, the OT milestone fields on the stage edit page will no longer mirror their input into the other start and end fields if there is no existing value in those inputs. A special handler will need to be created to reimplement this input mirroring in a way that designates the other input fields as "touched" for form submission.

mikezzb:
Fixes #3173

- If the value in the text box for the OT/DT desktop milestone start/end field is changed, and the other android/webview milestone start fields are empty or were the same value as the desktop milestone previously, their values are also changed to match.
- If this value mirroring occurs, the touched property on these fields are all changed to "true" (which is how we know to update the field once the form is submitted).

Demo:

https://github.com/GoogleChrome/chromium-dashboard/assets/23418428/9a00fe12-47b8-4bab-88b7-1445f89b27b0

### DIFF
--- a/client-src/elements/chromedash-guide-stage-page.js
+++ b/client-src/elements/chromedash-guide-stage-page.js
@@ -9,6 +9,7 @@ import {
 import './chromedash-form-table';
 import './chromedash-form-field';
 import {
+  COPY_ON_EDIT,
   FLAT_TRIAL_EXTENSION_FIELDS,
   formatFeatureForEdit,
   FORMS_BY_STAGE_TYPE} from './form-definition';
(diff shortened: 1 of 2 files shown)


DanielRyanSmith:
This implementation was close, but the fields only update once the input field focus changes. It is not intuitive to the user that the other fields are being changed, and there is a risk that the user updates one field and immediately clicks "Submit", which then would change the other fields just as the form is submitting, without letting the user see what is happening.

In the end, this change is not a high priority anymore and we're fine letting this go. Sorry for the response delay here.

PR #3742, opened 2024-03-25
Speed up permissions check

jrobbins:
One of the API Owners complained that the review dashboard was loading slowly, and specifically that the call to get the user permissions was taking almost 1000ms.

In this PR:
* Log the start and end of computing permission data so that we can keep an eye on performance in the future.
* When getting the set of features that the user can edit, just get the IDs since that is all we need.
* Cache the list of users who can review a given gate type.  This replaces some ndb queries with reds queries. 

These changes reduce the latency of this call from my desktop browser from around 850ms to about 350ms.

### DIFF
--- a/framework/permissions.py
+++ b/framework/permissions.py
@@ -97,10 +97,10 @@ def feature_edit_list(user: User) -> list[int]:
     return []
 
   # Query features to find which can be edited.
-  features_editable = feature_helpers.get_all(
-    filterby=('can_edit', user.email()))
+  editable_feature_keys = feature_helpers.get_all(
+      filterby=('can_edit', user.email()), keys_only=True)
   # Return a list of unique ids of features that can be edited.
-  return list(set([f['id'] for f in features_editable]))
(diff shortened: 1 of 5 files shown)


framework/permissions.py
```diff
@@ -97,10 +97,10 @@ def feature_edit_list(user: User) -> list[int]:
     return []
 
   # Query features to find which can be edited.
-  features_editable = feature_helpers.get_all(
-    filterby=('can_edit', user.email()))
+  editable_feature_keys = feature_helpers.get_all(
+      filterby=('can_edit', user.email()), keys_only=True)
```
DanielRyanSmith, line 101: It looks like the `get_all()` function's return type was mistyped (presumably by me 😅). and it actually is returning a list of keys or a list of verbose FeatureEntry dictionaries, so the return type needs to change to: `list[dict | ndb.Key]` - I think that should fix the mypy problems.

DanielRyanSmith (approved):

PR #3744, opened 2024-03-26
Speed up sorting of pending reviews.

jrobbins:
This improves the performance of feature search queries that are used on the chromestatus reviewers' dashboard.  

Specifically, it makes the step that gets sorted feature IDs use an async query rather than synchronous.  This allows some of the integer set union and intersection work to be done in parallel with the sorting query.  This is only a small improvement because the set logic stuff executes quickly anyway.

In this PR:
* Add more logging so that we can see where time is being spent during request processing.
* Make `sorted_by_pending_request_date()` and `sorted_by_review_date()` return futures rather than blocking while their query executes.  The ability to handle futures is already supported in the rest of the search pipeline.
* Make the final logging statement generate a shorter string

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -233,6 +233,7 @@ def process_query(
       permission_terms)
 
   # 2c. Create a parallel query for total sort order.
+  logging.info('creating total sort order for %r', sort_spec)
   total_order_promise = search_queries.total_order_query_async(sort_spec)
 
   # 3. Get the result of each future and combine them into a result ID set.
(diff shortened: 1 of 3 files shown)


DanielRyanSmith (approved):

jrobbins:
> > Specifically, it makes the step that gets sorted feature IDs use an async query rather than synchronous.
> 
> Could you explain how this is achieved by returning Future objects?

An NDB Query object has two methods `fetch()` and `fetch_async()`.  The `fetch()` method sends the query the datastore backend and waits for a response, which could take anywhere from 200ms to 4000ms for this query.  In contrast, the `fetch_async()` method sends a query to the backend, but then immediately returns a `Future` object, allowing execution of python code to proceed until we eventually wait on the value of the future.

In our main search method, `process_query()` there are several different queries run to (a) find matching features, (b) find the set of features that the user is allowed to view, and (c) make a list of all features sorted by the user's preferred sort order.  After all those pieces of data are available as python lists and sets, then the python code does (d) some set intersection and (e) lookup logic to produce the filtered, sorted list of feature IDs. 

Before this change, we were doing (d) and (e) after (a), (b), and (c) had all executed.
With this change, we do (d) after (a) and (b), so it can potentially start sooner, and then (e) after (a), (b) and (c).
So, basically (d) can run while the sorting query is still in-progress rather than waiting for it to finish.
The speed up is only in the case where query (c) is the slowest, and the amount of the speed up is only the time it takes to do (d) which is usually not very long, probably less than 50ms.

PR #3745, opened 2024-03-26
Turn feature links into a custom element.

jyasskin:
This puts the styles closer to where they're used and unifies the `rel` values. It should keep the fallbacks the same, except for the "(tracking bug)" link which I replaced with a `<sl-tag>`.


client-src/elements/chromedash-link.js
```diff
@@ -319,72 +321,169 @@ function enhanceGoogleDocsLink(featureLink) {
...
+            <sl-badge size="small" variant="${featureLink.http_error_code >= 500 ? 'danger' : 'warning'}">
+              ${featureLink.http_error_code}
+            </sl-badge>
+            ${this.fallback()}
+          </sl-tag>`;
+      }
+      return this.fallback();
+    }
+    try {
+      switch (featureLink.type) {
```
jyasskin, line 449: Switch github to [ignore whitespace](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#diff-view-options) to make most of this diff section go away.

jrobbins (approved):
Nice, this is a better organization.

PR #3749, opened 2024-03-27
Fancy links for GH issues representing external reviews

jyasskin:
For all open issues, this adds the amount of time they've been open: 
image

For open position issues, colors the links according to how long they've been open, with thresholds at 2 weeks and 1 month. https://www.chromium.org/blink/launching-features/#new-feature-prepare-to-ship says to give reviewers at least a month to review, so ... @chrishtr @miketaylr what colors do you want for what ages?
image

For resolved position issues, colors the links according to the position:
image

### DIFF
--- a/client-src/elements/chromedash-link.js
+++ b/client-src/elements/chromedash-link.js
@@ -133,9 +133,7 @@ function enhanceGithubIssueLink(featureLink, text) {
       if (externalReviewer) {
         // If this is an issue asking for external review, having it filed too recently is a warning
         // sign, which we'll indicate using the tag's color.
-        if (age < 2 * week) {
-          stateVariant = 'danger';
-        } else if (age < 4 * week) {
+        if (age < 4 * week) {
           stateVariant = 'warning';
         } else {
           // Still only neutral if the reviewer hasn't given a position yet.
--- a/client-src/elements/chromedash-link.js
+++ b/client-src/elements/chromedash-link.js
@@ -1,6 +1,5 @@
 // @ts-check
 
-import '@github/relative-time-element';
 import {css, html, LitElement} from 'lit';
 import {ifDefined} from 'lit/directives/if-defined.js';
 import {SHARED_STYLES} from '../css/shared-css';
@@ -128,7 +127,7 @@ function enhanceGithubIssueLink(featureLink, text) {
   if (stateVariant === undefined) {
     if (state === 'open') {
       const age = Date.now() - createdAt.getTime();
-      stateDescription = html`Opened <relative-time .date=${createdAt} format="relative" precision="day" threshold="P100Y">on ${_dateTimeFormat.format(createdAt)}</relative-time>`;
+      stateDescription = html`Opened <sl-relative-time date=${createdAt.toISOString()}>on ${_dateTimeFormat.format(createdAt)}</sl-relative-time>`;
       const week = 7 * 24 * 60 * 60 * 1000;
       stateVariant = 'success';
       if (externalReviewer) {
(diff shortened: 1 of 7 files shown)


client-src/elements/chromedash-link.js
```diff
@@ -109,6 +111,45 @@ function enhanceGithubIssueLink(featureLink, text) {
...
+        ({description: stateDescription, variant: stateVariant} = labelInfo);
+        break;
+      }
+    }
+  }
+
+  if (stateVariant === undefined) {
+    if (state === 'open') {
+      const age = Date.now() - createdAt.getTime();
+      stateDescription = html`Opened <relative-time .date=${createdAt} format="relative" precision="day" threshold="P100Y">on ${_dateTimeFormat.format(createdAt)}</relative-time>`;
```
jrobbins, line 131: We already have `<sl-relative-time>` as used in utils.js.  Could you use that? It might even just be wrapping the same underlying library.

jrobbins (approved):
LGTM after considering one possible change.

client-src/elements/chromedash-link.js
jyasskin: `<sl-relative-date>` is missing the `precision` argument that I'm using to turn "Opened 2 Hours Ago" into "Opened Today" (once https://github.com/github/relative-time-element/pull/280 goes in), but that's not a big deal if you'd rather avoid the dependency. I'll see if there are any other problems once my development machine finishes upgrading.

jyasskin: I have to undo the style from https://github.com/GoogleChrome/chromium-dashboard/blob/cffa57147e486fc1e44e972740306c461e884342/client-src/css/shared-css.js#L191-L193
but then it seems to work:
image

Here's a patch without `<relative-time>`.

miketaylr:
> so ... @chrishtr @miketaylr what colors do you want for what ages?

Hmm, this is trickier to consider than I would hope for. I do see the ~one month guidance for TAG, but not for other vendors, FWIW. I think the colors you have do make sense, once we document them. Red: "probably too soon for a fulsome response", orange: "proceed with caution" and gray/black: "YOLO 💀". Did I get that right?

> For resolved position issues, colors the links according to the position:

This looks great.

jyasskin:
@miketaylr That's about right. I'd s/YOLO/the API owners have to make a call/ though. ;) I'll send you a CL to copy the TAG's 1-month guideline to the other reviewers.

chrishtr:
> https://www.chromium.org/blink/launching-features/#new-feature-prepare-to-ship now asks for 1 month for vendors too.

It isn't up to us how long other vendors or the TAG take to respond though. Is this meant to represent whether the issue itself is overdue, or something whether a feature owner should consider it "timed out"?

jyasskin:
It's more about whether the external reviewer has clearly not had enough time to review it. The color here only goes to neutral gray after a month and not green because I don't think there's any automated way to decide whether the review has timed out. That's up to the human reviewers.

miketaylr:
> It isn't up to us how long other vendors or the TAG take to respond though.

Correct. But I think guidance to ask early enough to improve the odds of getting a response is good. And then API OWNERs can make a call (as we sometimes do "let's give it a week or two to bake and hope for a response", etc.)

chrishtr:
Ok makes aense. How about yellow for less than 1 months, meaning "in progress, shipping might block on it"? And after that gray.

jyasskin:
Works for me. I'll make that change Monday.

jyasskin:
Ok, everything less than a month is now yellow. I'll merge in a couple hours unless I hear concerns. Thanks!

PR #3751, opened 2024-03-27
Avoid getting permissions twice when signed out.

jrobbins:
When looking into chromestatus performance I noticed that the permissions API call was being done twice on some page loads.  It was only happening for signed-out users.  I think the cause was just the way that these comparisons are done.  This fix should make the site a bit faster for signed-out users, and also eliminate a potential distraction as I continue to work on performance for signed-in users.

### DIFF
--- a/client-src/elements/chromedash-header.js
+++ b/client-src/elements/chromedash-header.js
@@ -151,7 +151,7 @@ export class ChromedashHeader extends LitElement {
     this.googleSignInClientId = '',
     this.devMode = '';
     this.currentPage = '';
-    this.user = {};
+    this.user = undefined;
     this.loading = false;
   }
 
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-header.js
```diff
@@ -164,10 +164,10 @@ export class ChromedashHeader extends LitElement {
     }
 
     // user is passed in from chromedash-app
-    if (this.user && this.user.email) return;
+    if (this.user !== undefined) return;
 
     // user is passed in from chromedash-app, but the user is not logged in
-    if (!this.user) {
+    if (this.user === null) {
```
dlaliberte, line 170: Do you really mean "=== null".  Where could the null value be coming from?  Maybe add a comment about that.

dlaliberte (approved):

jrobbins:
The playwright timeouts were due to this change.
Figuring out how to do this right is going to be more complex that I thought.

jrobbins:
I'll come back to this after we phase out the old feature list page.

PR #3757, opened 2024-03-28
Intent detection for gates in Chromestatus URLs

DanielRyanSmith:
This PR allows the detection of gate IDs in intent threads to ensure we correlate the correct approvals and review requests with the correct gate and stage.

- The intent detection logic will still function as it previously did if no gate ID is detected.
- A check will occur to ensure the detected gate matches the intent type (e.g. an "Intent to Extend Experiment" intent thread should only update an origin trial extension gate).

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -77,19 +77,19 @@ def detect_field(subject):
 
   return None
 
-
-CHROMESTATUS_LINK_GENERATED_RE = re.compile(
+CHROMESTATUS_LINK_GENERATED_PATTERN = (
     r'Chrome( Platform)? ?Status(.com)?[ \w]*:?\s+'
     r'[> ]*https?://(www\.)?chromestatus\.com/'
-    r'(feature|guide/edit)/(?P<id>\d+)', re.I)
+    r'(feature|guide/edit)/(?P<id>\d+)')
+
+CHROMESTATUS_LINK_GENERATED_RE = re.compile(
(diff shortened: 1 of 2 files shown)


internals/detect_intent.py
```diff
@@ -86,6 +86,10 @@ def detect_field(subject):
     r'entry on the feature dashboard.*\s+'
     r'[> ]*https?://(www\.)?chromestatus\.com/'
     r'(feature|guide/edit)/(?P<id>\d+)', re.I)
+CHROMESTATUS_LINK_GENERATED_GATE_RE = re.compile(
```
jrobbins, line 89: You could reduce duplication here by defining a shared pattern string like
```
CHROMESTATUS_LINK_GENERATED_PATTERN = (
    r'Chrome( Platform)? ?Status(.com)?[ \w]*:?\s+'
    r'[> ]*https?://(www\.)?chromestatus\.com/'
    r'(feature|guide/edit)/(?P<id>\d+)')
```
And then use it for CHROMESTATUS_LINK_GENERATED_RE and CHROMESTATUS_LINK_GENERATED_GATE_RE with and appended part for the gate parameter.  You would need to rename `(?P<id>\d+)` for the gate ID to avoid name collision, like `(?P<gate_id>\d+)`.

jrobbins (approved):
LGTM after considering one suggestion.

internals/detect_intent.py
DanielRyanSmith: Great suggestion - updated

PR #3758, opened 2024-03-28
Add "OT extended" notification template

DanielRyanSmith:
This PR adds a new notification template that will be sent to the OT team when an origin trial is successfully extended. The template is not yet used, and will be implemented when the new automatic origin trial extension work has landed.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -722,6 +722,44 @@ class OriginTrialExtensionRequestHandler(basehandlers.FlaskHandler):
       'html': email_body,
     }
 
+
+class OriginTrialExtendedHandler(basehandlers.FlaskHandler):
+  """Notify about an origin trial extension being completed."""
+
+  IS_INTERNAL_HANDLER = True
+
+  def process_post_data(self, **kwargs):
+    extension_stage = self.get_param('stage')
+    ot_stage = self.get_param('ot_stage')
+    logging.info('Starting to notify about successful origin triale extension.')
+    send_emails([self.build_email(self, extension_stage, ot_stage)])
+
+  def build_email(self, extension_stage, ot_stage):
+    email_body = f"""
+<p>
+  A recent extension request was finalized and processed.
+  <br>
+  Requested by: {extension_stage["ot_owner_email"]}
+  <br>
+  Feature name: {ot_stage["ot_display_name"]}
+  <br>
+  Intent to Extend Experiment URL: {extension_stage["intent_thread_url"]}
+  <br>
+  New end milestone: {extension_stage["desktop_last"]}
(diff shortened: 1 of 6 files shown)


internals/notifier.py
```diff
@@ -722,6 +722,44 @@ def make_extension_request_email(self, extension_stage, ot_stage):
...
+
+class OriginTrialExtendedHandler(basehandlers.FlaskHandler):
+  """Notify about an origin trial extension being completed."""
+
+  IS_INTERNAL_HANDLER = True
+
+  def process_post_data(self, **kwargs):
+    extension_stage = self.get_param('stage')
+    ot_stage = self.get_param('ot_stage')
+    logging.info('Starting to notify about successful origin triale extension.')
```
jrobbins, line 734: Typo triale -> trial.

internals/notifier.py
```diff
@@ -722,6 +722,44 @@ def make_extension_request_email(self, extension_stage, ot_stage):
...
+    send_emails([self.build_email(self, extension_stage, ot_stage)])
+
+  def build_email(self, extension_stage, ot_stage):
+    email_body = f"""
+<p>
+  A recent extension request was finalized and processed.
+  <br>
+  Requested by: {extension_stage["ot_owner_email"]}
```
jrobbins, line 742: I think that you should use a Jinja2 template for this instead of a python f-string.  The reason is that Jinja2 will automatically do context-sensitive escaping.  We expect the emails to be safe but something might slip through. And, the display name is user controlled with no restrictions.  So, it would be better for all html-generating code to have a layer of defense against malicious input.

internals/notifier_helpers.py
```diff
@@ -209,3 +209,12 @@ def send_ot_notification(stage: Stage):
   else:
     cloud_tasks_helpers.enqueue_task(
         '/tasks/email-ot-creation-request',params)
+
+
+def send_trial_extended_notification(stage: Stage):
+  """Notify about """
```
jrobbins, line 215: Incomplete doctoring.

jrobbins (changes requested):
This generally looks good but I have one big request and a couple of small things.

internals/notifier_helpers.py
DanielRyanSmith: Oof, sorry about the careless errors here.

internals/notifier.py
DanielRyanSmith: You're definitely right. I made the temporary emails used for trial extension/creation notifications and followed that format, which was always meant to be removed. I've updated the implementation to use a template.

jrobbins (approved):

PR #3763, opened 2024-04-02
Fetch paired_user separately.

jrobbins:
This speeds up API calls to get permissions by about 200ms.

ChromeStatus has a UI detail that helps @google.com and @chromium.org users notice when they are signed into the wrong account.  If they are viewing a feature that has an owner with one account, but they are signed into the other account (called the "paired_user") then they see a warning.  This was implemented inefficiently in that the paired user is always looked up even when it is not needed.

In this PR:
* Change the permissions_api to only return permissions for the signed in user, not both users.  Add a `?returnPairedUser` mode that only returns the paired user.
* Change the JS code to explicitly fetch the paired user, but only on the feature detail page.
* Also, change permissions.py to use cached user objects in one function that was previously doing a direct NDB query.

When reloading the roadmap page or "my features page", the app currently spends about 400ms in the permissions call.   With this PR, those pages spend only about 180ms in that call. 

When reloading the feature detail page, the same savings are seen.  That is because even though the second call is made to get permissions for the paired user, and that call also takes about 180ms, it is done asynchronously.  So, the main page content loads sooner and the warning is the only part that is slightly delayed while waiting for the paired user.

The app does not repeatedly request permissions as the user is navigating around the site.  However, API Owners do frequently reload the page to get an updated list when they are all entering vote during their meeting.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -153,7 +153,8 @@ class ChromedashApp extends LitElement {
   fetchPairedUser() {
     if (this.paired_user !== undefined) {
       if (this.pageComponent) {
-        this.pageComponent.paired_user = pu;
+        this.pageComponent.paired_user = this.paired_user;
+        return;
       }
     }
     window.csClient.getPermissions(true).then((pu) => {
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -97,6 +97,7 @@ class ChromedashApp extends LitElement {
   static get properties() {
     return {
       user: {type: Object},
+      paired_user: {type: Object},
       loading: {type: Boolean},
       appTitle: {type: String},
       googleSignInClientId: {type: String},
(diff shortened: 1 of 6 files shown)


client-src/elements/chromedash-app.js
```diff
@@ -148,6 +150,22 @@ class ChromedashApp extends LitElement {
     });
   }
 
+  fetchPairedUser() {
+    if (this.paired_user !== undefined) {
+      if (this.pageComponent) {
+        this.pageComponent.paired_user = pu;
```
DanielRyanSmith, line 156: is `pu` defined in this context? Should this be `this.paired_user` instead?

DanielRyanSmith (approved):
Looks good, with 1 issue to address.

client-src/elements/chromedash-app.js
jrobbins: Whoops.  Fixed.

PR #3768, opened 2024-04-03
Fetch all review gates together (w/ new index).

jrobbins:
This should speed up the reviewer's dashboard loading time by reducing the variation in latency on the high end.

Previously, the way that the pending reviews box on the "my features" page was fetching data from the backend was to:
1. Run a feature query to get the features that should show up, and then after that,
2. Make N parallel API calls to get all the gates for each feature returned in step 1.
The performance was about 1s to 2s for the feature search, plus another 200ms to 500ms waiting for the majority of gate queries to return.  But, pretty often there was one gate query that took a whole second longer, maybe due to the request starting a new GAE instance.  It also led to a noticeable visual pop when it finally rendered.

In the new approach, the queries are both started immediately in parallel:
1. The features query performance is unchanged: it still takes 1s to 2s.
2. The /gates/pending query starts immediately and  takes 1.5s to 2.5s.
So, the overall average performance is about the same.  It's hard to tell because there is a lot of variation between trials.  However, doing the single larger request seems to have more consistent performance than the old approach, and much less of a visual pop.

In this PR:
* Define a new API endpoint to get pending gates.   It returns only the gates for the stages that have an active gate, which is about half of the total number of gates returned in the previous approach.  It requires a new index, I suppose that is so that the stage_id projection can be read from the index.
* Change the feature table to make the feature query request and the gates request in parallel.
* Add the new API endpoint to cs-client.js
* Change the logic for counting weekdays between the gate request and today.  Counting days one at a time in a loop was slow for a few long-abandoned gates.  Instead, now we keep the precise count up to 30 days and use a fast approximation for longer time ranges.

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -163,6 +163,52 @@ class GatesAPI(basehandlers.APIHandler):
         self.abort(400, 'Assignee is not a reviewer')
 
 
+class PendingGatesAPI(basehandlers.APIHandler):
+
+  def get_stage_ids_of_gates_pending_my_approval(self) -> list[int]:
+    """Return a list of stage_id needing approval by current user."""
+    user = self.get_current_user()
+    if not user:
+      return []
+
+    approvable_gate_types = approval_defs.fields_approvable_by(user)
+    if not approvable_gate_types:
+      logging.info('User has no approvable_gate_types')
+      return []
+    query = Gate.query(
+        Gate.state.IN(Gate.PENDING_STATES),
+        Gate.gate_type.IN(approvable_gate_types))
+    future_stage_ids = query.fetch_async(projection=['stage_id'])
+    return future_stage_ids
+
+  def do_get(self, **kwargs):
+    """Return a list of gates on features that have some active gates."""
+    # 1. Get the feature IDs of any active gates.  Also, prefetch approvers.
+    future_projections = self.get_stage_ids_of_gates_pending_my_approval()
+    prefetched_approvers = {
+        gate_type: approval_defs.get_approvers(gate_type)
+        for gate_type in approval_defs.APPROVAL_FIELDS_BY_ID}
+    if type(future_projections) == list:
+      stage_ids = set(future_projections)
+    else:
+      projections = future_projections.get_result()
+      stage_ids = set(proj.stage_id for proj in projections)
+    if not stage_ids:
+      return {'gates': []}
+
+    # 2. Fetch all the gates on those stages.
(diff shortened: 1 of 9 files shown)


api/reviews_api.py
```diff
@@ -163,6 +163,52 @@ def validate_assignees(self, assignees, fe, gate):
...
+      return []
+
+    approvable_gate_types = approval_defs.fields_approvable_by(user)
+    if not approvable_gate_types:
+      logging.info('User has no approvable_gate_types')
+      return []
+    query = Gate.query(
+        Gate.state.IN(Gate.PENDING_STATES),
+        Gate.gate_type.IN(approvable_gate_types))
+    future_stage_ids = query.fetch_async(projection=['stage_id'])
```
DanielRyanSmith, line 181: Could we call `get_result()` here at the end of the method in order to keep the return type consistently list[int], or does that eliminate the purpose of doing this asynchronously?

api/reviews_api.py
```diff
@@ -163,6 +163,52 @@ def validate_assignees(self, assignees, fe, gate):
         self.abort(400, 'Assignee is not a reviewer')
 
 
+class PendingGatesAPI(basehandlers.APIHandler):
+
+  def get_stage_ids_of_gates_pending_my_approval(self) -> list[int]:
```
DanielRyanSmith, line 168: This return type might not be correct if returning the result of `fetch_async()`.

api/reviews_api.py
```diff
@@ -163,6 +163,52 @@ def validate_assignees(self, assignees, fe, gate):
...
+    gates: list[Gate] = Gate.query(Gate.stage_id.IN(stage_ids)).fetch()
+
+    # 3. Convert to dicts and add possible assignees.
+    dicts = [converters.gate_value_to_json_dict(g) for g in gates]
+    for g in dicts:
+      g['possible_assignee_emails'] = prefetched_approvers.get(g['gate_type'], [])
+
+    return {
+        'gates': dicts,
+        }
```
DanielRyanSmith, line 209: Optional nit
```suggestion
    return {'gates': dicts}
```

DanielRyanSmith (approved):
Some small clarifications and nits, but otherwise LGTM

api/reviews_api.py
jrobbins: I updated several return types to be `list[int] | Future`.

api/reviews_api.py
jrobbins: Right, that would lose the benefit.

api/reviews_api.py
jrobbins: Done.

PR #3771, opened 2024-04-03
Send verify accuracy emails to enterprise users

DanielRyanSmith:
This change updates the logic for sending accuracy verification emails to ensure that owners of enterprise features also receive reminder emails.


jrobbins (approved):
This looks good.   Thanks for the quick turn-around on this.  

I was also thinking of this less-invasive approach of just checking the rollout fields rather than attempting a schema change and migration script at this time.

PR #3772, opened 2024-04-03
Reduce the number of parallel ndb queries (w/ new index).

jrobbins:
This speeds up the query for features pending review by 50-250ms on staging.

Google Cloud NDB requires an index to be defined for every combination of conditions that are used in each query sent from the app.  That makes it impossible for apps that have user-defined queries to translate the user query into a single complex NDB query because the needed index would not exist.  So, our approach to running user queries is to translate each term of the user query into a single-term NDB query, run those multiple NDB queries in parallel, and then do set union and intersection in python.

Each of these parallel queries is done with `fetch_async()` so that one does not need to wait for the others.  However, once the results are returned from NDB to python, they must be decoded into python objects.  This step is CPU intensive and does not run in parallel because python is not truly multi-threaded.  Decoding each result `feature_id` takes something like 0.1ms, which adds up if we do 5 NDB queries in parallel with each one having 3000 results.

In this PR:
* We still exclude enterprise queries by default, but now we turn off that default by using an explicit query-string parameter `showEnterprise` rather than adding a tautological term to the user query (thus avoiding one parallel NDB query).
* For the two most common permission terms, we define a built-in query for each.  These built-in queries use 2 or 3 combined conditions in a single NDB query (thus avoiding one or two parallel NDB queries).

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -35,6 +35,25 @@ MAX_TERMS = 6
 DEFAULT_RESULTS_PER_PAGE = 100
 
 
+def process_exclude_deleted_unlisted_query() -> Future:
+  """Return a future for all features, minus deleted and unlisted."""
+  query = FeatureEntry.query(
+      FeatureEntry.deleted == False,
+      FeatureEntry.unlisted == False)
+  future_feature_ids = query.fetch_async(keys_only=True)
+  return future_feature_ids
+
+
+def process_exclude_deleted_unlisted_enterprise_query() -> Future:
+  """Return a future for all features, minus deleted, unlisted, enterprise."""
+  query = FeatureEntry.query(
+      FeatureEntry.deleted == False,
+      FeatureEntry.unlisted == False,
+      FeatureEntry.feature_type <= core_enums.FEATURE_TYPE_DEPRECATION_ID)
+  future_feature_ids = query.fetch_async(keys_only=True)
+  return future_feature_ids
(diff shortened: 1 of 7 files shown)


internals/search.py
```diff
@@ -35,6 +35,25 @@
...
+  future_feature_ids = query.fetch_async(keys_only=True)
+  return future_feature_ids
+
+
+def process_exclude_deleted_unlisted_enterprise_query() -> Future:
+  """Return a future for all features, minus deleted, unlisted, enterprise."""
+  query = FeatureEntry.query(
+      FeatureEntry.deleted == False,
+      FeatureEntry.unlisted == False,
+      FeatureEntry.feature_type <= core_enums.FEATURE_TYPE_DEPRECATION_ID)
```
DanielRyanSmith, line 52: I don't know the answer to this, but could this be more efficient with the following change?

```suggestion
      FeatureEntry.feature_type != core_enums.FEATURE_TYPE_ENTERPRISE_ID)
```

DanielRyanSmith (approved):
LGTM besides 1 optional comment

internals/search.py
jrobbins: I don't think it would help.  My understanding is that NDB implements `!=` as `OR(query with <, query with >)`.  The second part of that `OR` should give an empty result, so it would be about the same.

PR #3775, opened 2024-04-04
Generalize /newfeatures page via QS params.

jrobbins:
This is a step toward being ready to split the /myfeatures page into separate pages that are basically just /newfeatures with a lot of query-string parameters.  

The reason for doing this is because with only a single table on the page, it can load faster than the /myfeatures page.  The feature owners never used any of the other boxes anyway.  With this change, this new functionality will be available at a specific URL that has no navigation link to it yet.  After I am sure that all four boxes can be replaced, I will replace the main menu navigation link to /myfeatures with four links to /newfeatures.

In this PR:
* Allow title, showQuery, columns, showEnterprise, and sort to be passed into the /newfeatures page via qs-params.
* Add a `refetch()` method so that if any editing it done via a gate column, the table content will reload (same as /myfeatures)
* Simplify the rendering functions since `renderBox()` was only called once.
* Add a property for the selected gate so that it can be highlighted after the user clicks on it, and display a message to explain why some users see more pending reviews than other users.

### DIFF
--- a/client-src/elements/chromedash-all-features-page.js
+++ b/client-src/elements/chromedash-all-features-page.js
@@ -15,7 +15,12 @@ export class ChromedashAllFeaturesPage extends LitElement {
   static get properties() {
     return {
       rawQuery: {type: Object},
+      title: {type: String},
+      showQuery: {type: Boolean},
       query: {type: String},
+      columns: {type: String},
+      showEnterprise: {type: Boolean},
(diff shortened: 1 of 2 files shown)


jrobbins:
Screenshot of the /newfeatures page looking and working exactly one box from the /myfeatures page.
image

KyleJu (approved):

PR #3781, opened 2024-04-05
Split /myfeatures into three pages.

jrobbins:
This pretty much finishes the KR to split the existing /myfeatures page into smaller, faster pages.

The problem with the /myfeatures page was that it had 4 boxes on it that made the page slow to load.   Also, we showed only 25 items in each box because we didn't want any one box to slow the page too much or push other content too far down the page.  That did not match user preferences.  The API owners are the heaviest users of that page they preferred it to load faster and show a longer list of pending reviews before requiring pagination.

The existing 4 boxes are becoming 3 pages because I am dropping the ability to see recently reviewed features.  In observing the API Owners, I never saw them use that box. 

I had planned to implement the split pages by using the /newfeatures URL with a bunch of query-string parameters.  However, that would make messy URLs and tempt users to do something funny with the title parameter.  Instead, I added SPA routes for short URLs and set the relevant properties via JS code.

In this PR:
* Adjust the new features page by adding a bit of padding and removing the ability to set the content title.
* In chromedash-app, reduce redundant code by refactoring some frequently occurring lines into `setupNewPage()`.
* In chromedash-app, add three new routes for `/myfeatures/review`, `/myfeatures/starred`, and `/myfeatures/editable`, and make each one set properties that configure the page to show just the relevant features.
* In main.py, add same SPA routes so that the user can reload the page there.
* In chromedash-drawer, reduce the padding a bit and replace the old "My features" item with a new "My features" section that has up to three items.
* In chromedash-feature-row, remove the `?context=myfeatures` query string parameter because it was left over from a previous approach that was no longer working anyway.  Now the back-arrow icon is handled by the app's `contextLink` property.  Reloading any page that had navigation context will lose that context and default to `/features` which is reasonable behavior.
* I don't remove the old `/myfeatures` routes or chromedash-myfeatures-page element yet, instead I mark them as deprecated.


jrobbins:
Screenshot showing new main menu items and highlighting for the currently active page:
image

DanielRyanSmith (approved):
Looking very smooth!

PR #3790, opened 2024-04-10
Validate Chromium inputs for OT creation (UI)

DanielRyanSmith:
This PR adds a number of checks when the user attempts to submit an OT Creation request that validates the input is in the correct format in accordance with the Chromium code. This PR does not include these checks server-side, which will come in a later PR.

Checks each of these params are set up in Chromium:
- Webfeature UseCounter
- Chromium trial name
- Third-party support setting
- Critical trial grace period setting

This is checked by obtaining the Chromium file contents and ensuring the files contain the given information as expected.

Some of the definitions of the form fields have been updated to include links to the files that require the data be added.

Note: These checks are handled on a submission attempt, so they are not handled through the semantic checks process, which is processed after each input change.

Screenshot 2024-04-10 at 11 59 16 AM

https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/0a7ee267-79e3-4ae4-bfed-540030fa1182

### DIFF
--- a/client-src/elements/chromedash-ot-creation-page.js
+++ b/client-src/elements/chromedash-ot-creation-page.js
@@ -274,13 +274,13 @@ export class ChromedashOTCreationPage extends LitElement {
       field.checkMessage = nothing;
       return false;
     }
-    const webfeatureUseCounterName = this.fieldValues.find(
-      field => field.name === 'ot_webfeature_use_counter').value;
+    const chromiumTrialName = this.fieldValues.find(
+      field => field.name === 'ot_chromium_trial_name').value;
     if (!this.gracePeriodFile) {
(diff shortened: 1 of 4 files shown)


jrobbins:
Is this PR ready for review?

DanielRyanSmith:
@jrobbins It is now, yes 🙂

jrobbins (approved):
LGTM.

How long does it take to download the files after the user hits "submit"?

DanielRyanSmith:
> How long does it take to download the files after the user hits "submit"?

The check process has felt very fast testing locally (<1 second). In the short video above, the file contents are requested and checks are completed each time the submit button is pressed, so that should be a relatively accurate reflection.

I'll need to make a change to the logic based on today's discussion of critical trials before merging.

PR #3791, opened 2024-04-11
Show feature links in the Consensus section

jyasskin:
This partially implements [a change to replace manual vendor-views status fields with standards-positions data](https://docs.google.com/document/d/1ZwoDdOaQs3ZBqhxztOmCi6e1QNQLKrIyRIOyaSf7-R8/edit?usp=sharing&resourcekey=0-gKVmDBqplbrSBeHQzuhv8w), enough to use the Feature Link data in the Consensus & Standardization section on top of the feature pages.

The formatting changes are the result of 'organize imports' and letting Ruff format "changed lines", where a "line" rounds up to the whole statement.


static/css/base.css
```diff
@@ -455,7 +455,7 @@
 }
 
 @supports not (scrollbar-gutter: stable) {
-  .sl-scroll-lock {
+  .sl-scroll-lock body {
```
jrobbins, line 458: What is this change for?

jrobbins (approved):
LGTM after considering a couple of questions.

static/css/base.css
jyasskin: It's the same situation as https://github.com/GoogleChrome/chromium-dashboard/pull/3639 -- starting the app changes this file -- but I didn't make a separate PR for it this time.

jrobbins:
Also, can you add comments with screenshots for the record?

internals/feature_links.py
```diff
@@ -116,6 +139,71 @@ def _remove_link(link: Link, fe: FeatureEntry) -> None:
...
+    if label.lower().startswith(position_prefix):
+      return label[len(position_prefix) :]
+  return None
+
+
+def _put_if_changed(model: ndb.Model, field: str, new_val) -> None:
+  old_val = getattr(model, field)
+  if old_val != new_val:
+    setattr(model, field, new_val)
+    logger.info(
```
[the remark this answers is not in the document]
jyasskin: This makes it a bit easier to tell where the message is coming from, and it's recommended by https://docs.python.org/3/library/logging.html, but I can undo it.

jyasskin (commented):
I've added screenshots to the OP.

client-src/elements/chromedash-feature-page_test.js
```diff
@@ -91,8 +91,8 @@ describe('chromedash-feature-page', () => {
         owners: ['fake chrome owner one', 'fake chrome owner two'],
         status: {text: 'fake chrome status text'},
       },
-      ff: {view: {text: 'fake ff view text'}},
-      safari: {view: {text: 'fake safari view text'}},
+      ff: {view: {val: 1, text: 'fake ff view text', url: 'fake ff url'}},
```
jcscottiii, line 94: The tests now render the case where val is truthy and will show the links. But now the tests don't assert the behavior of when the val is falsy.

client-src/elements/chromedash-feature-page_test.js
```diff
@@ -237,6 +237,16 @@ describe('chromedash-feature-page', () => {
...
 
+    const consensusSection = component.shadowRoot.querySelector('section#consensus');
+    assert.exists(consensusSection);
+    // spec link is clickable
+    assert.include(consensusSection.innerHTML, '<chromedash-vendor-views ');
+    assert.include(consensusSection.innerHTML, 'fake ff view text');
+    assert.include(consensusSection.innerHTML, 'href="fake ff url"');
+    assert.include(consensusSection.innerHTML, 'fake safari view text');
+    assert.include(consensusSection.innerHTML, 'href="fake safari url"');
+    assert.include(consensusSection.innerHTML, 'fake webdev view text');
```
jcscottiii, line 248: The web dev section always shows. (Or at least there is no conditional present like the firefox and Safari cases)

Add a separate comment saying WebDev text is present. This will be helpful since it is not a link

client-src/elements/chromedash-feature-page_test.js
jcscottiii, line 1: You could add a helper method to assert whether something is clickable (I know there are existing patterns that check innerHTML so we could follow up later in another PR)

But this helper could be focused on the Vendor links specifically so we could add it here:


```js
function assertClickableVendorLink(parentEl, href, text) {
  // Select chromedash-vendor-views element
  const vendorViewsEl = parentEl.querySelector(`chromedash-vendor-views[href="${href}"]`);
  assert.exists(vendorViewsEl);
  // Verify that the link's text content matches the expected display text
  assert.equal(vendorViewsEl.textContent, text);

  // Select the chromedash-link
  const link = vendorViewsEl.shadowRoot.querySelector('chromedash-link');
  assert.exists(link);

  // Check if the link's 'href' attribute matches the expected URL
  assert.equal(link.href, href);
}
```


Then below, you could do

```js
    // spec link is clickable
    assertClickableVendorLink(consensusSection, 'fake ff url', 'fake ff view text');
    assertClickableVendorLink(consensusSection, 'fake safari url', 'fake safari view text');
   // Webdev item is present
   assert.include(consensusSection.innerHTML, 'fake webdev view text'); // This is left as-is
```

jcscottiii (commented):
A few nits and one suggestion/future suggestion

jcscottiii (approved):

PR #3793, opened 2024-04-14
Implement a search input field with autocomplete for field names

jrobbins:
This implements typeahead functionality for the feature search box on the /newfeatures page.
It is commonly called "autocomplete", but since that term is also used for the browser's own autocomplete, I use the alternative accepted term "typeahead".

The basic concept is that the typeahead component has a `vocabulary` list of known words that can be used in its input field.  It looks at a `prefix` consisting of the characters that the user has already type, narrows down the vocabulary to a list of `candidates` and offers those in a menu.  The user can use Up/Down/Enter or click on a menu item to select it.  Selecting an item inserts it into the input field, overwriting the existing prefix and adding more characters, and leaving the text insertion caret at the end of the newly inserted text.

This is built using Shoelace components.  Class `chromedash-typeahead` uses an `<sl-dropdown>` component that manages a `<sl-input>` component and a `<sl-menu>` component that consists of menu items.  The problem with using these components is that they rely on changing the browser's keyboard focus to show which menu item is selected.  That is not appropriate for a typeahead widget because the user should be able to see the text insertion cursor and continue typing characters.   So, I have subclassed and modified the dropdown and menu item components. 

Our search terms consist of `field=values`, but at this point the autocomplete only helps the user specify the `field=` part.  Later PRs will improve the component to also support the `values` part of each term.


jrobbins:
This PR is a draft until I add unit tests.

jrobbins:
Screenshot after the user has typed "dev":
image

Screenshot of user in the process of changing desktop shipping milestone to devtrial start:
image

dlaliberte (approved):
Looks like a lot of work went into this.  Nice job.

client-src/elements/chromedash-typeahead.js
```diff
@@ -0,0 +1,411 @@
...
+import {SHARED_STYLES} from '../css/shared-css.js';
+import SlDropdown from '@shoelace-style/shoelace/dist/components/dropdown/dropdown.js';
+
+/* This file consists of 3 classes that together implement a "typeahead"
+   text field with autocomplete:
+
+   1. Chromedash-typeahead represents the overall UI widget, accepts a
+   `vocabulary` list of words, exposes a `value` string, and emits a
+   `sl-change` event when the user hits enter to submit the value.
+   Internally, it is responible for narrowing the vacabulary down to a list
```
KyleJu, line 13:
```suggestion
   Internally, it is responsible for narrowing the vocabulary down to a list
```

client-src/elements/chromedash-typeahead.js
```diff
@@ -0,0 +1,411 @@
...
+export class ChromedashTypeahead extends LitElement {
+  slDropdownRef = createRef();
+  slInputRef = createRef();
+
+  static get properties() {
+    return {
+      value: {type: String},
+      placeholder: {type: String},
+      candidates: {type: Array},
+      narrowCandidates: {attribute: false},
```
KyleJu, line 32: type definition for these properties?

KyleJu (approved):

KyleJu:
Impressive implementation on the UI!

client-src/elements/chromedash-typeahead.js
jrobbins: These are not types in terms of compiler type checking, but they are a configuration of how lit converts from attribute string values into component properties.    I actually was not using `narrowCandidates` so I removed it.  Instead I use `voacabulary` for the entire list of known words, and `candidates` for the narrowed list.

PR #3794, opened 2024-04-15
Validate OT creation arguments (API)

DanielRyanSmith:
This PR updates the existing OT creation form validation structure and moves that validation to the Chromestatus API endpoint. The endpoint will validate whether the inputs are valid and return messages to be displayed on the front end if any errors were found with the input.

This fixes an issue with the current validation implementation, in which CORS errors were occurring which were caused by  the fetching of Chromium file information from a different origin. This is no longer a problem now that the these validations are happening server-side.

Additionally, some small UI changes have been added to make a clearer submission process, as the creation submission can take some seconds when processing.

https://github.com/GoogleChrome/chromium-dashboard/assets/56164590/adc55d3f-7bb5-4f4e-a991-0e2c0358e2e9

- api/origin_trials_api.py: Moves OT creation input validation to API and creates a new POST endpoint for trial creation. Trial extension endpoint moved to PATCH request.
- api/stages_api.py: Move `update_stage()` method to `EntitiesAPIHandler` class, which then can be used in OriginTrialsAPI to update the OT stage.
- client-src/elements/chromedash-ot-creation-page.js: Add new overlay to be displayed with waiting for a response after OT submission, and display any error messages that are given in the response. Also, consider all fields as "touched", to ensure all fields are submitted even if the field has been filled previously and untouched during this session.
- client-src/js-src/cs-client.js: Add new OT creation endpoint and update extension endpoint to PATCH request.
- framework/basehandlers.py: Move `update_stage()` method to shared class.
- framework/origin_trials_client.py: Unrelated semantic fix.
- main.py: New OT creation endpoint.
- package.json: Removal of json5 JavaScript dependency.
- requirements.txt: Addition of new json5 Python dependency.


framework/basehandlers.py
```diff
@@ -331,6 +334,55 @@ def update_field_value(
     new_value = self.format_field_val(field, field_type, value)
     setattr(entity, field, new_value)
 
+  def update_stage(
```
jrobbins, line 337: Why move this into base handlers?  It is called from more than one file?

jrobbins (approved):

framework/basehandlers.py
DanielRyanSmith: This is called from the OriginTrialsAPI class to update the OT stage, and also in the StagesAPI class to update general stages. I thought this might be appropriate to not duplicate the code here 🙂

PR #3801, opened 2024-04-16
Obtain Chromium file contents asynchronously

DanielRyanSmith:
Updates the 3 requests to get Chromium file contents to be handled asynchronously in parallel, rather than one at a time.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import asyncio
 import json5
+import logging
 import requests
 import validators
 from base64 import b64decode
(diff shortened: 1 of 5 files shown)


jrobbins (approved):
This looks like it would work and it should be faster.

However, I feel that you have two levels of futures here when it should be possible to have only one.  Would it be simpler to have get_chromium_file() return a future but not be an async function itself?  Likewise, I don't think that _validate_creation_args() needs to be async.

DanielRyanSmith:
> However, I feel that you have two levels of futures here when it should be possible to have only one. Would it be simpler to have get_chromium_file() return a future but not be an async function itself? Likewise, I don't think that _validate_creation_args() needs to be async.

You're right, and I've updated get_chromium_file() as you mentioned 🙂 
However, I think that _validate_creation_args() needs to be async to await the `asyncio.gather()` command. Although, I am not terribly experienced with this kind of implementation and could be wrong. 😅

PR #3815, opened 2024-04-23
Utilize "OT Extension approved" email template

DanielRyanSmith:
This change adds the functionality to send the reminder email to finalize a trial extension after that trial extension request has been approved.

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -70,12 +70,14 @@ class VotesAPI(basehandlers.APIHandler):
     self.require_permissions(user, fe, gate, new_state)
 
     # Note: We no longer write Approval entities.
+    old_gate_state = gate.state
     new_gate_state = approval_defs.set_vote(feature_id, None, new_state,
         user.email(), gate_id)
 
+    recently_approved = (old_gate_state not in (Vote.APPROVED, Vote.NA) and
+                         new_gate_state in (Vote.APPROVED, Vote.NA))
     # Notify that trial extension has been approved.
-    if (gate.gate_type == GATE_API_EXTEND_ORIGIN_TRIAL and
-        new_gate_state == Vote.APPROVED):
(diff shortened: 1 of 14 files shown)


api/reviews_api.py
```diff
@@ -72,9 +70,16 @@ def do_post(self, **kwargs) -> dict[str, str]:
     self.require_permissions(user, fe, gate, new_state)
 
     # Note: We no longer write Approval entities.
-    approval_defs.set_vote(feature_id, None, new_state,
+    new_gate_state = approval_defs.set_vote(feature_id, None, new_state,
         user.email(), gate_id)
 
+    # Notify that trial extension has been approved.
+    if (gate.gate_type == GATE_API_EXTEND_ORIGIN_TRIAL and
+        new_gate_state == Vote.APPROVED):
```
jrobbins, line 78: You should treat either `Vote.APPROVED` or `Vote.NA` as approval.

Also, it is possible that there could be more votes after the one needed for approval.  E.g., two reviewers might both have a browser tab open on the same gate and both approve it without reloading.  So, checking that `gate.state not in (Vote.APPROVED, Vote.NA)` would avoid sending a double email in that situation.

internals/detect_intent.py
```diff
@@ -353,8 +354,12 @@ def create_approvals(self, feature: FeatureEntry, gate: Gate,
     if (detect_lgtm(body) and
         is_lgtm_allowed(from_addr, feature, approval_field)):
       logging.info('found LGTM')
-      approval_defs.set_vote(feature_id, approval_field.field_id,
+      new_state = approval_defs.set_vote(feature_id, approval_field.field_id,
           Vote.APPROVED, from_addr, gate.key.integer_id())
+      if (gate.gate_type == core_enums.GATE_API_EXTEND_ORIGIN_TRIAL and
+          new_state == Vote.APPROVED):
```
jrobbins, line 360: If you make the change above, make it here too.

jrobbins (approved):

PR #3819, opened 2024-04-25
Add a report on external review status

jyasskin:
There's a circular dependency in the generated openapi javascript, which is very similar to the one in https://github.com/OpenAPITools/openapi-generator/issues/6140 but caused by using the `discriminator` field in the `LinkPreview` object. It doesn't appear to cause any concrete problems other than a warning during rollup, but I've sent https://github.com/OpenAPITools/openapi-generator/pull/18503 to fix it in their repository.

> `gen/js/chromestatus-openapi/dist/esm/models/index.js -> gen/js/chromestatus-openapi/dist/esm/models/ExternalReviewsResponse.js -> gen/js/chromestatus-openapi/dist/esm/models/LinkPreview.js -> gen/js/chromestatus-openapi/dist/esm/models/index.js`

Screenshot of the WebKit report:
Screenshot of WebKit reviewer report


api/external_reviews_api.py
```diff
@@ -0,0 +1,246 @@
...
+        FeatureEntry.has_open_ff_review == True  # noqa: E712
+      ).fetch()
+      review_link = 'ff_views_link'
+      review_pattern = GECKO_REVIEW_URL_PATTERN
+    elif review_group == 'webkit':
+      unreviewed_features = FeatureEntry.query(
+        FeatureEntry.has_open_safari_review == True  # noqa: E712
+      ).fetch()
+      review_link = 'safari_views_link'
+      review_pattern = WEBKIT_REVIEW_URL_PATTERN
```
jcscottiii, line 163: If we anticipate adding more review groups in the future, consider moving this section into a separate helper method to help prevent this from growing too long.

client-src/elements/chromedash-report-external-reviews-page.js
```diff
@@ -0,0 +1,247 @@
...
+/** Array.sort() comparison function to order outstanding reviews by descending urgency.
+ *
+ * Reviews' features must be within the same stage.
+ *
+ * @param {OutstandingReview} a
+ * @param {OutstandingReview} b
+ * @returns {number}
+ */
+function compareOutstandingReview(a, b) {
+  console.assert(a.current_stage === b.current_stage);
```
jcscottiii, line 35: Is this console.assert needed?

client-src/elements/chromedash-report-external-reviews-page.js
```diff
@@ -0,0 +1,247 @@
...
+                              </td>
+                            </tr>
+                          `
+                        )}
+                      </table>
+                    </section>`
+                  : nothing
+              ),
+        error: e => {
+          console.error(e);
```
jcscottiii, line 234: Do you want to keep this console.error?

client-src/elements/chromedash-report-external-reviews-page.js
```diff
@@ -0,0 +1,247 @@
...
+                                      : nothing}${review.estimated_end_milestone
+                                      ? 'M' + review.estimated_end_milestone
+                                      : nothing}`}
+                              </td>
+                            </tr>
+                          `
+                        )}
+                      </table>
+                    </section>`
+                  : nothing
```
jcscottiii, line 231: Nit: Consider moving this a separate method. Something like `renderOutstandingReviews(reviews, links)`

client-src/elements/chromedash-report-external-reviews-page.js
jyasskin: I think I'd like to either keep it or turn it into an exception, although I'll add a better error message. The function doesn't do a reasonable sort on features whose current stages are different, so if the code ever changes to ask it to, we should notice that somehow. I lean toward a console message rather than an exception, because a bad sort doesn't make the page useless, and an exception probably would.

client-src/elements/chromedash-report-external-reviews-page.js
jyasskin: Yes, since this branch results in a useless page, and people who encounter it should have a message they can paste to us in their bug report. But I should give it a better error message.

client-src/elements/chromedash-report-external-reviews-page.js
jyasskin: Sure, done.

jcscottiii (approved):
LGTM

PR #3823, opened 2024-04-26
Fix Cc email attribute

DanielRyanSmith:
This is a fix for the new "Cc" attribute for sending emails.

- Attribute is now always passed as a list of email addresses or null.
- In staging environments, Cc addresses are instead sent to a new Google group for debugging, cr-status-staging-cc-emails@google.com.
- `cc` email attribute is now added after `mail.EmailMessage` is instantiated, rather than passing it into the constructor. This is to avoid passing in an invalid null value.
- Two notifier classes now return default success values, `{'message': 'OK'}`.

### DIFF
--- a/framework/sendemail.py
+++ b/framework/sendemail.py
@@ -62,8 +62,13 @@ def handle_outbound_mail_task():
   if settings.SEND_ALL_EMAIL_TO and to != settings.REVIEW_COMMENT_MAILING_LIST:
     to_user, to_domain = to.split('@')
     to = settings.SEND_ALL_EMAIL_TO % {'user': to_user, 'domain': to_domain}
-    # Don't add CC addresses for non-prod environments.
-    cc = None
+    if cc:
+      new_cc = []
+      for cc_addr in cc:
+        cc_user, cc_domain = cc_addr.split('@')
+        new_cc.append(
+            settings.CC_ALL_EMAIL_TO % {'user': cc_user, 'domain': cc_domain})
+      cc = new_cc
 
   sender = 'Chromestatus <admin@%s.appspotmail.com>' % settings.APP_ID
   if from_user:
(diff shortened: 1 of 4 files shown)


framework/sendemail.py
```diff
@@ -62,19 +62,26 @@ def handle_outbound_mail_task():
...
         from_user, from_user, settings.APP_ID)
 
   message = mail.EmailMessage(
-      sender=sender, to=to, cc=cc, subject=subject, html=email_html)
+      sender=sender, to=to, subject=subject, html=email_html)
   if reply_to:
     message.reply_to = reply_to
   message.check_initialized()
+  if cc:
+    message.cc = cc
```
KyleJu, line 84: Just to confirm, message.cc can be both cc from the param and cc after the processing? Because it is being reassigned on line 71

framework/sendemail.py
```diff
@@ -62,19 +62,26 @@ def handle_outbound_mail_task():
...
     to_user, to_domain = to.split('@')
     to = settings.SEND_ALL_EMAIL_TO % {'user': to_user, 'domain': to_domain}
-    # Don't add CC addresses for non-prod environments.
-    cc = None
+    if cc:
+      new_cc = []
+      for cc_addr in cc:
+        cc_user, cc_domain = cc_addr.split('@')
+        new_cc.append(
+            settings.CC_ALL_EMAIL_TO % {'user': cc_user, 'domain': cc_domain})
```
KyleJu, line 70: Do we have a test for this?

framework/sendemail.py
DanielRyanSmith: The reassignment on line 71 should only occur in staging environments, so this should be the exact input given for cc on prod.

framework/sendemail.py
DanielRyanSmith: We do, but I explicitly added a comparison to the existing test to make sure this is constructed as expected 👍

KyleJu: Thanks for doing it!

KyleJu (approved):

PR #3838, opened 2024-05-02
Add XFN gates to the PSA process.

jrobbins:
The API Owners requested this change because previous attempts to handle PSA feature reviews did not go far enough.

Originally, the "Web developer facing code change" feature type was envisioned as not having any cross-functional review gates because these changes were expected to be small and low-risk.  The documentation said that only a "PSA" email to blink-dev was needed, and an I2S was optional.

However, two things changed since that initial vision:
* Many feature owners seem to choose this feature type for changes that are more impactful, regardless of the wording of the future type description.
* It is now easier to request "N/A" votes on review gates.  So, it is less of a burden to have more gates, even in cases where not all of them are needed.


DanielRyanSmith (approved):
FYI this change will not add these gates to existing features - LGTM as long as that is not the intention

KyleJu (approved):

jrobbins:
> FYI this change will not add these gates to existing features - LGTM as long as that is not the intention

Yes, I think that is OK.  I will tell the API Owners.

PR #3843, opened 2024-05-03
OTs with multiple extensions will request API Owner reviews

DanielRyanSmith:
Recently, an OT extension was requested, but API Owners were not notified. This was because the logic that detects a new thread and triggers review requests checks if votes exist for a given gate TYPE, rather than a given gate ID. That meant that the extension was not considered a "new thread" because the feature previously had OT extension votes attributed to it from a previous extension.

This change updates the logic so that votes are detected based on the gate ID, and features with multiple trial extensions will still function correctly when detecting new threads.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -160,10 +160,12 @@ def is_lgtm_allowed(from_addr, feature, approval_field):
   return allowed
 
 
-def detect_new_thread(feature_id, approval_field):
-  """Return True if there are no previous approval values for this intent."""
-  existing_votes = Vote.get_votes(
-      feature_id=feature_id, gate_type=approval_field.field_id)
+def detect_new_thread(
+    feature_id: int,
+    gate_id: int,
+  ) -> bool:
+  """Return True if there are no previous approval values for this gate."""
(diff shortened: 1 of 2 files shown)


internals/detect_intent.py
```diff
@@ -160,10 +160,9 @@ def is_lgtm_allowed(from_addr, feature, approval_field):
   return allowed
 
 
-def detect_new_thread(feature_id, approval_field):
-  """Return True if there are no previous approval values for this intent."""
-  existing_votes = Vote.get_votes(
-      feature_id=feature_id, gate_type=approval_field.field_id)
+def detect_new_thread(feature_id: int, gate_id: int) -> bool:
+  """Return True if there are no previous approval values for this gate."""
+  existing_votes = Vote.get_votes(feature_id=feature_id, gate_id=gate_id)
```
jrobbins, line 165: This could just pass `gate_id=gate_id` because the gate also determines the feature.

jrobbins (approved):
LGTM after one comment

internals/detect_intent.py
DanielRyanSmith: Good call! I've made the change. 🙂

PR #3861, opened 2024-05-08
Origin trial creation cron job

DanielRyanSmith:
This implements a new cron job that will send detect trial creation requests and send those creation requests to the to the origin trials API. This cron job will not be formally implemented until endpoint access has been landed on the origin trials API.

Changes by file:
- cron.yaml: Add the cron definition, commented out until trial creation endpoints are available on the OT API.
- framework/origin_trials_client.py: Add new functions to create an origin trial and activate an origin trial. Remove some code in the trial extension function to move to a utils function.
- framework/utils.py: Add a new utility function to obtain chromium schedule date information.
- internals/maintenance_scripts.py: Write a new cron job class that checks for any origin trial stages that are flagged for origin trial creation and process them. Additionally, activates those trials if needed, or sets an activation date if that trial should be activated in the future.
- main.py: Add the route definition for trial creation cron job, commented out until trial creation endpoints are available on the OT API.


internals/maintenance_scripts.py
```diff
@@ -500,3 +505,74 @@ def get_template_data(self, **kwargs):
     ndb.put_multi(batch)
 
     return f'{count} Feature entities updated of {len(features_by_id)} available features.'
+
+
+class CreateOriginTrials(FlaskHandler):
+
+  def handle_creation(self, stage: Stage, stage_dict: StageDict) -> str | None:
+    """Send a flagged creation request for creation."""
```
jrobbins, line 513: This description seems circular.

jrobbins (approved):

internals/maintenance_scripts.py
DanielRyanSmith: Yes, it's not very clear. Updated it to be more specific. 🙂

PR #3864, opened 2024-05-08
Make the sidebar fill the vertical space.

### LINKED ISSUE #3852: The sidebar in the external reviews page doesn't use the full window height
The sidebar in the new external reviews page looks like this with a tall window:

Screenshot 2024-05-06 3 43 26 PM

Visiting one of the subpages with enough content shows the full sidebar. Conversely the spec mentors page uses the full height:

Screenshot 2024-05-06 3 44 20 PM

CC @jyasskin .

jyasskin:
Also remove a `margin-top` from the footer that was creating a gap with the sidebar in short windows. Then the footer's shadow disappeared because the `z-index` was being ignored, which required adding a `position` property. It's possible we'll want the `#content-flex-wrapper` to include some padding so things don't run right up to the footer, but nothing I've tested looks particularly bad.

Fixes #3852.

Most of the changed lines are unindenting a block after removing one `<div>` layer: look at the diff without whitespace changes.

| | Short Window | Tall Window |
|---|---|---|
| Before | image | image |
| After | image | image |


KyleJu (approved):
LGTM. Would you mind including a screenshot after the change?

KyleJu:
Also Playwright Tests need to be regenerated

jyasskin:
@dlaliberte, this change shifts the `packages/playwright/tests/__screenshots__/chromedash-guide-feature-page_pwtest.js/add-an-origin-trial-stage/create-origin-trial-stage-dialog-*` screenshots so that the mask covers a different part of the dialog under test. Is the relevant part of the dialog still visible so that it's being tested?

dlaliberte:
> @dlaliberte, this change shifts the `packages/playwright/tests/__screenshots__/chromedash-guide-feature-page_pwtest.js/add-an-origin-trial-stage/create-origin-trial-stage-dialog-*` screenshots so that the mask covers a different part of the dialog under test. Is the relevant part of the dialog still visible so that it's being tested?

Answered in chat earlier, but here is a bit more detail.  The top part of the dialog popup (dropdown) is covered by the mask, and it would be better to arrange that this didn't happen, somehow, if not too difficult.   Perhaps by scrolling the window content, resizing the window, or creating the mask in a way that is below the dialog.  Or maybe the screenshot could just focus on the popup element.  Not a big deal in any event.

dlaliberte (approved):

PR #3869, opened 2024-05-12
Origin trial activation cron job

DanielRyanSmith:
This implements a new cron job that will detect origin trials that are scheduled for trial activation and send those activation requests to the to the origin trials API. This cron job will not be formally implemented until endpoint access has been landed on the origin trials API.


internals/maintenance_scripts.py
```diff
@@ -576,3 +576,51 @@ def get_template_data(self, **kwargs):
...
+
+  def get_template_data(self, **kwargs):
+    """Activate any origin trials that are flagged for activation."""
+    self.require_cron_header()
+
+    success_count, fail_count = 0, 0
+    today = self._get_today()
+    # Get all OT stages.
+    ot_stages: list[Stage] = Stage.query(
+        ndb.OR(
```
jrobbins, line 594: Why not use the IN operator?

jrobbins (approved):

internals/maintenance_scripts.py
DanielRyanSmith: I didn't think to use it, but I think it reads better anyway. 🙂 Interestingly, after reading some more, it seems [using the IN and OR operators are functionally identical](https://cloud.google.com/datastore/docs/concepts/queries#in). I still think it's easier to interpret the meaning using the IN operator.

PR #3884, opened 2024-05-15
Make sending emails with CC more robust.

jrobbins:
We have some tasks with the CC field being a string and some with it being an array of strings.  This handles either.


DanielRyanSmith (approved):
Thanks for taking a look at this so quickly - hopefully this fixes the problem!

PR #3887, opened 2024-05-15
Fix detect-intent failures when gate_id was not set.

jrobbins:
When the gate type is found by looking at the subject line, but there is no gate_id specified in the chromestatus link, we still find the `gate` but `gate_id` was null.  That caused the query by gate_id to end up retrieving all gates on the site, causes the GAE instance to run out of RAM, and that results in a 500 without raising an exception.

This PR uses the ID of the found gate, so the query works and the task successfully completes.


DanielRyanSmith (approved):
Thanks for investigating this @jrobbins!

Issue #3883, opened 2024-05-14
Chrome Status Entry sends a lot of emails when an edit is submitted to the feature.


issackjohn:
**Describe the bug**
When I make a draft edit to a field in the chrome status entry feature, I am spammed with 10+ emails about a single change.

**To Reproduce**
1. Go to a feature you are an owner of editor for.
2. Click on 'Edit fields' in any of the Development stages.
3. Save the edit.
4. See lots of emails sent to you.

**Expected behavior**
One email which details the change is sent.

**Additional context**
![image](https://github.com/GoogleChrome/chromium-dashboard/assets/117320405/44b2e525-ec20-4f63-a064-bd2ce0d71893)

jrobbins:
Sorry about that.   We actually had an error-retry loop that would have (slowly) sent you an infinite number of the same message, but I deleted the task from our task queue.  I am looking for the root cause now.

jrobbins:
This should be resolved now.  I deployed a couple of fixes and I don't see any new email tasks accumulating in the past few hours.

issackjohn:
Thank you!

PR #3888, opened 2024-05-15
Give more RAM to notifier to avoid OOMs.

jrobbins:
This should resolve issue #3883.

After fixing some other adjacent defects, the repeated email continued.  The email-subscribers task was running part way, then running out of RAM, then failing and starting over.  Users who were the first ones to star a feature got emails for every run while more recent starrers got no emails.

I could not find anything in that part of the code that used an excessive amount of RAM.  Instead, I found that our notifier service did not specify a GAE instance class, which caused it to default to the smallest size instance with only 128MB of RAM.  Increasing it to 1G allowed the existing email-subscribers tasks to run to completion successfully.

### DIFF
--- a/notifier.yaml
+++ b/notifier.yaml
@@ -1,5 +1,6 @@
 runtime: python311
 service: notifier
+instance_class: F4_1G
 
 handlers:
 - url: /tasks/detect-intent


jrobbins:
FYI, this is already in prod.

DanielRyanSmith (approved):

PR #3892, opened 2024-05-20
OT extension stage dictionary bug fix

DanielRyanSmith:
Previously when creating trial extension information dictionaries for feature_entry_to_json_verbose(), each stage was converted one at a time, and extension stages were added as an attribute on the corresponding origin trial stage.

The problem with the original logic was that it was not guaranteed that origin trial stages would be converted before their extension stages, so occasionally extensions could be omitted from origin trial stage dicts. This change adds another step that associates extension stages with OT stage only after all stages have been converted to dictionaries.


jrobbins (approved):
Good catch.

PR #3897, opened 2024-05-20
Gate OT creation form to require approvals beforehand

DanielRyanSmith:
This change disables the OT creation request button until all gates and LGTMs have been approved. Additionally, navigating to the form directly without approvals will trigger a redirect, and sending a creation request directly to the API without approvals will return a bad request error.


jrobbins (approved):

DanielRyanSmith:
I'm going to leave this for when I return, as there's no hurry to land this just yet

PR #3907, opened 2024-05-23
Fix checking for usecounter

jrobbins:
This fixes a simple coding error that blocked a user from creating an OT request.

The other problem is that the unit tests for this file were not running because they are async.  Apparently it is possible to test async functions, but it gets complicated because we also need the NDB context.   Instead, I rewrote the file fetching function to be synchronous.  This will make the user wait an extra few seconds, but it should not add up to too much of a delay.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import asyncio
 import json5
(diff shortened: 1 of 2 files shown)


KyleJu (approved):

jrobbins:
Please re-review because I changed the code to make testing easier.

KyleJu (approved):

PR #3908, opened 2024-05-23
Mount the whole playwright tests directory so that pwtests-ui can watch for changes without a restart.

jyasskin:
This doesn't break pwtests-update; is there anything else I should check?

### DIFF
--- a/packages/playwright/Dockerfile
+++ b/packages/playwright/Dockerfile
@@ -22,5 +22,5 @@ WORKDIR /work
 # Cache the infrequent but time consuming changes early
 COPY --chown=pwuser:pwuser package.json /work/
 RUN npm install -D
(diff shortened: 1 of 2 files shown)


jyasskin:
Just removing that final COPY fails because `wait-for-app.sh` and `playwright.config.cjs` aren't present anymore. Expanding the mount to include all of `./` mounts it over the installed `node_modules` directory, so that the `playwright` executable isn't present anymore. I've uploaded a change that narrows the copy to just the 2 files that are needed, but I'm not certain that the efficiency gain in avoiding the extra copies is worth confusing ourselves when we add a third file in the root directory that needs to be present in the container. What do you think?

jcscottiii (approved):

PR #3909, opened 2024-05-23
Write a report on pending verifications

jyasskin:
This is pretty simple, and I can add more details in future PRs. @jstenback, does this look like enough to start being useful, and what other columns would you like me to include? I could imagine that it'd be useful to list the next planned stage with its planned milestone.

Screenshot: 
image


jcscottiii (commented):
I know we have other `list` APIs in openapi without pagination, but I am thinking we should utilize pagination for this API given this API revolves around features and our existing features API uses pagination. 

We could borrow what we do in the [features API](https://github.com/GoogleChrome/chromium-dashboard/blob/729c315d08a16a90eaa9a1e33397ab1f34466df7/api/features_api.py#L92-L93) (`start` and `num` query parameters)

These query parameters could be declared in the common `components/parameters` section so future `list` APIs can use it too

PR #3916, opened 2024-05-27
(OT process reminders) handle OT owner email being null

DanielRyanSmith:
This change fixes an issue with the OT process reminders in which the script fails if the OT owner email address is not set. Additionally, a new warning will be logged if a trial has no recorded contacts.

Some tests in maintenance_scripts_test.py have been updated to no longer output warning logs when running tests.


jrobbins (approved):
Thanks for cleaning up the test output.

DanielRyanSmith:
I think this is worth merging this soon to avoid any issues with the script when it runs next week, but it's up to the team if they'd like to do so

Issue #3903, opened 2024-05-22
A new enum field should be added to Stage to represent the creation status of the origin trial


DanielRyanSmith:
A new field ot_setup_status should be created for Stage entities. It should have 4 enum states ([which will be added in core_enums.py](https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/core_enums.py))
- `OT_NOT_CREATED`, default value for OT stages that have not yet been used for OT creation.
- `OT_READY_FOR_CREATION`, when the OT stage is ready to be created but hasn’t been attempted.
- `OT_CREATION_FAILED`, when the creation request has been attempted but was not successful.
- `OT_CREATED`, when the creation process was successful.

The cron jobs and requests currently set the boolean field ot_action_requested when there is a trial creation to handle. As Jason Chase stated in the design doc, this field needs to be more robust in order for the cron job to not attempt and fail trial creation multiple times.
- Set to `OT_READY_FOR_CREATION` during the OT creation form request step ([add value change in code here](https://github.com/GoogleChrome/chromium-dashboard/blob/main/api/origin_trials_api.py#L158-L159)).
- Set `OT_CREATION_FAILED` or `OT_CREATED` based on result in creation cron job ([add value changes in code here](https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/maintenance_scripts.py#L510-L577)).

DanielRyanSmith:
The work for this has been landed and will be deployed to production over the course of the next week.

PR #3917, opened 2024-05-27
new OT Stage field `ot_setup_status`

DanielRyanSmith:
Part of #3903.

Creates a new field to denote the status of origin trial creation.


internals/data_types.py
```diff
@@ -60,6 +60,7 @@ class StageDict(TypedDict):
   ot_is_deprecation_trial: bool
   ot_owner_email: str | None
   ot_require_approvals: bool
+  ot_setup_status: NotRequired[int]
```
KyleJu, line 63: Should this be a required field and backfilled for existing Trials?

KyleJu (approved):
LGTM with a question

KyleJu:
One more thing - adding a test for changes in converters.py

internals/data_types.py
DanielRyanSmith: We _could_ backfill this here for other origin trial stages, but there shouldn't be a need as long as we're querying specifically for OT stages with `OT_READY_FOR_CREATION` values, meaning that the old OT stages will be ignored.

But if we want to for data consistency purposes, it's safe to assume that any stage with a set `origin_trial_id` has already been created, so we could backfill that value this way.

PR #3918, opened 2024-05-27
OT creation process uses new enum field

DanielRyanSmith:
Part of #3903. Uses the new `ot_setup_status` field to process OT creation requests.


internals/maintenance_scripts.py
```diff
@@ -529,6 +529,8 @@ def handle_creation(self, stage: Stage, stage_dict: StageDict) -> str | None:
                       f'{stage.key.integer_id()}')
       cloud_tasks_helpers.enqueue_task(
           '/tasks/email-ot-creation-request-failed', {'stage': stage_dict})
+      stage.ot_setup_status = OT_CREATION_FAILED
+      stage.put()
```
KyleJu, line 533: I think the retry logic above can be improved:
- Based on the discussions in the design doc, the request should be retried only if there is a transient error from the OT server. The catch logic should be more granular, based on the exception types.
- attempted_requests += 1 under the `except error` seems unnecessary. It basically means that we are only gonna retry twice. Then we might as well remove the while loop and attempted_requests, as it is not providing any values
- create_origin_trial is throwing an exception when the response code is not 200, which means request_success is false. `request_success` can be removed and nested all the logging and email sending to `except error`.
- According the error handling section in the design doc, more granular error handling is needed here, based on the response code. Lets leave a TODO here

internals/maintenance_scripts.py
```diff
@@ -564,14 +566,14 @@ def get_template_data(self, **kwargs):
 
     # OT stages that are flagged to process a trial creation.
     ot_stages: list[Stage] = Stage.query(
-        Stage.stage_type.IN(ALL_ORIGIN_TRIAL_STAGE_TYPES),
-        Stage.ot_action_requested == True).fetch()
+        Stage.ot_setup_status == OT_READY_FOR_CREATION).fetch()
```
KyleJu, line 569: This will cause an issue if prepare_for_activation (or the trial activation) fails, but the OT trial is already marked as  OT_CREATED. I think we need a more robust error handling logic when the activation fails as well.

Questions: would this affect the new stage, e.g. stage2, that is linked to the previous OT? E.g. Different stages with the same OT?

internals/maintenance_scripts.py
DanielRyanSmith: I've removed the retry logic and added a TODO here 👍

internals/maintenance_scripts.py
DanielRyanSmith: It's true that we need to still set the activation date on the stage if the activation initially fails. I added some extra enums for this situation, namely `OT_ACTIVATED` and `OT_ACTIVATION_FAILED`, which now capture the activation process as well. I also switched the enum values around to organize them correctly, which should be fine since this enum is not yet set for any stage.

My thought was that we can use `ot_activation_date` to understand where in the process the stage is, but the extra enum values seem more readable. 🙂 

> Questions: would this affect the new stage, e.g. stage2, that is linked to the previous OT? E.g. Different stages with the same OT?

I may be misunderstanding the question, but any origin trial stages should not be associated with more than 1 actual origin trial, if that's what you mean

internals/maintenance_scripts.py
```diff
@@ -514,17 +514,12 @@ def handle_creation(self, stage: Stage, stage_dict: StageDict) -> str | None:
...
-      try:
-        new_id = origin_trials_client.create_origin_trial(stage)
-        request_success = True
-      except requests.RequestException:
-        attempted_requests += 1
-    if not request_success:
+    # TODO(DanielRyanSmith): This request should have a retry process based on
+    # the error code returned from the OT server.
+    try:
+      new_id = origin_trials_client.create_origin_trial(stage)
```
KyleJu, line 521: Checking the return response of Trials on RPCstudio, I think new_id is of int type?

internals/maintenance_scripts.py
```diff
@@ -539,10 +534,14 @@ def handle_activation(self, stage: Stage, stage_dict: StageDict) -> None:
...
           '/tasks/email-ot-activated', {'stage': stage_dict})
+      stage.ot_setup_status = OT_ACTIVATED
     except requests.RequestException:
       cloud_tasks_helpers.enqueue_task(
           '/tasks/email-ot-activation-failed', {'stage': stage_dict})
-
+      stage.ot_setup_status = OT_ACTIVATION_FAILED
+      # The activation still needs to occur,
+      # so the activation date is set for current date.
+      stage.ot_activation_date = date.today()
```
KyleJu, line 544: I think at this point, it would require investigations, otherwise it will be retried repeatedly?

internals/maintenance_scripts.py
```diff
@@ -557,6 +556,7 @@ def prepare_for_activation(self, stage: Stage, stage_dict: StageDict) -> None:
       self.handle_activation(stage, stage_dict)
     else:
       stage.ot_activation_date = date
+      stage.ot_setup_status = OT_CREATED
```
KyleJu, line 559: Should we set OT_CREATED as soon as the trial creation succeeds, in the `handle_creation` method?

internals/maintenance_scripts.py
DanielRyanSmith: > My question was mostly related to the previous issue where two stages are linked to the same trial [1]. I guess it still requires us to fix it manually?

Ah yeah, that should not be an issue with the creation process, and exactly 1 trial will be associated with 1 trial stage

internals/maintenance_scripts.py
DanielRyanSmith: That might be the case - it's returned as a string from the ListTrials endpoint. I can update this to anticipate an int data type

internals/maintenance_scripts.py
DanielRyanSmith: Yes, it will require investigation, and the request should not be retried again based on the logic. The date will not necessarily be used for the activation script if the activation failed here, but I thought it a good idea to set the activation date for any trial that still needs activating 😅

internals/maintenance_scripts.py
DanielRyanSmith: I think that's fine - it will need to be determined how to handle activation, and the ot_setup_status` value could change later down the road, but I've added an additional statement to set this value for clarity

internals/maintenance_scripts.py
KyleJu: I checked ListTrials endpoint and it also returns int. The string type has "" around it. Could you double check?

framework/origin_trials_client.py
```diff
@@ -151,7 +151,7 @@ def create_origin_trial(ot_stage: Stage) -> str | None:
         f'Failed to get response from origin trials API. {response.text}')
     raise e
 
-  return response.json()['id']
+  return str(response.json()['id'])
```
KyleJu, line 154: Modify the tests for it to reflect the data type

KyleJu (approved):
LGTM with a comment on tests to address

internals/maintenance_scripts.py
DanielRyanSmith: I'm not certain, since a direct RPC returns an int value, but when fetched from the OT API through Chromestatus, it's a string.  Either way, it's safer in this case to just assume that the data type is an int and convert from there

framework/origin_trials_client.py
DanielRyanSmith: Done!

PR #3920, opened 2024-05-28
Put the spinner under the "new features page" link.

jyasskin:
So that users can click to the new features page while the old is still loading.


KyleJu (approved):
LGTM

PR #3922, opened 2024-05-29
Enable JS type checking while linting.

jyasskin:
And fix a bug it found.


jrobbins (approved):
Nice.  This is will get a bit closer to converting things to TS this summer.

PR #3938, opened 2024-06-04
Add an `accurate_as_of` search field.

jyasskin:
This is suspiciously little code, but it seems to work. Is there anything else you'd like me to add a test for? Ordering nulls first when querying with `<` seems to be the default Datastore behavior, although I can't find it documented anywhere. 

image


jrobbins (approved):
Yep, that's it.  One-to-one FeatureEntry fields are very easy to expose as searchable, sortable fields.

PR #3942, opened 2024-06-05
Allow @google and @chromium users to request OTs.

jrobbins:
We had a user report that they could not create an OT, even though they saw the "Request Trial Creation" button.  Their POST to the server resulted in a 403 and they saw an endless spinner.

The underlying cause was that we offer the "Request Trial Creation" button to any user who can edit the feature or that has a @google.com or @chromium.org account.  However, our server was only accepting requests from feature editors.

The fix is to change the server-side logic to also allow @google.com and @chromium.org accounts to request OTs.  This is needed so that a googler can help with a feature being developed by non-gogglers.  The non-gogglers can request the OT themselves, but they might not have all the needed info.

In this PR:
* Refactor and rewrite the permission checking code in origin_trials_api.py to have the new desired behavior and to be more easily testable.
* Add unit tests, logging, and type annotations.  Specifically, calls to `flask.redirect()` are now type-checked as returning `flask.Response` rather than `requests.Response`.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import flask
 import json5
 import logging
 import requests
@@ -119,6 +120,15 @@ class OriginTrialsAPI(basehandlers.EntitiesAPIHandler):
 
     return validation_errors
 
+  def check_post_permissions(self, feature_id) -> flask.Response | dict:
+    """Raise an exception or redirect if the user cannot request OT."""
+    if permissions.is_google_or_chromium_account(self.get_current_user()):
+      return {}
+
+    redirect_resp = permissions.validate_feature_edit_permission(
+        self, feature_id)
+    return redirect_resp
+
   def do_post(self, **kwargs):
     feature_id = int(kwargs['feature_id'])
     # Check that feature ID is valid.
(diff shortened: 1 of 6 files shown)


api/origin_trials_api.py
```diff
@@ -119,6 +120,15 @@ def _validate_creation_args(
 
     return validation_errors
 
+  def check_post_permissions(self, feature_id) -> flask.Response | dict:
```
KyleJu, line 123: So as long as users have a Google or Chromium account, they can edit the feature? By edit, it is the permission to edit the feature or e.g. request an extension?

api/origin_trials_api.py
```diff
@@ -136,10 +146,9 @@ def do_post(self, **kwargs):
     if ot_stage is None:
       self.abort(404, msg=f'Stage {ot_stage_id} not found')
 
-    # Check that user has permission to edit the feature
-    # associated with the origin trial.
-    redirect_resp = permissions.validate_feature_edit_permission(
-        self, feature_id)
+    # Check that user has permission to edit the feature associated
+    # with the origin trial, or has a @google or @chromium account.
+    redirect_resp = self.check_post_permissions(feature_id)
```
KyleJu, line 151: Should we catch the exception and return an Auth error?

api/origin_trials_api.py
jrobbins: > So as long as users have a Google or Chromium account, they can edit the feature?

No, a couple of years ago it was the case that any @google.com or @chromium.org account could edit any issue, but now editing is limited to just users listed on the feature entry (and site admins and site editors).  This PR does not change the permission to edit any of the fields of the feature entry other than requesting an OT.

> By edit, it is the permission to edit the feature or e.g. request an extension?

In the source code "edit" permission refers to the ability to edit any of the fields of the feature entry.  Here, I am adding back the @google.com and @chromium.org rule just for requesting an OT.

api/origin_trials_api.py
jrobbins: We don't explicitly raise an exception, instead we do self.abort(403) which calls a flask function that raises an exception that is caught by flask.  The 403 code gets passed back to the browser.

For the case where the user is signed out, I suppose that you are correct.   While a UI request can respond with a 302 redirect, an API request should respond with 401.   However, that seems beyond the scope of this PR.  I'll add a TODO to consider that as part of converting to using OpenAPI.

PR #3954, opened 2024-06-11
Allow searching for an interval using "field=start..end".

jyasskin:
This only covers integers and dates, because string ranges aren't that useful, and the integers that enums map to aren't in a logical order.

image

### DIFF
--- a/internals/search_queries.py
+++ b/internals/search_queries.py
@@ -12,26 +12,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import datetime
-from typing import Union, Callable, Optional
+import logging
+from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


internals/search_queries.py
```diff
@@ -12,26 +12,39 @@
...
+from google.cloud.ndb.tasklets import Future  # for type checking only
 
 from framework import users
-from framework import utils
 from internals import core_enums
 from internals.core_models import FeatureEntry, Stage
 from internals.review_models import Gate
 
+T = TypeVar('T')
+QueryValue: TypeAlias = bool | int | str | datetime.datetime
```
jrobbins, line 31: Thanks for adding this type, it certainly simplifies some of the other code.

jrobbins (approved):

PR #3958, opened 2024-06-11
Typescript migration: misc files/introduce build as an intermediate step

markxiong0122:
1. Introduced a new build folder for outputting compiled ts/js files 
2. migrated some misc elements to typescript
- icons
- query field
- external reviewers
- autolink


client-src/elements/datatypes.ts
```diff
@@ -1,12 +1,30 @@
...
+
+export interface Component {
+  refRegs: RegExp[];
+  replacer: (match: any) => never[] | TextRun[];
+}
+
+//external-reviewers
+export type LabelInfo = {
+  description: string;
+  variant: 'primary' | 'success' | 'neutral' | 'warning' | 'danger';
```
jyasskin, line 29: I wonder if we can import this from Shoelace instead of redefining it ourselves.

package.json
```diff
@@ -73,12 +73,14 @@
     "@shoelace-style/shoelace": "^2.15.1",
     "@types/jest": "^29.5.12",
     "@types/node": "^20.13.0",
+    "@web/dev-server-esbuild": "^1.0.2",
```
jyasskin, line 76: Delete all the dependencies we don't need anymore.

package.json
```diff
@@ -21,16 +21,16 @@
     "start-emulator-persist": "gcloud beta emulators datastore start --host-port=:15606 --use-firestore-in-datastore-mode",
     "start-emulator": "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --use-firestore-in-datastore-mode",
     "stop-emulator": "curl -X POST 'http://localhost:15606/shutdown'; pkill -f \"CloudDatastore.jar\"",
-    "build": "gulp",
-    "watch": "gulp watch",
+    "build": "tsc && gulp",
+    "watch": "concurrently \"tsc --watch\" \"gulp watch\"",
```
jrobbins, line 25: This didn't work for me, but it doesn't seem to work on the main branch either, and I don't use it, but others might want to.

package.json
```diff
@@ -21,16 +21,16 @@
...
     "stop": "npm run pwtests-shutdown --workspace=playwright; kill `pgrep gunicorn`",
     "test": "(npm run start-emulator > /dev/null 2>&1 &); sleep 6; curl --retry 4 http://localhost:15606/ --retry-connrefused; npm run do-tests; status=$?; npm run stop-emulator; exit $status",
-    "webtest": "web-test-runner \"client-src/**/*_test.js\" --node-resolve --playwright --browsers chromium firefox",
-    "webtest:watch": "web-test-runner \"client-src/**/*_test.js\" --node-resolve --playwright --watch --browsers chromium",
-    "webtest-coverage": "web-test-runner \"client-src/**/*_test.js\" --node-resolve --playwright --coverage --browsers chromium firefox",
+    "webtest": "web-test-runner \"build/**/*_test.js\" --node-resolve --playwright --browsers chromium firefox",
+    "webtest:watch": "web-test-runner \"build/**/*_test.{js,ts}\" --node-resolve --playwright --watch --browsers chromium",
+    "webtest-coverage": "web-test-runner \"build/**/*_test.js\" --node-resolve --playwright --coverage --browsers chromium firefox",
```
jrobbins, line 33: For each of these where you reference files under build, I think you need to do `npm run build` first so that source files get compiled into the build/ directory.

package.json
jrobbins, line 76: Are these two rebuild dependencies needed?  npm start and webtest seem to work without them.

jrobbins (commented):
I am very happy that this is working.  I have a few comments.

client-src/elements/datatypes.ts
```diff
@@ -1,12 +1,30 @@
...
   [key: string]: any;
   0: number;
   1: string;
 }
 
-declare global {
-  interface Window {
-    google: any;
-  }
+//autolink
```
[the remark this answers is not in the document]
markxiong0122: i think Jason said we might want a separate files to handle those datatypes?

client-src/elements/datatypes.ts
markxiong0122: i dont see it on shoelace's documentation

package.json
markxiong0122: could you clarify what did not work? it seems fine when I run it

markxiong0122: oh i see. it a concurrency issue. sometimes gulp gets running before tsc compiles

client-src/elements/datatypes.ts
jyasskin: It's kinda documented in https://shoelace.style/components/badge#properties. The type is `SbBadge['variant']` after you have `import type SbBadge from '@shoelace-style/shoelace/dist/components/badge/badge.js';` in your imports.

We don't need to make this sort of change in these initial PRs, but it's the sort of thing we should try to clean up as we migrate the bulk of the code.

client-src/elements/external-reviewers.ts
```diff
@@ -108,9 +107,7 @@ export class ExternalReviewer {
    * @param {Record<string, LabelInfo>} labels
    */
   constructor(icon, labels) {
```
jyasskin, line 109: Also no strong need to do this in these initial PRs, but once you're migrating in bulk, you should replace `@param` comments with Typescript types on the parameters, like:

```suggestion
   */
  constructor(icon, labels: Record<string, LabelInfo>) {
```

package.json
```diff
@@ -21,16 +21,17 @@
     "start-emulator-persist": "gcloud beta emulators datastore start --host-port=:15606 --use-firestore-in-datastore-mode",
     "start-emulator": "gcloud beta emulators datastore start --host-port=:15606 --no-store-on-disk --use-firestore-in-datastore-mode",
     "stop-emulator": "curl -X POST 'http://localhost:15606/shutdown'; pkill -f \"CloudDatastore.jar\"",
-    "build": "gulp",
-    "watch": "gulp watch",
+    "build": "npm run tsc-clean && tsc && gulp",
```
jyasskin, line 24: I'd like to be able to take advantage of the incremental build, and also not to take the time to run `tsc` twice every time we have to start the server. The need to clean each time doesn't need to be fixed in this PR, but keep it on your radar as a future optimization.
```suggestion
    "build": "npm run tsc-clean && gulp",
```

package.json
```diff
@@ -79,6 +80,7 @@
     "color-string": ">=1.9.0",
     "del": "^7.1.0",
     "dot-prop": ">=9.0.0",
+    "esbuild": "^0.21.5",
```
jyasskin, line 83: I think we're not using some of the babel dependencies anymore after this change. See what all you can remove.

.gitignore
```diff
@@ -49,3 +49,6 @@ tsconfig.tsbuildinfo
 
 # rollup cache
 .rollup.cache/
+
+# build output
+build/
```
jyasskin, line 54: Add a newline at the end of the file.
```suggestion
build/
```

client-src/elements/autolink.ts
```diff
@@ -118,9 +124,9 @@ function replaceLinkRef(match) {
   }
   LINK_TRAILING_CHARS.forEach(([begin, end]) => {
     if (content.endsWith(end)) {
```
jyasskin, line 126: When you're adding `!` annotations, consider whether you can get the type checker to figure it out without the annotation. In this case, it's easy enough to

```suggestion
    if (end && content.endsWith(end)) {
```

jyasskin (approved):

PR #3960, opened 2024-06-11
Allow searching relative to the current stable milestone.

jyasskin:
Using a 'current_stable' keyword, to leave room for the possibility that
we'll want to be able to also search relative to the current beta or dev
versions.

This uses a different notion of "current" from [`reminders.py`](https://github.com/GoogleChrome/chromium-dashboard/blob/58b02bf7f31d663651516f76e0d4d78a73f5557a/internals/reminders.py#L47), but I couldn't think of a way to explain that "current" in a search term is the version _after_ the one currently being used by users. I think we can just use `current_stable+1` for the verification report.

This also depends on #3959 for the `x+y` syntax to work. (So did #3941, but `now+y` is much less common, so I didn't notice.) Testing that interaction would be tricky, so I didn't add a test.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -17,7 +17,7 @@ import dataclasses
 import datetime
 import logging
 import re
-from typing import Any, Optional, Union, Self
+from typing import Any, Optional, Self, Union
 
 from google.cloud.ndb import Key
 from google.cloud.ndb.tasklets import Future  # for type checking only
--- a/internals/search.py
+++ b/internals/search.py
@@ -13,10 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import dataclasses
 import datetime
 import logging
 import re
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, Self
 
 from google.cloud.ndb import Key
 from google.cloud.ndb.tasklets import Future  # for type checking only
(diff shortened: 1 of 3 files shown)


jrobbins:
Is this PR ready for review?  You have not selected reviewers yet.

jyasskin:
Yep, it's ready.

internals/search.py
```diff
@@ -91,12 +93,28 @@ def process_recent_reviews_query() -> list[int] | Future:
...
 
 
+@dataclasses.dataclass
+class QueryContext:
+  now: datetime.datetime
+  current_stable_milestone: int
+
+  @classmethod
+  def current(cls) -> Self:
+    """Computes the "current" QueryContext based on ambient information."""
```
jrobbins, line 103: I'll have to use the phrase "ambient information" in code comments myself sometime.

internals/search.py
```diff
@@ -91,12 +93,28 @@ def process_recent_reviews_query() -> list[int] | Future:
...
+    current_stable = None
+    for version in fetchchannels.get_omaha_data()[0]['versions']:
```
jrobbins, line 105: This will slow down every query a bit, but it should be OK since it is usually only hitting redis.

jrobbins (approved):

internals/search.py
jyasskin: We could turn this into a function instead of a constant value, so it's only computed for queries that need it, but yeah, my theory was that redis is fast enough.

PR #3961, opened 2024-06-11
Update the built CSS.

jyasskin:


jrobbins (approved):
Thanks for doing this.  I was thinking of sending this exact same PR.

PR #3964, opened 2024-06-12
Keep dependencies, including playwright package deps, up to date on each devcontainer startup.

jyasskin:
This opens a terminal-like window on each devcontainer startup to run the installations, but you can start editing files before that finishes. I was inspired to change this because `npm run webtests` complained that I didn't have the right packages installed for the current version of Playwright.


jcscottiii (approved):
This is a great improvement. LGTM!

PR #3965, opened 2024-06-12
Implemented dialog to inform users that the Guide UX page is obsolete.

jrobbins:
This is the first of the remaining steps on the mini-design doc for phasing out Guide UX pages.

I decided to use sl-dialog instead of a banner message so that it would be even more prominent and so that users would need to interact with it so that they get the idea.


jrobbins:
Screenshot
image

KyleJu (approved):

PR #3966, opened 2024-06-13
Fix date-dependent test

jrobbins:
I wrote some code 90 days ago that looked at a 90 day window of data.  But, I forgot to specify the `today` value in the unit test.  So, 90 days later, the test started failing.  This fixes it.

### DIFF
--- a/api/review_latency_api_test.py
+++ b/api/review_latency_api_test.py
@@ -60,7 +60,7 @@ class ReviewLatencyAPITest(testing_config.CustomTestCase):
 
   def test_do_get__nothing_requested(self):
     """When no reviews have been started, the result is empty."""
-    actual = self.handler.do_get()
+    actual = self.handler.do_get(today=self.today)
     self.assertEqual([], actual)
 
   def test_do_get__normal(self):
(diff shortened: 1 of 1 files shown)


api/review_latency_api_test.py
```diff
@@ -60,7 +60,7 @@ def tearDown(self):
 
   def test_do_get__nothing_requested(self):
     """When no reviews have been started, the result is empty."""
-    actual = self.handler.do_get()
+    actual = self.handler.do_get(today=self.today)
```
KyleJu, line 63: Isn't self.today is hardcoded to datetime(2024, 3, 22)? How will this resolve the out-of-date timestamp issue

jrobbins: Yes, self.today is hard coded, and the dates of the review requests are also hard coded, so they will be within 90 days even as we continue to run the same test months or years from now.     

The problem was that when I was not passing in a hard-coded value for self.today, then it was defaulting to the current datetime.datetime.now() and that was getting farther and farther away from the hard-coded review request date, so the test started failing once it was more than 90 days.

KyleJu: Aha I see. I think normally we can mock the clock if self.today cannot be hardcoded, but there is no need for it here.

PR #3967, opened 2024-06-13
Gray out pencil icons

jrobbins:
This is the second of the remaining steps on the mini design doc for phasing out Guide UX.
This makes the pencil icons gray and adds a tool tip explaining what the user should do instead of using the pencil icons.


client-src/elements/chromedash-feature-page.js
```diff
@@ -395,13 +399,19 @@ export class ChromedashFeaturePage extends LitElement {
           ${renderHTMLIf(
             canEdit && !this.feature.is_enterprise_feature,
             html`
-              <span class="tooltip" title="Edit this feature">
+              <span
+                class="tooltip"
+                title="Deprecated.  Please use 'Edit fields' buttons instead."
```
KyleJu, line 404:
```suggestion
                title="Deprecated. Please use 'Edit fields' buttons instead."
```

KyleJu (approved):

PR #3970, opened 2024-06-13
Fully remove sass

jrobbins:
We've had sass deprecated since last year.  Let's finally remove it.

How things work without sass:
* Our components use lit-css constants in each component file, or included from a shared JS file with lit-css.
* Our HTML templates (including spa.html) link to /static/css files that we generated from sass last year.


dlaliberte (approved):
Yeah!

gulpfile.babel.js
```diff
@@ -10,12 +10,9 @@ import babel from 'gulp-babel';
 import concat from 'gulp-concat';
 import license from 'gulp-license';
 import rename from 'gulp-rename';
-import gulpSass from 'gulp-sass';
```
jyasskin, line 13: Can you remove these from package.json too?

jrobbins: Good point, I'll do that.

Issue #3767, opened 2024-04-02
Store `rollout_milestone` field in the `milestones` field in Stage entities


DanielRyanSmith:
Almost all references to milestone fields are stored in Stage.milestones, which uses the MilestoneSet kind. The exception for this is `rollout_milestone` for enterprise features, which is a separate field from an older implementation with the milestone field unnecessarily separate from similar milestone references. Migrating this `rollout_milestone` field to instead live in the `milestones` field (as `milestones.desktop_first`), will consolidate this milestone field to be handled the same way all other milestone fields are.

jrobbins:
Daniel addressed this in PR #3771.

DanielRyanSmith:
The rollout_milestone field is now being stored in the `FeatureEntry.milestones.desktop_first` for all features. The next step would be to update references in the code to no longer use rollout_milestone.

PR #3972, opened 2024-06-14
Allow searching in all milestone fields at once.

jyasskin:
This nearly matches the criteria for sending verification emails in [`reminders.py`](https://github.com/GoogleChrome/chromium-dashboard/blob/c37b97c65a776a1d3870d117cdd3c0579b621512/internals/reminders.py#L224-L236) except that `reminders.py` is missing `browsers.chrome.ot.ios.start`, and search queries are missing `rollout_milestone`. I'll figure out how to reconcile that in another PR. Let me know if you notice any other discrepancies.
image
image

### DIFF
--- a/internals/search_queries.py
+++ b/internals/search_queries.py
@@ -39,6 +39,58 @@ class Interval(Generic[T]):
   high: T
 
 
+def validate_values(
+  field: ndb.Property, val_list: list[QueryValue | Interval[QueryValue]]
+):
+  """Checks that all of the values in val_list can be compared to the field's values."""
+  for val in val_list:
+    try:
+      # If the field can be set to val, then it can be compared to val.
+      if isinstance(val, Interval):
+        field._validate(val.low)
+        field._validate(val.high)
+      else:
+        field._validate(val)
+    except ndb.exceptions.BadValueError:
+      logging.info('Wrong type of value for %r: %r' % (field, val))
+      raise ValueError('Query value does not match field type')
+
+
+def build_filter(
+  field: ndb.Property, operator: str, val_list: list[QueryValue | Interval[QueryValue]]
+) -> ndb.Node:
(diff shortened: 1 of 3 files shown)


internals/search_queries.py
```diff
@@ -394,3 +408,45 @@ def sorted_by_review_date(descending: bool) -> Future:
...
+    stage_types = [
+      st for st in STAGE_TYPES_BY_QUERY_FIELD[field_name].values() if st is not None
+    ]
+    disjunction.append(
+      ndb.AND(
+        Stage.stage_type.IN(stage_types),
+        build_filter(STAGE_QUERIABLE_FIELDS[field_name], operator, val_list),
+      )
+    )
+  return Stage.query(ndb.OR(*disjunction)).fetch_async(
```
jrobbins, line 443: Datastore (and Firestore in Datastore mode) requires an index that exactly matches the set of conditions in the query.  So, you'll need to add something to `index.yaml`.  The local datastore emulator does not require this.

It can be tricky to create the right index, so the way that I often do this is to run in staging and look at the error message that is generated.

jyasskin: The new query is covered by the 
```yaml
- kind: Stage
  properties:
  - name: stage_type
  - name: milestones.desktop_first
  - name: feature_id
```
indices that support the existing filters that make up this disjunction. I checked by running the datastore emulator in persistent mode and looking at the index.yaml file that it generated.

DanielRyanSmith (approved):
This LGTM 🙂

As for the `rollout_milestone` issue, this is something that has come up and there's not discernible reason we should have it stored as a separate field rather than in the milestone fields, e.g. `milestones.desktop_first`. See https://github.com/GoogleChrome/chromium-dashboard/issues/3767. It might be worth migrating the values over to be stored in the same pattern as other stages at some point in the future.

PR #3975, opened 2024-06-15
Typescript migration: form 

markxiong0122:
migrated form-related files to typescript, including

- form-definition
- form-field-enums
- form-field-specs


client-src/elements/datatypes.ts
```diff
@@ -28,3 +29,148 @@ export type LabelInfo = {
...
+  later?: string;
+  allEarlier?: string;
+  allLater?: string;
+  warning?: string;
+  error?: string;
+}
+
+export interface Field {
+  type?: string;
+  name?: string;
```
jcscottiii, line 149: could you double check if we can use something like this:
```suggestion
  name?: keyof FormattedFeature;
```

I think we can do this. If we can, this would help tie back the possibilities of name back to FormattedFeature.

Currently, we do things like this:

```
      const featureJSONKey = ALL_FIELDS[field].name || field;
      const value = formattedFeature[featureJSONKey];
```

This reduces the benefit of typescript by doing this look up. We should narrow the type of name.


It also may need to be the union of multiple types.

```
name: keyof FormattedFeature | STAGE_FIELD_NAME_MAPPING[keyof STAGE_FIELD_NAME_MAPPING]
```

jcscottiii: By doing this, we could do further refactoring in the future easily by knowing what the possibilities are.

jcscottiii (commented):
I only reviewed the data types.ts file so far but wanted to give you some early feedback. Also, I see the use of `any` in multiple places. Anyway to reduce that?

client-src/elements/datatypes.ts
```diff
@@ -28,3 +29,148 @@ export type LabelInfo = {
   description: string;
   variant: 'primary' | 'success' | 'neutral' | 'warning' | 'danger';
 };
+
+//form-definition
+export interface Browser {
```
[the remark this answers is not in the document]
markxiong0122: Great suggestion. I'll work on it.

client-src/elements/datatypes.ts
markxiong0122: yes, this is possible. working on it rn

client-src/elements/form-field-enums.ts
```diff
@@ -396,10 +400,10 @@ export const GATE_TYPES = {
   TESTING_PLAN: 75,
 };
 
-export const GATE_PREPARING = 0;
-export const GATE_REVIEW_REQUESTED = 2;
-export const GATE_NA_REQUESTED = 9;
-export const VOTE_OPTIONS = {
+export const GATE_PREPARING: number = 0;
```
dlaliberte, line 403: These explicit types, when the type is obvious, are considered redundant and unnecessary in google3, so I wonder if the lint script (whatever we are using now) complained about the lack of the type here, and if so, should we change to configuration not complain.    I'm not sure about other cases where the value of a constant is assigned immediately given a literal or constructor that is clear, e.g. Set<string> and Record<string, number>.

markxiong0122: it did not complain... do you want me to remove them in the current files? I'm not sure about cases like Set and Record and would not those redundant annotations in further PRS

client-src/elements/datatypes.ts
dlaliberte: I suspect that feature names are, currently, arbitrary as long as they are unique.  There is statically defined list of feature names, and the valid feature names are only stored some data table that changes over time, which could possibly be retrieved from the server.    

So in terms of typing, I think it might be more correct to create a new type (e.g. type FeatureName = string) that happens to be an arbitrary string for now, and use that in all the places where feature names are used such as in this Field interface and in FormattedFeature, etc.  Some places it could be optional, (name: ?FeatureName) and other places it could be required (name: FeatureName).

markxiong0122: sounds good, just pushed new changes

client-src/elements/form-field-enums.ts
dlaliberte: I think it should be fine to remove explicit types where both humans and tools can easily figure out the inferred types, and especially when they are constants.   So it still depends on how smart the tools are, which I don't know.

markxiong0122: should we keep it for now then?

client-src/elements/datatypes.ts
```diff
@@ -1,6 +1,9 @@
+import {TemplateResult} from 'lit';
```
jcscottiii, line 1:
```suggestion
import type {TemplateResult} from 'lit';
```

You should be able to `import type` since it is only being used as a type here.

client-src/elements/form-field-enums.ts
jyasskin: I think we should remove these simple redundant types, and only include types when either Typescript demands it or it helps a human reader.

client-src/elements/form-definition.ts
```diff
@@ -1,7 +1,119 @@
...
+}
+
+interface Standards {
+  spec: string;
+  maturity: {val: number; text: string};
+}
+
+type FeatureName = string;
+
+interface Feature {
```
jyasskin, line 46: Eventually, this should be defined next to https://github.com/GoogleChrome/chromium-dashboard/blob/cfc58fc1e83763107f6c2a268db1a8ae568b0c29/client-src/js-src/cs-client.js#L277 or even in https://github.com/GoogleChrome/chromium-dashboard/blob/main/openapi/api.yaml. But I think defining it here is a reasonable intermediate state.

markxiong0122: you mean just the type or the feature interface as well?

jyasskin: The Feature interface.

markxiong0122: will fix this after migrating all element files

client-src/elements/form-field-specs.ts
```diff
@@ -2033,59 +2088,73 @@ function findMaxMilestone(fieldName, stageTypes, getFieldValue) {
...
   const stageTypes =
+    allEarlier &&
     // Only origin trials or dev trials, for now.
     OT_MILESTONE_START_FIELDS.has(allEarlier)
       ? STAGE_TYPES_ORIGIN_TRIAL
-      : DT_MILESTONE_FIELDS.has(allEarlier)
+      : allEarlier && DT_MILESTONE_FIELDS.has(allEarlier)
         ? STAGE_TYPES_DEV_TRIAL
         : null;
   // console.info(`stageTypes: ${stageTypes}`);
```
DanielRyanSmith, line 2135: Not related to this PR, but I believe these commented out console info statements can be removed.

client-src/elements/form-field-specs.ts
```diff
@@ -1552,7 +1599,7 @@ export const ALL_FIELDS = {
 
   all_platforms: {
     type: 'checkbox',
-    inital: false,
+    initial: false,
```
DanielRyanSmith, line 1602: Good typo find 🙂

client-src/elements/form-definition.ts
```diff
@@ -12,7 +123,7 @@ const COMMA_SEPARATED_FIELDS = [
   'i2s_lgtms',
 ];
 
-const LINE_SEPARATED_FIELDS = [
+const LINE_SEPARATED_FIELDS: string[] = [
```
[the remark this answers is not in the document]
DanielRyanSmith: Seconding @jyasskin here

DanielRyanSmith (approved):
This is looking good to me, as long as the considerations from the other comments are handled (it looked like most of them were). 🙂

PR #3978, opened 2024-06-17
TypeScript Migration: Toast

markxiong0122:
migrated chromedash-toast to typescript

### DIFF
--- a/client-src/elements/chromedash-toast.ts
+++ b/client-src/elements/chromedash-toast.ts
@@ -1,9 +1,10 @@
 import {LitElement, css, html} from 'lit';
 import {SHARED_STYLES} from '../css/shared-css.js';
-import {property} from 'lit/decorators.js';
+import {customElement, state, property} from 'lit/decorators.js';
 
 const DEFAULT_DURATION = 7000;
 
+@customElement('chromedash-toast')
 class ChromedashToast extends LitElement {
   @property({type: String})
   msg = '';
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-toast.ts
```diff
@@ -74,35 +71,39 @@ class ChromedashToast extends LitElement {
   /**
    * Shows a message in the toast.
    * @method showMessage
```
jyasskin, line 73: I think we can also delete the `@method`, which is redundant with the method name.
```suggestion
```

client-src/elements/chromedash-toast.ts
```diff
@@ -1,27 +1,24 @@
 import {LitElement, css, html} from 'lit';
 import {SHARED_STYLES} from '../css/shared-css.js';
+import {property} from 'lit/decorators.js';
 
 const DEFAULT_DURATION = 7000;
 
 class ChromedashToast extends LitElement {
```
jyasskin, line 7: You can also replace the `customElements.define` at the bottom of the file with
```suggestion
@customElement('chromedash-toast')
class ChromedashToast extends LitElement {
```

client-src/elements/chromedash-toast.ts
```diff
@@ -1,27 +1,24 @@
...
-    this.waitingForTransition = false;
-  }
+  @property({type: Boolean, reflect: true})
+  open = false;
+
+  @property({attribute: false})
+  actionLabel = '';
+
+  @property({type: Number})
+  currentTimeout: number | null = null;
```
jyasskin, line 18: Looking at the code using this, it seems like it shouldn't be settable using an attribute on the element, and instead should be 
```suggestion
  @state
  currentTimeout: number | null = null;
```
But I haven't checked all the calling code to make sure that's a correct change, and you've correctly translated the `static properties()` method.

jyasskin (approved):

PR #3979, opened 2024-06-17
completely removing datatypes

markxiong0122:
completely removed datatypes.ts and defining types in individual files now. Will add it back when needed


dlaliberte (approved):
Seems very reasonable to move types to associated files.  Maybe others have feedback on this?

However, I would (later) look into replacing that Property type with something named more specific to this context, like TimelineProperty, and maybe change how it is defined and used, which seems like kind of a hack.

client-src/elements/chromedash-timeline.ts
```diff
@@ -2,7 +2,18 @@ import {LitElement, css, html, nothing} from 'lit';
 import {SHARED_STYLES} from '../css/shared-css.js';
 import {showToastMessage} from './utils.js';
 import {property, state} from 'lit/decorators.js';
-import {Property} from './datatypes.js';
+
+declare global {
+  interface Window {
+    google: any;
+  }
+}
```
jyasskin, line 10: I've been wondering if we really need this declaration. It looks like https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.visualization/index.d.ts provides it for us. Do we just need to s/window.google/google/g?

markxiong0122: yes because sometimes window calls it. in cases such as following removing that would throw errors: `new window.google.visualization.DataView(groupedTable);`

dlaliberte: > yes because sometimes window calls it. in cases such as following removing that would throw errors: `new window.google.visualization.DataView(groupedTable);`

Ah, that's something I happen to know about.    In webstatus.dev, we are importing the google.visualization types with this "code":
```
// See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.visualization/index.d.ts
/// <reference types="@types/google.visualization" />
```

This should probably be before other imports in each file that needs it.

markxiong0122: i think we do have this as well

jyasskin: We don't need to remove the `window.`: tsc realizes that's a no-op. The real issue is that there are lots of type errors once we're actually checking the types of calls into the charts library:
```
client-src/elements/chromedash-timeline-page_test.ts:18:9 - error TS2322: Type '() => void' is not assignable to type '{ (options: LoadOptions): Promise<void>; (version: string | number, options: LoadOptions): Promise<void>; (visualization: "visualization", version: string | number, options: LoadOptions): Promise<...>; }'.
  Type 'void' is not assignable to type 'Promise<void>'.

18         load: () => {},
           ~~~~

  node_modules/@types/google.visualization/index.d.ts:11:18
    11         function load(options: LoadOptions): Promise<void>;
                        ~~~~
    The expected type comes from property 'load' which is declared here on type 'typeof charts'

client-src/elements/chromedash-timeline.ts:176:7 - error TS2554: Expected 2-3 arguments, but got 4.

176       [{column: 2, type: 'string'}]
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

client-src/elements/chromedash-timeline.ts:188:7 - error TS2740: Type 'DataView' is missing the following properties from type 'DataTable': addColumn, addRow, addRows, clone, and 24 more.

188       view = new window.google.visualization.DataView(groupedTable);
          ~~~~

client-src/elements/chromedash-timeline.ts:189:12 - error TS2339: Property 'setRows' does not exist on type 'DataTable'.

189       view.setRows(
               ~~~~~~~

client-src/elements/chromedash-timeline.ts:195:61 - error TS2345: Argument of type 'Element | null' is not assignable to parameter of type 'Element'.
  Type 'null' is not assignable to type 'Element'.

195     const chart = new window.google.visualization.LineChart(chartEl);
                                                                ~~~~~~~

client-src/elements/chromedash-timeline.ts:209:7 - error TS2322: Type 'string' is not assignable to type 'number'.

209       width: '100%',
          ~~~~~

  node_modules/@types/google.visualization/index.d.ts:819:13
    819             width?: number | undefined;
                    ~~~~~
    The expected type comes from property 'width' which is declared here on type 'LineChartOptions'

client-src/elements/chromedash-timeline.ts:210:7 - error TS2322: Type 'string' is not assignable to type 'number'.

210       height: '100%',
          ~~~~~~

  node_modules/@types/google.visualization/index.d.ts:796:13
    796             height?: number | undefined;
                    ~~~~~~
    The expected type comes from property 'height' which is declared here on type 'LineChartOptions'


Found 7 errors in 2 files.

Errors  Files
     1  client-src/elements/chromedash-timeline-page_test.ts:18
     6  client-src/elements/chromedash-timeline.ts:176
```

That's probably worth fixing in a separate PR, in which you can investigate which bits are real bugs.

client-src/elements/autolink.ts
```diff
@@ -3,7 +3,18 @@
...
+
+interface TextRun {
+  content: string;
+  tag?: string;
+  href?: string;
+}
+
+interface Component {
+  refRegs: RegExp[];
+  replacer: (match: any) => never[] | TextRun[];
```
jyasskin, line 15: We should refine this `any` time in a later PR.

jyasskin (approved):
Looks good to me, as an incremental improvement. There are at least 3 things to clean up in future PRs, plus the general need for types to have comments describing what they mean.

client-src/elements/chromedash-timeline.ts
markxiong0122: already put those down, will fix later

markxiong0122:
close this bc conflicts are addressed in another PR

PR #3985, opened 2024-06-18
Specify the --project for the datastore emulator.

jyasskin:
Without this I get `ERROR: (gcloud.beta.emulators.datastore.start) The required property [project] is not currently set.` I suspect other folks have run `gcloud config set project VALUE` to set a default, but putting it in the `package.json` will make non-devcontainer operation smoother.


jcscottiii (approved):
LGTM

We should create a separate issue to use a non-real project name (e.g. `local`). This will prevent a user from accidentally doing things to the live staging db in case DATASTORE_EMULATOR_HOST is not set.  I recently [set](https://github.com/GoogleChrome/chromium-dashboard/blob/5e05f61f07a137708a0f3f7d83b83074a70289c6/.devcontainer/db-docker-compose.yml#L16) the project to cr-status-staging when setting up the Devcontainer. But I did it because the start up script has the project [hardcoded](https://github.com/GoogleChrome/chromium-dashboard/blob/5e05f61f07a137708a0f3f7d83b83074a70289c6/scripts/start_server.sh#L8). But we should eventually revisit that.

PR #3988, opened 2024-06-19
Typescript Migration: Activity

markxiong0122:
migrated activity component and related page and test files


jcscottiii:
Take a look at some of these suggestions in this commit https://github.com/jcscottiii/chromium-dashboard/commit/cf0b6bffc0412c4452d9b0b027e8372ac35d4776

DanielRyanSmith (approved):
Making great progress!

client-src/js-src/types.d.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+interface FeatureLink {
+  url: string;
+  type: string;
+  information: any;
+  http_error_code: number;
+}
+
+declare interface ChromeStatusClient {
```
jyasskin, line 25: Now that #3991 is in, I think you can replace this file with `import type {ChromeStatusClient} from '../js-src/cs-client.js';` in the places that use that type.

client-src/js-src/types.d.ts
```diff
@@ -0,0 +1,54 @@
...
+  constructor(
+    message: string,
+    resource: string,
+    method: string,
+    status: number
+  );
+}
+
+declare class FeatureNotFoundError extends Error {
+  featureID: number;
```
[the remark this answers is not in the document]
markxiong0122: good point. I just checked and it should be ID indeed

client-src/js-src/types.d.ts
markxiong0122: made my life easier. Thank you for doing this

PR #3991, opened 2024-06-20
Unify all JS building into the Rollup pipeline with module output.

jyasskin:
This should make it easier to Typescriptify js-src and any JS depending on files in that directory.

Build times before this change: [0m24.487s, 0m26.391s, 0m25.525s, 0m25.144s]
Build times after this change: [0m24.524s, 0m24.357s, 0m24.286s, 0m24.027s]

[The error in the Python tests is from https://bugzilla.mozilla.org/show_bug.cgi?id=1314686 timing out. I think it's unrelated to the changes in this PR.]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ node_modules
 # Generated files
 static/css/
 static/dist/
-static/js/
 
 # OS files
 *.DS_Store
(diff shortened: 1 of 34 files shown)


jrobbins (approved):
Thanks for doing this.  This is exactly the kind of cleanup that this code has been needing.

jcscottiii (approved):
LGTM. 
I see you removed various IIFE too. Thanks for that!

DanielRyanSmith (approved):
Thanks for handling this!

PR #3993, opened 2024-06-21
completely remove datatypes.ts

markxiong0122:
raised this PR from my main branch as the conflicts are resolved here


jyasskin (approved):
I think we'll want to rename some of these to be more descriptive, or to comment them to explain what they're for, but this is good as an interim step.

DanielRyanSmith (approved):

PR #3994, opened 2024-06-21
Typescript Migration: add stage dialog

markxiong0122:
migrated add-stage-dialog


jyasskin (approved):
Make sure `chromedash-add-stage-dialog.ts` is the only changed file after you merge from `main`, but then this looks good.

PR #3995, opened 2024-06-21
Typescript Migration: banner

markxiong0122:
migrated banner


client-src/elements/chromedash-banner.ts
```diff
@@ -1,19 +1,13 @@
...
-    };
-  }
-
-  constructor() {
-    super();
-    this.timestamp = null; // Unix timestamp: seconds since 1970-01-01.
-    this.message = '';
-  }
+  @property({type: Number})
+  timestamp: null | number = null; // Unix timestamp: seconds since 1970-01-01.
```
jyasskin, line 8: While you're moving things anyway, it'd be good to use the jsdoc format for this comment:
```suggestion
  /** Unix timestamp: seconds since 1970-01-01. */
  timestamp: null | number = null;
```

jyasskin (approved):
Make sure `chromedash-banner.ts` is the only changed file after you merge from `main`, but then this looks good.

PR #3996, opened 2024-06-21
Typescript Migration: callout

markxiong0122:
migrated callout


client-src/elements/chromedash-callout.ts
```diff
@@ -1,30 +1,25 @@
...
-    this.hidden = true;
-    this.top = 0;
-    this.left = 0;
-    this.signedIn = false;
-    this.dismissedCues = [];
-  }
+  @property({type: String})
+  targetId = '';
+  @property({type: String})
+  side: 'north' | 'south' | 'east' | 'west' = 'south';
```
jyasskin, line 10: Preserve the TODO:
```suggestion
  /** TODO(jrobbins) Support sides other than "south". */
  @property({type: String})
  side: 'north' | 'south' | 'east' | 'west' = 'south';
```

JSDoc does have a `@todo` tag if you want to see what that looks like in your editor's tooltips: https://jsdoc.app/tags-todo

client-src/elements/chromedash-callout.ts
```diff
@@ -1,30 +1,25 @@
...
+  @property({type: String})
+  cue = '';
```
jyasskin, line 12: Preserve the comment:
```suggestion
  /** String to send to the server when dismissed. */
  @property({type: String})
  cue = '';
```

jyasskin (approved):
Make sure `chromedash-callout.ts` is the only changed file after you merge from `main`, but then this looks good (with the preserved comments).

PR #4007, opened 2024-06-25
Avoid OOMs on staging notifier

jrobbins:
I had made this change for prod in #3888, but I failed to do it for staging.


DanielRyanSmith (approved):
Thanks Jason!

PR #4008, opened 2024-06-25
Typescript Migration: admin blink component

markxiong0122:
migrated the following component:

1. chromedash-admin-component-listing
2. chromedash-admin-component-listing_test
3. chromedash-admin-blink-page
4. chromedash-admin-blink-page_test


client-src/elements/chromedash-admin-blink-component-listing.ts
```diff
@@ -1,8 +1,19 @@
...
+import {html, css, LitElement, TemplateResult} from 'lit';
 import {SHARED_STYLES} from '../css/shared-css.js';
 import {VARS} from '../css/_vars-css.js';
 import {LAYOUT_CSS} from '../css/_layout-css.js';
+import {customElement, property, query} from 'lit/decorators.js';
+import '../js-src/openapi-client';
+import {DefaultApiInterface} from 'chromestatus-openapi';
+import {get} from 'http';
 
+interface User {
```
jyasskin, line 10: From the call site (https://github.com/GoogleChrome/chromium-dashboard/blob/e497b9f5b42603d4bc5e9a5a2a6fcbd62453cdb8/client-src/elements/chromedash-admin-blink-page.js#L79-L80), it looks like this is `import type {ComponentsUser} from 'chromestatus-openapi;`

client-src/elements/chromedash-admin-blink-component-listing_test.ts
```diff
@@ -4,8 +4,10 @@ import {assert, expect, fixture, oneEvent} from '@open-wc/testing';
 import sinon from 'sinon';
 import {DefaultApi} from 'chromestatus-openapi';
 
-/** @type {Map<number, import('chromestatus-openapi').ComponentsUser>} */
-const testUsersMap = new Map([
+const testUsersMap = new Map<
+  number,
+  import('chromestatus-openapi').ComponentsUser
```
jyasskin, line 9: Move this import up to the file level as `import type {ComponentsUser} from 'chromestatus-openapi';`

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -62,29 +65,21 @@ export class ChromedashAdminBlinkPage extends LitElement {
...
+  user = null;
+  @property({type: Boolean})
+  _editMode = false;
+  @property({
+    type: Array<import('chromestatus-openapi').OwnersAndSubscribersOfComponent>,
+  })
   components;
-
-  /** @type {Map<int, import('chromestatus-openapi').ComponentsUser}> */
+  @property({type: Map<number, import('chromestatus-openapi').ComponentsUser>})
```
jyasskin, line 79: Move these imports up to the file level.

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -191,8 +186,8 @@ export class ChromedashAdminBlinkPage extends LitElement {
     if (location.hash) {
       const hash = decodeURIComponent(location.hash);
       if (hash) {
-        const el = this.shadowRoot.querySelector(hash);
-        el.scrollIntoView(true, {behavior: 'smooth'});
+        const el = this.shadowRoot!.querySelector(hash);
+        el!.scrollIntoView({behavior: 'smooth'});
```
jyasskin, line 190: This could be a behavior change, but it's probably the intended behavior.

client-src/elements/chromedash-admin-blink-component-listing.ts
```diff
@@ -81,49 +92,51 @@ export class ChromedashAdminBlinkComponentListing extends LitElement {
...
+  @property({type: Boolean, reflect: true})
+  editing = false;
+  @property({type: Object})
+  component = {};
+  @property({type: Number})
+  index = 0;
+  @property({type: Object})
+  usersMap: Map<number, User> = new Map();
+  @property({type: Number})
+  componentId = 0;
```
jyasskin, line 105: Why did you rename `id` to `componentId`? It's not necessarily wrong, but surprising as part of the Typescript refactoring.

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -191,8 +186,8 @@ export class ChromedashAdminBlinkPage extends LitElement {
...
       }
     }
   }
```
jyasskin, line 193: You missed a use of `<chromedash-admin-blink-component-listing .id=...` in the section below this that won't match up with your renamed `.componentId` property.

client-src/elements/chromedash-admin-blink-page.ts
markxiong0122: yeah it only accepts 1 argument, and should be the configs

client-src/elements/chromedash-admin-blink-component-listing.ts
markxiong0122: it was because there's user.id down below, and TypeScript does not allow this overlap

client-src/elements/chromedash-admin-blink-page.ts
markxiong0122: good catch. thanks

client-src/elements/chromedash-admin-blink-component-listing.ts
markxiong0122: oh sorry i messed this up with another error, I changed this because `Property 'id' in type 'ChromedashAdminBlinkComponentListing' is not assignable to the same property in base type 'LitElement'.` as it extends from litelement

jyasskin: Aha! Because LitElement is an HTMLElement, and `id` there is https://developer.mozilla.org/en-US/docs/Web/API/Element/id. Yep, good change.

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -203,7 +202,7 @@ export class ChromedashAdminBlinkPage extends LitElement {
           (component, index) => html`
             <li class="layout horizontal" id="${component.name}">
               <chromedash-admin-blink-component-listing
-                .id=${component.id}
+                .id=${component.componentId}
```
jyasskin, line 205: When you change the name of an attribute or property in an element, you have to update the attributes and property assignments in uses of that element. You also need to make sure that when you access a field on an object, that field actually exists: `component` is an `OwnersAndSubscribersOfComponent`, which has an `id` field and not a `componentId` field. So this needs to change to

```suggestion
                .componentId=${component.id}
```

You also need to find a way to test this and any other uses of the property, to be sure you've correctly understood what needs to match. @jrobbins, what's a good way for Mark to test this?

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -62,29 +69,21 @@ export class ChromedashAdminBlinkPage extends LitElement {
...
+  @property({type: Boolean})
+  loading = true;
+  @property({type: Object})
+  user = null;
+  @property({type: Boolean})
+  _editMode = false;
+  @property({
+    type: Array<OwnersAndSubscribersOfComponent>,
+  })
   components;
```
jyasskin, line 82: This syntax doesn't work to tell Typescript what type a field has: https://lit.dev/docs/components/properties/#property-options:~:text=reactive%20state.-,type,-When%20converting%20a

The VSCode `runem.lit-plugin` extension notices this and underlines the incorrect type in yellow. Do you have that installed?

You'll want
```suggestion
  @state()
  components: Array<OwnersAndSubscribersOfComponent> | undefined;
```

(I changed `@property({attribute:false})` to `@state()` because these fields are assigned in `fetchData()` rather than in a surrounding template.)

client-src/elements/chromedash-admin-blink-page.ts
```diff
@@ -62,29 +69,21 @@ export class ChromedashAdminBlinkPage extends LitElement {
...
-
-  /** @type {Map<int, import('chromestatus-openapi').ComponentsUser}> */
+  @property({type: Map<number, ComponentsUser>})
   usersMap;
```
jyasskin, line 84: Similar to `components`, but without the `|undefined`
```suggestion
  @state()
  usersMap!: Map<number, ComponentsUser>;
```

jyasskin (changes requested):

client-src/elements/chromedash-admin-blink-page.ts
markxiong0122: this results in a lot of `object might be undefined` errors. I manually added ! after each but not sure if there's a better way to do so

jyasskin: I think the `!`s are correct for this, at least for this change. We might want to add some more explicit conditions to handle when `listComponentUsers` returns an undefined `components` field, or it might be that https://github.com/GoogleChrome/chromium-dashboard/blob/1b192c0ecfbb4f0e1eda57c4387bd9b7df5c8bbb/openapi/api.yaml#L192-L201 should change to list its properties as required (@jcscottiii?), but we can do either of those in a future PR.

jcscottiii: We can definitely do that given that those two keys are always returned

https://github.com/GoogleChrome/chromium-dashboard/blob/1b192c0ecfbb4f0e1eda57c4387bd9b7df5c8bbb/api/components_users.py#L63-L64

jyasskin: And the syntax for that is in https://swagger.io/docs/specification/data-models/dictionaries/#fixed-keys / https://json-schema.org/learn/getting-started-step-by-step#define-required-properties.

DanielRyanSmith (approved):

markxiong0122:
migrated the following component:

chromedash-admin-component-listing
chromedash-admin-component-listing_test
chromedash-admin-blink-page
chromedash-admin-blink-page_test

PR #4011, opened 2024-06-25
Extension initiation email updates

DanielRyanSmith:
This change updates two parts of the behavior found when directed to the feature detail page to finalize an extension.

- Checks that the stage has a pending OT action request before showing the extension finalization dialog, so that the dialog will not be displayed a 2nd time if the action has already been taken.

- Removes an erroneous extra argument that is provided to the dialog.


DanielRyanSmith:
This is probably a good option for a new Playwright test. I'll look into that today.

jrobbins (approved):

PR #4013, opened 2024-06-25
Typescript Migration: admin feature links

markxiong0122:
migrated chromedash-admin-feature-links-page


client-src/elements/chromedash-admin-feature-links-page.ts
```diff
@@ -3,7 +3,26 @@ import {showToastMessage} from './utils.js';
...
+interface FeatureLinksSummary {
+  total_count: number;
+  covered_count: number;
+  uncovered_count: number;
+  error_count: number;
+  http_error_count: number;
+  link_types: {key: string; count: number}[];
+  uncovered_link_domains: {key: string; count: number}[];
+  error_link_domains: {key: string; count: number}[];
+}
```
jyasskin, line 23: I think these types belong in cs-client.js, near the operation that returns them. You can do that while it's still a JS file, by copying the (other) `FeatureLink` example. You'll need to pick a different name for the type.

client-src/elements/chromedash-admin-feature-links-page.ts
```diff
@@ -28,21 +47,26 @@ export class ChromedashAdminFeatureLinksPage extends LitElement {
...
+  featureLinksSummary: FeatureLinksSummary = {
+    total_count: 0,
+    covered_count: 0,
+    uncovered_count: 0,
+    error_count: 0,
+    http_error_count: 0,
+    link_types: [],
+    uncovered_link_domains: [],
+    error_link_domains: [],
+  };
```
jyasskin, line 69: I'd avoid the default where you have to invent field values, and just use

```suggestion
  featureLinksSummary!: FeatureLinksSummary;
```

client-src/elements/chromedash-admin-feature-links-page.ts
```diff
@@ -28,21 +47,26 @@ export class ChromedashAdminFeatureLinksPage extends LitElement {
...
-  }
 
-  constructor() {
-    super();
-    this.featureLinksSummary = {};
-    this.featureLinksSamples = [];
-  }
+  @property({type: String})
+  sampleId = '';
+  @property({type: Boolean})
```
jyasskin, line 53: I think these are `@state()`s too, but double-check the element's call sites.

client-src/elements/chromedash-admin-feature-links-page.ts
markxiong0122: pick a different name for the one in cs-client? or this one? could you elaborate more? I think the one in cs-client.js is missing features_ids, so we could not just use that without modifying a bit or introducing a new type

markxiong0122: and also there's no FeatureLinksSummary

jyasskin: You need all 3 types defined, they should all be defined in `cs-client.js` near the functions that return them, and you'll need 3 distinct names for them. I don't have strong opinions on what the names are, although copying the name of the function that returns them could be a good strategy, leading perhaps to `SampleFeatureLink`.

markxiong0122: would it be ok if we put them in type.d.ts for now, as cs-client is still a javascript file?

jyasskin: As I said above, you can define them in `cs-client.js` using the `/** @typedef {Object} ... @property ... */` syntax that's already used to define a FeatureLink type there.

https://github.com/GoogleChrome/chromium-dashboard/blob/5ca89c828aba07197557d49d0ec64d15dfe22242/client-src/js-src/cs-client.js#L17-L23

I'd prefer to put them in the right file now, so that nobody has to remember to go back and clean up `types.d.ts` later.

markxiong0122: will do

jyasskin (approved):

PR #4014, opened 2024-06-25
Typescript Migration: all features page

markxiong0122:
migrated chromedash-all-features-page and its test file


client-src/elements/chromedash-all-features-page.ts
```diff
@@ -15,37 +28,34 @@ export class ChromedashAllFeaturesPage extends LitElement {
...
-    this.sortSpec = '';
-    this.showEnterprise = false;
-    this.user = {};
-    this.start = 0;
-    this.num = 100;
-    this.starredFeatures = new Set();
-    this.selectedGateId = 0;
-  }
+  @state()
+  rawQuery: RawQuery | undefined = {};
```
jyasskin, line 41: This needs to be an `@property` since it's set from outside the element, at https://github.com/GoogleChrome/chromium-dashboard/blob/9bff5bc3b05d091b5f1e7270175091222ba72aa9/client-src/elements/chromedash-app.js#L343. Double-check the other fields too.

client-src/elements/chromedash-all-features-page.ts
```diff
@@ -15,37 +28,34 @@ export class ChromedashAllFeaturesPage extends LitElement {
...
-      starredFeatures: {type: Object},
-      selectedGateId: {type: Number},
-    };
-  }
+  @property({type: String})
+  title = 'Features';
+  @property({type: Boolean})
+  showQuery = true;
+  @property({type: Object})
+  user: User = null!;
```
jyasskin, line 36: `null!` doesn't make much sense. Do you mean
```suggestion
  @property({type: Object})
  user!: User;
```

client-src/elements/chromedash-all-features-page.ts
```diff
@@ -2,7 +2,20 @@ import {LitElement, css, html, nothing} from 'lit';
...
+import {User} from './chromedash-activity-log.js';
 
+interface RawQuery {
+  q?: string;
+  columns?: string;
+  showEnterprise?: boolean;
+  sort?: string;
+  start?: string;
+  num?: string;
+  [key: string]: any;
```
jyasskin, line 15: Define this in `utils.js` next to https://github.com/GoogleChrome/chromium-dashboard/blob/9bff5bc3b05d091b5f1e7270175091222ba72aa9/client-src/elements/utils.js#L357.

All fields should actually be strings, since that's what comes out of URLSearchParams.getAll().

client-src/elements/chromedash-all-features-page.ts
```diff
@@ -59,28 +69,28 @@ export class ChromedashAllFeaturesPage extends LitElement {
     }
 
     if (this.rawQuery.hasOwnProperty('q')) {
-      this.query = this.rawQuery['q'];
+      this.query = this.rawQuery['q'] ?? this.query;
```
jyasskin, line 72: I'd use `!` here instead of implying that the field could be null. Or Typescript should understand that this pattern guarantees non-null-ness:
```suggestion
    if (this.rawQuery.q != null) {
      this.query = this.rawQuery.q;
```
I'm not sure why the original code used `hasOwnProperty`; maybe an old worry about prototype pollution.

client-src/elements/chromedash-all-features-page.ts
```diff
@@ -59,28 +69,28 @@ export class ChromedashAllFeaturesPage extends LitElement {
...
       this.showEnterprise = true;
     }
     if (this.rawQuery.hasOwnProperty('sort')) {
-      this.sortSpec = this.rawQuery['sort'];
+      this.sortSpec = this.rawQuery['sort'] ?? this.sortSpec;
     }
     if (
       this.rawQuery.hasOwnProperty('start') &&
-      !Number.isNaN(parseInt(this.rawQuery['start']))
+      !Number.isNaN(parseInt(this.rawQuery['start'] as string))
```
jyasskin, line 85: You shouldn't need the `as string`, once you update the `[key: string]:any` definition to `[key: string]:string`.

client-src/elements/chromedash-all-features-page_test.ts
```diff
@@ -5,28 +5,28 @@ import './chromedash-toast';
 import {ChromeStatusClient} from '../js-src/cs-client';
 import sinon from 'sinon';
 
+window.csClient = new ChromeStatusClient('fake_token', 1);
+let getStarsStub = sinon.stub(window.csClient, 'getStars');
```
jyasskin, line 9: I'm surprised it works to create the stub once, restore it in the `afterEach`, and then expect to get the stubbed behavior in later tests.

DanielRyanSmith (approved):

PR #4021, opened 2024-06-26
Change endpoint name for OT creation requests

DanielRyanSmith:
This change updates the endpoint targets for OT creation and activation.


KyleJu (approved):
Also use the new /v1/trials/{trial_id}:extend endpoint for trial extension

KyleJu:
> Also use the new /v1/trials/{trial_id}:extend endpoint for trial extension

Actually don't worry about it in this PR. It can be done in a separate PR. The endpoint needs to be tested a bit more

PR #4025, opened 2024-06-27
Typescript Migration: drawer

markxiong0122:
migrated chromedash-drawer and its test file

google.accounts -> window.google.accounts is made because the namespace google is taken by google.visualization package. this avoids conflicts


client-src/elements/chromedash-drawer.ts
```diff
@@ -177,13 +174,13 @@ export class ChromedashDrawer extends LitElement {
       fetch('/dev/mock_login', {method: 'POST'})
         .then(response => {
           if (!response.ok) {
-            throw new Error('Sign in failed! Response:', response);
+            throw new Error('Sign in failed! Response: ${resposne.status}');
```
DanielRyanSmith, line 177:
```suggestion
            throw new Error('Sign in failed! Response: ${response.status}');
```

client-src/elements/chromedash-activity-log.ts
```diff
@@ -77,11 +77,13 @@ export class ChromedashAmendment extends LitElement {
 customElements.define('chromedash-amendment', ChromedashAmendment);
 
 export interface User {
+  //will move to chromedash-app.js after migration
```
DanielRyanSmith, line 80: It might help to formalize this more as a TODO.

```suggestion
  // TODO(markxiong0122): Move this interface to chromedash-app.js after migration.
```

client-src/elements/chromedash-drawer.ts
```diff
@@ -177,13 +174,13 @@ export class ChromedashDrawer extends LitElement {
       fetch('/dev/mock_login', {method: 'POST'})
         .then(response => {
           if (!response.ok) {
-            throw new Error('Sign in failed! Response:', response);
+            throw new Error('Sign in failed! Response: ${response.status}');
```
jyasskin, line 177: This also needs backticks in order to be evaluated:
```suggestion
            throw new Error(`Sign in failed! Response: ${response.status}`);
```

jyasskin:
> google.accounts -> window.google.accounts is made because the namespace google is taken by google.visualization package. this avoids conflicts

Install https://www.npmjs.com/package/@types/google.accounts instead.

client-src/elements/chromedash-drawer.ts
jyasskin: You missed this change.

client-src/elements/chromedash-activity-log.ts
```diff
@@ -72,11 +72,13 @@ export class ChromedashAmendment extends LitElement {
 customElements.define('chromedash-amendment', ChromedashAmendment);
 
 export interface User {
+  // TODO(markxiong0122): Move this interface to chromedash-app.js after migration.
```
jyasskin, line 75: The interfaces should be next to the functions that create instances in the first place, which is in cs-client in this case.

```suggestion
  // TODO(markxiong0122): Move this interface near cs-client.js:getPermissions() after migration.
```

client-src/elements/chromedash-activity-log.ts
```diff
@@ -72,11 +72,13 @@ export class ChromedashAmendment extends LitElement {
...
   can_create_feature: boolean;
   can_edit_all: boolean;
   can_comment: boolean;
   is_admin: boolean;
   email: string;
+  approvable_gate_types?: any;
```
jyasskin, line 81: The type itself is initialized in https://github.com/GoogleChrome/chromium-dashboard/blob/b0049ed27615bb01a4063924557bf8a0f5f36806/api/permissions_api.py#L45-L65
which implies that
```suggestion
  approvable_gate_types: int[];
  editable_features: int[];
```
finishes the type.

client-src/elements/chromedash-drawer.ts
```diff
@@ -177,13 +174,13 @@ export class ChromedashDrawer extends LitElement {
...
           }
         })
         .then(() => {
           setTimeout(() => {
-            const url = window.location.href.split('?')[0];
-            window.location = url;
+            const url = window.location.href.split('?')[0] as string & Location;
```
jyasskin, line 182: Try to avoid `as` clauses when you can, since the typechecker doesn't check them. I think you just want

```suggestion
            const url = window.location.href.split('?')[0]!;
```
here.

DanielRyanSmith (approved):

PR #4030, opened 2024-06-28
Typescript Migration: enterprise release notes

markxiong0122:
migrated chromedash-enterprise-release-notes-page and its test file


client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -1,4 +1,4 @@
-import {html, LitElement, css} from 'lit';
+import {html, LitElement, css, TemplateResult} from 'lit';
```
jyasskin, line 1: Run 'Organize Imports' to sort these.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -13,28 +13,50 @@ import {
...
-  static get properties() {
-    return {
-      currentFeatures: {type: Array},
-      upcomingFeatures: {type: Array},
-      features: {type: Array},
-      channels: {type: Object},
-      selectedMilestone: {type: Number},
-    };
-  }
+export interface Feature {
```
jyasskin, line 20: Please don't add multiple Feature interfaces. Instead, define a single one in an appropriate place (cs-client.js), and extend it as you discover more fields. If the tests for a particular file want to provide fake data for a Feature, and you don't want to include all the fields that the API provides, you can define a local subset using `Pick<Feature, 'id'|'name'|'summary'|...>`: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys. You can define a local alias for that subset using 

```typescript
type AliasName = Pick<...>;
```

But I'd generally avoid exporting those aliases.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -13,28 +13,50 @@ import {
...
-  }
+interface Channels {
+  stable: {
+    version: number;
+  };
+}
+
+@customElement('chromedash-enterprise-release-notes-page')
+export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
+  //TODO(markxiong0122): add types for the following properties, import Feature from chromedash-feature after migration
```
jyasskin, line 49: I think this TODO isn't accurate anymore.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -205,7 +227,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
       ([milestone, platforms]) => ({
         stage_type: STAGE_ENT_ROLLOUT,
         rollout_milestone: Number(milestone),
-        rollout_platforms: Array.from(platforms),
+        rollout_platforms: Array.from(platforms as Set<string>),
```
jyasskin, line 230: Adjust the `new Set()` call above to include a type, rather than casting down here.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -252,13 +274,13 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
       .sort((a, b) => {
         // Highest impact of the stages from feature A.
         const impactA = Math.max(
-          a.stages
+          ...a.stages
```
jyasskin, line 277: Lol, so this never worked. :)

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -298,7 +324,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
         const queryParams = parseRawQuery(window.location.search);
         if (milestoneQueryParamKey in queryParams) {
           this.selectedMilestone = parseInt(
-            queryParams[milestoneQueryParamKey],
+            queryParams[milestoneQueryParamKey] as string,
```
jyasskin, line 327: It would probably be better to fix the return type of `parseRawQuery` in `utils.js` rather than adding a cast here. I think it returns a `Record<string, string>`?

client-src/elements/chromedash-enterprise-release-notes-page_test.ts
```diff
@@ -256,17 +260,17 @@ describe('chromedash-enterprise-release-notes-page', () => {
       <chromedash-enterprise-release-notes-page></chromedash-enterprise-release-notes-page>`);
     assert.exists(component);
     assert.instanceOf(component, ChromedashEnterpriseReleaseNotesPage);
-    assert.equal(parseRawQuery(window.location.search).milestone, 100);
+    assert.equal((parseRawQuery(window.location.search) as QueryParams).milestone, 100);
```
jyasskin, line 263: Try to fix `parseRawQuery` instead of adding all these casts.

client-src/elements/chromedash-enterprise-release-notes-page_test.ts
```diff
@@ -280,8 +284,8 @@ describe('chromedash-enterprise-release-notes-page', () => {
 
     // invalid feature requests would trigger the toast to show message
     const toastEl = document.querySelector('chromedash-toast');
-    const toastMsgSpan = toastEl.shadowRoot.querySelector('span#msg');
-    assert.include(toastMsgSpan.innerHTML,
+    const toastMsgSpan = toastEl!.shadowRoot!.querySelector('span#msg');
```
jyasskin, line 287: Consider using `?` instead of `!` to fix cases where the object could actually be null.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -8,33 +10,32 @@ import {
...
+  @state()
+  currentFeatures: Feature[] = [];
+  @state()
+  upcomingFeatures: Feature[] = [];
+  @state()
+  features: Feature[] = [];
+  @state()
+  channels!: Channels;
+  @state()
+  selectedMilestone?: number | undefined = undefined;
```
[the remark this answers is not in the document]
markxiong0122: my bad, should've spotted this earlier

DanielRyanSmith (approved):
Sorry for the delay here

PR #4048, opened 2024-07-02
Moved type Feature into cs-client 

markxiong0122:
moved type Feature into cs-client


client-src/elements/form-definition.ts
```diff
@@ -165,7 +103,9 @@ export function formatFeatureForEdit(feature: Feature): FormattedFeature {
     search_tags: feature.tags,
 
     // from feature.browsers.chrome
-    blink_components: feature.browsers.chrome.blink_components[0],
+    blink_components: feature.browsers.chrome.blink_components
+      ? feature.browsers.chrome.blink_components[0]
+      : null,
```
jyasskin, line 108: Check out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining for a better way to write this sort of condition, but I think you should go the other way on these fields: if they didn't already check for null/undefined, you should set the type of their input fields as required. So this one would imply that `FeatureDictInnerChromeBrowserInfo.blink_components` is not nullable or optional.

client-src/js-src/cs-client.js
```diff
@@ -41,6 +41,236 @@
...
 
+/**
+ * @typedef {Object} StageDict
+ * @property {number} id
+ * @property {string} created
+ * @property {number} feature_id
+ * @property {number} stage_type
+ * @property {string} display_name
+ * @property {number} intent_stage
+ * @property {string|null} intent_thread_url
```
jyasskin, line 52: Mark these as optional rather than nullable (https://jsdoc.app/tags-property#examples:~:text=A%20type%20definition%20with%20required%20and%20optional%20property):

```suggestion
 * @property {string} [intent_thread_url]
```

client-src/elements/form-definition.ts
```diff
@@ -1,103 +1,41 @@
...
+import {Feature} from '../js-src/cs-client';
 
 interface FormattedFeature {
   category: number;
   enterprise_feature_categories: string[];
   feature_type: number;
   intent_stage: number;
   accurate_as_of: boolean;
-  spec_link: string;
+  spec_link: string | null;
```
jyasskin, line 11: To the extent that these are actually optional properties, they'll become

```suggestion
  spec_link?: string;
```

jyasskin (changes requested):

client-src/elements/form-definition.ts
markxiong0122: but it is nullable in the original type definition tho, should we keep it as `blink_components: feature.browsers.chrome.blink_components?.[0] ?? null,`

jyasskin: @DanielRyanSmith wrote the type definition in https://github.com/GoogleChrome/chromium-dashboard/pull/2844, and I think he was being conservative there. For example, `blink_components` is set by https://github.com/GoogleChrome/chromium-dashboard/blob/b6dc809b7a0c465bea3efbfe999edb2eacdd6728/api/converters.py#L338, which ensures it's always set.

@DanielRyanSmith, what do you think we should do with the types here?

Even for the ones that are optional, I don't think there's a benefit from coercing `undefined` to `null` with `?? null`.

DanielRyanSmith: I'm in agreement that coercing to null doesn't seem necessary here. I also think some of the old type definitions I wrote need to be re-examined and a number of the fields should be made not required, but that is separate work for another day. 😅

markxiong0122: that sounds good. i put as `    blink_components: feature.browsers.chrome.blink_components?.[0],` now and it seems working properly. could you check and approve if there's no additional issue

client-src/elements/form-definition.ts
```diff
@@ -1,103 +1,41 @@
...
+  ff_views?: number;
+  ff_views_link?: string;
+  ff_views_notes?: string;
+  safari_views?: number;
+  safari_views_link?: string;
+  safari_views_notes?: string;
+  web_dev_views?: number;
+  web_dev_views_link?: string;
+  web_dev_views_notes?: string;
+  other_views_notes?: string | null;
```
jyasskin, line 38:
```suggestion
  other_views_notes?: string;
```

I think?

jyasskin (approved):
Looks good, thanks!

client-src/elements/form-definition.ts
markxiong0122: oops left this behind

PR #4049, opened 2024-07-02
Typescript Migration: typeahead

markxiong0122:
migrated chromedash-typeahead and its test file


client-src/elements/chromedash-typeahead.ts
```diff
@@ -77,35 +75,36 @@ export class ChromedashTypeahead extends LitElement {
...
-    this.slDropdownRef.value.show();
+    this.slDropdownRef.value!.show();
   }
 
   findPrefix() {
-    const inputEl = this.slInputRef.value.input;
+    const inputEl = this.slInputRef.value?.input;
+    if (!inputEl) return;
     const wholeStr = inputEl.value;
     const caret = inputEl.selectionStart;
```
jrobbins, line 98: Could we avoid repeatedly using `!` below by changing this to something like this?
`const caret: Number = inputEl.selectionStart || 0;`

client-src/elements/chromedash-typeahead.ts
```diff
@@ -123,37 +122,40 @@ export class ChromedashTypeahead extends LitElement {
 
   async handleCandidateSelected(e) {
     const candidateValue = e.detail.item.value;
-    const inputEl = this.slInputRef.value.input;
-    const wholeStr = inputEl.value;
+    const inputEl = this.slInputRef.value?.renderRoot.querySelector('input');
```
jrobbins, line 125: I think we could avoid repeated use of `!` below by bailing out here if `inputEl`is not found.
`if (!inputEl) return;`

client-src/elements/chromedash-typeahead_test.ts
```diff
@@ -199,13 +202,13 @@ describe('chromedash-typeahead-item', () => {
...
     `);
-    const valueEl = component.shadowRoot.querySelector('#value');
-    assert.include(valueEl.innerHTML, 'aVal');
-    assert.include(valueEl.innerHTML, 'ue');
-    const bold = component.shadowRoot.querySelector('b');
-    assert.include(bold.innerHTML, 'aVal');
-    const docEl = component.shadowRoot.querySelector('#doc');
-    assert.include(docEl.innerHTML, 'Docs about it');
+    const valueEl = component.shadowRoot!.querySelector('#value');
+    assert.include(valueEl!.innerHTML, 'aVal');
```
jrobbins, line 206: Can we avoid the repeated `!`s below by adding:
`assert.exists(valueEl);`?

jrobbins (approved):
Looks good to me after considering a few comments.

client-src/elements/chromedash-typeahead.ts
markxiong0122: sounds good

client-src/elements/chromedash-typeahead_test.ts
markxiong0122: yeah that's a good idea, rendorRoot is not applicable to those so I used ! in the first place

PR #4050, opened 2024-07-02
Remove OT request note field from form

DanielRyanSmith:
The "Anything else?" request note field for OT creation is being deprecated with the automation of the process, so this field will be removed entirely from the form.

### DIFF
--- a/client-src/elements/form-definition.ts
+++ b/client-src/elements/form-definition.ts
@@ -628,7 +628,6 @@ export const ORIGIN_TRIAL_CREATION_FIELDS: MetadataFields = {
         'ot_has_third_party_support',
         'ot_is_critical_trial',
         'ot_require_approvals',
-        'ot_request_note',
       ],
     },
   ],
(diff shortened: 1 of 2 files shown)


KyleJu:
Did a quick search and it is used elsewhere across the [codebase](https://github.com/search?q=repo%3AGoogleChrome%2Fchromium-dashboard%20ot_request_note&type=code). Should they be removed as well?

DanielRyanSmith:
> Did a quick search and it is used elsewhere across the [codebase](https://github.com/search?q=repo%3AGoogleChrome%2Fchromium-dashboard%20ot_request_note&type=code). Should they be removed as well?

My plan was to remove this from the form first so that it is not visible to users, then we can retire the rest of the references in a separate change (sorry, I didn't explicitly state that though 😅)

KyleJu (approved):

KyleJu:
> > Did a quick search and it is used elsewhere across the [codebase](https://github.com/search?q=repo%3AGoogleChrome%2Fchromium-dashboard%20ot_request_note&type=code). Should they be removed as well?
> 
> My plan was to remove this from the form first so that it is not visible to users, then we can retire the rest of the references in a separate change (sorry, I didn't explicitly state that though 😅)

I see! Thanks for the explanation!

PR #4057, opened 2024-07-03
Typescript Migration: featurelist

markxiong0122:
migrated chromedash-featurelist

### DIFF
--- a/client-src/elements/chromedash-featurelist.ts
+++ b/client-src/elements/chromedash-featurelist.ts
@@ -19,15 +19,15 @@ class ChromedashFeaturelist extends LitElement {
   @property({type: Array})
   editableFeatures!: Feature[];
   @property({type: Object})
-  metadataEl = document.querySelector('chromedash-metadata'); // The metadata component element. Directly edited in template/features.html
+  metadataEl; // The metadata component element. Directly edited in template/features.html
   @property({attribute: false})
-  _onFeatureToggledBound = this._onFeatureToggled.bind(this);
+  _onFeatureToggledBound;
   @property({attribute: false})
-  _onStarToggledBound = this._onStarToggled.bind(this);
+  _onStarToggledBound;
   @state()
   features!: Feature[]; // Directly edited and accessed in template/features.html
   @state()
-  searchEl!: HTMLInputElement; // The search input element. Directly edited in template/features.html
+  searchEl; // The search input element. Directly edited in template/features.html
   @state()
   filtered!: Feature[];
   @state()
@@ -39,17 +39,18 @@ class ChromedashFeaturelist extends LitElement {
   @state()
   _hasScrolledByUser = false; // Used to set the app header state.
   @state()
-  _featuresUnveilMetric = new Metric('features_unveil').start();
+  _featuresUnveilMetric = new Metric('features_unveil');
   @state()
-  _featuresFetchMetric = new Metric('features_loaded').start();
+  _featuresFetchMetric = new Metric('features_loaded');
 
-  constructor() {
-    super();
-    this._loadData();
-  }
   connectedCallback() {
     super.connectedCallback();
-    this.searchEl = document.querySelector('.search input')!;
+    this.searchEl = document.querySelector('.search input');
+    this.metadataEl = document.querySelector('chromedash-metadata');
+    this._featuresFetchMetric.start();
+    this._featuresUnveilMetric.start();
+    this._onFeatureToggledBound = this._onFeatureToggled.bind(this);
+    this._onStarToggledBound = this._onStarToggled.bind(this);
     this._loadData();
   }
 
--- a/client-src/elements/chromedash-featurelist.ts
+++ b/client-src/elements/chromedash-featurelist.ts
@@ -17,30 +17,29 @@ class ChromedashFeaturelist extends LitElement {
   @property({type: String})
   signedInUser = '';
   @property({type: Array})
-  editableFeatures!: Feature[];
-  @property({type: Object})
-  metadataEl; // The metadata component element. Directly edited in template/features.html
+  editableFeatures!: number[];
   @state()
-  features!: Feature[]; // Directly edited and accessed in template/features.html
-  @state()
-  searchEl; // The search input element. Directly edited in template/features.html
+  features!: Feature[];
+  /** The search input element. */
   @state()
   filtered!: Feature[];
   @state()
   openFeatures = new Set<number>();
   @state()
   starredFeatures = new Set<number>();
+  searchEl: HTMLInputElement;
+  metadataEl: HTMLElement;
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-featurelist.ts
```diff
@@ -214,7 +210,7 @@ class ChromedashFeaturelist extends LitElement {
   }
 
   // Directly called from template/features.html
-  filter(val, shouldPushState) {
+  filter(val, shouldPushState = false) {
```
DanielRyanSmith, line 213: This is the change we talked about in the meeting yesterday, and the change should have no functional ramifications.

client-src/elements/chromedash-featurelist.ts
```diff
@@ -374,7 +370,7 @@ class ChromedashFeaturelist extends LitElement {
     let filteredWithState = this.filtered.map(feature => {
       const editable =
         this.isSiteEditor ||
-        (this.editableFeatures && this.editableFeatures.includes(feature.id));
+        (this.editableFeatures && this.editableFeatures.includes(feature));
```
DanielRyanSmith, line 373: Is there a reason to change checking `feature` vs. `feature.id` here?

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,47 @@ const MAX_FEATURES_SHOWN = 500;
...
+  @state()
+  starredFeatures = new Set<number>();
+  @state()
+  _hasInitialized = false; // Used to check initialization code.
+  @state()
+  _hasScrolledByUser = false; // Used to set the app header state.
+  @state()
+  _featuresUnveilMetric = new Metric('features_unveil');
+  @state()
+  _featuresFetchMetric = new Metric('features_loaded');
```
jyasskin, line 44: Similarly, these weren't [reactive](https://lit.dev/docs/components/properties/#internal-reactive-state) before your change, so you shouldn't add reactivity in your change.
```suggestion
  _hasInitialized = false; // Used to check initialization code.
  _hasScrolledByUser = false; // Used to set the app header state.
  _featuresUnveilMetric = new Metric('features_unveil');
  _featuresFetchMetric = new Metric('features_loaded');
```

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,47 @@ const MAX_FEATURES_SHOWN = 500;
...
-    this._featuresUnveilMetric = new Metric('features_unveil');
-    this._featuresFetchMetric = new Metric('features_loaded');
-    this._featuresUnveilMetric.start();
+    this.metadataEl = document.querySelector('chromedash-metadata');
     this._featuresFetchMetric.start();
-
-    /* The running context `this` inside `renderItem` is `lit-virtualizer`
-     * rather than `chromedash-featurelist`, so all function calls inside need
-     * to be bound to `this`. */
+    this._featuresUnveilMetric.start();
```
jyasskin, line 51: You should probably keep these `.start()` calls in the constructor, to avoid changing any timings in this PR. It's possible that `connectedCallback()` is actually a better place, but that bug fix should be separate from the Typescript conversion.

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,47 @@ const MAX_FEATURES_SHOWN = 500;
...
-
-    this._featuresUnveilMetric = new Metric('features_unveil');
-    this._featuresFetchMetric = new Metric('features_loaded');
-    this._featuresUnveilMetric.start();
+    this.metadataEl = document.querySelector('chromedash-metadata');
     this._featuresFetchMetric.start();
-
-    /* The running context `this` inside `renderItem` is `lit-virtualizer`
-     * rather than `chromedash-featurelist`, so all function calls inside need
-     * to be bound to `this`. */
```
jyasskin, line 49: Don't delete comments like this.

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,47 @@ const MAX_FEATURES_SHOWN = 500;
...
-    this.filtered = [];
-    this.metadataEl = document.querySelector('chromedash-metadata');
+  @property({type: Object})
+  user!: User;
+  @property({type: Boolean})
+  isSiteEditor = false;
+  @property({type: String})
+  signedInUser = '';
+  @property({type: Array})
+  editableFeatures!: Feature[];
```
jyasskin, line 20: I believe this type should be 
```suggestion
  editableFeatures!: number[];
```
Because it's set at https://github.com/GoogleChrome/chromium-dashboard/blob/55e6a2b8c07d92995541d970e0d9a983d971acca/templates/features.html#L35, which can't assign an object type. (On the other hand, that line appears to come from https://github.com/GoogleChrome/chromium-dashboard/blob/55e6a2b8c07d92995541d970e0d9a983d971acca/framework/basehandlers.py#L497 and so always be an empty array. But this PR isn't the place to fix that. FYI @DanielRyanSmith)

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,47 @@ const MAX_FEATURES_SHOWN = 500;
...
+  @property({type: Object})
+  metadataEl; // The metadata component element. Directly edited in template/features.html
+  @property({attribute: false})
+  _onFeatureToggledBound;
```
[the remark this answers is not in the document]
markxiong0122: yep, deleting them would also pass the test

client-src/elements/chromedash-featurelist.ts
markxiong0122: id does not exist on the type, so I assumed that it was initially a typo

markxiong0122: but according to the updated type we should indeed include .id. my bad

client-src/elements/chromedash-featurelist.ts
```diff
@@ -8,48 +10,36 @@ const MAX_FEATURES_SHOWN = 500;
...
-
-    /* The running context `this` inside `renderItem` is `lit-virtualizer`
-     * rather than `chromedash-featurelist`, so all function calls inside need
-     * to be bound to `this`. */
-    this._onFeatureToggledBound = this._onFeatureToggled.bind(this);
-    this._onStarToggledBound = this._onStarToggled.bind(this);
-
+    this._featuresUnveilMetric.start();
+    this.searchEl = document.querySelector('.search input')!;
+    this.metadataEl = document.querySelector('chromedash-metadata')!;
```
jyasskin, line 42: No need to change this now, but in future changes it can be helpful to keep unchanged lines in the same order as they were in the original code, to minimize the apparent diff for reviewers.

jyasskin (approved):

PR #4059, opened 2024-07-03
Typescript Migration: form-field

markxiong0122:


client-src/elements/chromedash-form-field.ts
```diff
@@ -57,9 +58,8 @@ export class ChromedashFormField extends LitElement {
   connectedCallback() {
     super.connectedCallback();
     this.fieldProps = ALL_FIELDS[this.name] || {};
-
     // Register this form field component with the page component.
-    const app = document.querySelector('chromedash-app');
+    const app: ChromedashApp | null = document.querySelector('chromedash-app');
```
dlaliberte, line 62: There must be a chromedash-app, so it would be reasonable to create a utility for getting this element, and throwing an error if it does not exist.

markxiong0122: this is a good point. added a utility function to get the element, but it still doesn't pass the webtest with the same error `      AssertionError: expected '\n      <!---->\n      <!--?lit$227800399$-->\n            <tr class="">\n              <th colspan="2">\n                <b><!--?lit$227800399$-->Unlisted:</b>\n              </th>\n            </tr>\n          \n      <tr class="">\n        <td><!--?lit$227800399$-->\n        <sl-checkbox size="small" name="unlisted" id="id_unlisted">\n          <!--?lit$227800399$-->A specific label\n        </sl-checkbox>\n       <!--?lit$227800399$--></td>\n        <td>\n          <!--?lit$227800399$--><span class="helptext"> <!--?lit$227800399$--> Check this box to hide draft features in list views. Anyone\n    with a link will be able to view the feature\'s detail page. </span>\n          <!--?lit$227800399$-->\n        </td>\n      </tr>\n\n      <!--?lit$227800399$-->\n    ' to include 'checked'`

client-src/elements/chromedash-form-field.ts
```diff
@@ -504,5 +513,3 @@ function getFieldValueWithStage(fieldName, stageOrId, formFieldValues) {
   const value = getFieldValueFromFeature(fieldName, feStage, feature);
   return value;
 }
```
jyasskin, line 515: Look for functions with `@param {...}` jsdoc, and convert them to Typescript syntax. Otherwise you actually lose typechecking during the conversion to TS.

client-src/elements/chromedash-form-field.ts
```diff
@@ -1,49 +1,54 @@
...
-    this.checkMessage = '';
-  }
+@customElement('chromedash-form-field')
+export class ChromedashFormField extends LitElement {
+  @property({type: String})
+  name = '';
+  @property({type: Number}) // Represents which field this is on the form.
+  index = -1;
+  @property({type: Object}) // All other field value objects in current form.
+  fieldValues!: FieldValues;
```
jyasskin, line 24: This property is an array.

client-src/elements/chromedash-form-field.ts
```diff
@@ -179,7 +187,7 @@ export class ChromedashFormField extends LitElement {
     }
     // There should only be one now.
     const formFieldElement = formFieldElements[0];
-
+    let checkResult;
```
jyasskin, line 190: That's interesting: if `checkResult` didn't exist here, then the below line would either read `window.checkResult` (which I don't think is created anywhere) or throw. @dlaliberte, what's going on here?

Instead of creating a local variable that'll always be `undefined`, this PR should fix or delete the logic.

client-src/elements/chromedash-form-field.ts
```diff
@@ -273,9 +281,10 @@ export class ChromedashFormField extends LitElement {
     // form field name can be specified in form-field-spec to match DB field name
     const fieldName = this.fieldProps.name || this.name;
     // choices can be specified in form-field-spec or fetched from API
-    const choices = this.fieldProps.choices || this.componentChoices;
+    const choices: [number, string][] | [number, string, string][] =
+      this.fieldProps.choices || this.componentChoices;
 
-    let fieldHTML = '';
+    let fieldHTML: TemplateResult | string = '';
```
jyasskin, line 287: I'd probably fix the types by changing the initialization to "html``" instead, since that matches the other assignments.

```suggestion
    let fieldHTML = html``;
```

client-src/elements/chromedash-form-field.ts
markxiong0122: will fix when @dlaliberte throw some hints here :))

client-src/elements/chromedash-form-field.ts
```diff
@@ -1,49 +1,57 @@
...
 import {ref} from 'lit/directives/ref.js';
+
+import {ChromedashApp} from './chromedash-app';
 import './chromedash-textarea';
 import {ALL_FIELDS, resolveFieldForFeature} from './form-field-specs';
 import {getFieldValueFromFeature, showToastMessage} from './utils.js';
+import {Feature, StageDict} from '../js-src/cs-client';
+import {FormattedFeature} from './form-definition';
+
+interface FormFieldValue {
```
markxiong0122, line 13: @jyasskin Could you check if the type is defined correctly here? it seems like the latest change wont pass playwright test

client-src/elements/chromedash-form-field.ts
```diff
@@ -205,7 +216,7 @@ export class ChromedashFormField extends LitElement {
     };
     // Attach the feature to the getFieldValue function, which is needed to
     // iterate through stages not in the form.
-    getFieldValue.feature = this.fieldValues?.feature;
```
jyasskin, line 208: In general, you can't change what members get assigned without checking that those members are never used. In this case, `getFieldValue.feature` is used at https://github.com/GoogleChrome/chromium-dashboard/blob/a92a06d3219c6d5ac4435fd67c4db781f3247f05/client-src/elements/form-field-specs.ts#L2091. That area of bug matches the test failure.

To get Typescript to accept this member on a function, you'll have to give `getFieldValue` a type. Maybe 
```typescript
{ (fieldName: string, stageOrId: string): any;
  feature?: Feature; }
```
Based on https://www.typescriptlang.org/docs/handbook/2/functions.html#call-signatures

client-src/elements/chromedash-form-field.ts
```diff
@@ -1,49 +1,57 @@
...
-    this.stageType = undefined;
-    this.componentChoices = {};
-    this.checkMessage = '';
-  }
+  @property({type: String})
+  name = '';
+  @property({type: Number}) // Represents which field this is on the form.
+  index = -1;
+  @property({type: Array}) // All other field value objects in current form.
+  fieldValues: FormFieldValue[] = [];
```
jyasskin, line 27: It turns out this isn't just an array; lines like https://github.com/GoogleChrome/chromium-dashboard/blob/a92a06d3219c6d5ac4435fd67c4db781f3247f05/client-src/elements/chromedash-guide-metadata-page.js#L205 assign a `.feature` property to it. You can capture that as `FormFieldValue[] & {feature: Feature}`.

client-src/elements/chromedash-form-field.ts
```diff
@@ -465,12 +481,12 @@ export class ChromedashFormField extends LitElement {
...
- * @param {number|Object|undefined} stageOrId
- * @param {Array<{name:string, value:*}>} formFieldValues
- * @return {*}
  */
-function getFieldValueWithStage(fieldName, stageOrId, formFieldValues) {
+function getFieldValueWithStage(
+  fieldName: string,
+  stageOrId: number | StageDict | undefined,
+  formFieldValues: FormFieldValue[] & {feature?: Feature}
+): any {
```
jyasskin, line 489: I'd leave off the explicit `:any` return type so that if and when the contained functions get more precise return types, this one will automatically propagate them.
```suggestion
) {
```

jyasskin (commented):
Looks good once @dlaliberte says what to do with the `let checkResult`.

client-src/elements/chromedash-form-field.ts
jyasskin: Ok, what's going on is that this function runs before the `<sl-input>` element finishes upgrading, so `formFieldElement.input` is always false. As far as I can see, that means `setupSemanticCheck()` is always a no-op, so we should just delete it instead of fixing up its types. Any objections?

dlaliberte: You appear to be correct, Jeffrey.  I don't recall what happened regarding this checkResult value.  But I also don't recall leaving it as a noop.  If the semantic checks are still working, and assuming tests are sufficient, then this function can go.  I do recall having difficulty synchronizing the initial semantic check call when first loading a page that has pre-existing errors in the data.  So be sure to check for that situation.

client-src/elements/chromedash-form-field.ts
jyasskin: Ping.

jyasskin (approved):
Thanks @dlaliberte. @markxiong0122, to test the semantic checks like Dan's saying, you can set up a feature with milestones like at the bottom of https://cr-status-staging.appspot.com/guide/stage/6158217113763840/3/5695674368917504, and check that the errors there still appear.

PR #4060, opened 2024-07-05
Typescript Migration: gate

markxiong0122:
migrated
chromedash-gate-chip
chromedash-gate-column and its test file


client-src/elements/chromedash-gate-column.ts
```diff
@@ -838,8 +883,7 @@ export class ChromedashGateColumn extends LitElement {
           <th>Review status</th>
         </tr>
         ${responses.map(v => this.renderVoteRow(v, canVote))}
-        ${othersPending.map(ae => this.renderPendingVote(ae, canVote))}
-        ${addVoteRow}
+        ${othersPending.map(ae => this.renderPendingVote(ae))} ${addVoteRow}
```
jrobbins, line 886: Nice. Catching that unused parameter is exactly the kind of fix that this Typescript conversion was supposed to catch.

jrobbins (approved):

PR #4061, opened 2024-07-05
Add new OT approvals field `ot_approval_buganizer_custom_field_id`

DanielRyanSmith:
This adds a new OT approvals field `ot_approval_buganizer_custom_field_id` that allows the user to supply the custom field ID that dictates whether an OT registration approval is approved or denied.

This is a field that was added in a bit later in the OT approvals work and has not been captured in the Chromestatus form. This change has come up with the advent of the automated OT creation process, and is needed to fully complete that work.


client-src/elements/chromedash-ot-creation-page.js
```diff
@@ -142,6 +142,13 @@ export class ChromedashOTCreationPage extends LitElement {
         stageId: this.stage.id,
         isApprovalsField: true,
       },
+      {
+        name: 'ot_approval_buganizer_custom_field_id',
+        touched: true,
+        value: '',
```
KyleJu, line 148: Is this the default value? Should this be 0 instead?

KyleJu (approved):
LGTM with a comment

client-src/elements/chromedash-ot-creation-page.js
DanielRyanSmith: This is the default value in the input form, which will always be cast to a string. An empty string here denotes an empty initial field value

PR #4068, opened 2024-07-08
Typescript Migration: feature row & table

markxiong0122:
migrated chromedash-feature-row and chromedash-feature-row

### DIFF
--- a/client-src/elements/chromedash-feature-row.ts
+++ b/client-src/elements/chromedash-feature-row.ts
@@ -1,34 +1,36 @@
 import {LitElement, css, html, nothing} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
 import {SHARED_STYLES} from '../css/shared-css.js';
+import {Feature, StageDict} from '../js-src/cs-client.js';
 import {
+  GATE_ACTIVE_REVIEW_STATES,
   GATE_TEAM_ORDER,
   STAGE_SHORT_NAMES,
-  GATE_ACTIVE_REVIEW_STATES,
 } from './form-field-enums';
 
-class ChromedashFeatureRow extends LitElement {
-  static get properties() {
-    return {
-      feature: {type: Object},
-      columns: {type: String},
-      signedIn: {type: Boolean},
-      canEdit: {type: Boolean},
-      starredFeatures: {type: Object},
-      approvals: {type: Object},
-      gates: {type: Object},
-      selectedGateId: {type: Number},
-    };
-  }
+interface ActiveStagesAndGates {
+  stage: StageDict;
+  gates: any[]; // TODO(markxiong0122): Add Gate type when PR#4060 is merged.
+}
 
-  constructor() {
-    super();
-    this.starredFeatures = new Set();
-    this.feature = null;
-    this.canEdit = false;
-    this.approvals = {};
-    this.gates = {};
-    this.selectedGateId = 0;
-  }
+@customElement('chromedash-feature-row')
+class ChromedashFeatureRow extends LitElement {
+  @property({type: Object})
+  feature!: Feature;
+  @property({type: String})
+  columns;
+  @property({type: Boolean})
+  canEdit = false;
+  @property({type: Boolean})
+  signedIn;
+  @property({type: Object})
+  starredFeatures = new Set<number>();
+  @property({type: Object})
+  approvals; // TODO(markxiong0122): define approvals type
+  @property({type: Object})
+  gates;
+  @property({type: Number})
+  selectedGateId = 0;
 
   static get styles() {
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-feature-row.ts
```diff
@@ -1,34 +1,36 @@
...
+  @property({type: String})
+  columns;
+  @property({type: Boolean})
+  canEdit = false;
+  @property({type: Boolean})
+  signedIn;
+  @property({type: Object})
+  starredFeatures = new Set<number>();
+  @property({type: Object})
+  approvals; // TODO(markxiong0122): define approvals type
```
jrobbins, line 29: I think this field can be deleted.  It is left over from an earlier implementation of reviews.  Approval objects basically got replaced with Gate objects.

jrobbins (approved):
Seems pretty straightforward.  LGTM after making one suggested change.

client-src/elements/chromedash-feature-table.ts
```diff
@@ -1,52 +1,49 @@
...
-    this.start = 0;
-    this.num = 100;
-    this.alwaysOfferPagination = false;
-    this.noResultsMessage = 'No results';
-    this.canEdit = false;
-    this.gates = {};
-    this.selectedGateId = 0;
-  }
+  @state()
+  query = '';
```
jyasskin, line 12: This can't be `@state` because it's set as an attribute at https://github.com/GoogleChrome/chromium-dashboard/blob/6e446f44ce481954865849d14c6f5cf06b222216/client-src/elements/chromedash-myfeatures-page.js#L86-L87

Please double-check the [3 uses of this element](https://github.com/search?q=repo%3AGoogleChrome%2Fchromium-dashboard%20%22%3Cchromedash-feature-table%22&type=code) to make sure none of them are setting fields that you've marked as `@state`. If they're setting it as a `.property` (especially objects and arrays that can't be safely passed as attributes), you can use `@property({attribute:false})`.

client-src/elements/chromedash-feature-table.ts
```diff
@@ -1,52 +1,49 @@
...
+  @property({type: Boolean})
+  canEdit = false;
+  @property({type: Object})
+  gates: any; // TODO(markxiong0122): Add Gate type when PR#4060 is merged.
+  @property({type: Number})
+  selectedGateId = 0;
+  @property({type: String})
+  columns;
+  @property({type: Boolean})
+  signedIn;
```
jyasskin, line 46: Ideally, these would include `: boolean` or `: string` so that uses get type-checked. (`columns` might even be `"normal"|"approvals"` rather than `string`, but I haven't checked how it's used.)

jyasskin (changes requested):

client-src/elements/chromedash-feature-row.ts
```diff
@@ -1,34 +1,34 @@
...
+  feature!: Feature;
+  @property({type: String})
+  columns;
+  @property({type: Boolean})
+  canEdit = false;
+  @property({type: Boolean})
+  signedIn;
+  @property({type: Object, attribute: false})
+  starredFeatures = new Set<number>();
+  @property({type: Object, attribute: false})
```
markxiong0122, line 28: should this be an array of GateDict?

jrobbins: Yes, that would be good.

jrobbins (approved):

client-src/elements/chromedash-feature-table.ts
```diff
@@ -88,7 +86,7 @@ class ChromedashFeatureTable extends LitElement {
     window.csClient
       .getPendingGates()
       .then(res => {
```
jyasskin, line 88: To find the right types for internal values, start by adding a type to the result of the API call. I think this one is

```suggestion
      .then((res: {gates: GateDict[]}) => {
```
But please double-check that.

client-src/elements/chromedash-feature-row.ts
```diff
@@ -1,34 +1,35 @@
...
+@customElement('chromedash-feature-row')
+class ChromedashFeatureRow extends LitElement {
+  @property({type: Object, attribute: false})
+  feature!: Feature;
+  @property({type: String})
+  columns;
+  @property({type: Boolean})
+  canEdit = false;
+  @property({type: Boolean})
+  signedIn;
```
jyasskin, line 26: Remember to add types and initializers to fields. I think these are:
```suggestion
  @property({type: String})
  columns: undefined | 'approvals';
  @property({type: Boolean})
  canEdit = false;
  @property({type: Boolean})
  signedIn = false;
```

PR #4069, opened 2024-07-08
Typescript Migration: header

markxiong0122:


client-src/elements/chromedash-header.ts
```diff
@@ -243,13 +239,13 @@ export class ChromedashHeader extends LitElement {
...
           if (!response.ok) {
-            throw new Error('Sign in failed! Response:', response);
+            throw new Error(`Sign in failed! Response: ${response}`);
           }
         })
         .then(() => {
           setTimeout(() => {
             const url = window.location.href.split('?')[0];
-            window.location = url;
+            window.location = url as string & Location;
```
jrobbins, line 248: Why not `window.location.href = url;`?

client-src/elements/chromedash-header.ts
```diff
@@ -274,12 +270,12 @@ export class ChromedashHeader extends LitElement {
...
         setTimeout(() => {
           const url = window.location.href.split('?')[0];
-          window.location = url;
+          window.location = url as string & Location;
         }, 1000);
       })
       .catch(() => {
         console.error('Sign in failed, so signing out to allow retry');
-        signOut();
+        this.signOut();
```
jrobbins, line 278: Good fix.

client-src/elements/chromedash-header_test.ts
```diff
@@ -32,9 +32,9 @@ describe('chromedash-header', () => {
 
     // invalid feature requests would trigger the toast to show message
     const toastEl = document.querySelector('chromedash-toast');
-    const toastMsgSpan = toastEl.shadowRoot.querySelector('span#msg');
+    const toastMsgSpan = toastEl?.shadowRoot?.querySelector('span#msg');
```
jrobbins, line 35: Should this one reference `renderRoot` as well?  
What's the rule for choosing shadowRoot vs renderRoot?

markxiong0122: choosing renderRoot over shadowRoot is because if we assert instance it'll automatically figure out that component.renderRoot isn't null, so we dont have to add `!` or `?`. However, for this one we're accesing shadowDOM of an element that's outside of the class so renderRoot is not applicable

DanielRyanSmith (approved):

PR #4072, opened 2024-07-08
Reorganize Explainers help text, and say that explainers shouldn't be in Google Docs.

jyasskin:
Based on a discussion with @chrishtr about how to avoid some explainer faux pas.

image

image

The existing text about I2P timing was redundant with the warning in the I2P dialog that 
> Before you Draft Intent to Prototype email, it is strongly recommended that you do the following:
> * Start incubating: Explainer"


client-src/elements/form-field-specs.ts
```diff
@@ -690,11 +690,23 @@ export const ALL_FIELDS: Record<string, Field> = {
...
-    extra_help: html` <p>
+    help_text: html`Link to explainer(s) (one URL per line). See the
+      <a
+        target="_blank"
+        href="https://www.chromium.org/blink/launching-features/#start-incubating"
+        >launch process</a
+      >
+      for detailed advice, or expand the extra help here for a summary.`,
+    extra_help: html`<p>
+        Host your explainer somewhere like Github that makes it easy to file
```
chrishtr, line 701:
```suggestion
        Host your explainer somewhere like Github (and *not* Google Docs) that makes it easy to file
```

PR #4084, opened 2024-07-09
Typescript Migration: link

markxiong0122:
migrated chromedash-link and its test file


client-src/elements/chromedash-link.ts
```diff
@@ -117,7 +120,7 @@ function enhanceChromeStatusLink(featureLink, text) {
   </a>`;
 }
 
-function enhanceGithubIssueLink(featureLink, text) {
+function enhanceGithubIssueLink(featureLink, text?) {
```
DanielRyanSmith, line 123: This text param doesn't seem to be used in the code, and the all invocations of this function are called with 1 arg. An unrelated issue here, but maybe something worth fixing in a separate PR.

DanielRyanSmith (approved):

PR #4085, opened 2024-07-09
Typescript Migration: metadata

markxiong0122:
chromedash-metadata


client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,33 @@
...
+    REMOVED: 7,
+    ORIGINTRIAL: 8,
+    INTERVENTION: 9,
+    ON_HOLD: 10,
+    NO_LONGER_PURSUING: 1000,
+  };
+
+  @state()
+  selected;
+  _className;
```
jyasskin, line 27: Did you mean to include `@state()` for each of these members? I think at least some of them do need to be reactive.

It'd also be good to give each one a `: type` if you can find an appropriate one. `_className` might be as precise as `: undefined | 'canaryisdev' | 'betaisdev'`.

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,33 @@
...
+  _fetchError;
+  _channels;
```
jyasskin, line 29: I think at least `_channels` needs to keep its initializer.

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,33 @@
...
+    DEPRECATED: 6,
+    REMOVED: 7,
+    ORIGINTRIAL: 8,
+    INTERVENTION: 9,
+    ON_HOLD: 10,
+    NO_LONGER_PURSUING: 1000,
+  };
+
+  @state()
+  selected;
```
jyasskin, line 26: This is directly modified at https://github.com/GoogleChrome/chromium-dashboard/blob/c975330f24e750a4adc96de955a50be8f8c11a23/client-src/js-src/features-page.js#L54 so it needs to stay a `@property({attribute:false})`.

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,33 @@
 import {LitElement, css, html} from 'lit';
+import {customElement, property, state} from 'lit/decorators.js';
 import {ifDefined} from 'lit/directives/if-defined.js';
 import {SHARED_STYLES} from '../css/shared-css.js';
 
+@customElement('chromedash-metadata')
 class ChromedashMetadata extends LitElement {
```
jyasskin, line 7: It looks like this element is fetched from other files using `document.querySelector('chromedash-metadata')`, so check out https://lit.dev/docs/components/defining/#typescript-typings for how to give those calls the right types.

client-src/elements/chromedash-metadata.ts
markxiong0122: oh yeah i apologize should've checked before sending PR

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,42 @@
...
+  @property({type: String, attribute: false})
+  selected!: string;
+  @state()
+  _className: 'canaryisdev' | 'betaisdev' | undefined;
+  @state()
+  _fetchError!: boolean;
+  @state()
+  _channels: Channels = {};
+  @state()
+  _versions!: (string | number)[];
```
markxiong0122, line 39: @jyasskin i'm not sure if this is correct but based on the assignment below it seems to be correct. Could you help double check this? thanks

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,42 @@
...
-      implStatuses: {type: Array}, // Read in chromedash-featurelist.
-      status: {attribute: false}, // Read in chromedash-featurelist.
-      selected: {attribute: false}, // Directly edited in /templates/features.html
-      _className: {attribute: false},
-      _fetchError: {attribute: false},
-      _channels: {attribute: false},
-      _versions: {attribute: false},
-    };
-  }
+interface InplStatus {
```
DanielRyanSmith, line 7:
```suggestion
interface ImplStatus {
```

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,42 @@
...
-      ON_HOLD: 10,
-      NO_LONGER_PURSUING: 1000,
-    };
-    this._channels = {};
-    this._versions = [];
-  }
+@customElement('chromedash-metadata')
+class ChromedashMetadata extends LitElement {
+  @property({type: Array})
+  implStatuses: InplStatus[] = [];
```
DanielRyanSmith, line 15:
```suggestion
  implStatuses: ImplStatus[] = [];
```

DanielRyanSmith (changes requested):

DanielRyanSmith (approved):
Adjusting my review status, since my only suggestions were nits here

client-src/elements/chromedash-metadata.ts
markxiong0122: good catch, thanks!

client-src/elements/chromedash-metadata.ts
```diff
@@ -1,39 +1,42 @@
...
+    NO_LONGER_PURSUING: 1000,
+  };
+  @property({type: String, attribute: false})
+  selected!: string;
+  @state()
+  _className: 'canaryisdev' | 'betaisdev' | undefined;
+  @state()
+  _fetchError!: boolean;
+  @state()
+  _channels: Channels = {};
```
jyasskin, line 37: I think the right type here is 
```suggestion
  _channels: {
    canary?: number,
    dev?: number,
    beta?: number,
    stable?: number,
  } = {};
```

It also looks like it's never re-assigned, so it may as well be a plain member and not `@state()`, since it'll never cause a re-render.

client-src/elements/chromedash-metadata.ts
```diff
@@ -85,15 +88,17 @@ class ChromedashMetadata extends LitElement {
...
+    ).selectedIndex = index; // log null
   }
 
   _processResponse(response) {
     // TODO(ericbidelman): Share this data across instances.
     const windowsVersions = response[0];
     for (let i = 0, el; (el = windowsVersions.versions[i]); ++i) {
       // Include previous version if current is foobar'd.
-      this._channels[el.channel] =
+      this._channels[el.channel].version =
```
jyasskin, line 101: Nothing assigns `this._channels[el.channel]` to be an object, so I'd expect this to throw when it's executed.

client-src/elements/chromedash-metadata.ts
markxiong0122: should we initialize them with 0s as they are used in calculating the number of interactions below

jyasskin: I would leave the default `undefined` to avoid changing the code's behavior. It'll put some `undefined` values into `_versions`, which I haven't traced through to see what they'll do, but unless you do trace it through, I don't think it's a good idea to change it.

markxiong0122: `    this._versions = [
      this.implStatuses[this.status.PROPOSED - 1].val,
      this.implStatuses[this.status.IN_DEVELOPMENT - 1].val,
      this.implStatuses[this.status.DEPRECATED - 1].val,
      this._channels.canary.version,
      this._channels.dev.version,
      this._channels.beta.version,
      this._channels.stable.version,
    ];` and `    for (let i = this._channels.canary.version - 1; i >= 1; i--) ` if we just leave it undefined it won't pass the test or even just start it as there're a lot of errors of `Object is possibly 'undefined'.`

jyasskin: @DanielRyanSmith might know the code paths better, but I didn't see a reason to think that 0 would have better behavior than undefined. Omitting the value might, but you'd need to trace where the values go to be sure of that.

You can add undefined to the type of _versions, and put `if (this._channels.canary !== undefined)` around the for loop. That's the current behavior since undefined-1 is NaN, which isn't >=1.

PR #4088, opened 2024-07-10
Typescript Migration: userlist

markxiong0122:
chromedash-userlist


client-src/elements/chromedash-userlist.ts
```diff
@@ -1,17 +1,13 @@
...
-      users: {attribute: false},
-    };
-  }
-
-  constructor() {
-    super();
-    this.users = [];
-  }
+  @state()
+  users!: User[];
```
jyasskin, line 10: This needs to be initialized for the uses inside this file to work:
```suggestion
  users: User[] = [];
```

client-src/elements/chromedash-userlist.ts
```diff
@@ -75,14 +71,17 @@ class ChromedashUserlist extends LitElement {
   }
 
   _onAdminToggle() {
-    const formEl = this.shadowRoot.querySelector('form');
-    const adminCheckbox = formEl.querySelector('input[name="is_admin"]');
-    const siteEditorCheckbox = formEl.querySelector(
+    const formEl: HTMLFormElement | null =
```
jyasskin, line 74: I believe you don't need the explicit type here. `querySelector` has a smart enough type to get it right:
```suggestion
    const formEl =
```

client-src/elements/chromedash-userlist.ts
```diff
@@ -75,14 +71,17 @@ class ChromedashUserlist extends LitElement {
...
+      this.renderRoot.querySelector('form');
+    const adminCheckbox = formEl?.querySelector(
+      'input[name="is_admin"]'
+    ) as SlCheckbox;
```
jyasskin, line 78: Are you sure `SlCheckbox` is the right type for an `<input>` element?

client-src/elements/chromedash-userlist.ts
```diff
@@ -75,14 +71,17 @@ class ChromedashUserlist extends LitElement {
...
+    const siteEditorCheckbox = formEl?.querySelector(
       'input[name="is_site_editor"]'
-    );
+    ) as SlCheckbox;
     // Admins will always be site editors, so if the admin box is checked,
     // the site editor box is also checked and disabled.
-    if (adminCheckbox.checked) {
+    if (adminCheckbox?.checked) {
```
jyasskin, line 84: I'd use `querySelector(...)!` instead of `?.checked` to deal with the nullability here, because the target element is unconditionally in the rendered HTML.

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +92,31 @@ class ChromedashUserlist extends LitElement {
...
     e.preventDefault();
-    const formEl = this.shadowRoot.querySelector('form');
+    const formEl: HTMLFormElement | null =
+      this.renderRoot.querySelector('form');
 
-    if (formEl.checkValidity()) {
-      const email = formEl.querySelector('input[name="email"]').value;
-      const isAdmin = formEl.querySelector('input[name="is_admin"]').checked;
-      const isSiteEditor = formEl.querySelector(
+    if (formEl && formEl.checkValidity()) {
```
jyasskin, line 98: Instead of `x && x.y()`, use `x?.y()`. Except here, because the form is unconditionally in the rendered HTML, I'd just make the line above `...querySelector('form')!`.

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +92,31 @@ class ChromedashUserlist extends LitElement {
...
+      const emailInput: HTMLInputElement | null = formEl.querySelector(
+        'input[name="email"]'
+      );
+      const adminCheckbox: SlCheckbox | null = formEl.querySelector(
+        'input[name="is_admin"]'
+      );
+      const siteEditorCheckbox: SlCheckbox | null = formEl.querySelector(
```
jyasskin, line 105: Check your types here too.

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +92,31 @@ class ChromedashUserlist extends LitElement {
...
+    const formEl: HTMLFormElement | null =
+      this.renderRoot.querySelector('form');
 
-    if (formEl.checkValidity()) {
-      const email = formEl.querySelector('input[name="email"]').value;
-      const isAdmin = formEl.querySelector('input[name="is_admin"]').checked;
-      const isSiteEditor = formEl.querySelector(
+    if (formEl && formEl.checkValidity()) {
+      const emailInput: HTMLInputElement | null = formEl.querySelector(
+        'input[name="email"]'
```
jyasskin, line 100: `querySelector()`'s typing has an overload that lets you treat it as any type, even if the selector can't possibly return an element of that type. As a matter of taste, I'd use `as HTMLInputElement` in that case, to emphasize to the reader that you're asserting a type, not writing a type that the compiler will check.

client-src/elements/chromedash-userlist.ts
markxiong0122: it seems like SlCheckbox is used throughout similar use cases and also here in https://github.com/GoogleChrome/chromium-dashboard/blob/3f8901bc4cf7cb2dd841066f120389ed6ef76e8e/client-src/elements/chromedash-form-field.ts#L264

jyasskin: Yeah, when you see `<sl-checkbox>`, it'll have the `SlCheckbox` type, but when you see `<input>` it'll be `HTMLInputElement`.

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +91,30 @@ class ChromedashUserlist extends LitElement {
...
     if (formEl.checkValidity()) {
-      const email = formEl.querySelector('input[name="email"]').value;
-      const isAdmin = formEl.querySelector('input[name="is_admin"]').checked;
-      const isSiteEditor = formEl.querySelector(
+      const emailInput = formEl.querySelector(
+        'input[name="email"]'
+      ) as HTMLInputElement;
+      const adminCheckbox = formEl.querySelector(
+        'input[name="is_admin"]'
+      ) as SlCheckbox;
```
jyasskin, line 102:
```suggestion
      ) as HTMLInputElement;
```

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +91,30 @@ class ChromedashUserlist extends LitElement {
...
+      const siteEditorCheckbox = formEl.querySelector(
         'input[name="is_site_editor"]'
-      ).checked;
+      ) as SlCheckbox;
```
jyasskin, line 105:
```suggestion
      ) as HTMLInputElement;
```

client-src/elements/chromedash-userlist.ts
```diff
@@ -93,21 +91,30 @@ class ChromedashUserlist extends LitElement {
...
+      const email = emailInput?.value;
+      const isAdmin = adminCheckbox?.checked;
+      const isSiteEditor = siteEditorCheckbox?.checked;
       window.csClient.createAccount(email, isAdmin, isSiteEditor).then(json => {
         if (json.error_message) {
           alert(json.error_message);
         } else {
           this.addUser(json);
           formEl.reset();
-          formEl.querySelector('input[name="is_site_editor"]').disabled = false;
```
jyasskin, line 110: Don't move these queries out of a callback, because the DOM can change between the call to `createAccount` and the execution of the stuff in `.then()`. It _might_ not in this case, depending on what's in `render()`, but it's safest to just keep the query in its original location.

client-src/elements/chromedash-userlist.ts
jrobbins: This one should be `as HTMLInputElement`.

You can see on line 151 that this element is just an old fashioned `<input>` element.   I believe that this form was created before we started using Shoelace and we never upgraded it to use Shoelace.

jrobbins (changes requested):

client-src/elements/chromedash-userlist.ts
```diff
@@ -75,11 +71,13 @@ class ChromedashUserlist extends LitElement {
...
-    const adminCheckbox = formEl.querySelector('input[name="is_admin"]');
-    const siteEditorCheckbox = formEl.querySelector(
+    const formEl = this.renderRoot.querySelector('form');
+    const adminCheckbox = formEl?.querySelector(
+      'input[name="is_admin"]'
+    ) as SlCheckbox;
+    const siteEditorCheckbox = formEl?.querySelector(
       'input[name="is_site_editor"]'
-    );
+    ) as SlCheckbox;
```
jrobbins, line 80: This one should also be `as HTMLInputElement`.

jrobbins (changes requested):

jrobbins (approved):

PR #4089, opened 2024-07-10
Typescript Migration: feature detail

markxiong0122:
migrated chromedash-feature-detail

### DIFF
--- a/client-src/elements/chromedash-feature-detail.ts
+++ b/client-src/elements/chromedash-feature-detail.ts
@@ -533,29 +533,6 @@ class ChromedashFeatureDetail extends LitElement {
     return gatesForStage.map(g => this.renderGateChip(feStage, g));
   }
 
-  // Create an extension stage for an origin trial stage on button click.
-  createExtensionStage(feStage: StageDict, extensionAlreadyExists) {
-    if (feStage.stage_type in OT_EXTENSION_STAGE_MAPPING) {
-      return;
-    }
-    const confirmText =
-      'Trial extension fields are already available for this trial. ' +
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-feature-detail.ts
```diff
@@ -24,24 +22,30 @@ import {
   STAGE_TYPES_ORIGIN_TRIAL,
   VOTE_OPTIONS,
 } from './form-field-enums';
+import {makeDisplaySpecs} from './form-field-specs';
+import {getFieldValueFromFeature, hasFieldValue, isDefinedValue} from './utils';
 
+import '@polymer/iron-icon';
+import {property, state} from 'lit/decorators.js';
+import {SHARED_STYLES} from '../css/shared-css.js';
```
jyasskin, line 30: Run 'Organize Imports' again.

client-src/elements/chromedash-feature-detail.ts
```diff
@@ -71,37 +75,33 @@ export const DETAILS_STYLES = [
...
-    this.progress = {};
-    this.anyCollapsed = true;
-    this.previousStageTypeRendered = 0;
-    this.sameTypeRendered = 0;
-    this.selectedGateId = 0;
-    this.openStage = 0;
-  }
+  @property({type: String})
+  appTitle = '';
+  @property({type: Array, attribute: false})
```
jyasskin, line 80: As far as I can tell, the `type` field is only used when an attribute is defined (to convert string attribute values into typed property values), so we should generally have either `attribute:false` or `type: Something` but not both.

jyasskin (approved):

PR #4095, opened 2024-07-10
Typescript Migration: stack rank

markxiong0122:
migrated chromedash-stack-rank and it's page & test file


client-src/elements/chromedash-stack-rank-page.ts
```diff
@@ -34,22 +35,14 @@ export class ChromedashStackRankPage extends LitElement {
...
-    this.tempList = [];
-  }
+  @property({type: String, attribute: false})
+  type!: 'css' | 'feature';
+  @property({type: String, attribute: false})
+  view!: 'popularity' | 'animated';
+  @property({type: Array, attribute: false})
+  viewList: any[] = [];
+  @property({type: Array, attribute: false})
+  tempList: any[] = [];
```
DanielRyanSmith, line 45: Per [@jyasskin's comment](https://github.com/GoogleChrome/chromium-dashboard/pull/4089#discussion_r1672824243), this should likely be changed to having 1 of either `type` or `attribute`.

client-src/elements/chromedash-stack-rank.ts
```diff
@@ -1,150 +1,147 @@
...
+          visibility: visible;
         }
 
         .stack-rank-item-result {
-          margin-right: 5px;
+          display: flex;
+          align-items: center;
+          gap: 15px;
+          margin-right: 15px;
+        }
```
DanielRyanSmith, line 106: Was there a consolidation of CSS here? It looks like there might be some small behavior changes based on the rearranging (e.g. `margin-right` here was probably set at 5px originally but is now 15px).

markxiong0122: i dont think i've ever chanegd styles(). I think that difference was bc there's another `.stack-rank-item-result {
          margin-right: 5px;` at the very bottom https://github.com/GoogleChrome/chromium-dashboard/blob/362e6710044624c417cf782b5c2ec4dd120c361d/client-src/elements/chromedash-stack-rank.js#L144C7-L144C34 and one in the middle, maybe it was a mistake by the changes detection code of GitHub?

markxiong0122: i double checked that styles was not modified

DanielRyanSmith (approved):

PR #4103, opened 2024-07-11
Update stage edit path to not require intent stage

DanielRyanSmith:
This change fixes an apparent oversight in which a possible path for the stage edit page was using the "intentStage" argument instead of the "stageId" argument. The current path `/guide/stage/{featureId}/{intentStage}` is not used anywhere on the site, and navigating to this URL will show a mostly blank page with an error toast message. The error is caused because no stage ID is available to obtain relevant stage data to populate the fields.

This change should have no effect on current site behavior.

The original plan for this path is what is being fixed. The path is now `/guide/stage/{featureId}/{stageId}`. The stage edit page does not use the intentStage arg for anything relevant to the page. This will now allow some of our redirect URLs to no longer pass in a (mostly irrelevant) intent stage variable. 

For reference, the reason there is even a URL that adds the intent stage argument for the stage edit page is a relic of the schema migration. Previously, the fields that were shown on the page were specific to this intent stage arg, but this is no longer the case.

### DIFF
--- a/client-src/elements/chromedash-app.js
+++ b/client-src/elements/chromedash-app.js
@@ -429,10 +429,10 @@ class ChromedashApp extends LitElement {
       this.pageComponent.featureId = parseInt(ctx.params.featureId);
       this.pageComponent.appTitle = this.appTitle;
     });
-    page('/guide/stage/:featureId(\\d+)/:intentStage(\\d+)', ctx => {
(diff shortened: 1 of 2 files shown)


DanielRyanSmith:
It looks like there are unrelated failures due to a Chromium outage

jrobbins (approved):

PR #4106, opened 2024-07-11
Clarify the names and descriptions of the feature types

jyasskin:
... to try to direct developers to the right one more often.

Changes 
image
to
image

The [discussion on blink-api-owners-discuss](https://groups.google.com/a/chromium.org/g/blink-api-owners-discuss/c/uxVqqPJHENo) was generally positive, with some future improvements identified.

This does not change the 
image
description in features' metadata sections because that string comes from https://github.com/GoogleChrome/chromium-dashboard/blob/7f37bd7c9df24d3861cb694abe85d6bc8c6cb2e5/internals/processes.py#L410-L414 instead of the JS form definition.


DanielRyanSmith (approved):
I think a change like this, while not a complete solution, will go far in avoiding wrong choices for feature type in the future.

jyasskin:
The [discussion on blink-api-owners-discuss](https://groups.google.com/a/chromium.org/g/blink-api-owners-discuss/c/uxVqqPJHENo) suggested a couple changes, which I've adopted. The "New or changed feature" vs "catches up" descriptions could do more to explain exactly what's different between those options, but I don't have time to figure that out before going on vacation, so I'm going to merge this for the incremental improvement, and come back to it in another PR.

PR #4107, opened 2024-07-12
Typescript Migration: guide-editall-page

markxiong0122:


client-src/elements/chromedash-guide-editall-page.ts
```diff
@@ -280,7 +285,7 @@ export class ChromedashGuideEditallPage extends LitElement {
         <chromedash-form-field
           name=${field}
           index=${index}
-          stageId=${stageId}
+          stageId=${ifDefined(ifDefined(stageId))}
```
DanielRyanSmith, line 288: Do we need the nested `ifDefined()` invocation here?

DanielRyanSmith (approved):

client-src/elements/chromedash-guide-editall-page.ts
markxiong0122: seems like we dont need it as i tried it out. It's nested because my editor showed an error of this potentially being "null" string. I'll remove them

jrobbins (approved):
It looks like you addressed Jeffrey's comment.  And, this looks good to me.

PR #4111, opened 2024-07-12
Typescript Migration: guide verify accuracy page

markxiong0122:
chromedash-guide-verify-accuracy-page and test file


client-src/elements/chromedash-guide-verify-accuracy-page_test.ts
```diff
@@ -46,21 +46,6 @@ describe('chromedash-guide-verify-accuracy-page', () => {
...
-          milestone_str: 'No active development',
-          text: 'No active development',
-          val: 1,
-        },
-      },
-      ff: {view: {text: 'No signal', val: 5}},
-      safari: {view: {text: 'No signal', val: 5}},
-      webdev: {view: {text: 'No signal', val: 4}},
-      other: {view: {}},
-    },
```
DanielRyanSmith, line 63: I think you're right here that this is not the correct format for the feature that is handled with the accuracy verification page

DanielRyanSmith (approved):

PR #4112, opened 2024-07-12
[1/3] Intent template adjustments

DanielRyanSmith:
This change updates the `intent_to_implement.html` template with small fixes and changes, as well as making the template contain only the body contents of a draft intent. This is to enable its use for creating intent drafts and using them for notification submissions, as well as displaying the draft intent contents on the client side SPA component (coming in subsequent PR).

**Functionality should be exactly the same after landing this change, with the exception of some fixes/adjustments to the intent contents.**

Separate table elements were merged into a single table due to the intent containing offset columns for milestone information. Some milestone fields were moved to list the "first" milestone before the "last" milestone.

### Before converting to a single table:
Screenshot 2024-07-12 at 4 53 46 PM

### After converting to a single table:
Screenshot 2024-07-12 at 4 54 43 PM


templates/blink/intent_to_implement.html
```diff
@@ -1,31 +1,3 @@
-<p>Email to</p>
```
jrobbins, line 1: Since we are deleting this, we just need to make sure that all 3 of your changes go out in the same deployment.  It should be OK if a user has a browser tab open on a previous version of the app because as soon as the click the button it will open a new tab that does a full page load and gets the latest version of the SPA.

templates/estimated-milestones-table.html
```diff
@@ -26,57 +26,41 @@
...
-
-  <table>
+    {% endif %}{% if stage.milestones.webview_last %}
+        <tr><td>Origin trial WebView last</td>
+        <td>{{stage.milestones.webview_last}}</td></tr>
 
+    {% endif %}{% endfor %}
     {% for stage in stage_info.ship_stages %}{% if stage.milestones.ios_first %}
-        <tr><td>Shipping on WebView</td>
+        <tr><td>Shipping on iOS</td>
```
jrobbins, line 63: Thanks for catching this mistake.

internals/reminders_test.py
```diff
@@ -174,7 +174,7 @@ def test_build_email_tasks_feature_accuracy__enterprise(self):
     self.assertEqual('feature_owner@example.com', task['to'])
     self.assertEqual('[Action requested] Update feature one', task['subject'])
     self.assertEqual(None, task['reply_to'])
-    TESTDATA.make_golden(task['html'], 'test_build_email_tasks_feature_accuracy_enterprise.html')
+    # TESTDATA.make_golden(task['html'], 'test_build_email_tasks_feature_accuracy_enterprise.html')
```
jrobbins, line 177: Thanks for catching this mistake.  That effectively disabled a test.

jrobbins (approved):

templates/blink/intent_to_implement.html
DanielRyanSmith: Actually, this PR should be available to deploy on its own, as this piece of the intent page has been moved to `launch.html`. The existing page should look exactly the same, but this `intent_to_implement.html` only contains the template for intent body contents, and no aspects of the surrounding intent draft preview page. 🙂

internals/reminders_test.py
DanielRyanSmith: I believe I was the last one to edit this, so naturally I'm fixing my own mistake. 😅

templates/blink/intent_to_implement.html
jrobbins: Ah, OK.

PR #4113, opened 2024-07-13
[2/3] New Intents API

DanielRyanSmith:
This change adds a new Intents API. It facilitates new endpoints to the Chromestatus API.

`features/{feature_id}/{stage_id}/intent`
(GET): Returns the contents on an intent body based on the given stage.
(POST): Posts an intent draft directly to blink-dev with a number of specified CC recipients.

Some types used for these new endpoints are generated by OpenAPI.


DanielRyanSmith:
@jcscottiii This was my first attempt using OpenAPI - maybe you could check if I made any silly errors? 😅

internals/notifier.py
```diff
@@ -864,6 +866,43 @@ def build_email(
...
 
 
+class IntentToBlinkDevHandler(basehandlers.FlaskHandler):
+  """Submit an intent email directly to blink-dev."""
+  IS_INTERNAL_HANDLER = True
+  EMAIL_TEMPLATE_PATH = 'blink/intent_to_implement.html'
+
+  def process_post_data(self, **kwargs):
+    self.require_task_header()
+    feature_id = self.get_param('feature_id', required=True)
```
jrobbins, line 876: Future work idea: Maybe we could add API entry points in a second openapi.yaml file for all our tasks.  We don't need to generate code or docs for them, but it would be useful to have typed parameter objects rather than plain python dicts.

internals/notifier.py
```diff
@@ -864,6 +866,43 @@ def build_email(
...
+    feature = FeatureEntry.get_by_id(feature_id)
+    if not feature:
+      self.abort(400, 'Feature not found.')
+    json_data = self.get_json_param_dict()
+    send_emails([self.build_email(feature, json_data)])
```
jrobbins, line 881: It would be useful to first call build_email, then log some details of it, then call send_emails.  That way we have some clues in the logs for debugging.

internals/notifier.py
```diff
@@ -1130,8 +1169,8 @@ def send_emails(email_tasks):
       try:
         cloud_tasks_helpers.enqueue_task(
             '/tasks/outbound-email', task)
-      except e:
-        logging.exception('could not enqueue.')
+      except Exception as e:
+        logging.exception(f'could not enqueue. {e}')
```
jrobbins, line 1173: Doesn't `logging.exception()` automatically add exception info?

openapi/api.yaml
```diff
@@ -146,6 +146,67 @@ paths:
...
+          in: path
+          description: Stage ID
+          required: true
+          schema:
+            type: integer
+      responses:
+        '200':
+          description: Intent draft body.
+          content:
+            text/plain:
```
jrobbins, line 170: Instead of `text/plain`, let's do something more like `MessageResponse` where there is a single string field, but we can more easily add other fields later.

jrobbins (changes requested):

openapi/api.yaml
```diff
@@ -189,6 +250,18 @@ components:
...
   schemas:
+    PostIntentRequest:
+      properties:
+        gate_id:
+          type: integer
+        intent_cc_emails:
+          type: array
+          items:
+            type: string
+    MessageResponse:
```
jcscottiii, line 261: MessageResponse seems very generic but I don't think this component will be used for other endpoints correct?

api/intents_api.py
```diff
@@ -0,0 +1,135 @@
...
+from framework import permissions
+from internals import processes
+from internals import stage_helpers
+from internals.core_enums import INTENT_STAGES_BY_STAGE_TYPE
+from internals.core_models import FeatureEntry, Stage
+from internals.review_models import Gate
+from pages.intentpreview import compute_subject_prefix
+
+
+class IntentOptions(TypedDict):
```
jcscottiii, line 32: Agree with the earlier comment that we should use openapi for the tasks API. It would prevent us from having to maintain this manually. Can you add a comment about the location of the task code that consumes these parameters?

openapi/api.yaml
DanielRyanSmith: Added a new `GetIntentResponse` definition with the `subject` and `email_body` fields.

internals/notifier.py
DanielRyanSmith: Ah, you're right! TIL 🙂 
However, VS Code was highlighting this as syntactically problematic.
Screenshot 2024-07-17 at 12 53 50 PM

This is not related to the actual changes in this PR however, and I'll revert things.

internals/notifier.py
DanielRyanSmith: I've added some logging that logs the subject and recipients. The body contents get logged before submission (at least on staging), and I thought it might be long-winded to print the intent body in the logs again separately.

api/intents_api.py
```diff
@@ -0,0 +1,135 @@
...
+    if stage is None:
+      self.abort(404, msg=f'Stage {stage_id} not found')
+
+    # Check that the user has feature edit permissions.
+    redirect_resp = permissions.validate_feature_edit_permission(
+        self, feature_id)
+    if redirect_resp:
+      return redirect_resp
+
+    body = PostIntentRequest(**self.request.get_json())
```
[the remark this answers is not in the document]
DanielRyanSmith: Agreed that it's ambiguous. I've gone with `parsed_args` here.

api/intents_api.py
```diff
@@ -0,0 +1,135 @@
...
+    if body.gate_id:
+      gate = Gate.get_by_id(body.gate_id)
+    if gate:
+      default_url += f'?gate={body.gate_id}'
+
+    subject = f'{compute_subject_prefix(feature, intent_stage)}: {feature.name}'
+    cc_emails = body.intent_cc_emails or []
+    # Make sure emails are not empty and are unique.
+    cc_emails = sorted(list(set([email for email in cc_emails if email])))
+    params: IntentOptions = {
```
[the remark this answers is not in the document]
DanielRyanSmith: Done!

api/intents_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+    return render_template('blink/intent_to_implement.html', **template_data)
+
+  def do_post(self, **kwargs) -> Response|dict|MessageResponse:
+    """Submit an intent email directly to blink-dev."""
+    feature_id = int(kwargs['feature_id'])
+    # Check that feature ID is valid.
+    if not feature_id:
+      self.abort(400, msg='No feature specified.')
+    feature: FeatureEntry|None = FeatureEntry.get_by_id(feature_id)
```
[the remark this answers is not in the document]
DanielRyanSmith: Great idea! The `get_specified_feature()` method only checks if the user has view access however, so that permission check will still need to be maintained in the intents method. I've added a `get_specified_stage()` helper now 🙂

api/intents_api.py
```diff
@@ -0,0 +1,135 @@
...
+    template_data = {
+      'feature': converters.feature_entry_to_json_verbose(feature),
+      'stage_info': stage_helpers.get_stage_info_for_templates(feature),
+      'should_render_mstone_table': stage_info['should_render_mstone_table'],
+      'should_render_intents': stage_info['should_render_intents'],
+      'intent_stage': intent_stage,
+      'default_url': default_url,
+    }
+
+    return render_template('blink/intent_to_implement.html', **template_data)
```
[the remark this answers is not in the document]
DanielRyanSmith: Agreed that this is a good idea. I've made the change to accommodate this

openapi/api.yaml
```diff
@@ -146,6 +146,67 @@ paths:
...
+        '200':
+          description: Intent draft body.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/MessageResponse'
+        '400':
+          description: No feature or stage ID specified.
+        '404':
+          description: Feature or stage not found based on given ID.
```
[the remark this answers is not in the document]
DanielRyanSmith: I see! Makes sense - I've updated the definition. 🙂

openapi/api.yaml
DanielRyanSmith: There are a number of endpoints that will likely use this response structure, e.g. [FeaturesAPI](https://github.com/GoogleChrome/chromium-dashboard/blob/main/api/features_api.py#L314-L341)

api/intents_api.py
DanielRyanSmith: Added a comment noting what this was being used for. dicts with varying properties are sent to  the `enqueue_task()` helper - I'd be curious to know what is the best way to document this in OpenAPI, since the props are so variable. Make an OpenAPI type definition for each possible dict structure passed to `enqueue_task()`?

internals/notifier.py
jrobbins: Or maybe just typedicts to gain type safety while avoiding anyone thinking that that second api.yaml is actually a public API.

internals/notifier.py
jrobbins: OK

internals/notifier.py
jrobbins: I think it could be fixed as `except Exception:`'

api/intents_api.py
jcscottiii: Could you give me an example of how it varies?

If it can be complex but they all share similar fields, you could use composition via `allOf` and `discriminator` More about it here: https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/ (Check the `allOf` section)

openapi/api.yaml
jcscottiii: Thanks for verifying that other endpoints will use it. Makes sense then.

openapi/api.yaml
```diff
@@ -189,6 +237,24 @@ components:
       in: header
       name: X-Xsrf-Token
   schemas:
+    GetIntentResponse:
+      properties:
```
jcscottiii, line 241: Could you take a look at each of the properties in these new schemas and check if any of them are required / will always be provided?

If so, you can mark them as required. This will help with the TypeScript.

Example

image

api/intents_api.py
DanielRyanSmith: Most of the invocations of enqueue_task() pass in a dict for `task_params` that have various key-value pairs that are unique fields for populating the email task, e.g. this [invocation in feature_links.py](https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/feature_links.py#L253-L257) uses completely different keys.

However, I see what you're saying on how we could handle this. Perhaps this is just a matter of making new schemas for each possible task_param structure and using `anyOf`. Like this example from the link:
```yaml
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              anyOf:
                - $ref: '#/components/schemas/PetByAge'
                - $ref: '#/components/schemas/PetByType'
```

openapi/api.yaml
DanielRyanSmith: Great idea! I marked fields "required" as needed. 🙂

api/intents_api.py
```diff
@@ -0,0 +1,137 @@
...
+    template_data = {
+      'feature': converters.feature_entry_to_json_verbose(feature),
+      'stage_info': stage_helpers.get_stage_info_for_templates(feature),
+      'should_render_mstone_table': stage_info['should_render_mstone_table'],
+      'should_render_intents': stage_info['should_render_intents'],
+      'intent_stage': intent_stage,
+      'default_url': default_url,
+    }
+
+    return {
```
jrobbins, line 80: I was thinking that instead of creating a dict, this would instantiate a `GetIntentResponse` object (therefore getting the type checking in the model class generated by OpenAPI), and then `return get_intent_response.to_dict()`.  

For an existing example in our code, see
https://github.com/GoogleChrome/chromium-dashboard/blob/main/api/external_reviews_api.py#L268

jrobbins (approved):

api/intents_api.py
DanielRyanSmith: Ah right, this is how I handled the POST request, but forgot to do so with the GET request. Good call! I've made the change. 🙂

jrobbins (approved):

jcscottiii (approved):

PR #4114, opened 2024-07-13
[3/3] New Intent SPA component and direct intent posting

DanielRyanSmith:
This change converts the old intent page to a SPA component and adds the functionality for posting intent drafts directly to blink-dev.


client-src/elements/chromedash-intent-content.js
```diff
@@ -0,0 +1,212 @@
...
+        }
+      `,
+    ];
+  }
+
+  setCopyEmailListener() {
+    const copyEmailBodyEl = this.shadowRoot.querySelector('#copy-email-body');
+    const emailBodyEl = this.shadowRoot.querySelector('.email');
+    if (copyEmailBodyEl && emailBodyEl) {
+      copyEmailBodyEl.addEventListener('click', () => {
```
jrobbins, line 143: Now that this is in lit, can we use `@click=...` in the `render()` method rather than using `addEventListener()`?  

It is OK to do this in a follow-up PR.  Also, this needs to be typescript.

client-src/elements/chromedash-intent-preview-page.js
```diff
@@ -0,0 +1,270 @@
...
+
+  renderFeatureUnlistedAlert() {
+    if (!this.displayFeatureUnlistedWarning) return nothing;
+    return html`<div class="alertbox">
+      Important: This feature is currently unlisted. Please only share feature
+      details with people who are collaborating with you on the feature.
+    </div>`;
+  }
+
+  computeSubjectPrefix() {
```
jrobbins, line 147: This could also be computed server-side to reduce duplicate code.

client-src/js-src/cs-client.js
```diff
@@ -646,6 +646,14 @@ export class ChromeStatusClient {
     return this.doGet(`/features/${featureId}/process`);
   }
 
+  // Intents API
```
jrobbins, line 649: Once you have OpenAPI endpoints defined (in your other PR), I think that you can call generated code rather than adding code to cs-client.

jrobbins (approved):
LGTM after considering some potential follow-up work.

client-src/js-src/cs-client.js
DanielRyanSmith: I think you're right, and this is something that I will implement in a subsequent PR once moving these files to TypeScript. 🙂

jrobbins (approved):

PR #4115, opened 2024-07-13
Typescript Migration: report external reviews

markxiong0122:
chromedash-report-external-reviews-page
chromedash-report-external-reviews-dispatch-page


client-src/elements/chromedash-report-external-reviews-page.ts
```diff
@@ -1,23 +1,23 @@
 // @ts-check
 import {Task} from '@lit/task';
 import '@shoelace-style/shoelace';
+import {
+  DefaultApiInterface,
+  OutstandingReview,
+  OutstandingReviewCurrentStageEnum as Stage,
```
jrobbins, line 7: `Stage` sounds like an object that respresents a stage.  In fact, we have such objects used in this app.  This is actually an enum, so call it something like `StageEnum`.

jrobbins (changes requested):

client-src/elements/chromedash-report-external-reviews-page.ts
markxiong0122: Good point. thanks!

DanielRyanSmith (approved):
This might need your review again @jrobbins

jrobbins (approved):

PR #4119, opened 2024-07-14
Typescript Migration: app

markxiong0122:
chromedash-app


client-src/elements/chromedash-app.ts
```diff
@@ -100,42 +104,36 @@ export class ChromedashApp extends LitElement {
...
+  @property({type: String})
+  bannerMessage = '';
+  @property({type: Number})
+  bannerTime: number | null = null;
+  @state()
+  loading = true;
+  @state()
+  currentPage = '';
+  @state()
+  pageComponent: any = null;
```
markxiong0122, line 126: couldn't find a reference to this so I left it as any. Any directions to type this?

DanielRyanSmith: I see:
https://github.com/GoogleChrome/chromium-dashboard/blob/fe67f68e4831f5a2d776ae86b42a24ae3bd21d04/client-src/elements/chromedash-app.js#L280-L282

and
https://github.com/GoogleChrome/chromium-dashboard/blob/fe67f68e4831f5a2d776ae86b42a24ae3bd21d04/client-src/elements/chromedash-app.js#L534-L536

Would this correlate to `HTMLElement`? As long as we can confirm that `componentName` is typed as a string as an argument for `setupNewPage()` (which should always be the case)

markxiong0122: but if we type it as `HTMLElement` it would give `does exist on` error as we have `this.pageComponent.unsavedChanges` and `this.pageComponent.contextLink` such cases in the code below. I thought it was the just `chromedash-app` itself but it seems like `unsavedChanges` doesn't exist on it and I couldn't find other components that have it either. Should I just extend `HTMLElement` with the properties shown in the code?

jrobbins: Let's stick with `any` and add a TODO comment in this PR. Then, do a follow-up PR to redesign this code to be type safe by defining a new base class `ChromeStatusPage` that declares the expected fields and changes all our page classes to subclass from it.

jrobbins (approved):

client-src/elements/chromedash-app.ts
DanielRyanSmith: I see - agreed with @jrobbins 🙂

DanielRyanSmith (approved):

Issue #4100, opened 2024-07-10
Automated origin trial creation UI enhancements


DanielRyanSmith:
Some additional UI enhancements are still needed for the automated OT creation process.

- [x] Redirect user away from the OT creation form and show OT creation request button as disabled if approvals have not been obtained (added in #3897).

The OT creation form should be closed after user submission. This includes:
  - [x] Disabling the button for navigating to the OT creation request form.
  - [x] Redirecting the user if the OT creation form is visited when a creation request is in progress.

PR #4121, opened 2024-07-15
Users cannot resubmit OT creation form after submission

DanielRyanSmith:
Addresses #4100 

This change disables the button to navigate to the OT creation form when the form has already been submitted, and navigating directly to the creation form after submission will result in a redirect. This is to avoid users from changing data  and the submissions status of the trial while creation is in progress, and will be required for automated OT creation process.

### DIFF
--- a/client-src/elements/chromedash-feature-detail.ts
+++ b/client-src/elements/chromedash-feature-detail.ts
@@ -17,7 +17,7 @@ import {
   FORMS_BY_STAGE_TYPE,
 } from './form-definition';
 import {
-  OT_EXTENSION_STAGE_MAPPING,
+  OT_SETUP_STATUS_OPTIONS,
   STAGE_SHORT_NAMES,
   STAGE_TYPES_ORIGIN_TRIAL,
   VOTE_OPTIONS,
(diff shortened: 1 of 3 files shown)


KyleJu (commented):
Code looks great! Would it be possible to have some screenshots so I can understand this change better?

DanielRyanSmith:
Screenshot 2024-07-17 at 11 25 53 AM

This change will disable the button that navigates to the OT creation request form if the user has already submitted the form once. When the user submits the form, the `ot_setup_status` for the stage is set to `OT_READY_FOR_CREATION`, which will flag the trial creation as ready to be processed. When the automated creation process is active, `ot_setup_status` will be set to `OT_CREATED`, `OT_ACTIVATED`,  or `OT_CREATION_FAILED`, etc. during the creation process.

With this change, we check `ot_setup_status` and make sure the trial is not in already in the process of being created, or flagged for creation. If it is, the user has already submitted the OT creation form, and so we disable navigation to the form.

KyleJu (approved):
Thanks for the screenshots!

PR #4122, opened 2024-07-15
Add setting for automated OT creation flag

DanielRyanSmith:
Addresses #4100 

This change adds a flag that controls whether the entire automated OT creation process is active. This will give an easy rollback option in case some issues arise with the process on the final deployment.

The cron jobs that handle automated trial creation will now run on their regular schedule, but will not perform any actions until the flag is set to run the automated process.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -37,7 +37,7 @@ cron:
   schedule: every day 6:00
 - description: Send origin trial process reminder emails.
   url: /cron/send-ot-process-reminders
-  schedule: every monday 10:00
+  schedule: every monday 4:00
 - description: Check if any origin trials require creation
   url: /cron/create_origin_trials
   schedule: every 5 minutes
--- a/cron.yaml
+++ b/cron.yaml
@@ -37,12 +37,10 @@ cron:
   schedule: every day 6:00
 - description: Send origin trial process reminder emails.
   url: /cron/send-ot-process-reminders
-  schedule: every monday 4:00
-
-# TODO(DanielRyanSmith): Add this job when OT creation is fully implemented.
-# - description: Check if any origin trials require creation
-#   url: /cron/create_origin_trials
-#   schedule: every 5 minutes
-# - description: Check if any origin trials require activation
-#   url: /cron/activate_origin_trials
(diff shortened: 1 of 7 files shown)


cron.yaml
```diff
@@ -38,11 +38,9 @@ cron:
...
-# TODO(DanielRyanSmith): Add this job when OT creation is fully implemented.
-# - description: Check if any origin trials require creation
-#   url: /cron/create_origin_trials
-#   schedule: every 5 minutes
-# - description: Check if any origin trials require activation
-#   url: /cron/activate_origin_trials
-#   schedule: every day 4:00
+- description: Check if any origin trials require creation
+  url: /cron/create_origin_trials
+  schedule: every 5 minutes
```
KyleJu, line 43: This is the most frequent job in this file at 12 runs/hour. Would you know the Cron job quotas

KyleJu (commented):
LGTM the code. Question about the PR description - how does this flag enable "an easy rollback option" in production?

cron.yaml
DanielRyanSmith: > LGTM the code. Question about the PR description - how does this flag enable "an easy rollback option" in production? Also maybe is this documented in the design doc? Might be useful for the poc on rotation

Great idea to document in the design doc - it's not currently there. This change makes it so that if something is wrong with the automated OT creation process, the flag can be unset and the old manual process will be used instead until the flag is switched back, so it's the only setting that needs to be changed to completely revert to the old process.

> This is the most frequent job in this file at 12 runs/hour. Would you know the Cron job quotas

This is a good question. Looking at [the docs](https://cloud.google.com/appengine/docs/flexible/scheduling-jobs-with-cron-yaml), it looks like we can have a total of 250 total cron jobs running, but also, this job just sends a request to the desired path of and handles the process as if it were an app engine request, so all of our [App Engine processing limits](https://cloud.google.com/appengine/docs/standard/quotas) apply here, which should be more than sufficient.

KyleJu: Good point about app engine requests!

KyleJu (approved):

Issue #4102, opened 2024-07-11
The "Feature type" field is user-hostile


mfreed7:
So the fact that the "feature type" field can never be changed is a fairly big deal. Three times (at least) now, I have used the wrong feature type, only to discover that later and have to recreate the entire entry, copy/paste by copy/paste.

I would recommend the following changes:
1. Make feature type the **very first thing** you have to enter, before you're allowed to enter anything else. That includes the title of the entry. That should help notify folks that this step matters. Akin to when I created this bug - Github first asked me if I was reporting a bug or requesting a feature. There should be a **giant red banner** indicating that this selection can never be changed, so take some time and get it right.
2. List, explicitly and in great detail, the consequences of each choice. For example, which choices will result in an "Intent to Ship" which requires 3 LGTMs to ship, vs a "PSA" which does not. For example, I have no idea what material differences I will encounter when selecting "New feature incubation" vs "Existing feature implementation". Are there any? If so, what? If not, why are these two separate choices?
3. Rename "New feature incubation" and "Existing feature implementation" to better describe them. An "Existing feature implementation" is an oxymoron - if it already exists, why am I implementing it? I presume "existing" means "in a spec" somewhere, but that's not very clear.

As an illustrative example, I recently filed [this entry](https://chromestatus.com/feature/6567965055778816) as a "Web developer-facing change to existing code" because that was very precisely what it was. A developer-facing change to existing code. I did not realize that this option would not trigger the need for 3 LGTMs from the owners. I was therefore [forced to copy/paste](https://groups.google.com/a/chromium.org/g/blink-dev/c/GU9zSycWOP8/m/iigilH_VAwAJ) into [a new entry](https://chromestatus.com/feature/5142143019253760).

@yoavweiss

jyasskin:
https://github.com/GoogleChrome/chromium-dashboard/pull/4106 does (3) and part of (2). I'll try to send another PR for (1) on Monday.

DanielRyanSmith:
I believe these suggestions have been implemented for the most part. It seems like these changes have reduced the amount of wrong choices for the feature type field (although there is likely still progress to be made). Maybe we're safe to close this issue?

PR #4123, opened 2024-07-15
Move the 'feature type' field to the top of the 'new feature' page.

jyasskin:
@mfreed7 suggested this in #4102, and I think it makes sense.

image


jrobbins (approved):
This seems like an improvement.  Let's try it.

PR #4129, opened 2024-07-16
Typescript Migration: guide new page

markxiong0122:


client-src/elements/chromedash-guide-new-page_test.ts
```diff
@@ -54,7 +54,7 @@ describe('chromedash-guide-new-page', () => {
     assert.include(featureTypeFormField.outerHTML, 'Feature removal');
 
     // submit button exists
-    const submitButton = component.shadowRoot.querySelector(
+    const submitButton = component.renderRoot.querySelector(
       'input[type="submit"]'
     );
```
DanielRyanSmith, line 59: Is there a reason that `!` is needed in the previous querySelector invocations, but not the one for the submit button?

DanielRyanSmith (approved):

client-src/elements/chromedash-guide-new-page_test.ts
markxiong0122: it was bc we inserted this to exist so the editor knows it's not null while the previous ones do not have such insertions

PR #4137, opened 2024-07-18
Add `trial_contacts` arg to OT creation request

### LINKED ISSUE #3902: "OT Contacts" argument needs to be added to OT creation Chromestatus request
[The OT Creation request code](https://github.com/GoogleChrome/chromium-dashboard/blob/main/framework/origin_trials_client.py#L105-L154) in Chromestatus should send over the "OT Contacts” argument.
OT Contacts are the combined values of ot_owner_email (one email) and ot_emails (multiple emails) fields.
The ot_emails field should be filtered to remove any non-google.com emails, as they cannot be added to internal groups and ganpati groups. This is how we handle the manual process as well.

DanielRyanSmith:
Fixes #3902 

This change adds a new argument to the creation request sent to the OT API, which contains a list of "@google.com" contacts associated with an origin trial. These contacts will be used to create groups for data access purposes.

Additionally, this updates the `internal_registration_config` argument that was renamed in a recent CL.

### DIFF
--- a/framework/origin_trials_client.py
+++ b/framework/origin_trials_client.py
@@ -54,7 +54,8 @@ class InternalRegistrationConfig(TypedDict):
 
 class CreateOriginTrialRequest(TypedDict):
   trial: RequestTrial
-  internal_registration_config: InternalRegistrationConfig
+  registration_config: InternalRegistrationConfig
+  trial_contacts: list[str]
 
 
 def get_trials_list() -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


framework/origin_trials_client.py
```diff
@@ -163,20 +164,29 @@ def create_origin_trial(ot_stage: Stage) -> str | None:
...
+    'registration_config': {'approval_type': 'NONE'},
+    'trial_contacts': []
   }
+
+  # Get a list of all OT @google.com contacts (ot_owner_email must be a google
+  # contact).
+  unique_contacts = [ot_stage.ot_owner_email]
+  unique_contacts.extend(ot_stage.ot_emails)
+  unique_contacts = [email for email in set(unique_contacts)
+                     if email.endswith('@google.com')]
```
KyleJu, line 176: Group contacts need to be filtered as well, since they cannot be added to Ganpati or Google group APIs. But that is much harder to do here. We need to do it on the OT side

KyleJu (approved):

PR #4143, opened 2024-07-18
Avoid a 404 because there is no ot_requests.css.

jrobbins:
This reference was added in https://github.com/GoogleChrome/chromium-dashboard/pull/3352/files
but there was no file added for ot_requests.css.


DanielRyanSmith (approved):

DanielRyanSmith:
Looks like this deployment is helping us find more than just problems with recent changes. 😅 
Thanks for the fix!

PR #4145, opened 2024-07-18
Avoid errors accessing `this.feature` when request not finished.

DanielRyanSmith:
Not sure if this is from recent changes, but some locations with the `renderSubheader` method did not wait for relevant properties to be populated (namely, the feature to be displayed), which caused some log errors. The page would render correctly after the data was fetched, however. Regardless, waiting until the data has finished loading will fix this problem.

### DIFF
--- a/client-src/elements/chromedash-guide-metadata-page.ts
+++ b/client-src/elements/chromedash-guide-metadata-page.ts
@@ -153,15 +153,18 @@ export class ChromedashGuideMetadataPage extends LitElement {
   }
 
   renderSubheader() {
-    return html`
-      <div id="subheader">
-        <h2 id="breadcrumbs">
+    const link = this.loading ? nothing : html`
           <a href=${this.getNextPage()}>
-            <iron-icon icon="chromestatus:arrow-back"></iron-icon>
-            Edit feature: ${this.feature.name}
-          </a>
-        </h2>
-      </div>
+              <iron-icon icon="chromestatus:arrow-back"></iron-icon>
+              Edit feature: ${this.feature.name}
+           </a>
+         `;
+     return html`
+        <div id="subheader">
+          <h2 id="breadcrumbs">
+            ${link}
+          </h2>
+        </div>
     `;
   }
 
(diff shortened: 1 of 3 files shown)


jrobbins (changes requested):
Good idea, but I think this will lead to a bit of vertical shifting as the page loads that we can avoid pretty easily.  Instead of not calling renderSubheader(), call it, but put the conditional inside of it.  And, instead of returning `nothing`, return the same elements but without the text.  Something like:
```
renderSubheader() {
  const link = this.loading ? nothing : html`
        <a href=${this.getNextPage()}>
            <iron-icon icon="chromestatus:arrow-back"></iron-icon>
            Request origin trial extension: ${this.feature.name}
         </a>
       `;
   return html`
      <div id="subheader">
        <h2 id="breadcrumbs">
          ${link}
        </h2>
      </div>
  `;
}
```

client-src/elements/chromedash-guide-metadata-page.ts
```diff
@@ -230,7 +233,7 @@ export class ChromedashGuideMetadataPage extends LitElement {
 
   render() {
     return html`
-      ${this.renderSubheader()}
+      ${this.loading ? nothing : this.renderSubheader()}
```
jrobbins, line 236: This line should be reverted so that it always calls `renderSubheader()`.

client-src/elements/chromedash-ot-creation-page.ts
```diff
@@ -442,7 +445,7 @@ export class ChromedashOTCreationPage extends LitElement {
 
   render() {
     return html`
-      ${this.renderSubheader()}
+      ${this.loading ? nothing : this.renderSubheader()}
```
jrobbins, line 448: Same

client-src/elements/chromedash-ot-extension-page.ts
```diff
@@ -369,7 +372,7 @@ export class ChromedashOTExtensionPage extends LitElement {
 
   render() {
     return html`
-      ${this.renderSubheader()}
+      ${this.loading ? nothing : this.renderSubheader()}
```
jrobbins, line 375: Same.

jrobbins (changes requested):

client-src/elements/chromedash-guide-metadata-page.ts
DanielRyanSmith: Ah, sorry about that - I've removed these changes.

jrobbins (approved):

PR #4146, opened 2024-07-19
Fix showEnterprise on enterprise list page.

jrobbins:
The enterprise feature list page was just showing regular features rather than enterprise features.

The code path is a little complicated, but basically whenever a query is set for the chromedash-feature-table, and that query mentions `feature_type`, then `showEnterprise` should be set to true, and the query should be set. 

However, those properties had `attribute: false`, so the values passed to them via attributes was not being set.  That cause the enterprise feature list to show all features rather than sending any query string to the server, and it did not ask the server to include enterprise features.


DanielRyanSmith (approved):
Thanks for the investigation and fix!

PR #4165, opened 2024-07-23
OpenAPI: channels_api 

markxiong0122:


markxiong0122:
close it as data structure change is needed

PR #4168, opened 2024-07-23
Avoid rendering roadmap before channel info loads.

jrobbins:
The render() method can be called before the channel info loads, so in that case just return nothing.  That data loads pretty fast, so there is no need for skeletons.


jrobbins:
Daniel has a better PR.

PR #4169, opened 2024-07-23
Avoid accessing roadmap data before loaded

DanielRyanSmith:


client-src/elements/chromedash-roadmap.ts
```diff
@@ -314,7 +314,7 @@ export class ChromedashRoadmap extends LitElement {
         milestone => html`
           <chromedash-roadmap-milestone-card
             style="width:${this.cardWidth}px;"
-            .channel=${this.milestoneInfo[milestone]}
+            .channel=${this.milestoneInfo ? this.milestoneInfo[milestone] : null}
```
jrobbins, line 317: I think this could be shorter as:
`.channel=${this.milestoneInfo?.[milestone]}`

jrobbins (approved):
LGTM after considering one comment in three places.

PR #4194, opened 2024-07-30
Fix undefined var in early call to render() in provote dialog.

### LINKED ISSUE #4172: Unable to approve request, javascript error ChromedashPrevoteDialog.render
**Describe the bug**
The UI for approving requests do not let me approve anything. My name was not prefilled, but could be added with assign reviewers, and then clicking save after setting status "Approved", nothing visibly happens. 

There appears an error and a stack trace in the  console though:

```
chromedash-gate-column.js:63 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'map')
    at ChromedashPrevoteDialog.render (chromedash-gate-column.js:63:1)
    at ChromedashPrevoteDialog.update (lit-element.js:6:235)
    at ChromedashPrevoteDialog.performUpdate (reactive-element.js:6:4891)
    at ChromedashPrevoteDialog.scheduleUpdate (reactive-element.js:6:4427)
    at ChromedashPrevoteDialog._$ET (reactive-element.js:6:4335)
    at async openPrevoteDialog (chromedash-gate-column.js:36:1)
```

**To Reproduce**
Steps to reproduce the behavior:  (Might have to be on my account or an API Owner)
1. Go to https://chromestatus.com/myfeatures/review?gate=5891529720201216
2. Click on the API Owners button for 'Permissions API' (I assume others will also work)
3. Select "Approved" in the dropdown.
4. Click "Save"

**Expected behavior**
Status of "Approved" should be saved.

**Actual behavior**
Nothing visibly happens.

**Additional context**
Opera, Chromium 125.

jrobbins:
This should resolve #4172.

An undefined variable was being referenced whenever an API Owner tried to vote Approved on their gate on a stage that also had other pending gates.  This triggers the pre-vote dialog to remind them about the need for those other approvals.

Basically, the dialog render() method is called implicitly before the dialog is shown, so that caused an error before the list of gates was set to a defined value.  The fix is to make the render code more robust in the case where that variable is undefined.

### DIFF
--- a/client-src/elements/chromedash-prevote-dialog.ts
+++ b/client-src/elements/chromedash-prevote-dialog.ts
@@ -97,6 +97,10 @@ class ChromedashPrevoteDialog extends LitElement {
   }
 
   render() {
+    if (this.pendingGates === undefined) {
+      return html`Loading gates...`;
+    }
+
     return html` <sl-dialog label="Prerequsites for API Owner approval">
       <div id="prereqs-header">
         The following prerequisite gates are missing approvals:


DanielRyanSmith (approved):
Thanks for the quick investigation and fix!

PR #4195, opened 2024-07-31
OpenAPI: comments_api

markxiong0122:
implemented comments_api with OAS


api/comments_api.py
```diff
@@ -114,22 +122,26 @@ def do_post(self, **kwargs) -> dict[str, str]:
           feature, gate_id, post_to_thread_type, user.email(), comment_content)
 
     # Callers don't use the JSON response for this API call.
-    return {'message': 'Done'}
+    return SuccessMessage(message='Done').to_dict()
 
   def do_patch(self, **kwargs) -> dict[str, str]:
-    comment_id = self.get_param('commentId', required=True)
-    comment: Activity = Activity.get_by_id(comment_id)
+    json_data = self.request.json
```
markxiong0122, line 128: @jrobbins @DanielRyanSmith Do we want to do such parsing to cast it to the auto-generated model or just keep what it was previously?

jrobbins: Yes, I'd like to see us use the auto-generated model for the request parameters.

api/comments_api.py
```diff
@@ -72,18 +80,24 @@ def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
...
-    post_to_thread_type = self.get_param(
-        'postToThreadType', required=False)
 
-    comment_content = self.get_param('comment', required=False)
+    json_data = self.request.json
+    comment_request_data = {
+        'comment': json_data['comment'],
+        'post_to_thread_type': json_data.get('postToThreadType', None),
+    }
+    comment_request = CommentsRequest(**comment_request_data)
```
jrobbins, line 97: I looked into the generated controller and model classes and it looks like the `from_dict()` method is the easiest way to have the generated code validate the input.   So, could these 6 lines be just one line?
`comment_request = CommentRequest.from_dict(self.request.json)`

api/comments_api.py
```diff
@@ -24,28 +32,28 @@
 from internals import slo
 
 
-def amendment_to_json_dict(amendment: Amendment) -> dict[str, Any]:
-  return {
-      'field_name': amendment.field_name,
-      'old_value': amendment.old_value.strip('[]'),
-      'new_value': amendment.new_value.strip('[]'),
-      }
+def amendment_to_json_dict(amendment: Amendment) -> AmendmentModel:
```
jrobbins, line 35: This still seems to return `dict[str, Any]` or more precisely `dict[str, str]`.  You could probably make it return `AmendmentModel` and I think that would work with `ActivityModel`.

api/comments_api.py
markxiong0122: this works and is a lot easier! Thanks

api/comments_api.py
markxiong0122: That would work but requires changing the test file too. I'll push some changes and see if that is what we want

jrobbins (approved):

api/comments_api.py
```diff
@@ -72,18 +80,19 @@ def do_get(self, **kwargs) -> dict[str, list[dict[str, Any]]]:
       lambda c: self._should_show_comment(c, user_email, is_admin), comments))
 
     dicts = [activity_to_json_dict(c) for c in comments]
-    return {'comments': dicts}
+    return GetCommentsResponse(comments=dicts).to_dict()
```
jrobbins, line 83: #4213 has merged, so I think that this can return the `GetCommentResponse` directly without calling to_dict().

jrobbins (approved):
One more suggestion

api/comments_api.py
markxiong0122: The dict conversion does not work and still gives such errors when removing it `======================================================================
ERROR: test_patch__undelete_comment (api.comments_api_test.CommentsAPITest.test_patch__undelete_comment)
Handler unmarks a comment as deleted as requested by authorized user.
----------------------------------------------------------------------
  File "/Users/mark/chromium-dashboard/api/comments_api_test.py", line 249, in test_patch__undelete_comment
    self.assertEqual(get_resp['comments'][0]['deleted_by'], None)
                     ~~~~~~~~^^^^^^^^^^^^
TypeError: 'GetCommentsResponse' object is not subscriptable

======================================================================
ERROR: test_post__comment_only (api.comments_api_test.CommentsAPITest.test_post__comment_only)
Handler adds a comment only, does not require approval permission.
----------------------------------------------------------------------
  File "/opt/homebrew/Cellar/python@3.11/3.11.8/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/mock.py", line 1375, in patched
    return func(*newargs, **newkeywargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mark/chromium-dashboard/api/comments_api_test.py", line 268, in test_post__comment_only
    self.assertEqual(actual, {'message': 'Done'})
  File "/opt/homebrew/Cellar/python@3.11/3.11.8/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 873, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/homebrew/Cellar/python@3.11/3.11.8/Frameworks/Python.framework/Versions/3.11/lib/python3.11/unittest/case.py", line 863, in _baseAssertEqual
    if not first == second:
           ^^^^^^^^^^^^^^^
  File "/Users/mark/chromium-dashboard/cs-env/lib/python3.11/site-packages/chromestatus_openapi/models/base_model.py", line 64, in __eq__
    return self.__dict__ == other.__dict__
                            ^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute '__dict__'` It looks like in the test file it still expect the response to be returned by the do_get method defined in the specific handler

jrobbins: Sure, the tests would need to be updated to validate the returned model object instead of a returned dict.  It's OK with me if that is left for a future PR.

PR #4197, opened 2024-08-01
Rename field_id to gate_type.

jrobbins:
This should reduce potential confusion when reading the code.  It eliminates some of the older terminology of "approval field IDs" and uses the newer term "gate type".  Also, ApprovalFieldDef is now GateInfo.


DanielRyanSmith (approved):
Oops, missed reviewing this -

Thanks for taking care of this! It's a relic from before the schema migration 🙂

PR #4199, opened 2024-08-01
OpenAPI: component_users

markxiong0122:
there are pre-defined specifications on this api, so this PR just casted requests into models


api/component_users.py
```diff
@@ -48,15 +50,17 @@ def do_get(self, **kwargs):
   @permissions.require_admin_site
   def do_put(self, **kwargs) -> tuple[dict, int]:
     params = self.request.get_json(force=True)
+    component_users_request = ComponentUsersRequest(**params)
```
jrobbins, line 53: Let's try `ComponentUsersRequest.from_dict(params)` because that method does validation of the type of the user input.

api/component_users.py
```diff
@@ -48,15 +50,17 @@ def do_get(self, **kwargs):
   @permissions.require_admin_site
   def do_put(self, **kwargs) -> tuple[dict, int]:
     params = self.request.get_json(force=True)
+    component_users_request = ComponentUsersRequest.from_dict(params)
```
jrobbins, line 53: If we have the `params` local variable, there will be a temptation to use it, which we should discourage.  So how about eliminating `params` and just writing the call to the request model constructor the same way in all API handlers, e.g., `ComponentUsersRequest.from_dict(self.request.json)`?

jrobbins (approved):

PR #4211, opened 2024-08-05
OpenAPI: cues_api

markxiong0122:


api/cues_api.py
```diff
@@ -33,12 +35,14 @@ class CuesAPI(basehandlers.APIHandler):
 
   def do_post(self, **kwargs):
     """Dismisses a cue card for the signed in user."""
-    cue = self.get_param('cue', allowed=ALLOWED_CUES)
```
DanielRyanSmith, line 36: The `allowed` param is no longer being used to abort in the context of receiving disallowed values for cue. We probably still want to use this helper method for obtaining the value from the request.

markxiong0122: I obtained values the request by casting the input to the generated model in new line 31 (and it also has enum to limit the choice to only ALLOWED_CUES)

DanielRyanSmith: What is the behavior if a request is sent with a bad value? Currently, the request will be aborted with a 400 status code and this mesage:

https://github.com/GoogleChrome/chromium-dashboard/blob/796b60b4232ba5e0dee0bbcd1b01583f5609e0d1/framework/basehandlers.py#L106-L107

markxiong0122: it handled with an extra condition check, which throws a 400 error if it's not allowed as we discussed yesterday

DanielRyanSmith: Perfect, thanks!

DanielRyanSmith (approved):

PR #4214, opened 2024-08-07
Separate creation and setup requests in OT creation process

DanielRyanSmith:
Per discussion, this change updates the OT creation logic to have to separate requests - one to create the origin trial and one to handle post-creation setup steps.


KyleJu (commented):
Will come back to this once the server side is complete

jrobbins (approved):

PR #4218, opened 2024-08-08
Disable changing OT milestones when OT creation process is in progress (client-side)

DanielRyanSmith:
This change disables OT desktop milestones from being changed while an OT creation request is in progress of being processed. This only disables inputs on client-side forms, and the equivalent API changes will come in a subsequent PR.

Screenshot 2024-08-08 at 3 19 55 PM


client-src/elements/chromedash-guide-stage-page.ts
```diff
@@ -264,13 +265,16 @@ export class ChromedashGuideStagePage extends LitElement {
...
+      const disabledHelpText = getDisabledHelpText(field, feStage);
       return html`
         <chromedash-form-field
           name=${field}
           index=${index}
           value=${value}
           .fieldValues=${this.fieldValues}
           .feature=${formattedFeature}
+          helpText="${disabledHelpText}"
+          ?disabled=${disabledHelpText}
```
jrobbins, line 277: The code that you have is fine, but one idea to make this more compact is to use one attribute instead of two, like `disabledReason=${getDisabledHelpText(...)}` and then have that be handled inside chromedash-form-field to do both the disabling and the help text.

jrobbins (approved):

client-src/elements/chromedash-guide-stage-page.ts
DanielRyanSmith: This is a great suggestion, and I've updated the implementation with this approach 🙂

jrobbins (approved):

PR #4220, opened 2024-08-09
OpenAPI: login logout

markxiong0122:
login/logout api using OAS


api/login_api.py
```diff
@@ -31,7 +30,10 @@ def do_get(self, **kwargs):
     self.abort(405, valid_methods=['POST'])
 
   def do_post(self, **kwargs):
-    token = self.get_param('credential')
+    request = SignInRequest.from_dict(self.request.json)
+    token = request.credential
+    if not token:
+        raise werkzeug.exceptions.BadRequest(description="Missing required field 'credential'")
```
DanielRyanSmith, line 36: It might be worth catching the error `from_dict` will raise if they json does not match what is needed to create the object here (this is done in your previous changes in cues_api)

DanielRyanSmith (commented):
1 comment here

api/login_api.py
markxiong0122: good point. thanks!

DanielRyanSmith (approved):

PR #4226, opened 2024-08-11
OpenAPI: OT

markxiong0122:
implemented OAS on origin_trials_api


api/origin_trials_api.py
```diff
@@ -158,7 +161,8 @@ def do_post(self, **kwargs):
       if gate.state not in (Vote.APPROVED, Vote.NA):
         self.abort(400, 'Unapproved gate found for trial stage.')
 
-    body = self.get_json_param_dict()
+    #TODO(markxiong0122): remove to_dict() when PR#4213 is merged
```
jrobbins, line 164: #4213 has has merged already.

jrobbins (commented):
This looks OK to me, but I did not check all the fields in the OT OpenAPI models.  Daniel knows that better.

DanielRyanSmith (approved):
This looks okay to me as well, but we should test these changes locally to be sure everything looks as expected (I can help with that during deployment).

markxiong0122:
> This looks okay to me as well, but we should test these changes locally to be sure everything looks as expected (I can help with that during deployment).

I didn't find all OT pages so this indeed needs further testing. Thank you for helping with this! Should I merge this PR or wait until we test it locally?

PR #4239, opened 2024-08-13
OpenAPI: processes_api

markxiong0122:


openapi/api.yaml
```diff
@@ -325,6 +345,88 @@ components:
...
+        description:
+          type: string
+        gate_type:
+          type: integer
+        rule:
+          type: string
+        approvers:
+          type: array
+          items:
+            type: string
```
DanielRyanSmith, line 419: This field failed when I ran the tests myself, and the tests were actually looking for a string type, rather than a list of strings

markxiong0122: I set it to array because the in approval_defs.py it's defined as either string / list[str]. It seems like it won't be a list given the tests we had. I typed them as string and it worked.

DanielRyanSmith: This may be an issue we need to look at before landing. Are there different circumstances where a string is passed over a list of strings? And do we need to facilitate both?

markxiong0122: yes in [approval_def.py](https://github.com/GoogleChrome/chromium-dashboard/blob/dd4a0a4d6b0fe292b51411bb1d84a36da0205857/internals/approval_defs.py#L114), approvers could either be a single email address or a list of such, but it seems like test cases are only looking for a single string

DanielRyanSmith (approved):

PR #4247, opened 2024-08-16
Implement phrase search via post processing

jrobbins:
This implements one of the missing details on our list before making the new search page the default.  

Basically, if the user searches for `"one two three"` inside quotes: 
1. first do the search separate words `one`, `two`, and `three` using the datastore indexes to get feature_ids,
2. then we get those feature entires
3. then we iterate through them and look for the phrase `one two three` in the strings of each entry.

### DIFF
--- a/internals/search_fulltext.py
+++ b/internals/search_fulltext.py
@@ -161,7 +161,7 @@ def index_feature(fe: FeatureEntry) -> None:
   updated_fw_list[0].put()
 
 
-def _canonicalize_string(s: str) -> str:
+def canonicalize_string(s: str) -> str:
   """Return a string of lowercase words separated by single spaces."""
   lower_s = s.lower().replace("'", "")
   words = WORD_RE.findall(lower_s)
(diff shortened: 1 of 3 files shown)


internals/search_fulltext.py
```diff
@@ -22,6 +22,10 @@
 
 from framework.basehandlers import FlaskHandler
 from internals.core_models import FeatureEntry
+from internals.feature_helpers import (
+    get_future_results,
+    get_entries_by_id_async,
```
KyleJu, line 27:
```suggestion
    get_entries_by_id_async
```
And fix formatting after

internals/search_fulltext.py
```diff
@@ -157,6 +161,29 @@ def index_feature(fe: FeatureEntry) -> None:
   updated_fw_list[0].put()
 
 
+def canonicalize_string(s: str) -> str:
+  """Return a string of lowercase words separated by single spaces."""
+  lower_s = s.lower().replace("'", "")
```
KyleJu, line 166: Optional: maybe trim the white spaces using .strip() as well

internals/search_fulltext.py
```diff
@@ -157,6 +161,29 @@ def index_feature(fe: FeatureEntry) -> None:
...
+  words = WORD_RE.findall(lower_s)
+  words = [w for w in words if w not in STOP_WORDS]
+  canonicalized = ' '.join(words)
+  return ' ' + canonicalized + ' '
+
+
+def post_process_phrase(phrase: str, feature_ids: list[int]) -> list[int]:
+  """Fetch the given features and check if the really have the phrase."""
```
KyleJu, line 174:
```suggestion
  """Fetch the given features and check if they really have the phrase."""
```

KyleJu (approved):
LGTM with nits

internals/search_fulltext.py
jrobbins: Done

internals/search_fulltext.py
jrobbins: That's not needed because the next line `WORD_RE.findall()` finds words in the string regardless of any whitespace.

internals/search_fulltext.py
jrobbins: Fixed.

PR #4253, opened 2024-08-19
OpenAPI: settings_api

markxiong0122:
implemented OAS on settings_api


api/settings_api.py
```diff
@@ -27,16 +32,26 @@ def do_post(self, **kwargs):
     user_pref = user_models.UserPref.get_signed_in_user_pref()
     if not user_pref:
       self.abort(403, msg='User must be signed in')
-    new_notify = self.get_bool_param('notify')
-    user_pref.notify_as_starrer = new_notify
+    raw_data = self.request.json
+    new_notify = raw_data.get('notify')
+
+    if not isinstance(new_notify, bool):
```
jrobbins, line 38: What happens if we don't have this explicit input validation and we just rely on `from_dict()`?

api/settings_api.py
```diff
@@ -27,16 +32,26 @@ def do_post(self, **kwargs):
...
+        raise werkzeug.exceptions.BadRequest(
+            f"Expected boolean for 'notify', got {type(new_notify).__name__}"
+        )
+
+    request = PostSettingsRequest.from_dict(raw_data)
```
jrobbins, line 43: We should probably only use the variable name `request` for the HTTPRequest object.  Instead this should be something like `settings_request` or `request_model`.

api/settings_api.py
DanielRyanSmith: Similarly, doesn't `get_bool_param()` have this check intrinsically?

markxiong0122: without this validation, tests would fail as bad requests caused by invalid input would not be raised (I think OAS would abort instead of raising a BadRequest exception). `get_bool_param()` helps get `notify` directly, but not from the model we generated

jrobbins: We could always update the tests.  But, that can be deferred to a future PR.

jrobbins (approved):

PR #4254, opened 2024-08-19
OpenAPI: stars_api

markxiong0122:
implemented OAS on stars api


package.json
```diff
@@ -45,7 +45,7 @@
     "deploy-rc": "npm run build && ./scripts/deploy_site.sh rc",
     "openapi": "npm run openapi-backend && npm run openapi-frontend",
     "openapi-frontend": "rm -rf gen/js/chromestatus-openapi && openapi-generator-cli generate -i openapi/api.yaml -g typescript-fetch  -o gen/js/chromestatus-openapi --config openapi/js-config.yaml && npm install",
-    "openapi-backend": ". cs-env/bin/activate; rm -rf gen/py/chromestatus_openapi && openapi-generator-cli generate -i openapi/api.yaml -g python-flask  -o gen/py/chromestatus_openapi --additional-properties=packageName=chromestatus_openapi && pip install -r requirements.txt",
+    "openapi-backend": ". cs-env/bin/activate; rm -rf gen/py/chromestatus_openapi && openapi-generator-cli generate -i openapi/api.yaml -g python-flask  -o gen/py/chromestatus_openapi --additional-properties=packageName=chromestatus_openapi,modelPropertyNaming=original && pip install -r requirements.txt",
```
jcscottiii, line 48: I get the same result without the additional flag

markxiong0122: I'll remove it. thanks

jrobbins:
I'm thinking that we can change the API to use snake case, but we will need backwards compatibility for 30 days to prevent breaking users who have a tab open on the site.  Here's the steps:
1. Go ahead and define the openapi parameter using `feature_ids`.
2. In the API handler, construct a GetStarsResponse object using `feature_ids`, and call to_dict(), but just store the result in a `result` variable rather than returning it.
3. Add a line for backwards compatibility: `result['featureIds'] = result['feature_ids']` and then return that.  Add a TODO comment to delete it after 30 days.
4. Change the tests to check for  `actual['feature_ids']`
5. On the client-side, in cs-client.js getStars() return `res.feature_ids || res.featureIds`.  Again with a TODO saying that the `featureIds` part can be deleted in 30 days.


The backwards compatibility on the client side will likely never be used, but it is there in case we deploy a new version of the server, and some users open tabs on the new version of the client, and then we need to roll back the server.  We don't want to break the clients that were opened during the time before we rolled back.

jrobbins (approved):

PR #4262, opened 2024-08-20
Implement fulltext search within a specific field.

jrobbins:
This addresses another checklist item in our plan to make the new search page the default page.

Support for full text search within a specific field works much like searching for a phrase:
1. First, search results are calculated based solely on the word mentioned, regardless of the field.
2. Then, those results are fetched and post-processed to filter for only the ones that contain the word or phrase within the desired field.

### DIFF
--- a/internals/search_fulltext.py
+++ b/internals/search_fulltext.py
@@ -46,75 +46,89 @@ class FeatureWords(ndb.Model):
   words = ndb.StringProperty(repeated=True)
 
 
-def get_strings(fe : FeatureEntry) -> list[str]:
+def _get_strings_dict(fe: FeatureEntry) -> dict[str, list[str|None]]:
+  return {
+      'creator_email': [fe.creator_email],
+      'updater_email': [fe.updater_email],
+      'owner_emails': fe.owner_emails,
+      'editor_emails': fe.editor_emails,
+      'cc_emails': fe.cc_emails,
+
+      'name': [fe.name],
+      'summary': [fe.summary],
+      # TODO: category
+      'blink_components': fe.blink_components,
+      'tags': fe.search_tags,
+      'feature_notes': [fe.feature_notes],
+
+      'bug_url': [fe.bug_url],
+      'launch_bug_url': [fe.launch_bug_url],
+
+      # TODO: impl_status_Chrome
+      'flag_name': [fe.flag_name],
+      'finch_name': [fe.finch_name],
+      'non_finch_justification': [fe.non_finch_justification],
+      'ongoing_constraints': [fe.ongoing_constraints],
+
+      'motivation': [fe.motivation],
+      'devtrial_instructions': [fe.devtrial_instructions],
+      'activation_risks': [fe.activation_risks],
+      'measurement': [fe.measurement],
+
+      'initial_public_proposal_url': [fe.initial_public_proposal_url],
+      'explainer_links': fe.explainer_links,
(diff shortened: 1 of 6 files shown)


internals/search_fulltext.py
```diff
@@ -46,77 +46,91 @@ class FeatureWords(ndb.Model):
...
-  strings.extend(fe.devrel_emails)
-  strings.append(fe.debuggability)
-  strings.extend(fe.doc_links)
-  strings.extend(fe.sample_links)
+  strings_dict = _get_strings_dict(fe)
+
+  if field_name and field_name in strings_dict:
+    strings = strings_dict[field_name]
+  else:
+    strings = [s for s_list in strings_dict.values() for s in s_list]
```
KyleJu, line 123: Do we have to check None for  s_list? 

Also nit: if it is not more memory efficient to write two for loops in one [] statement, write two separate for loops statement instead

```suggestion
    strings = [s for s_list in strings_dict.values() for s in s_list]
```

KyleJu: Actually None is checked below. Ignore my first comment

internals/search_fulltext.py
```diff
@@ -168,21 +182,24 @@ def canonicalize_string(s: str) -> str:
   return ' ' + canonicalized + ' '  # Avoids matching partial words.
 
 
-def post_process_phrase(phrase: str, feature_ids: list[int]) -> list[int]:
+def post_process_phrase(
+    phrase: str, feature_ids: list[int],
+    field_name: str|None = None) -> list[int]:
   """Fetch the given features and check if they really have the phrase."""
```
KyleJu, line 188: Nit: update the comment to incorporate  field_name

internals/search_fulltext.py
```diff
@@ -168,21 +182,24 @@ def canonicalize_string(s: str) -> str:
...
                      for fs in feature_strings)
     if has_phrase:
       result.append(fe.key.integer_id())
   return result
 
 
-def search_fulltext(textterm: str) -> Optional[list[int]]:
+def search_fulltext(
+    textterm: str, field_name: str|None = None) -> Optional[list[int]]:
   """Return IDs of features that have some word(s) from phrase."""
```
KyleJu, line 203: Update the comment to incorporate field_name

internals/search_queries.py
```diff
@@ -106,6 +107,8 @@ def single_field_query_async(
   field_name = field_name.lower()
   if field_name in COMPLEX_FIELDS:
     return COMPLEX_FIELDS[field_name](operator, val_list, limit)
+  elif operator == ':' and field_name in FULLTEXT_FIELDS:
```
KyleJu, line 110: Missing test cases for this?

internals/search_fulltext.py
jrobbins: I wrote two because I am iterating over a nested list to flatten the list.  I believe that this is a python idiom for flattening a 2-level list.

internals/search_fulltext.py
jrobbins: Done.

internals/search_fulltext.py
jrobbins: Done.

internals/search_queries.py
jrobbins: Added

KyleJu (approved):

PR #4266, opened 2024-08-21
Obtain Chromium source files concurrently

DanielRyanSmith:
This change updates the logic in which Chromium source file contents are obtained in order to fetch files asynchronously. This approach is based on the [example of using concurrent.futures in the docs](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example) and may be handled with less issues than the previous implementation.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -95,8 +95,6 @@ class OriginTrialsAPI(basehandlers.EntitiesAPIHandler):
       validation_errors['ot_chromium_trial_name'] = (
           'Chromium trial name is already used by another origin trial')
 
-    print(chromium_files['enabled_features_text'])
-    print(type(chromium_files['enabled_features_text']))
     enabled_features_json = json5.loads(chromium_files['enabled_features_text'])
     if (not any(feature.get('origin_trial_feature_name') == chromium_trial_name
                 for feature in enabled_features_json['data'])):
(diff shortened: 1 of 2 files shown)


api/origin_trials_api.py
```diff
@@ -65,12 +66,18 @@ def do_get(self, **kwargs):
...
     """Check that all provided OT creation arguments are valid."""
-    try:
-      enabled_features_text = get_chromium_file(ENABLED_FEATURES_FILE_URL)
-      webfeature_file = get_chromium_file(WEBFEATURE_FILE_URL)
-      grace_period_file = get_chromium_file(GRACE_PERIOD_FILE)
-    except requests.exceptions.RequestException:
-      self.abort(500, 'Error obtaining Chromium file for validation')
+    chromium_files = {}  # Chromium source file contents.
+    with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+      future_to_url = {executor.submit(get_chromium_file, f['url']): f['name']
```
jrobbins, line 71: Thank you for linking to the example and following the example closely.  I think the name `future_to_url` might be leftover from the example.  A more accurate variable name for what this code is doing might be `future_to_name`.

jrobbins (approved):

api/origin_trials_api.py
DanielRyanSmith: Ah, good catch! Yes, a remnant of following the example 🙂

PR #4274, opened 2024-08-22
Handle OpenAPI models correctly for GET requests

DanielRyanSmith:
This change fixes a bug where OpenAPI models were not handled properly for GET requests. The conditional logic to handle OpenAPI models has been changed to properly detect them by checking if the handler_data has a `to_dict` attribute, and to use it accordingly.


jrobbins (approved):
Hmm, I'm curious as to why the previous approach did not work.

DanielRyanSmith:
> Hmm, I'm curious as to why the previous approach did not work.

I'm not certain, but I also had trouble with an implementation using `issubclass()`. I'd imagine there's a simple solution using these variations, but this PR's approach also worked. 😅

PR #4277, opened 2024-08-22
Add simple warning header text to OT creation form

DanielRyanSmith:
This is a small change to update the OT creation form to add some text that warns users to be sure that the data in the form is correct, and to contact the OT support team if they have any questions. The reason for adding this is to prevent users from attempting to give bad data (e.g. a fake use counter) to submit the form. We've had some instances of users adding these fake data or filler inputs, and this should discourage that behavior in the future.

The wording can be changed as suggested.
Screenshot 2024-08-22 at 3 16 27 PM

### DIFF
--- a/client-src/elements/chromedash-ot-creation-page.ts
+++ b/client-src/elements/chromedash-ot-creation-page.ts
@@ -43,6 +43,18 @@ export class ChromedashOTCreationPage extends LitElement {
           transform: translate(-50%, -50%);
           height: 300px;
         }
+        .warning-div {
+          background-color: #fbfba7;
+          border: 2px solid #ffe735;
+          border-radius: 10px;
+          padding: 4px;
+          margin-bottom: 32px;
+        }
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-ot-creation-page.ts
```diff
@@ -43,6 +43,18 @@ export class ChromedashOTCreationPage extends LitElement {
           transform: translate(-50%, -50%);
           height: 300px;
         }
+        .warning-div {
```
jrobbins, line 46: There is already a `class="warning"` defined for the whole site that you can use instead of defining your own.  It uses an orange background with no border.  I think it would be good to use it so that things stay consistent and we can change it one place.  The colors are defined in static/css/main.css.

jrobbins (approved):
LGTM.  Users will still try odd things occasionally, but I suppose that this will help.

PR #4300, opened 2024-08-27
Make detection of intent_thread_url more robust

jrobbins:
This should resolve a failure that was discovered by Kyle today.  The problem was that an intent thread URL was not clickable because it lacked the final `.org` part of the URL.

Logs showed that the URL was correctly parsed in the first message that we detected on this thread.  However, we also detected a quoted version of it in a reply to the original email.  The quoted version of the URL is a little more complicated.  E.g.,
`https://example.com/a-long-path-ending-with.org.` in the first email gets becomes
`https://example.com/a-long-path-ending-with.org <https://example.com/a-long-path-ending-with.org>.`
in the quoted version of the text.  And, that can be word-wrapped like:
'https://example.com/a-long-path-ending-with.org\r\n <https://example.com/a-long-path-ending-with.org>.'

I had covered the `\n` case for unix users, but not the `\r\n` case for windows users.

The fix is to also detect `\r` to indicate the end of the URL.

Also, there is no point to ever set these URLs again if they already have a value, in fact that could overwrite data that a user entered manually.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -124,11 +124,11 @@ def detect_gate_id(body) -> int | None:
 THREAD_LINK_RE = re.compile(
     r'To view this discussion on the web visit\s+'
     r'(> )?(https://groups\.google\.com/a/chromium.org/d/msgid/blink-dev/'
-    r'\S+)[\n> .]', re.MULTILINE)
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (commented):
LGTM with a question on implementation

DanielRyanSmith (approved):

PR #4309, opened 2024-08-29
Clarify that the security gate is for WP security.

jrobbins:
One of the feedback items we got from a ChromeStatus user survey was that the Security review gate in ChromeStatus seems to duplicate the Security review bit in our internal launch tool.  So, this helps clarify that the ChromeStatus gate is for Web Platform security, whereas the internal one is for Chrome Browser Security.

I'm using the acronym "WP" for "Web Platform" because space is limited in the gate chips.  It should be pretty familiar to our users.  But, it will also be clarified in some changes to the gate column text coming in a separate PR.


jstenback:
Would it take up an unreasonable amount of space if we spelled out "Web Platform" rather than abbreviating it?

jstenback:
Oops, I see you addressed that question in the initial comment. Apologies for not reading that fully before commenting.

KyleJu (approved):

PR #4316, opened 2024-08-30
Record an Activity when assigning a reviewer.

### LINKED ISSUE #4273: Assigning a reviewer should create an Activity record on the gate
If a review is assigned to someone, that should have the side-effect of creating an Activity record on the gate so that we can see how and when the assignment was done.  That can be helpful if the team desires to assign it back to a previously assigned person.

jrobbins:
This resolves #4273.

There are two ways to assign a reviewer: via a UI menu, or by requesting a review which triggers auto-assignment.  In each case we now create an Activity records that shows up as a comment on the gate.


DanielRyanSmith (approved):
Looks great!

Issue #4246, opened 2024-08-15
Add gate descriptions that clarify their purpose


past:
One consistent feedback we received in the recent survey is that people aren't sure why there are all these approval gates slowing down their launches. It would be helpful to at least add some explanation of the value of the gate (if not the context of its original addition) in a description field. I imagine a short intro right after the panel header would work here.

jrobbins:
We are drafting the text in an internal doc:
https://docs.google.com/document/u/0/d/17mZl9Ml3MyEes7oW585YUj39mz_Wk6xSB2Khh7ux-fc/edit

PR #4319, opened 2024-08-30
Add gate rationale

jrobbins:
This addresses #4246.  We got some feedback from a user survey that said that users were not clear on why each gate was needed.  We add a "Why this gate?" section that can be expanded to show the rationale for each gate.

In this PR:
* Refactor CSS for `<details>` to a shared file
* Refactor client-side gate text to a new file
* Change from using hover text to using `<details>`
* Add logic to display the rationale if it is defined


client-src/elements/chromedash-gate-column.ts
```diff
@@ -669,14 +648,36 @@ export class ChromedashGateColumn extends LitElement {
...
+      <details>
+        <summary>Why this gate?</summary>
+        ${rationale}
+      </details>
+    `;
+  }
+
   renderWarnings() {
-    if (this.gate && this.gate.team_name == 'Privacy') {
+    if (this.gate && ['Privacy', 'WP Security'].includes(this.gate.team_name)) {
```
DanielRyanSmith, line 680: I tried this locally and was not able to see this additional warning on Security gates

Screenshot 2024-09-04 at 11 13 23 AM

I was able to see it on privacy gates, however:

Screenshot 2024-09-04 at 11 13 29 AM

DanielRyanSmith (approved):
Ignore previous comment, this LGTM 🙂

PR #4325, opened 2024-09-02
Pressing slash focuses on feature searchbox.

jrobbins:
This is a very common UI detail that I am also implementing for webstatus.dev.

### DIFF
--- a/client-src/elements/chromedash-typeahead.ts
+++ b/client-src/elements/chromedash-typeahead.ts
@@ -91,6 +91,16 @@ export class ChromedashTypeahead extends LitElement {
     this.slDropdownRef.value!.show();
   }
 
+  focus() {
+    const slInput: SlInput = this.slInputRef.value as SlInput;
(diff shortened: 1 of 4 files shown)


client-src/elements/chromedash-typeahead.ts
```diff
@@ -91,6 +91,16 @@ export class ChromedashTypeahead extends LitElement {
     this.slDropdownRef.value!.show();
   }
 
+  focus() {
+    const slInput: SlInput = this.slInputRef.value as SlInput;
+    slInput?.focus();
+  }
+
+  blur() {
```
KyleJu, line 99: Is this used?

KyleJu (approved):
LGTM with a comment

client-src/elements/chromedash-typeahead.ts
jrobbins: No, I just felt that if I added focus(), it should also have blur().

PR #4328, opened 2024-09-03
Limit pagination length

KyleJu:
Limit the pagination length to avoid overflow. 

It shows 4 pages + and - from the current page on the pagination and the first/last page, if exist. For pages not displayed, it shows ... instead. E.g.

Pages at the back missing:
Screenshot 2024-09-03 2 49 42 PM

Pages at the front missing:
Screenshot 2024-09-03 2 49 53 PM

4 to the right exactly + last page, without ...:
Screenshot 2024-09-03 2 50 16 PM

4 to the left exactly + first page without ...:
Screenshot 2024-09-03 2 50 26 PM

Missing front and back:
Screenshot 2024-09-03 2 50 32 PM

Less than 10 pa
Screenshot 2024-09-03 2 00 10 PM
ges, and showed fully:

### DIFF
--- a/client-src/elements/chromedash-feature-pagination.ts
+++ b/client-src/elements/chromedash-feature-pagination.ts
@@ -14,7 +14,14 @@
  * limitations under the License.
  */
 
-import {LitElement, type TemplateResult, CSSResultGroup, css, html, nothing} from 'lit';
+import {
+  LitElement,
+  type TemplateResult,
+  CSSResultGroup,
+  css,
+  html,
+  nothing,
+} from 'lit';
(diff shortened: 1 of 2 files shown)


KyleJu:
The missing front and back case has a bit UI issue

KyleJu:
> The missing front and back case has a bit UI issue. The page size drop-down bar is squeezed

I will reduce it to 4 each size, so it will show 4 *2 (each side) + 1 (current page) + 2 (first/last buttons) = 11 buttons max, instead of 13

jrobbins (approved):

PR #4332, opened 2024-09-04
Add main menu link for annual roadmap.

jrobbins:
This adds just one link to the main nav to show a query for shipping_year=2024 (or whatever the current year is).  From there, the user can change the query in the search box.

Also, fix `@state` vs `@property` so that the typeahead value can be passed in from the page (which gets it from the URL).


KyleJu (approved):

client-src/elements/chromedash-drawer.ts
```diff
@@ -261,6 +261,12 @@ export class ChromedashDrawer extends LitElement {
     const myFeaturesMenu = this.renderMyFeaturesMenu();
     const adminMenu = this.renderAdminMenu();
 
+    const year = new Date().getFullYear();
```
jcscottiii, line 264: Just a heads-up that on Jan 1, this will cause a lot of the playwright screenshot tests to break when the link changes from `Shipping 2024` to `Shipping 2025`.

jrobbins: Good point. I'll follow up with a PR to elide that.

PR #4337, opened 2024-09-05
Load feature names only for the /newfeatures page

KyleJu:
Add a boolean query param `name_only` to retrieve data subsets - feature name and feature ID for a faster page load. Also cache this data using a new cache prefix, `FEATURE_NAME_CACHE_KEY`.

- `get_feature_names_by_ids` returns the same data type as get_by_ids, but with only `id` and `name` entries.
- `FEATURE_NAME_CACHE_KEY|*` cache is deleted when there are changes to a FeatureEntry
- `name_only` param should only use at `/newfeatures` page, as it might break other pages

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -351,7 +351,7 @@ def get_feature_names_by_ids(feature_ids: list[int],
                      if f is not None and f.get('id')}
 
   for feature_id in feature_ids:
-    if result_dict.get(feature_id) is None:
+    if feature_id not in result_dict:
       futures_by_id[feature_id] = FeatureEntry.get_by_id_async(feature_id)
 
   for future in futures_by_id.values():
@@ -362,7 +362,6 @@ def get_feature_names_by_ids(feature_ids: list[int],
         'id': feature_id,
         'name': fe.name,
       }
-
       result_dict[feature_id] = feature_name_dict
 
   if update_cache:


internals/feature_helpers.py
```diff
@@ -332,6 +332,52 @@ def get_all(limit: Optional[int]=None,
...
+            FeatureEntry.FEATURE_NAME_CACHE_KEY, feature_id)
+        for feature_id in feature_ids]
+    cached_features = rediscache.get_multi(lookup_keys)
+    if cached_features:
+      result_dict = {f['id']: f
+                     for f in cached_features.values()
+                     if f is not None and f.get('id')}
+
+  for feature_id in feature_ids:
+    if result_dict.get(feature_id) is None:
```
jrobbins, line 354: Slightly more pythonic:
`if feature_id not in result_dict:`

internals/feature_helpers.py
```diff
@@ -332,6 +332,52 @@ def get_all(limit: Optional[int]=None,
...
+
+  for future in futures_by_id.values():
+    fe: Optional[FeatureEntry] = future.get_result()
+    if fe and not fe.deleted:
+      feature_id = fe.key.integer_id()
+      feature_name_dict = {
+        'id': feature_id,
+        'name': fe.name,
+      }
+
```
jrobbins, line 365: I would not use a blank line here.  The next line of code is logically closely associated with the line above.

jrobbins (approved):
LGTM after considering a couple of nits.

PR #4357, opened 2024-09-13
Add script to backfill shipping_year

jrobbins:
This PR adds a maintenance script that can backfill the shipping_year field for all the feature entries in our DB.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -19,7 +19,7 @@ from typing import Any
 from google.cloud import ndb  # type: ignore
 import requests
 
-from api import converters
+from api import converters, channels_api
 from framework.basehandlers import FlaskHandler
 from framework import cloud_tasks_helpers
 from framework import origin_trials_client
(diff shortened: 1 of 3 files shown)


internals/maintenance_scripts.py
```diff
@@ -699,3 +700,106 @@ def get_template_data(self, **kwargs) -> str:
       ndb.delete_multi(keys_to_delete)
 
     return  (f'{counter} empty extension stages deleted.')
+
+
+LAST_MILESTONE_TO_YEAR: dict[int, int] = {
```
DanielRyanSmith, line 705: Nit: Since this is always converted to a list of tuples, it could feasibly be that data type to begin with

DanielRyanSmith (approved):
LGTM besides a nit that doesn't need to be changed since this is a temporary measure.

internals/maintenance_scripts.py
jrobbins: I just thought that the dictionary syntax is a little cleaner, with fewer parentheses.

PR #4374, opened 2024-09-18
Add name_only for data fetch in /newfeatures

KyleJu:
Add the name_only param when fetching FeatureEntries in `/newfeatures`.  Speeds up the page load.

A follow-up for https://github.com/GoogleChrome/chromium-dashboard/pull/4337.

See responses in /newfeatures page:
Screenshot 2024-09-18 1 41 50 PM


jrobbins (approved):
Awesome.  Let's try to measure the speedup on staging.

PR #4375, opened 2024-09-19
add security checks to report apis

jrobbins:
We offer some report pages that present tables of info.   Anyone could compute these by scraping data that we present elsewhere on the site, but we still want to limit access to these reports because the API calls do a lot of work.

Also, there has been a long-standing TODO to avoid redirecting to the login page when the request is an API request.  This implements that part.

In this PR:
* Add permission check for feature latency report and review latency report
* Update existing unit tests and add new ones that show non-authorized requests being rejected
* Implement the TODO for avoiding redirect for signed-out API GETs


packages/playwright/tests/__screenshots__/chromedash-guide-feature-page_pwtest.js/add-an-origin-trial-stage/origin-trial-panels-chromium-mobile-linux.png
DanielRyanSmith, line 1: It looks like the only difference here is the underlined "File an issue" link. I'm not quite sure what change here would cause that. 🤔 It looks it's a matter of where the "mouse" is hovering while testing, but I'm no Playwright expert

framework/permissions.py
```diff
@@ -142,10 +142,16 @@ def can_review_gate(
 
 
 def _maybe_redirect_to_login(handler_obj) -> flask.Response | dict:
-  # TODO(jrobbins): For API calls, we should return 401 rather than a redirect.
+  # Don't redirect if the handler is not an UI page (it is an API handler).
+  if not hasattr(handler_obj, 'get_common_data'):
+      return {}
```
DanielRyanSmith, line 147: Thanks for taking care of this. I wasn't sure the best way to implement this, but this approach seems like a simple solution. 🙂

DanielRyanSmith (approved):

PR #4379, opened 2024-09-20
Align feature_type enum names on client and server.

jrobbins:
A few months back we changed the feature_type enum display names to be more clear.  But, the change was only made on the client side.  This meant that the autocomplete was offering `feature_type="Chrome catches up"` but the server didn't recognize that because it was looking for `"Existing feature implementation"`.  This PR aligns them so that users can use what the search autocomplete suggests.  It also has the effect of using the new terms in notification emails for newly created features.

### DIFF
--- a/internals/core_enums.py
+++ b/internals/core_enums.py
@@ -128,10 +128,10 @@ FEATURE_TYPE_DEPRECATION_ID = 3
 FEATURE_TYPE_ENTERPRISE_ID = 4
 
 FEATURE_TYPES = {
-    FEATURE_TYPE_INCUBATE_ID: 'New feature incubation',
-    FEATURE_TYPE_EXISTING_ID: 'Existing feature implementation',
-    FEATURE_TYPE_CODE_CHANGE_ID: 'Web developer facing change to existing code',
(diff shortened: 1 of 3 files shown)


DanielRyanSmith (approved):

KyleJu (approved):

internals/core_enums.py
```diff
@@ -128,10 +128,10 @@
 FEATURE_TYPE_ENTERPRISE_ID = 4
 
 FEATURE_TYPES = {
-    FEATURE_TYPE_INCUBATE_ID: 'New feature incubation',
-    FEATURE_TYPE_EXISTING_ID: 'Existing feature implementation',
-    FEATURE_TYPE_CODE_CHANGE_ID: 'Web developer facing change to existing code',
-    FEATURE_TYPE_DEPRECATION_ID: 'Feature deprecation',
+    FEATURE_TYPE_INCUBATE_ID: 'New or changed feature',
+    FEATURE_TYPE_EXISTING_ID: 'Chrome cathes up',
```
KyleJu, line 132:
```suggestion
    FEATURE_TYPE_EXISTING_ID: 'Chrome cathes up',
```

catches?

PR #4382, opened 2024-09-21
Cache full search results.

jrobbins:
As we chatted about, this adds redis caching of full search results.

In this PR:
* Add a new function `process_query_using_cache()` that wraps around around `process_query()`, and some helper functions to decide which queries are cacheable.
* Add cache invalidation calls at points where features (and their parts) are created or updated.
* Clarified that `rediscache.delete_keys_with_prefix()` should take a prefix string rather than a pattern.  This better fits the name of the function and the way that that it was being used.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -22,6 +22,7 @@ from typing import Any, Optional, Self, Union
 from google.cloud.ndb import Key
 from google.cloud.ndb.tasklets import Future  # for type checking only
 
+from framework import rediscache
 from framework import users
 from internals import (
   approval_defs,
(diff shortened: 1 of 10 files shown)


internals/search.py
```diff
@@ -311,6 +312,73 @@ def _sort_by_total_order(
   return sorted_id_list
 
 
+def make_cache_key(
+  user_query: str, sort_spec: str | None, show_unlisted: bool,
+  show_deleted: bool, show_enterprise: bool, start: int, num: int,
+  name_only: bool) -> str:
+  """Return a redis key string to store cached search results."""
+  return '|'.join([
```
KyleJu, line 320: I wonder if we can make the key name more readable, since the max allowed key size is quite big [1]? e.g. FeatureSearch||unlisted=None|deleted=True|enterprise=False|...'.

It would also prevent issues if we modify the parameters of a cache key, e.g. changing the sequence of these params, etc.

[1]https://redis.io/docs/latest/develop/use/keyspace/

internals/search.py
```diff
@@ -311,6 +312,73 @@ def _sort_by_total_order(
...
+) -> tuple[list[dict[str, Any]], int]:
+  """"""
+  cache_key = make_cache_key(
+      user_query, sort_spec, show_unlisted, show_deleted, show_enterprise,
+      start, num, name_only)
+  if is_cacheable(user_query, name_only):
+    logging.info('Checking cache at %r', cache_key)
+    cached_result = rediscache.get(cache_key)
+    if cached_result:
+      logging.info('Found cached search result')
```
KyleJu, line 366: nit: add the cache key name here: `Found cached search result for x`

internals/search.py
```diff
@@ -311,6 +312,73 @@ def _sort_by_total_order(
...
+
+  logging.info('Computing search result')
+  computed_result = process_query(
+      user_query, sort_spec=sort_spec, show_unlisted=show_unlisted,
+      show_deleted=show_deleted, show_enterprise=show_enterprise,
+      start=start, num=num, context=context, name_only=name_only)
+
+  if is_cacheable(user_query, name_only):
+    logging.info('Storing search result in cache: %r', cache_key)
+    rediscache.set(cache_key, computed_result)
```
KyleJu, line 377: Let's set a short expiry date here. Less chance for issues

internals/search.py
```diff
@@ -311,6 +312,73 @@ def _sort_by_total_order(
...
+
+  if ('is:recently-reviewed' in user_query or
+      'now' in user_query or
+      'current_stable' in user_query):
+    logging.info('Search query not cached: time-based')
+    return False
+
+  logging.info('Search query can be cached')
+  return True
+
```
KyleJu, line 346:
```suggestion


```

KyleJu: Two lines between methods

internals/search.py
```diff
@@ -311,6 +312,73 @@ def _sort_by_total_order(
...
+  name_only=False,
+) -> tuple[list[dict[str, Any]], int]:
+  """"""
+  cache_key = make_cache_key(
+      user_query, sort_spec, show_unlisted, show_deleted, show_enterprise,
+      start, num, name_only)
+  if is_cacheable(user_query, name_only):
+    logging.info('Checking cache at %r', cache_key)
+    cached_result = rediscache.get(cache_key)
+    if cached_result:
```
KyleJu, line 365: Should we check cached_result  is None specifically instead? It is to distinguish an empty result, () vs cache miss , None

internals/search.py
jrobbins: Done

internals/search.py
jrobbins: Done

internals/search.py
jrobbins: Done.  I set it to 1 hour.  That is probably about the longest that it will actually stay cached during a normal day anyway, because we clear the cache when people do edits.

internals/search.py
jrobbins: Done

internals/search.py
jrobbins: Sure.  The empty result is still a non-empty tuple, so it is still considered true, but checking for None is more explicit.

internals/search.py
```diff
@@ -319,9 +319,14 @@ def make_cache_key(
...
-      str(show_deleted), str(show_enterprise), str(start), str(num),
-      str(name_only),
+      user_query,
+      'sort_spec=' + str(sort_spec),
+      'show_unlisted=' + str(show_unlisted),
+      'show_deleted=' + str(show_deleted),
+      'show_enterprise=' + str(show_enterprise),
+      'start=' + str(start),
+      'num=' + str(num),
+      'name_only=' + str(name_only),
```
KyleJu, line 329: Note: we only cache name_only queries for now. This part of the key is optional

KyleJu (approved):

internals/search.py
jrobbins: I think it is best to include it, just in case we decide to broaden what we cache.

PR #4391, opened 2024-09-24
Make the new feature search page the default

jrobbins:
I believe that we have checked off everything that we had in mind for this feature, so it is time to finally make it the default this week.

In this PR:
* Route traffic to the new feature list page at /features
* Redirect from /newfeatures to /features
* Move old feature list to /oldfeatures
* Update banner links, but keep them, and de-emphasize the link back to the old page.

### DIFF
--- a/client-src/elements/chromedash-app.ts
+++ b/client-src/elements/chromedash-app.ts
@@ -358,7 +358,8 @@ export class ChromedashApp extends LitElement {
       this.pageComponent.showQuery = false;
       this.pageComponent.rawQuery = parseRawQuery(ctx.querystring);
     });
-    page('/newfeatures', ctx => {
+    page('/newfeatures', () => page.redirect('/features'));
+    page('/features', ctx => {
       if (!this.setupNewPage(ctx, 'chromedash-all-features-page', true)) return;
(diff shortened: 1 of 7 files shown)


client-src/elements/chromedash-app.ts
```diff
@@ -358,7 +358,8 @@ export class ChromedashApp extends LitElement {
       this.pageComponent.showQuery = false;
       this.pageComponent.rawQuery = parseRawQuery(ctx.querystring);
     });
-    page('/newfeatures', ctx => {
+    page('/newfeatures', () => page.redirect('/features'));
+    page('/features', ctx => {
```
KyleJu, line 362: Question about this - I know the old /features page was not a part of SPA. Is that the reason why it is not routed here before? e.g. there was no `page('/features', ctx => {...`

KyleJu (approved):

client-src/elements/chromedash-app.ts
jrobbins: Yes, exactly.  And, now the old page is at `/oldfeatures` but that is not in the SPA routing because it is still an MPA page.

PR #4393, opened 2024-09-25
Convert intent preview components to TypeScript

DanielRyanSmith:
This change converts the last Lit elements in the repository to TypeScript files.

Additionally, a part of the `lint` and `lint-fix` commands have been removed for eslint, because an error occurs when specifying a path with no JavaScript files.


jrobbins (approved):
Woo-hoo, no more JS files in this directory.

PR #4397, opened 2024-09-27
Update shipping_year when milestones are edited.

jrobbins:
Once a feature owner fills in some shipping milestones, the feature's shipping_year should be calculated based on those milestones.  This PR adds that calculation to the features_api.

In this PR:
* Refactor `fetch_chrome_release_info()` into fetchchannels.py to avoid a circular import.
* Implement the new logic in features_api.py
    *  `_patch_update_stages()` now returns the list of stages that were updated so that we can check their milestones later.
    * Those updated stages and `changed_fields` are now passed into `_patch_update_special_fields()` so that we can check milestones and add a changed field item if `shipping_year` is set.
* Refactor three functions from maintenance_scripts.py to stage_helpers to avoid having APIs depend on maintenance stuff.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -259,8 +259,11 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
     if should_remove_first_notice_milestone(feature, feature_changes):
       feature.first_enterprise_notification_milestone = None
 
-    # Set shipping_year based on milestones.
-    if updated_stages:
+    # If a shipping stage was edited, set shipping_year based on milestones.
+    updated_shipping_stages = [
+        us for us in updated_stages
+        if us.stage_type in STAGE_TYPES_SHIPPING.values()]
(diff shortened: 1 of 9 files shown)


api/features_api.py
```diff
@@ -255,6 +259,30 @@ def _patch_update_special_fields(
     if should_remove_first_notice_milestone(feature, feature_changes):
       feature.first_enterprise_notification_milestone = None
 
+    # Set shipping_year based on milestones.
+    if updated_stages:
+      existing_shipping_stages = (
+          stage_helpers.get_all_shipping_stages_with_milestones(
```
DanielRyanSmith, line 265: This seems like a large check for any time any stage is updated. Is it safe to check only when a shipping milestone is changed?

jrobbins: Good point.  Done.

DanielRyanSmith (approved):

PR #4400, opened 2024-09-27
Fix population of search box from QS.

jrobbins:
This fixes the case where the user visits a URL that has a `?q=something` query-string parameter by making that `something` appear in the search box.  It was being passed in from the parent element, but that was not having any effect because the query property was defined using `@state` instead of `@property`.


jcscottiii (approved):

jrobbins:
Thanks for the quick review.  Actually, this was part of #4332 already, but I forgot it was in there.  It's better to merge it as a separate fix.  So, I'll do that.

PR #4401, opened 2024-09-27
Only highlight main nav items on exact match.

jrobbins:
This avoids the possibility of both "All features" and "Shipping 2024" main menu items both being highlighted at the same time.  It also means that the "All features" item will not be highlighted if the user specifies a search term, but that is arguably correct since they are no longer looking at *all* features.

### DIFF
--- a/client-src/elements/chromedash-drawer.ts
+++ b/client-src/elements/chromedash-drawer.ts
@@ -231,7 +231,7 @@ export class ChromedashDrawer extends LitElement {
   }
 
   isCurrentPage(href) {
-    return this.currentPage.startsWith(href);
+    return this.currentPage === href;
   }
 
   toggleDrawerActions() {


client-src/elements/chromedash-drawer.ts
```diff
@@ -231,7 +231,7 @@ export class ChromedashDrawer extends LitElement {
   }
 
   isCurrentPage(href) {
-    return this.currentPage.startsWith(href);
+    return this.currentPage === href;
```
jcscottiii, line 234: Something to consider in the future: I think all of the urls on the menu have 0 or 1 query parameters now so it doesn't matter. But if that number grows, we will need to parse the urls and then [sort](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/sort) the query string parameters before comparing.

This will help in case a user types in the URL directly into the browser but with the query parameters out of order.

jcscottiii (approved):
LGTM. I had one thought for the future

PR #4454, opened 2024-10-18
Install gcloud in devcontainer

jcscottiii:
Similar to https://github.com/GoogleChrome/webstatus.dev/blob/main/.devcontainer/Dockerfile


past (approved):
Confirmed that this works. Thanks!

Issue #4455, opened 2024-10-18
Implement gcs_helpers w/ local emulator


jrobbins:
Write a helper file to allow us to access Google Cloud Storage.  

- [x] It should store attachment content and mime-type.  
- [x] It should create thumbnails for images at the time of attachment creation for image attachments.
- [x] It should generate URLs that can redirect the user to the viewable content or thumbnail.  (No download deposition for now)  
- [x] It should include an emulator for developers who are running locally.

PR #4459, opened 2024-10-19
Add GCS helper functions.

jrobbins:
This should resolve issue #4455.  This is based on the Monorail code that stores attachments in Google Cloud Storage and serves that content directly from GCS.

There is no GCS emulator, so I am using NDB to store attachments when running locally.

For example, when we add a field named `screenshots` to `FeatureEntry`, that field will be a repeated string field.  The uploaded screenshot images will go into bucket `cr-status-staging.appspot.com-screenshots` and have blobs with names that are UUIDs, like `'a031-e031-0302-2c71'`.  A feature entry that has a screenshot might have a single string in the `screenshots` field with a value like `'/attachments/screenshots/a031-e031-0302-2c71'`.  When the user views the feature detail page, it would include a thumbnail like ` and that would be inside a link to
`/attachments/screenshots/a031-e031-0302-2c71`.  Those images would be served by a new handler that returns the image bytes directly when running locally or gives a redirect to GCS when running on GAE.

### DIFF
--- a/framework/gcs_helpers.py
+++ b/framework/gcs_helpers.py
@@ -1,0 +1,186 @@
+# Copyright 2024 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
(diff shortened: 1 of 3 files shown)


jcscottiii:
We used https://github.com/fsouza/fake-gcs-server as a GCS emulator and it worked well

jrobbins:
> We used https://github.com/fsouza/fake-gcs-server as a GCS emulator and it worked well in webstatus.dev 

Good suggestion, but since we are using so little of GCS's functionality, I think that it is simpler to use NDB for local dev and avoid adding a dependency.

jrobbins:
I'm going to try an alternative approach.  I'm closing this PR for now but there's a chance I could re-open it.

jcscottiii:
I also found https://github.com/oittaa/gcp-storage-emulator

This might be easier to use for this project since everything can be installed via pip.

Varryme (commented):
resolved

PR #4482, opened 2024-10-25
Show a link for shipping next year after Sept 1.

jrobbins:
A chat with Panos, Chris H, and Nandu led to the conclusion that we should show the next year starting each September.  In other words, January through August, it will only show one link for the current year, but in September through the end of December, it will show two links.


KyleJu (approved):

jrobbins:
Hmm, since this is a date-based visible change, it is going to cause playwright failures every year in January and September.  There is a way to redact a rectangle in the UI, but this change has a variable-sized impact on the UI.  I suppose that we have until January to think about it :).

PR #4499, opened 2024-10-30
Use Pillow and the uploaded mime_type.

jrobbins:
This is a follow-up to #4473.  In that review, it was suggested to limit the upload size and use the Pillow library.  

In this PR:
* Use Pillow instead of the GAE legacy image library
* Specify a maximum upload size (bigger than the max image size in case we ever combine images with other data)
* Remove the filename-to-mime-type logic and use the mime-type that the browser supplies.  (I think that was not available when the Monorail upload code was first written.)

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -798,6 +798,10 @@ def FlaskApplication(import_name, routes, pattern_base='', debug=False):
   # Flask apps also have a debug setting that can be used to auto-reload
   # template source code. TODO: investigate using the setting.
 
+
+  # Reject any huge POSTs
+  app.config['MAX_CONTENT_LENGTH'] = 16 * 1024 * 1024  # 16 MB
+
   # Set the CORS HEADERS.
   CORS(app, resources={r'/data/*': {'origins': '*'}})
 
--- a/internals/attachments.py
+++ b/internals/attachments.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import io
 import logging
-from google.appengine.api import images
+from PIL import Image
 from google.cloud import ndb  # type: ignore
 from typing import Tuple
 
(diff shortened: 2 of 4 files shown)


framework/basehandlers.py
```diff
@@ -798,6 +798,10 @@ def FlaskApplication(import_name, routes, pattern_base='', debug=False):
   # Flask apps also have a debug setting that can be used to auto-reload
   # template source code. TODO: investigate using the setting.
 
+
+  # Reject any huge POSTs
+  app.config['MAX_CONTENT_LENGTH'] = 16 * 1024 * 1024  # 16 MB
```
jcscottiii, line 803: This integer and MAX_ATTACHMENT_SIZE should reference the same constant instead of declaring 16MB twice.

internals/attachments.py
```diff
@@ -76,18 +62,17 @@ def store_attachment(feature_id: int, content: bytes, mime_type: str) -> str:
...
+      im.save(thumb_buffer, 'PNG')
+      thumb_content = thumb_buffer.getvalue()
     except Exception as e:
       # Do not raise exception for incorrectly formed images.
-      # See https://bugs.chromium.org/p/monorail/issues/detail?id=597 for more
-      # detail.
       logging.exception(e)
     if thumb_content:
       attachment.thumbnail = thumb_content
+      logging.info('Thumbnail is %r bytes', len(thumb_content))
```
jcscottiii, line 75: Did you mean to leave this log statement here?

jcscottiii (approved):
LGTM with one nit and one question

framework/basehandlers.py
jrobbins: They are not the same value.  The overall post can be bigger than the max for one attachment.

internals/attachments.py
jrobbins: Yes, I'd like to keep this for a while to make sure that things are working as intended.

framework/basehandlers.py
jcscottiii: Whoops, you're right! Could you add a comment to MAX_ATTACHMENT_SIZE that MAX_ATTACHMENT_SIZE should always be less than MAX_CONTENT_LENGTH ? It would prevent this case:
- Developer A: Increases MAX_ATTACHMENT_SIZE above MAX_CONTENT_LENGTH. But they don't increase MAX_CONTENT_LENGTH since it's in another file.
- Developer A: Expects a larger file to be accepted but requests get rejected due to the old MAX_CONTENT_LENGTH

PR #4509, opened 2024-10-31
OT requested clean up/changes

### LINKED ISSUE #4156: No way to modify obsolete "Origin Trial documentation link" in the "Intent to Extend Experiment" e-mail
In my feature, the Intent to Extend Experiment e-mail contains an obsolete "Origin Trial documentation link" field.

I couldn't find a way to modify the field in the Chromestatus entry. It's not under the "Origin trial" development stage and I couldn't find it anywhere else either.

My feature:

https://chromestatus.com/feature/5100466238652416

Obsolete field in the e-mail:

chromestate2

DanielRyanSmith:
This fixes #4156 

This allows users to update the OT documentation link that surfaces in Intent to Ship emails.

Additionally, this adds the "Origin trial Name" to intent templates if it exists.


pages/intentpreview_test.py
```diff
@@ -248,14 +248,17 @@ class IntentEmailPreviewTemplateTest(testing_config.CustomTestCase):
...
+        id=1, name='feature one', summary='sum',
+        owner_emails=['user1@google.com'],
         category=1, intent_stage=core_enums.INTENT_IMPLEMENT)
     # Hardcode the key for the template test
     self.feature_1.key = ndb.Key('FeatureEntry', 234)
     self.feature_1.wpt = True
     self.feature_1.wpt_descr = 'We love WPT!'
     self.feature_1.put()
 
+    self.stage_1 = Stage(feature_id=1, stage_type=150, ot_display_name="Test 123")
```
jrobbins, line 260: Is `Test 123` supposed to show up in the new golden, because I don't see it there.

DanielRyanSmith: Sorry about this, I shipped this over to you in a hurry without vetting it and it seems there are some issues that need to be handled. I'll re-request a review once this is taken care of.

jrobbins (approved):

DanielRyanSmith:
After some digging, it seems the Playwright tests were hanging because the OT documentation field is "required" by default (since it used to only be on the OT creation form page). I've added a new variation of the same field that is not required for the normal field-editing pages.

Issue #4355, opened 2024-09-12
Move the action in escalation emails closer to the top


past:
**Is your feature request related to a problem? Please describe.**
Feedback suggests that people sometimes scan the escalation emails and if the information looks correct assume there is nothing for them to do. 

**Describe the solution you'd like**
Perhaps moving the action button towards the top, before the explanation of the current status (like a tl;dr) would anchor their mind towards acting.

**Additional context**
Further confusion arises when clicking the button takes you to a page with much more information than the email, which increases the mental overhead of confirming the data. Maybe we should make the button click take you to a simple interstitial or just a "success" message?

KyleJu:
@jrobbins I will make progress on this issue this week since it is a part of my KR

Varryme:
I found the wiki feed that deprecated I took care it

KyleJu:
A few ideas in the email:
1. Title can be changed to: `[Action requested] Verify your webstatus feature before launch - Insecure form warnings on iOS`. Emails rarely use the format of `[webstatus] ESCALATED: [Action requested]` in email subjects.
2. Move the `verify feature accuracy` to the top, and replace the position of the `Insecure form warnings on iOS` link
3. Show the `accurate_as_of date` in the email to highlight when the last update was
4. Highlight it is scheduled to ship **soon** in the first sentence
5. Merge the 4 boxes of texts into two, with one being action items on top and another one for brief explanations. It might reduce the mental overhead to read the email and prompt more actions

![Screenshot 2024-10-31 3 29 07 PM](https://github.com/user-attachments/assets/50e1d6d1-e32a-4176-986f-a2cb3aca28ec)


In the feature accuracy page
1. Move the `confirm accuracy` to the top of the page. Authors can avoid reading through the page and verify the accuracy right away if they are confident about the information.

![Screenshot 2024-10-31 3 38 15 PM](https://github.com/user-attachments/assets/c39114b8-4a68-4b0b-836e-b0e348bf1afc)

KyleJu:
This is complete

PR #4510, opened 2024-10-31
 Simplify feature accuracy notification email

KyleJu:
I have a few ideas in https://github.com/GoogleChrome/chromium-dashboard/issues/4355#issuecomment-2450979970. Overall, the idea is to simplify the sentences and reduce the overhead of reading it. Instead highlight the actions on top.

Remove use of extra []s and change wording in email subjects:
- From `[webstatus] [Action requested] Update Saved queries in sharedStorage.selectURL`  to  `[webstatus] Action requested - Verify Saved queries in sharedStorage.selectURL` 
- When escalation, from `[webstatus] ESCALATED: [Action requested] Update Saved queries in sharedStorage.selectURL
`  to  `[webstatus] Escalation request - Verify Saved queries in sharedStorage.selectURL` 

Rearrange the email body:
- Move `Verify feature accuracy`  button to the top
- Merge four sections into three 
- Highlight the feature is launched soon in the first sentence

Regular accuracy notification (Note that the email subject in prod will start with [webstatus]):
Screenshot 2024-11-01 7 44 06 PM

Escalation email:
Screenshot 2024-11-01 7 46 01 PM

Before this change:
Screenshot 2024-11-01 7 48 32 PM

### Other ideas
On top of it, wdyt if I grey the section under `id=why-triggered`? Normally the explanation for "why I receive this email" is always greyed in the email

Screenshot 2024-10-31 4 59 48 PM

### DIFF
--- a/internals/reminders.py
+++ b/internals/reminders.py
@@ -40,6 +40,7 @@ CHROME_RELEASE_SCHEDULE_URL = (
 WEBSTATUS_EMAIL = 'webstatus@google.com'
 CBE_ESCLATION_EMAIL = 'cbe-releasenotes@google.com'
 STAGING_EMAIL = 'jrobbins-test@googlegroups.com'
+EMAIL_SUBJECT_PREFIX = 'Action requested'
 
 
 def get_current_milestone_info(anchor_channel: str):
@@ -101,7 +102,10 @@ def build_email_tasks(
     html = render_template(body_template_path, **body_data)
     subject = subject_format % fe.name
     if is_escalated:
-      subject = f'ESCALATED: {subject}'
+      if EMAIL_SUBJECT_PREFIX in subject:
+        subject = subject.replace(EMAIL_SUBJECT_PREFIX, "Escalation request")
+      else:
+        subject = f'ESCALATED: {subject}'
     recipients = choose_email_recipients(fe, is_escalated)
     for recipient in recipients:
       email_tasks.append({
@@ -223,7 +227,7 @@ class FeatureAccuracyHandler(AbstractReminderHandler):
   # This grace period needs to be consistent with
   # ACCURACY_GRACE_PERIOD in client-src/elements/utils.ts.
   ACCURACY_GRACE_PERIOD = timedelta(weeks=4)
-  SUBJECT_FORMAT = '[Action requested] Update %s'
+  SUBJECT_FORMAT = EMAIL_SUBJECT_PREFIX + ' - Verify %s'
   EMAIL_TEMPLATE_PATH = 'accuracy_notice_email.html'
   FUTURE_MILESTONES_TO_CONSIDER = 2
   MILESTONE_FIELDS = [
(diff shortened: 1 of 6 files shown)


DanielRyanSmith:
> 1. Where does the [webstatus] prefix come from in the email subject?

I'm not certain here, but this could be a byproduct of webstatus@google.com Google group sending these emails to our inbox. E.g. [None of the actual emails have the prefix in the group](https://groups.google.com/a/google.com/g/webstatus)

> 2. Is there a way to update test_build_email_tasks_escalated_feature_accuracy.html test automatically?

Yes - [uncommenting the line here](https://github.com/GoogleChrome/chromium-dashboard/blob/main/internals/reminders_test.py#L200-L202) and then running the test will update the contents of the file that is being compared (be sure to re-comment the line back so it's not trying to rewrite the contents each time)

templates/accuracy_notice_email.html
```diff
@@ -34,33 +46,5 @@
...
-    {% include "estimated-milestones-table.html" %}
-  </p>
-  <p>
-    Your feature entry is an important resource for
-    cross functional teams that help drive adoption of new features and
-    enterprise IT admins who might be affected by web platform changes.
-  </p>
-
-  <p>
-    We need to know whether your plans are changing or staying
```
past, line 50: How about "The Blink launch process needs to know" instead of "We need to know"? It's not just our team that needs this information, as we're constantly reminded by others, and this framing could add to the sense of urgency and legitimacy that the notification should have.

past (approved):
This is much improved, thanks! I didn't look at the code too closely as I assume Jason will, but I have one more nit: can we refer to the "Chromestatus team" instead of the "Webstatus team" everywhere, including the email subject if possible?

Nowadays with webstatus.dev being a separate thing, the confusion becomes a bit more concerning.

templates/accuracy_notice_email.html
KyleJu: Addressed the comment. The email subject prefix [webstatus] comes from somewhere else. Let me chat with Jason tomorrow

internals/reminders.py
```diff
@@ -101,7 +102,10 @@ def build_email_tasks(
     html = render_template(body_template_path, **body_data)
     subject = subject_format % fe.name
     if is_escalated:
-      subject = f'ESCALATED: {subject}'
+      if EMAIL_SUBJECT_PREFIX in subject:
+        subject = subject.replace(EMAIL_SUBJECT_PREFIX, "Escalation request")
```
jrobbins, line 106: Nit: use single quotes.

templates/accuracy_notice_email.html
jrobbins: I still prefer "we".  It's the simplest way to write it, which allows the emphasis to be on the action that is needed.

jrobbins:
> On top of it, wdyt if I grey the section under `id=why-triggered`?

I'd prefer not to use gray text.  The fact that it comes last should be enough.

jrobbins (approved):
LGTM after one nit.

templates/accuracy_notice_email.html
KyleJu: Agree with "we" here. It also sounds more personal, which always works better in a request

internals/reminders.py
KyleJu: Should have been caught by the lint! I will add it to the Thursday meeting

PR #4536, opened 2024-11-06
Detect revised Groups footer link.

jrobbins:
This should report a problem reported to us via chat by one of the API owners: recent review requests were detected by our server, but the link to the intent thread was not recorded.  

The reason was that the Google Groups footer format changed on Nov 1.
Old footer included `To view this discussion on the web visit https://...`
New footer includes `To view this discussion visit https://...`

In this PR:
* Change the regex and associated tests

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -124,11 +124,11 @@ def detect_gate_id(body) -> int | None:
 
 
 THREAD_LINK_RE = re.compile(
-    r'To view this discussion on the web visit\s+'
+    r'To view this discussion visit\s+'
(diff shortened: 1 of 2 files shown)


jcscottiii (approved):
LGTM with one question: With the recent change to the template, do we need to support both the old and new regexs for a period of time?

jrobbins:
> LGTM with one question: With the recent change to the template, do we need to support both the old and new regexs for a period of time?

I'm seeing the new footer consistently on new messages.  Of course, the Groups team could always change it again, but I don't know that they would change it back.  If anything, I think they should add a comma before "visit", but the old one was missing the comma too, so they probably never will.

PR #4541, opened 2024-11-07
Implement attachment_api and serving of attachments

jrobbins:
This PR implements the server side logic for uploading and serving attachments.  Specifically, 
* Uploading immediately stores the attachment and returns a URL that can be placed into a UI field that has a list of URLs (some of which are stored on our server and some of which might link to other sites)
* Serving checks that the request was made with the expected URL (including a domain that does not have our main domain cookies) and then returns the attachment content or thumbnail

Also in the PR,
* Added an openapi declaration, and the associated generated files.  This is our first openapi entry point that accepts binary data.
* Changed the thumbnail size to match the size of images in the existing enterprise release notes page.
* Fixed a comment formatting error in permission.py


api/attachments_api.py
```diff
@@ -0,0 +1,94 @@
...
+    """Handle POST requests to create a single feature."""
+    feature_id = kwargs.get('feature_id', None)
+
+    # Validate the user has edit permissions and redirect if needed.
+    redirect_resp = permissions.validate_feature_edit_permission(
+        self, feature_id)
+    if redirect_resp:
+      self.abort(403, msg='User lacks permission to edit')
+
+    files = kwargs.get('mock_files', self.request.files)
```
jcscottiii, line 38: We should not have this logic that looks for mock_files in the keywords in the live code. Instead, we should adjust the tests. More about this in the test file.
```suggestion
    files = self.request.files
```

api/attachments_api_test.py
```diff
@@ -0,0 +1,175 @@
...
+        self.handler.do_post(feature_id=self.feature_id)
+
+  def test_do_post__valid_file(self):
+    """With a valid user and valid request, we store the attachment."""
+    testing_config.sign_in('owner@chromium.org', 111)
+    mock_files = {'uploaded-file': testing_config.Blank(
+        filename='hello_attach.txt',
+        read=lambda: b'hello attachments!',
+        mimetype='text/plain')}
+    with test_app.test_request_context(self.request_path):
```
jcscottiii, line 96:
```suggestion
    mock_file = (io.BytesIO(b'hello attachments!'), 'test.txt')
    with test_app.test_request_context(path=self.request_path, data={'uploaded-file': mock_file}):
```

internals/attachments.py
```diff
@@ -105,11 +106,18 @@ def get_attachment(feature_id: int, attachment_id: int) -> Attachment|None:
...
 
-def get_attachment_uri(attachment: Attachment) -> str:
-  """Return the URI path that will serve tis attachment."""
-  uri = '/feature/%r/attachment/%r' % (
+def get_attachment_url(attachment: Attachment) -> str:
+  """Return the URL path that will serve this attachment."""
+  if settings.DEV_MODE or settings.UNIT_TEST_MODE:
+    origin = settings.SITE_URL
+  else:
+    digits = attachment.key.integer_id() % 1000
```
jcscottiii, line 114: Could you add a comment as to why we are doing modulo 1000?

jcscottiii (commented):
A few comments to consider.

Also, I see that you have API tests that:
1. assert the POST behavior
2. assert serving from an entity manually added to the database.

You should add a test that does the POST and serves the entity afterwards. This can come later in a playwright test when we have the UI. But I just wanted to bring that up now.

api/attachments_api.py
```diff
@@ -0,0 +1,94 @@
...
+from framework import permissions
+from internals import attachments
+
+
+class AttachmentsAPI(basehandlers.EntitiesAPIHandler):
+  """Features are the the main records that we track."""
+
+  @permissions.require_create_feature
+  def do_post(self, **kwargs) -> dict[str, str]:
+    """Handle POST requests to create a single feature."""
```
jcscottiii, line 29: Also, I think you meant to adjust this comment too

api/attachments_api.py
```diff
@@ -0,0 +1,94 @@
...
+      return redirect_response
+
+    if is_thumb and attachment.thumbnail:
+      content = attachment.thumbnail
+      headers = self.get_headers()
+      headers['Content-Type'] = 'image/png'
+    else:
+      content = attachment.content
+      headers = self.get_headers()
+      headers['Content-Type'] = attachment.mime_type
```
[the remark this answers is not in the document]
jrobbins: Done.

api/attachments_api.py
```diff
@@ -0,0 +1,94 @@
...
+
+from chromestatus_openapi.models import AddAttachmentResponse
+
+from framework import basehandlers
+from framework import permissions
+from internals import attachments
+
+
+class AttachmentsAPI(basehandlers.EntitiesAPIHandler):
+  """Features are the the main records that we track."""
```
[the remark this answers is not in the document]
jrobbins: Fixed.

api/attachments_api.py
jrobbins: Done.

api/attachments_api_test.py
jrobbins: Thanks for that suggestion.  I had spent half the day struggling with passing in a complete POST body as a binary string.

internals/attachments.py
jrobbins: Done.

jrobbins:
> A few comments to consider.
> 
> Also, I see that you have API tests that:
> 
> 1. assert the POST behavior
> 2. assert serving from an entity manually added to the database.
> 
> You should add a test that does the POST and serves the entity afterwards. This can come later in a playwright test when we have the UI. But I just wanted to bring that up now.

I went ahead and added a round-trip unit test now.

jcscottiii (approved):

PR #4558, opened 2024-11-13
Treat intent threads as new based on gate.status.

jrobbins:
This should resolve a problem that Panos brought up in chat today.  

When the review functionality was first designed, there was no `gate.state` variable, instead the state of the gate was always derived from the set of votes on that gate.  So, in detect_intent, an incoming email message is considered to be the start of an intent thread if there were no votes on its gate.  The problem with that is that a reviewer can manually vote REVIEW_STARTED or a vote of NO_RESPONSE can be entered for an assigned reviewer.  So, the list of votes might not be empty even before the intent thread starts.

Now that we have `gate.state` and it has been backfilled for all existing gates, we can use that to make a better decision about when an incoming email message is actually the start of an intent thread.  If the gate is in the initial PREPARING state, regardless of Votes, then we treat the thread as new, and so the gate becomes REVIEW_REQUESTED.  If the gate was anything else, we treat the message as an ongoing discussion, which basically means that we leave the gate state as-is.

### DIFF
--- a/internals/detect_intent.py
+++ b/internals/detect_intent.py
@@ -169,10 +169,9 @@ def is_lgtm_allowed(from_addr, feature, gate_info):
   return allowed
 
 
-def detect_new_thread(gate_id: int) -> bool:
-  """Return True if there are no previous approval values for this gate."""
-  existing_votes = Vote.get_votes(gate_id=gate_id)
-  return not existing_votes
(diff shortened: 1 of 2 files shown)


KyleJu (approved):

vonidorboy24:
Why aren't i receiving messages to my devics

PR #4567, opened 2024-11-18
Add webstatus group contacts for escalated feature accuracy emails

KyleJu:
Add webstatus group contacts for escalated feature accuracy emails. It was missing in the recent change, https://github.com/GoogleChrome/chromium-dashboard/pull/4528

### DIFF
--- a/internals/reminders.py
+++ b/internals/reminders.py
@@ -57,25 +57,27 @@ def get_current_milestone_info(anchor_channel: str):
 def choose_email_recipients(
     feature: FeatureEntry, is_escalated: bool, is_accuracy_email: bool) -> list[str]:
   """Choose which recipients will receive the email notification."""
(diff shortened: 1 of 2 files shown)


internals/reminders.py
```diff
@@ -57,25 +57,27 @@ def get_current_milestone_info(anchor_channel: str):
 def choose_email_recipients(
     feature: FeatureEntry, is_escalated: bool, is_accuracy_email: bool) -> list[str]:
   """Choose which recipients will receive the email notification."""
+  ws_group_emails = []
+  if is_escalated:
```
jrobbins, line 61: I think we want to include these mailing lists regardless of is_escalated. Johnny was checking on how many regular notifications went out each week, not just escalations.

jrobbins (changes requested):

PR #4579, opened 2024-11-21
Implement attachments widget with thumbnails

jrobbins:
This is a step toward hosting screenshots on the site.  It enhances the current approach where the user enters URLs for screenshots by dynamically adding `` elements to display a thumbnail-sized version of the linked image.

In this PR:
* Define a new form element chromedash-attachments, which contains a chrome dash-textarea and a set of thumbnails.
* Add dispatch logic to our form field container class to create a chromedash-attachments element when specified.
* Change the form spec for the screenshots field to use a widget of type `attachments`.
* Remove the existing logic that loaded images and displayed an error message in case of an unexpected content type.  Now the image either shows up or a broken-image icon shows up.

In an upcoming PR, I will add an "upload" button that immediately uploads and image and appends its URL in the text box.  So, users will be able to use a mix of screenshots hosted on our server and screenshots hosted elsewhere.

### DIFF
--- a/client-src/elements/chromedash-attachments.ts
+++ b/client-src/elements/chromedash-attachments.ts
@@ -1,0 +1,104 @@
+import {LitElement, css, html, nothing} from 'lit';
+import {customElement, property} from 'lit/decorators.js';
+import {createRef, ref, Ref} from 'lit/directives/ref.js';
+import {styleMap} from 'lit-html/directives/style-map.js';
+import {ChromedashTextarea} from './chromedash-textarea.js';
+import {URL_REGEX} from './form-field-specs.js';
+
+const DEFAULT_PLACEHOLDER = 'https://...\nhttps://...';
+const DEFAULT_SPLIT_PATTERN = String.raw`\s+`;
+
+@customElement('chromedash-attachments')
+export class ChromedashAttachments extends LitElement {
+  textareaRef: Ref<ChromedashTextarea> = createRef();
+
+  @property({type: String})
+  name;
+  @property({type: String})
+  value;
+  @property({type: String})
+  pattern;
+  @property({type: String})
+  chromedash_single_pattern = URL_REGEX;
+  @property({type: String})
+  chromedash_split_pattern = DEFAULT_SPLIT_PATTERN;
+
+  // Note: Form element components cannot have their own styles.
+
+  // Must render to light DOM, so sl-form-fields work as intended.
+  createRenderRoot() {
+    return this;
+  }
+
+  validate() {
+    if (this.textareaRef?.value?.validate) {
+      this.textareaRef.value.validate();
+    }
+  }
+
+  handleFieldUpdated(e) {
+    let fieldValue: string = e.target.value;
+    this.value = fieldValue;
+  }
+
+  renderOneThumbnail(url) {
+    if (url == '') {
+      return nothing;
+    }
+
+    const linkStyles = {
+      display: 'inline-block',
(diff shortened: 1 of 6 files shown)


client-src/elements/chromedash-attachments.ts
```diff
@@ -0,0 +1,102 @@
...
+  @property({type: String})
+  value;
+  @property({type: String})
+  pattern;
+  @property({type: String})
+  chromedash_single_pattern = URL_REGEX;
+  @property({type: String})
+  chromedash_split_pattern = DEFAULT_SPLIT_PATTERN;
+
+  // Note: Form element components cannot have their own styles.
```
KyleJu, line 26: This comment doesn't belong here?

client-src/elements/chromedash-attachments.ts
```diff
@@ -0,0 +1,102 @@
...
+      border: `2px solid var(--link-color)`,
+      cursor: `zoom-in`,
+    };
+    const imgStyles = {
+      maxHeight: '150px',
+      maxWidth: '200px',
+    };
+
+    return html`
+    <a href=${url} target="_blank" style=${styleMap(linkStyles)}>
```
KyleJu, line 62: There should be indentation

client-src/elements/chromedash-attachments.ts
```diff
@@ -0,0 +1,102 @@
...
+    return items.map(url => this.renderOneThumbnail(url));
+  }
+
+  render() {
+    return html`
+      <chromedash-textarea
+        ${ref(this.textareaRef)}
+        size="small"
+        name=${this.name}
+        .value=${this.value}
```
KyleJu, line 82:
```suggestion
        value=${this.value}
```
Should be this an attribute binding?

client-src/elements/chromedash-form-field.ts
```diff
@@ -345,6 +346,19 @@ export class ChromedashFormField extends LitElement {
...
         </chromedash-textarea>
       `;
+    } else if (type === 'attachments') {
+      fieldHTML = html`
+        <chromedash-attachments
+          ${ref(this.updateAttributes)}
+          name="${fieldName}"
+          id="id_${this.name}"
+          size="small"
+          .value=${fieldValue}
```
KyleJu, line 356: ditto?
```suggestion
          value=${fieldValue}
```

client-src/elements/chromedash-attachments.ts
jrobbins: It's in the location where we would normally have the CSS styles.  In fact, I spent some time adding lit-css styles, only to realize that they are ignored.   I could put the comment on the next set of lines that avoids creating a renderRoot or on the lines that do the styleMap, but this is the location that would have helped me avoid wasted time.

client-src/elements/chromedash-attachments.ts
jrobbins: Done.  I had followed the example from existing code, but the `.` is not actually needed.

client-src/elements/chromedash-form-field.ts
jrobbins: Done.

client-src/elements/chromedash-attachments.ts
jrobbins: Done.  I had thought lint-fix would do that, but apparently not.

KyleJu (approved):

PR #4607, opened 2024-12-04
Implement screenshot upload button

jrobbins:
This PR should fully expose the screenshot upload functionality to users who are editing enterprise features.

In this PR:
* The existing `chromedash-attachments` element is enhanced to offer an upload button.  It consists of a hidden `<input type=file>` element and an associated `sl-button` that calls `click` on the hidden form field.  The reason for this is that the `<input type=file>` element cannot be easily styled to simply look like a button.
* A change handler is called when the user selects a file.  It calls the existing API endpoint to immediately upload the selected file. It appends the resulting URL to the end of the `textarea` that holds a list of URLs.
* When running locally, uploaded images are served from `http://127.0.0.1:7777`, which was failing form field URL validation, so we add that as a possibility in the URL validation regex.
* Changed the form field label to correct double plural and added brief help.
* Set max attachment size to 1MB because that is the maximum for an datastore entity (the size of a field is unlimited only up to that max).


client-src/elements/form-field-specs.ts
```diff
@@ -109,9 +109,12 @@ export function resolveFieldForFeature(
...
  * Also removing ! and % because they have special meaning for some older email routing systems. */
 const USER_REGEX: string = String.raw`[A-Za-z0-9_#$&*+\/=?\{\}~^.\-]+`;
-const DOMAIN_REGEX: string = String.raw`(([A-Za-z0-9\-]+\.)+[A-Za-z]{2,6})`;
+const DOMAIN_NAME_REGEX: string = String.raw`(([A-Za-z0-9\-]+\.)+[A-Za-z]{2,6})`;
+const LOCALHOST_REGEX: string = String.raw`(localhost|127\.0\.0\.1)`;
+const DOMAIN_REGEX: string =
+  '(' + DOMAIN_NAME_REGEX + '|' + LOCALHOST_REGEX + ')';
 
-const EMAIL_ADDRESS_REGEX: string = USER_REGEX + '@' + DOMAIN_REGEX;
+const EMAIL_ADDRESS_REGEX: string = USER_REGEX + '@' + DOMAIN_NAME_REGEX;
```
KyleJu, line 117: One question - how is this regex being used and how is it related to the LOCALHOST_REGEX?

KyleJu (approved):

client-src/elements/form-field-specs.ts
jrobbins: When developing locally, attachments are served from the local host rather than a regular domain name.   So, the URL validation regex that was looking for something like `google.com` was rejecting `127.0.0.1`.  I implemented it as a separate regex because we don't need that for email addresses.

PR #4613, opened 2024-12-04
Actually display thumbnails rather than full images.

jrobbins:
This PR makes the gallery of small images actually request the thumbnail-sized images for self-hosted images.  

In the case where a thumbnail does not exist for whatever reason, the full-sized image is returned, so the client-side can always safely request a thumbnail.


KyleJu:
CI has lint issues

KyleJu (approved):

Issue #4587, opened 2024-11-25
Deleted entry still showing in ChromeStatus database


npm1:
**Describe the bug**
https://chromestatus.com/feature/5169880146640896 was deleted but still shows in the database. The new entry is https://chromestatus.com/feature/5134341540413440

**Expected behavior**
The deleted entry should not show up in the database?

**Additional context**
This is tripping up our privacy reviewer bot.

jrobbins:
This change is live now.

PR #4619, opened 2024-12-06
Exclude deleted features from API responses.

jrobbins:
This should resolve issue #4587.

Basically, we should exclude deleted feature info from API responses, unless it is specifically requested.

In this PR:
* Delete old duplicate feature deletion code that was left over from when we were doing double writes for `Feature` and `FeatureEntry` two years ago.
* When requesting the gates of a feature, reply with an empty list if the feature was deleted, unless the request has query string parameter include_deleted=1.
* Make our API always request gates with include_deleted=1 so that a user who knows the URL of a deleted feature entry can view a complete feature entry and possibly decide to undelete it.
* Exclude deleted features from the `features_v2.json` response.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -352,7 +352,7 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
     """Delete the specified feature."""
     # TODO(jrobbins): implement undelete UI.  For now, use cloud console.
     feature_id = kwargs.get('feature_id', None)
-    feature = self.get_specified_feature(feature_id=feature_id)
+    feature: FeatureEntry = self.get_specified_feature(feature_id=feature_id)
     if feature is None:
       return {'message': 'ID does not match any feature.'}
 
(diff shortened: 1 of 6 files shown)


api/features_api.py
```diff
@@ -352,7 +352,7 @@ def do_delete(self, **kwargs) -> dict[str, str]:
     """Delete the specified feature."""
     # TODO(jrobbins): implement undelete UI.  For now, use cloud console.
     feature_id = kwargs.get('feature_id', None)
-    feature = self.get_specified_feature(feature_id=feature_id)
+    feature: FeatureEntry = self.get_specified_feature(feature_id=feature_id)
     if feature is None:
       return {'message': 'ID does not match any feature.'}
```
DanielRyanSmith, line 357: Nit: this check is not needed, since get_specified_feature() will abort if the feature is not found.

internals/feature_helpers.py
```diff
@@ -476,6 +476,7 @@ def get_features_by_impl_status(limit: int | None=None, update_cache: bool=False
     feature_list = []
     for section in query_results:
       if len(section) > 0:
+        section = [f for f in section if not f.deleted]
         section = [f for f in section if f.feature_type != FEATURE_TYPE_ENTERPRISE_ID]
```
DanielRyanSmith, line 480: Nit: these list comprehensions can be combined
```suggestion
        section = [f for f in section if not f.deleted]
        section = [f for f in section if not f.deleted
                   and f.feature_type != FEATURE_TYPE_ENTERPRISE_ID]
```

DanielRyanSmith (approved):
LGTM, small nits

api/features_api.py
jrobbins: Good point.  Gone.

internals/feature_helpers.py
jrobbins: True, but I kind of like having them separate.  Skipping deleted features seems like a separate idea than getting in to the details of the feature types.

DanielRyanSmith: That's fine with me - no worries about making this change then. 🙂

PR #4624, opened 2024-12-08
OT endpoints use new `ot_use_counter_bucket_number` field

DanielRyanSmith:
This field is needed for the purpose of WebFeature and WebDXFeature use counter monitoring for origin trials. Its value will be inferred based on the number value that use counters are listed as in [web_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/web_feature.mojom;l=1?) and [webdx_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom;l=1).


KyleJu:
> This field is needed for the purpose of WebFeature and WebDXFeature use counter monitoring for origin trials. Its value will be inferred based on the number value that use counters are listed as in [web_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/web_feature.mojom;l=1?) and [webdx_feature.mojom](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom;l=1).

DanielRyanSmith:
> How often are the OT endpoints used? Should we fetch these files every time vs store them through a separate cron job daily?

These are only fetched when an origin trial creation form is submitted, which is usually less than once a day, so I think it's fine to keep fetching them as needed. Additionally, the files will block submission if the contents are not up-to-date, so it's useful to get the most recent version if a feature owner just landed the requirements for submitting the origin trial

api/origin_trials_api.py
```diff
@@ -47,6 +48,48 @@ def get_chromium_file(url: str) -> str:
     return  b64decode(conn.read()).decode('utf-8')
 
 
+def get_chromium_files_for_validation() -> dict:
+  """Get all chromium file contents stored in a dictionary"""
+  chromium_files = {}  # Chromium source file contents.
+  with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
```
KyleJu, line 54: Do we need 5 workers? CHROMIUM_SRC_FILES only has 4 files so far

KyleJu (approved):
LGTM with a comment to consider

api/origin_trials_api.py
DanielRyanSmith: It seems fine to drop it to 4 to match 🙂

Issue #4634, opened 2024-12-14
Track SLO for both initial response and resolution, not counting time spend in NEEDS_WORK


jrobbins:
Right now we have an SLO for a reviewer's initial response based on the `requested_on` timestamp and the current time vs. a limit `slo_initial_response` defined in the `GateType`.  When a reviewer gives any vote or comment, we set `responded_on` at that time.  

Now, we will also calculate a time for resolution and set `resolved_on` when a gate becomes NA, APPROVED, or DENIED.  That will be compared against a new `slo_resolution` limit defined in the `GateType`.

Unlike the initial response, the resolution can be delayed by work needing to be done by the feature owner rather than the reviewers.  To track the time spent in the NEEDS_WORK state, we will use new fields `needs_work_started_on` and `needs_work_elapsed`. 

Steps:

- [x] Define fields `resolved_on`, `needs_work_started_on`, and `needs_work_elapsed` and the logic to set them.
- [x] Use the new fields to calculate a value for the elapsed days spent out of a new `slo_resolution` limit.
- [x] Display both SLOs at the top of the gate column, and highlight reviews that are overdue by either measure.
- [x] Send notification emails to reviewers for reviews that have reached the limit for the resolution SLO.

PR #4635, opened 2024-12-14
Track SLO fields for review resolution

jrobbins:
This is progress on #4634.

In this PR:
* Add new Gate fields `resolved_on`, `needs_work_started_on`, and `needs_work_elapsed`.
* Add logic to `record_vote()` to set those fields based on the gate state.

### DIFF
--- a/internals/slo.py
+++ b/internals/slo.py
@@ -63,22 +63,61 @@ def remaining_days(requested_on: datetime.datetime, slo_limit: int) -> int:
   return slo_limit - weekdays_between(requested_on, now_utc())
 
 
-def record_vote(gate: Gate, votes: list[Vote]) -> bool:
+def record_vote(gate: Gate, votes: list[Vote], old_gate_state: int) -> bool:
   """Record a Gate SLO response time if needed.  Return True if changed."""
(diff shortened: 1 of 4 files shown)


internals/slo.py
```diff
@@ -63,22 +63,68 @@ def remaining_days(requested_on: datetime.datetime, slo_limit: int) -> int:
...
+  if latest_state in Vote.RESPONSE_STATES:
+    if gate.responded_on is None:
+      logging.info('SLO: Got reviewer vote as initial response')
+      gate.responded_on = latest_vote.set_on
+      if gate.requested_on is None:
+        logging.info('SLO: Reviewer is the person initiating the review')
+        gate.requested_on = latest_vote.set_on
+      changed = True
+
+  sent_back_for_rework = (
```
KyleJu, line 92: nit: are sent_back_for_rework, sent_back_for_rework and sent_back_for_rework mutually exclusive? If so, maybe we should define right before they are used and return early for readability:
```
sent_back_for_rework = ...
if sent_back_for_rework:
  return true
finished_rework = ...
```

KyleJu (approved):

PR #4643, opened 2024-12-17
Add a confidential field to policies

yanndago:
- Add a new confidential field to features
- This new field makes it impossible to:
    -  access the feature via direct url
    -  find the feature using search
    - See the feature in the release notes
- The only users who can see a confidential feature are admins, owners of a google.com and chromium.org account, feature owners, editors and creators.


client-src/elements/form-field-specs.ts
```diff
@@ -2125,6 +2125,14 @@ export const ALL_FIELDS: Record<string, Field> = {
     take action to continue using some existing functionaity.`,
   },
 
+  confidential: {
+    type: 'checkbox',
+    label: 'Confidential',
+    initial: false,
+    help_text: html `This is a confidential feature that should not be publicly
```
jrobbins, line 2132: It would be more clear if this explained who could view the feature.  E.g., "@google.com accounts, @chromium.org accounts, the feature owners, and editors."

framework/permissions.py
```diff
@@ -44,12 +44,40 @@ def is_google_or_chromium_account(user: User) -> bool:
...
+
+def can_view_feature(user, feature) -> bool:
+  """Return True if the user is allowed to view the given feature."""
+  if not feature:
+    return False
+
+  return not feature.confidential or can_view_confidential_feature(user, feature)
+
+def can_view_confidential_feature(user, feature):
+  ''' Check if the user is an owner, editor, spec mentor, or creator
```
jrobbins, line 62: Nit: all our other docstrings use three double-quotes.

internals/feature_helpers.py
```diff
@@ -45,6 +46,27 @@ def filter_unlisted(feature_list: list[dict]) -> list[dict]:
   return listed_features
 
 
+def filter_confidential_ids(feature_ids: list[int]) -> list[dict]:
+  """Filters a feature id list to return only the ids of features the user should see."""
+  user = users.get_current_user()
+  visible_ids = []
+  for id in feature_ids:
+    if permissions.can_view_feature_with_id(user, id):
```
jrobbins, line 54: This seems like it would be too slow because each call to `can_view_feature_with_id()` makes a separate round-trip to NDB.  It would be better to call `can_view_feature()` given that the calling code has already loaded the `FeatureEntry` objects.  I think the functions that call this need to move the filtering logic ahead of the conversion-to-dict step.  That probably means that we should change the cache to hold `FeatureEntry` objects rather than the converted python dicts.  So, that might best be done with a prequel PR for just that change.

internals/search.py
```diff
@@ -459,7 +459,7 @@ def process_query(
   result_id_set.intersection_update(permission_ids)
   logging.info('got %r result IDs with permissions', len(result_id_set))
 
-  result_id_list = list(result_id_set)
+  result_id_list = feature_helpers.filter_confidential_ids(list(result_id_set))
```
jrobbins, line 462: This line also looks like it would be too slow.  Rather than post-filtering, the permission check should be part of `permission_terms` query that runs in parallel with the user-entered query terms.  There is already logic there for excluding unlisted feature entries.  However, there is a TODO comment for adding back personalized results (the unlisted items that the current user is actually allowed to see).  Would it be OK to also leave that aspect as a TODO for confidential results at first?

jrobbins (approved):
LGTM, but there is still the nit about a docstring using single quotes.

PR #4644, opened 2024-12-18
Add enterprise product category field to features

yanndago:
* This PR adds a new field to categorize enterprise features into basic Chrome browser update, Chrome enterprise core and Chrome enterprise premium feature.
* This field will later be used to categorize the features in the enterprise release notes


yanndago:
Screenshot 2024-12-18 at 09 42 29

internals/core_models.py
```diff
@@ -88,6 +88,7 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
   name = ndb.StringProperty(required=True)
   summary = ndb.TextProperty(required=True)
   category = ndb.IntegerProperty(required=True)
+  enterprise_product_category = ndb.IntegerProperty(required=True, default=ENTERPRISE_PRODUCT_CATEGORY_CHROME_BROWSER_UPDATE)
```
jrobbins, line 91: Could this not be required?  It has a default defined for when a value is not specified.

client-src/elements/form-definition.ts
```diff
@@ -192,6 +195,7 @@ export const FLAT_METADATA_FIELDS: MetadataFields = {
         'name',
         'summary',
         'unlisted',
+        'enterprise_product_category',
```
jrobbins, line 198: Do you mean for this field to show up on web platform features?  I was thinking that it only applied to features with the enterprise feature_type.

yanndago: ha just removed it, it should only be for enterprises

jrobbins (approved):

PR #4650, opened 2024-12-20
Fix confidential and product category fields not being set when creating enterprise feature

yanndago:


jrobbins (approved):
Make sure to fill in the PR description.

PR #4657, opened 2024-12-27
Add web_feature for feature creation backend

KyleJu:
Integrate web-features with [chromestatus.com](http://chromestatus.com/).  Add `web_feature` in the backend.


internals/core_models.py
```diff
@@ -94,6 +94,7 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
   star_count = ndb.IntegerProperty(default=0)
   search_tags = ndb.StringProperty(repeated=True)
   feature_notes = ndb.TextProperty()  # copy from comments
+  web_feature = ndb.TextProperty()
```
jrobbins, line 97: Why not StringProperty?  A TextProperty will not be index in NDB so we would not be able to query on its value.

KyleJu: Oops it should be TextProperty. Thanks for catching it!

KyleJu:
I need to rebase once https://github.com/GoogleChrome/chromium-dashboard/pull/4663 is merged

jrobbins (approved):

PR #4662, opened 2025-01-01
Display SLO for review resolutions.

jrobbins:
Display the SLO info for review resolutions.  Previously we were only displaying SLO info for the reviewers' initial response.  We recently implemented and deployed logic to track more fields related to review resolution.  This PR exposes those fields in the API and UI.

In this PR:
+ Add resolution-related fields to the API converters and openapi spec.
+ Fix some of the client-side types that are expected for these fields.
+ Rewrite the rendering logic.  Instead of displaying one line of SLO info, it will now display two lines.  The first is always the SLO for the initial response.  The second line can give SLO info for resolution, or show that the gate is in the NEEDS_WORK state since a specific date.


KyleJu (approved):

jrobbins:
This relates to #4634.

PR #4664, opened 2025-01-03
Add hyperlink on the bottom

KyleJu:
A follow-up for #4658. Link selected feature group to webstatus.dev.

### UI
Screenshot 2025-01-06 11 49 48 AM

### DIFF
--- a/client-src/elements/chromedash-form-field.ts
+++ b/client-src/elements/chromedash-form-field.ts
@@ -470,12 +470,14 @@ export class ChromedashFormField extends LitElement {
         ? html`
             <tr>
               <td colspan="2" class="webdx">
+                See web feature
                 <a
                   href="https://webstatus.dev/features/${this.value}"
                   target="_blank"
                 >
-                  See feature ${this.value} in Web Platform Status
+                  ${this.value}
                 </a>
+                in webstatus.dev
               </td>
             </tr>
           `
--- a/client-src/elements/chromedash-form-field.ts
+++ b/client-src/elements/chromedash-form-field.ts
@@ -466,6 +466,20 @@ export class ChromedashFormField extends LitElement {
             </tr>
           `
         : nothing}
+      ${this.name === 'web_feature' && this.value
+        ? html`
+            <tr>
+              <td colspan="2" class="webdx">
+                <a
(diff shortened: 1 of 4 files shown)


client-src/elements/chromedash-form-field.ts
```diff
@@ -466,6 +466,20 @@ export class ChromedashFormField extends LitElement {
...
         : nothing}
+      ${this.name === 'web_feature' && this.value
+        ? html`
+            <tr>
+              <td colspan="2" class="webdx">
+                <a
+                  href="https://webstatus.dev/features/${this.value}"
+                  target="_blank"
+                >
+                  See feature ${this.value} in Web Platform Status
```
jrobbins, line 477: The feature name might blend in to this sentence.  E.g., in your PR description screenshot it shows "See feature alerts in Web Platform Status".  While it is true that the user must have selected or typed "alerts" just before that moment, the screenshot could be more clear out-of-context if the name of the feature was visually differentiated.  The cleanest way to do that might be to only link the feature name rather than the whole sentence.

Also, it might be better to refer to it as a "web feature" here.  And maybe refer to the other site as "webstatus.dev" to avoid confusion over the similar site names "Chrome Platform Status" and "Web Platform Status".

jrobbins (approved):
LGTM after considering some tweaks.

client-src/elements/chromedash-form-field.ts
KyleJu: Done! Thanks for the suggestion

KyleJu:
Note to myself: update the [playwright tests](https://github.com/GoogleChrome/chromium-dashboard/blob/main/packages/playwright/tests/__screenshots__/chromedash-guide-new-page_pwtest.js/enter-feature-group/feature-group-firefox-mobile-linux.png) before merge

KyleJu:
> Note to myself: update the [playwright tests](https://github.com/GoogleChrome/chromium-dashboard/blob/main/packages/playwright/tests/__screenshots__/chromedash-guide-new-page_pwtest.js/enter-feature-group/feature-group-firefox-mobile-linux.png) before merge

Tried a number of test setup but still cannot get the link to show. I left a TODO and added  a unit test instead. @jrobbins

PR #4665, opened 2025-01-03
Offer a "Re-request Review" button

jrobbins:
This is another step toward having more fine-grained SLO measurements.  

Already, when a gate is in NEEDS_WORK state, that time is not counted against the reviewer's SLO.   Instead of the old assumption that the reviewers would periodically check on any NEEDS_WORK gates to see if the work had been done, or the assumption that the feature owners would contact the reviewer outside of chromestatus, the new workflow is that the feature owner must press the "Re-request review" button to explicitly end their turn and put the gate back in the reviewer's hands.

In this PR:
* Offer a new button to "Re-request review" when a gate is in the NEEDS_WORK state.  This button simply allows the feature owner to vote REVIEW_REQUESTED again.  

A feature owner voting REVIEW_REQUESTED is already allowed by the permission checks in reviews_api.

There is already logic to handle the vote and do the needed gate state transitions from NEEDS_WORK to REVIEW_REQUESTED, and for that to stop counting time spent in the NEEDS_WORK state so that additional time is counted against the reviewer's SLO again.

There is no new kind of email notification sent to the reviewers because they will already get an email notification as a side-effect of feature owner's vote.

### DIFF
--- a/client-src/elements/chromedash-gate-column.ts
+++ b/client-src/elements/chromedash-gate-column.ts
@@ -550,6 +550,26 @@ export class ChromedashGateColumn extends LitElement {
     `;
   }
 
+  renderReviewStatusNeedsWork() {
+    const rereviewButton = !this.userCanRequestReview()
+      ? nothing
+      : html`
+          <div>
+            <sl-button
+              size="small"
+              variant="primary"
+              @click=${this.handleReviewRequested}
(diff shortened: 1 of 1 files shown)


KyleJu (approved):

jrobbins:
This relates to #4634.

PR #4670, opened 2025-01-06
Send SLO reminders for review resolution

jrobbins:
This is progress toward implementing improved SLOs for both an initial response and a review resolution.  Now that we can calculate when a review is due for a resolution, we will start sending email reminders about review resolutions that are due or overdue.

In this PR:
* Consider all gates that have active reviews and organize any due or overdue ones into four lists: initial response due, initial response overdue, resolution due, and resolution overdue.
* Generate four sets of email notifications based on those four lists.
* Make the wording of the reminder email vary based on initial response vs. resolution.
* Add unit tests and golden data files.


KyleJu (approved):

jrobbins:
This relates to #4634.

PR #4671, opened 2025-01-07
Automatically expand stages that have reviews in progress.

jrobbins:
This is a step toward supporting more fine-grained SLO measurements.  Now that feature owners need to explicitly re-request reviews that were marked NEEDS_WORK, it is important that they can easily notice such reviews on the feature detail page.

Previously, the only open stage accordion sections were for the stage that the user had marked as active, or the stage containing a gate mentioned in the URL.  Now, the app will also automatically expand any section that has a gate in an active review state, or a mixed combination of PREPARING and APPROVED gates.

In this PR:
* Add the conditional logic and helper functions
* Add unit tests


KyleJu (approved):

jrobbins:
This relates to #4634.

PR #4679, opened 2025-01-09
Use the origin trial display name in trial extension emails

DanielRyanSmith:
This change fixes a small problem where the feature name is being used in the origin trial extension emails rather than the origin trial name. The origin trial name is now passed with the email task parameters to populate the subject line in the email.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -840,21 +840,22 @@ class OTExtensionApprovedHandler(basehandlers.FlaskHandler):
   def process_post_data(self, **kwargs):
     self.require_task_header()
     feature = self.get_param('feature')
-    if feature is None:
-      self.abort(400, 'No feature provided.')
     gate_id = self.get_param('gate_id')
-    if gate_id is None:
-      self.abort(400, 'Extension gate ID not provided.')
     requester_email = self.get_param('requester_email')
-    if not requester_email:
-      self.abort(400, 'Extension requester\'s email address not provided.')
+    ot_display_name = self.get_param('ot_display_name')
     logging.info('Starting to notify about successful origin trial extension.')
-    send_emails([self.build_email(feature, requester_email, gate_id)])
+    send_emails([self.build_email(
+        feature, requester_email, gate_id,ot_display_name)])
 
     return {'message': 'OK'}
 
   def build_email(
-      self, feature: FeatureEntry, requester_email: str, gate_id: int):
+      self,
+      feature: FeatureEntry,
+      requester_email: str,
+      gate_id: int,
+      ot_display_name: str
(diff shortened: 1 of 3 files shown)


internals/notifier.py
```diff
@@ -840,21 +840,22 @@ class OTExtensionApprovedHandler(basehandlers.FlaskHandler):
...
     self.require_task_header()
     feature = self.get_param('feature')
-    if feature is None:
-      self.abort(400, 'No feature provided.')
     gate_id = self.get_param('gate_id')
-    if gate_id is None:
-      self.abort(400, 'Extension gate ID not provided.')
     requester_email = self.get_param('requester_email')
-    if not requester_email:
-      self.abort(400, 'Extension requester\'s email address not provided.')
```
DanielRyanSmith, line 850: These abort statements were unnecessary, since the get_param method contains an abort statement in the event the parameter is not provided

internals/notifier.py
```diff
@@ -840,21 +840,22 @@ class OTExtensionApprovedHandler(basehandlers.FlaskHandler):
...
+    ot_display_name = self.get_param('ot_display_name')
     logging.info('Starting to notify about successful origin trial extension.')
-    send_emails([self.build_email(feature, requester_email, gate_id)])
+    send_emails([self.build_email(
+        feature, requester_email, gate_id,ot_display_name)])
```
jrobbins, line 848: Style: need space after comma.

internals/notifier.py
```diff
@@ -840,21 +840,22 @@ class OTExtensionApprovedHandler(basehandlers.FlaskHandler):
...
     feature = self.get_param('feature')
-    if feature is None:
-      self.abort(400, 'No feature provided.')
     gate_id = self.get_param('gate_id')
-    if gate_id is None:
-      self.abort(400, 'Extension gate ID not provided.')
     requester_email = self.get_param('requester_email')
-    if not requester_email:
-      self.abort(400, 'Extension requester\'s email address not provided.')
+    ot_display_name = self.get_param('ot_display_name')
```
jrobbins, line 845: This will fail for any tasks currently in the queue that lack `ot_display_name`, but that should be OK because those tasks are short-lived and happen at a different time of day than we usually do deployments.

jrobbins (approved):

internals/notifier.py
DanielRyanSmith: Good catch!

internals/notifier.py
DanielRyanSmith: This should not be a problem, since the `ot_display_name` field is required to be populated on the OT creation form 🙂

PR #4680, opened 2025-01-09
Add clarification about mailing lists as contacts for OT creation

### LINKED ISSUE #4531: [Chromestatus-OT] Better instructions for team mailing address
Currently if authors include a team mailing contact, e.g. random-google-team@google.com, the creation will fail at the backend. Maybe create a better instruction for it. It will be addressed on the OT side as well

DanielRyanSmith:
Fixes #4531 

Adds an additional caveat to the instructions about supplying mailing lists as OT contacts.

### DIFF
--- a/client-src/elements/form-field-specs.ts
+++ b/client-src/elements/form-field-specs.ts
@@ -1462,9 +1462,11 @@ export const ALL_FIELDS: Record<string, Field> = {
     label: 'Origin trial contacts',
     help_text: html` List any other individuals or groups to include on the
       contact list (e.g. for reminders on trial milestones). Mailing list emails
-      can be used here, but only email addresses of individuals
-      will receive access to view registrant data on <a target="_blank"
-      href="http://go/ot-registrants-dashboard">go/ot-registrants-dashboard</a>.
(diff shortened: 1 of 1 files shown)


client-src/elements/form-field-specs.ts
```diff
@@ -1461,7 +1461,12 @@ export const ALL_FIELDS: Record<string, Field> = {
     required: false,
     label: 'Origin trial contacts',
     help_text: html` List any other individuals or groups to include on the
-      contact list (e.g. for reminders on trial milestones).
+      contact list (e.g. for reminders on trial milestones). Mailing list emails
+      can be used here, but only email addresses of individuals will receive
+      access to view registrant data on
+      <a target="_blank" href="http://go/ot-registrants-dashboard"
```
KyleJu, line 1467: http://go/ot-registrants-dashboard is a Google specific URL (go/ links), not sure if this should be linked here

DanielRyanSmith: We have a check in place to ensure only google and chromium accounts have access to the OT creation request form, so this already assumes that the user fulfills that request

KyleJu (approved):

Issue #4702, opened 2025-01-15
Allow feature owners to supply CSS property use counters for origin trials


DanielRyanSmith:
Currently, users can supply WebFeature and WebDXFeature use counters for monitoring origin trial usage, but no option is available for [CSS property use counters](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/css_property_id.mojom). The Origin Trials API can accept CSS property use counters, but feature owners cannot supply them through the origin trial creation request form.

- [ ] The request form should allow the user to provide this type of use counter.
- [ ] When submitting the origin trial creation request form, the correct [mojom file](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/css_property_id.mojom) should be checked to verify the use counter exists in the file.
- [ ] The format of the request sent to the OT API should contain the correct use counter type information.

PR #4703, opened 2025-01-15
New histogram ID enum for differentiating use counter types

DanielRyanSmith:
This change is a follow-up to #4624 that updates the implementation to use new histogram ID enums to differentiate use counters rather than having different config types. Changes were made to the origin trials API and how it represents use counter types (histogram ID), and this change mirrors those changes.

This is not a breaking change and can be successfully deployed without request interruption between ChromeStatus and the OT API.

Note that the `css_property_id` histogram ID enum is not yet used, which will come at a later time when addressing #4702.

### DIFF
--- a/framework/origin_trials_client.py
+++ b/framework/origin_trials_client.py
@@ -23,6 +23,7 @@ import requests
 
 from framework import secrets
 from framework import utils
+from internals.core_enums import BlinkHistogramID
 from internals.core_models import Stage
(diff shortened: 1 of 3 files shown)


KyleJu (approved):

AlipAbdullah:
Claiming namespace @AlipMagical

AlipMagical:
Approve

AlipMagical:
#closed

PR #4705, opened 2025-01-15
Allow bypassing OT Chromium checks in non-prod environments

### LINKED ISSUE #4428: Allow bypassing of name-in-file requirements for OT creation requests on staging
Currently, when submitting an origin trial creation request on staging, the Chromium Trial Name, use counter name, and whether the trial allows third party origins is checked against the actual data in the Chromium codebase. This means that only actual origin trial names can be used on staging. Additionally, the OT API has a check that disallows origin trials that have the same Chromium Trial Name. This means that, with the current implementation, there are a finite amount of trial names (that mirror actual trial names) we can provide on staging that are considered valid and can be submitted.

Ideally, this check can be bypassed on staging so that any value can be provided for these fields. However, this would cause an issue that the logic for the name-in-file validation checks cannot be tested manually. We'll need some way to be able to verify both on staging.

DanielRyanSmith:
Fixes #4428 

This change adds a checkbox to the OT creation form in non-prod environments to skip the Chromium file checks. The server-side code is implemented so that a bypass request would always be ignored in prod environments.


api/origin_trials_api.py
```diff
@@ -66,6 +67,10 @@ def get_chromium_files_for_validation() -> dict:
 def find_use_counter_value(
     body: dict, chromium_files_dict: dict) -> int | None:
   """Find where the use counter is defined and return its value."""
+  # Chromium file checks can be bypassed, but only in non-prod environments.
```
KyleJu, line 70: optional nit: Should this be factored and live in settings?

gen/js/chromestatus-openapi/src/models/CreateOriginTrialRequest.ts
```diff
@@ -140,6 +140,12 @@ export interface CreateOriginTrialRequest {
      * @memberof CreateOriginTrialRequest
      */
     ot_action_requested?: FieldInfo;
+    /**
+     * 
+     * @type {FieldInfo}
+     * @memberof CreateOriginTrialRequest
+     */
+    ot_creation__bypass_file_checks?: FieldInfo;
```
KyleJu, line 148: Are these generated files?

client-src/elements/chromedash-ot-creation-page.ts
```diff
@@ -199,15 +199,29 @@ export class ChromedashOTCreationPage extends LitElement {
...
       value: true,
       stageId: this.stage.id,
       alwaysHidden: true,
     });
+
+    // For non-prod environments, add a checkbox to bypass the checks ensuring
+    // that all values exist in Chromium files.
+    // TODO(DanielRyanSmith): Identify non-prod environments using a flag rather
+    // than the app's title.
+    if (this.appTitle !== 'Chrome Platform Status') {
```
KyleJu, line 216: optional nit: again if this should be factored to settings so the flag for non-prod is consistent across the app

KyleJu (approved):
LGTM with nits

gen/js/chromestatus-openapi/src/models/CreateOriginTrialRequest.ts
DanielRyanSmith: Yes, these are generated with the `npm run openapi` command

api/origin_trials_api.py
DanielRyanSmith: This definitely should be - and I'll write an issue to make this change and land that in a separate PR

DanielRyanSmith:
#4707  created

PR #4706, opened 2025-01-15
Allow sending of OT creation/activation emails manually

### LINKED ISSUE #4533: Add a way to manually send an "Origin trial has been created/activated" email through Chrome Status
We have a number of email templates to send to users when the automated OT creation process completes successfully, and also email templates to notify the support team if something fails. However, if the automated process fails, and the support member completes the last steps of the trial creation manually, we have no good way of sending the "Creation Successful"  emails to trial owners through Chrome Status as the final step. Having a way to manually send one of these emails would be useful in this case, although the frequency this is needed should continue to decrease as the automated process becomes more reliable.

DanielRyanSmith:
Fixes #4533 

This change allows two new script paths that manually create email tasks to send notice to origin trial owners regarding two scenarios:
- An origin trial has been created and its activation is pending a later branch date.
- An origin trial has been created and activated and is ready for use.

These will  be available in the event that a failure occurs during the automated OT creation/activation process, so that OT support members can send these emails out after following the manual process.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -801,3 +801,37 @@ class BackfillGateDates(FlaskHandler):
 
     return max(v.set_on for v in votes
                if v.state == Vote.NEEDS_WORK)
+
+
+class SendManualOTCreatedEmail(FlaskHandler):
+  """Manually send an email to origin trial contacts that an origin trial has
+  been created but not yet activated."""
+  def get_template_data(self):
+    self.require_cron_header()
+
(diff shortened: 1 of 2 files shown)


jrobbins (approved):
We should document this somewhere, otherwise the engineer on rotation will not know that it is available.

DanielRyanSmith:
> We should document this somewhere, otherwise the engineer on rotation will not know that it is available.

Yes, good call - very important. I've sent a CL to you to update our documentation.

PR #4709, opened 2025-01-17
Implement storage and api for gate survey questions

### LINKED ISSUE #4695: Implement storage and API for gate survey questions
For self-certified reviews, we will need to store the survey question answers in NDB and communicate them to the frontend by adding new fields to the Gates API.

jrobbins:
This should resolve #4695.  It is the first part of support self-certification of review gates.  Basically, each review team can define some survey questions to screen feature owner review requests.  If the answers to those questions indicate that the change is not a risk, then the feature owner will be able to click a button to get approval for that gate rather than wait for human review.

In this PR:
* Define a new sub-model of Gate to hold survey question answers.
* Add the survey answers to the converter logic and response openapi models when doing a GET on a Gate.
* Add a SurveyAnswers openapi model to the requests that edit a gate.
* Add logic to the GatesAPI to set either the existing assignee field or the new survey question answers field.
* Add logic to determine when a gate is eligible for self certification based on the Privacy team survey questions.
* Also, change the gate editing request type from POST to PATCH because that is more appropriate for editing, but keep backward compatibility by still handling POST for now.


internals/self_certify.py
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from chromestatus_openapi.models import SurveyAnswers as OASurveyAnswers
+
+from internals.core_enums import GATE_PRIVACY_ORIGIN_TRIAL, GATE_PRIVACY_SHIP
+from internals.review_models import Gate, SurveyAnswers
+
+
+def update_survey_answers(gate: Gate, new_answers: OASurveyAnswers):
```
DanielRyanSmith, line 21: Optional nit: This file could be named `self_certify_helpers.py` since it contains no classes and consists only of helper functions

DanielRyanSmith (approved):

internals/self_certify.py
jrobbins: Hmm, it would seem odd to have an X_helpers file without having the X file first.  I'll stick with the shorter name for now, but keep _helpers in mind.

Issue #4666, opened 2025-01-03
Create a WebDX features data ingestion pipeline 


KyleJu:
Address TODOs in https://github.com/GoogleChrome/chromium-dashboard/pull/4658. Create a WebDX features data ingestion pipeline and validate the feature group in the feature creation form

PR #4722, opened 2025-01-24
Implement Webdx Feature IDs datastore and endpoint

KyleJu:
A part of https://github.com/GoogleChrome/chromium-dashboard/issues/4666. Implement `/webdxfeatures` backend API and return a sorted dictionary of feature IDs. Also create a singleton ndb model to store the feature ID list.

### DIFF
--- a/api/webdx_feature_api.py
+++ b/api/webdx_feature_api.py
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 from collections import OrderedDict
 
 from framework import basehandlers
 from internals.webdx_feature_models import WebdxFeatures
 
 MISSING_FEATURE_ID = 'N/A'
-MISSING_FEATURE_TEXT = 'Missing feature'
+TBD_FEATURE_ID = 'TBD'
 
 
 class WebdxFeatureAPI(basehandlers.APIHandler):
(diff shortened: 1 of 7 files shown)


api/webdx_feature_api.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class WebdxFeatureAPI(basehandlers.APIHandler):
+  """The list of ordered webdx feature ID populates the "Web Feature ID" select field
+  in the guide form"""
+
+  def do_get(self, **kwargs):
+    """Returns an ordered dict with Webdx feature id as both keys and values."""
+    webdx_features = WebdxFeatures.get_webdx_feature_id_list()
+    if not webdx_features:
```
jrobbins, line 31: What about the cases where (1) we deploy this to prod for the first time, and (2) a developer starts using this code locally for the first time?  Would get_webdx_feature_id_list return None in those cases?

api/webdx_feature_api.py
```diff
@@ -0,0 +1,43 @@
...
+        self.abort(500, 'Error obtaining Webdx feature ID list.')
+
+    feature_ids_dict = OrderedDict()
+    # The first key, value pair is the id when features are missing from the list.
+    feature_ids_dict.update({MISSING_FEATURE_ID: [MISSING_FEATURE_TEXT, MISSING_FEATURE_ID]})
+
+    feature_list = webdx_features.feature_ids
+    feature_list.sort()
+    for id in feature_list:
+      feature_ids_dict.update({id: [id, id]})
```
jrobbins, line 41: Why update{} rather than just setting one dictionary item?

api/webdx_feature_api.py
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import OrderedDict
+
+from framework import basehandlers
+from internals.webdx_feature_models import WebdxFeatures
+
+MISSING_FEATURE_ID = 'N/A'
```
jrobbins, line 20: I was thinking that we would offer both `N/A` and `Missing feature` as separate items that the user could choose.  N/A would mean that this feature entry will never map to a WebDX feature, and Missing would mean that it just hasn't been figured out yet.   In fact, maybe 'missing' is not quite right and 'TBD' would capture the idea that it needs to be filled in later.

api/webdx_feature_api.py
KyleJu: Good point. I will return {} instead, and logging the message.

api/webdx_feature_api.py
KyleJu: Fixed!

api/webdx_feature_api.py
KyleJu: Good idea. Fixed1

jrobbins (approved):

PR #4740, opened 2025-01-28
Update URL for usecounter enums

jrobbins:
The location of the use counter enums that we used was changed yesterday, so this PR updates our code to get them from their new URL.  Also, I improved error handling a bit.

See: https://chromium.googlesource.com/chromium/src/+/5d15720921afb24de54a64b766138c45962cbcef


jcscottiii (approved):
LGTM thanks for catching this. We have a similar failure in webstatus and this is the fix.

PR #4744, opened 2025-01-29
Implement LFC survey question form

### LINKED ISSUE #4696: Implement a gate survey form
Review gates for teams that have survey questions will need to display those questions with form fields for answers.  When the user edits the form, the answers should be saved to the server.

Rather than yet another submit button, I am thinking that these form fields will save automatically as soon as each form field is changed.  We can display a tiny spinner during the saving process.

jrobbins:
This should resolve #4696.  The privacy gates will offer self-certification instead of review, but only if the user enters survey question answers that cause the server to set self_certify_eligible to True.  This PR just displays the form and saves any value changes to the server.  A future PR will change the behavior of the "Request review" button for the eligible case.

In this PR:
* Refactor the display of the "Survey questions" part of the gate column into a new component.
* Display the specific questions for the privacy form.
* Handle UI element changes by doing a PATCH to the server.


api/reviews_api.py
```diff
@@ -160,8 +160,6 @@ def do_patch(self, **kwargs) -> dict[str, str]:
     new_assignees = request.assignees
     new_answers = request.survey_answers
 
-    logging.info('@@@ request is %r', request)
```
DanielRyanSmith, line 163: Is this meant to be removed?

client-src/elements/chromedash-survey-questions.ts
```diff
@@ -0,0 +1,177 @@
...
+      <p class="instructions">&nbsp;</p>
+    `;
+  }
+
+  handleFieldChange(name: string, value: string | boolean) {
+    window.csClient
+      .updateGate(this.feature.id, this.gate.id, null, {[name]: value})
+      .then(() => {
+        this.refetch();
+      });
```
DanielRyanSmith, line 77: This will send a request for every time any field changes state, including for each new character change in an input field. Would a "Submit" option be a better choice to avoid the potential of redundant requests?

client-src/elements/chromedash-survey-questions.ts
```diff
@@ -0,0 +1,177 @@
...
+      return this.renderPrivacyForm();
+    }
+    const questionnaireText = GATE_QUESTIONNAIRES[this.gate.gate_type];
+    if (!questionnaireText) return html`No questions`;
+    const markup =
+      typeof questionnaireText == 'string'
+        ? autolink(questionnaireText)
+        : questionnaireText;
+    return html`
+      <!-- prettier-ignore -->
```
DanielRyanSmith, line 163: What is the ignore needed here for?

DanielRyanSmith (commented):
Some questions to clear things up for me 🙂

api/reviews_api.py
jrobbins: Yeah, that was left-over debugging that is not needed.

client-src/elements/chromedash-survey-questions.ts
jrobbins: Actually, the sl-change event is only sent once the user un-focuses on the input field or navigates away.  So, it is not one for every keystroke.  I am thinking that automatic updates are better than having a submit button because there are already enough buttons to compete for the user's attention.  Specifically, after filling in these answers, the user will be often be clicking on the "Request review" button.  I don't want them to have to fill-in and press submit before being able to press the "Request review" button.

client-src/elements/chromedash-survey-questions.ts
jrobbins: Gone.  I think I had that because in a previous iteration I was using `white-space: pre-line`.  All the textual questions are written as markup now rather than text with line-breaks, so it should not matter if prettier decides to break the next line into three.  And, prettier doesn't seem to want to do that now anyway.

client-src/elements/chromedash-survey-questions.ts
DanielRyanSmith: Ah, you are correct about the unfocus aspect 🙂  - although, maybe there is a risk of the user not unfocusing for something - it's like you said about the likelihood of the "Request review" button being the next action - so that sounds fine to me

DanielRyanSmith (approved):

PR #4752, opened 2025-01-31
Implement self certify dialog

### LINKED ISSUE #4697: Offer self-certification of review gates when appropriate
For those teams that offer survey questions, we will define a self-certification criteria for the answers. 

Questions will be phrased in such a way that default values do not make the gate eligible for self-certification.

Pressing the "Request review" or "Request N/A" button will:
* If the gate is not eligible for self-certification, then display a dialog asking the user to (a) fill in or edit the survey questions or (b) request the review as normal.
* If the gate is eligible based on the survey answers, then offer the user a choice to (a) self-certify or (b) go ahead with the full review.

The server will need to be modified to permit a feature owner to self-certify when the gate is eligible.

jrobbins:
This should resolve #4697.  For the privacy team's two gates, we will now offer the feature editors the ability to self-certify the gate if the survey answers make it eligible.

In this PR:
* Define another boolean in the OpenAPI response when getting Gates: `self_certify_possible` is true for gates where it is ever possible to self-certify.  For now, that is only the privacy gates.
* Change the VotesAPI to allow a feature editor to post an APPROVED or NA vote on a gate iff it is eligible.  This results in a normal vote Activity generation and email notification, as if a reviewer had voted.
* Change the `refetch()` method on the feature detail page to not set `this.loading` so as to avoid excessive blinking when page content is updated.  The whole page needs to refetch rather than just the survey questions form because the "Request review" button logic depends on it.
* Implement a self-certification dialog and have it open when the user presses "Request review" or "Request NA", but only if the self-certification is possible on that gate.  The buttons offered inside the dialog depend on whether the gate is eligible.

### DIFF
--- a/client-src/elements/chromedash-feature-page.ts
+++ b/client-src/elements/chromedash-feature-page.ts
@@ -346,7 +346,6 @@ export class ChromedashFeaturePage extends LitElement {
   }
 
   refetch() {
-    this.loading = true;
     Promise.all([
       window.csClient.getFeature(this.featureId),
       window.csClient.getGates(this.featureId),
@@ -356,7 +355,6 @@ export class ChromedashFeaturePage extends LitElement {
         this.feature = feature;
         this.gates = gatesRes.gates;
         this.comments = commentRes.comments;
-        this.loading = false;
       })
       .catch(error => {
         if (error instanceof FeatureNotFoundError) {
--- a/client-src/elements/chromedash-survey-questions.ts
+++ b/client-src/elements/chromedash-survey-questions.ts
@@ -50,15 +50,13 @@ export class ChromedashSurveyQuestions extends LitElement {
   @state()
   loading = true;
 
-  refetch() {
-    window.csClient.getGates(this.feature.id).then(resp => {
-      for (const g of resp.gates) {
-        if (g.id === this.gate.id) {
-          this.gate = g;
-          break;
-        }
-      }
+  _fireEvent(eventName, detail) {
+    const event = new CustomEvent(eventName, {
+      bubbles: true,
+      composed: true,
+      detail,
     });
+    this.dispatchEvent(event);
   }
 
   renderQuestionnaireSkeleton(): TemplateResult {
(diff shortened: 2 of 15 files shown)


jrobbins:
Screenshot when the gate is possible to self-certify, but the answers do not make it eligible:
image

Screenshot when both possible and eligible:
image

client-src/elements/chromedash-feature-page.ts
```diff
@@ -346,7 +346,6 @@ export class ChromedashFeaturePage extends LitElement {
   }
 
   refetch() {
-    this.loading = true;
```
DanielRyanSmith, line 349: Does this change any existing behavior?

client-src/elements/chromedash-survey-questions.ts
```diff
@@ -72,7 +70,8 @@ export class ChromedashSurveyQuestions extends LitElement {
     window.csClient
       .updateGate(this.feature.id, this.gate.id, null, {[name]: value})
       .then(() => {
-        this.refetch();
+        this._fireEvent('refetch-needed', {});
+        // this.refetch();
```
DanielRyanSmith, line 74: Safe to removd

DanielRyanSmith (approved):

DanielRyanSmith:
Thanks for the screenshots! Very helpful to understand the implementation 🙂

client-src/elements/chromedash-feature-page.ts
jrobbins: Yes, it makes the page less jumpy after reviewer votes.  It used to be that the whole page would show skeletons briefly, and then new info would show up in the form of a gate changing color and the message at the top of the gate column changing.  Now, those change happen in a much smoother way.  

I suppose what is lost is the visual feedback in situation where the refetch takes a long time.  If we find that that happens a lot or users mention it, we could add a more subtle loading indicator.

client-src/elements/chromedash-survey-questions.ts
jrobbins: Gone

PR #4753, opened 2025-01-31
"OT approved" email bug fix

DanielRyanSmith:
This change fixes a bug related to the automated email sent when an origin trial is approved. The incorrect arguments were supplied for the `build_email` method.


jrobbins (approved):
Why didn't mypy catch this?

jrobbins:
> Why didn't mypy catch this?

Ah, the body of `process_post_data()` was not checked because that function does not have any type info in its signature.  We should probably add a bunch more return types throughout our existing code.

DanielRyanSmith:
> Why didn't mypy catch this?

I think this is because the `feature` and `contacts` variables are technically returned as untyped. This may have been flagged by mypy if I explicitly gave a type to these two vars.

PR #4780, opened 2025-02-14
Allow reviewers to override a self-certified NA.

jrobbins:
Feature owners can self-certify an NA, but it needs to be possible to a reviewer to look at it afterwards and override the gate state.

In this PR:
* Make feature owners who are doing self-certifications vote NA_SELF instead of NA.  But, still allow permissions to vote NA for backward compatibility.
* Count NA_SELF as an NA for the gate state only if it is the only vote on the gate.  Otherwise, use the regular logic, which ignores NA_SELF.
* Render NA_SELF as "N/a (self-certified)" in the UI and "na (self-certified)" in notifications.

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -71,7 +71,6 @@ class VotesAPI(basehandlers.APIHandler):
     old_state = old_votes[0].state if old_votes else Vote.NO_RESPONSE
     self.require_permissions(user, fe, gate, new_state)
 
-    # Note: We no longer write Approval entities.
     old_gate_state = gate.state
     new_gate_state = approval_defs.set_vote(feature_id, None, new_state,
         user.email(), gate_id)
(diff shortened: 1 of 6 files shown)


DanielRyanSmith (approved):
LGTM after considering this question:

I know that OT gates cannot be self-certified, but is there a risk of the app's logic being incorrect somewhere because of this new gate state?

For example:
https://github.com/GoogleChrome/chromium-dashboard/blob/aa3f06348fab0d1a9f2b29979964bfde3ae1cb9d/internals/notifier_helpers.py#L226-L228

jrobbins:
> I know that OT gates cannot be self-certified, but is there a risk of the app's logic being incorrect somewhere because of this new gate state?

Thanks for the reminder.  That is correct that the Gate.state would still be NA, only individual votes can be NA_SELF.   I grepped through the code again to make sure that nothing else was checking for vote values.

Issue #4799, opened 2025-02-24
Use WebDXFeature use counter instead of Web Feature ID


KyleJu:
In new feature creation, a Web Feature ID is recorded upon creation (N/A or TBD if it does not exist) . WebDXFeature use counter should be stored underlying in the database, in case of feature name changes or split.

**Describe the solution you'd like**
Update the histogram code to ingest the new histogram by expanding [here](https://github.com/GoogleChrome/chromium-dashboard/blob/66e03cac2653e15a890f390c0e3cecccd6731186/internals/fetchmetrics.py#L228-L231) with `WebDXFeatureObserver`. WebDXFeatureObserver has the mapping between feature ids and chromium usecounter (hash numbers).

KyleJu:
APIs:
Get /webdxfeatures API: a map that returns webdx UseCounter name and webDX feature IDs mapping information
Feature creation post request: translate the UseCounter name to enum

Database:
Create a new webDX use counter field
Delete/clean existing feature ID fields

KyleJu:
Unclear from the meetings: Should the feature creation UI show the list of WebDXFeature UseCounter strings instead of Web Feature IDs?

PR #4807, opened 2025-02-25
Store Webdx feature through Webdx UseCounter enum

KyleJu:
https://github.com/GoogleChrome/chromium-dashboard/issues/4799. Update the backend APIs to return a list of available Web features and its UseCounter enum. And create a new field `webdx_usecounter_enum`. The list is available from [usecounter enum](https://source.corp.google.com/piper///depot/google3/analysis/uma/ukm/aggregation/webdx_features/enums.xml;bpv=1).

### Follow-up
Clean up the old web_feature fields

### DIFF
--- a/api/api_specs.py
+++ b/api/api_specs.py
@@ -48,6 +48,7 @@ FEATURE_FIELD_DATA_TYPES: FIELD_INFO_DATA_TYPE = [
   ('explainer_links', 'links'),
   ('feature_notes', 'str'),
   ('web_feature', 'str'),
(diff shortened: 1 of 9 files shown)


jrobbins (changes requested):
I'd prefer that you include some of the follow-up in this PR.  I think that if this was deployed without the follow-ups, a user could enter a web feature ID and then see it as blank on the feature detail page.

jrobbins (approved):

PR #4809, opened 2025-02-25
Change `yyyymmdd` to `date` in `blink_features` example query

tunetheweb:
Due to some changes on the HTTP Archive side, the `blink_features` example query needs a slight tweak.


jrobbins (approved):

jrobbins:
Thanks for this change.  We typically deploy a new version to prod on Fridays.

PR #4810, opened 2025-02-25
Add explanation field to survey questions

jrobbins:
This is a field that the privacy team requested on their survey question form.  The explanation is a text field where the feature owner explains why they checked off some of the boxes to self-certify the gate.  The gate is only eligible if something is filled in.

In this PR:
* Add the explanation field to the SurveyAnswers NDB model, and add it to converters.py so that it is sent to the frontend.
* Add explanation to the openapi spec so that the frontend can save after the user edits that field.
* Change the eligibility rule so that explanation is required.
* Also, update an existing test to have the user vote NA_SELF instead of Approval because in a previous PR we introduced the NA_SELF vote value.

### DIFF
--- a/client-src/elements/chromedash-survey-questions.ts
+++ b/client-src/elements/chromedash-survey-questions.ts
@@ -170,8 +170,8 @@ export class ChromedashSurveyQuestions extends LitElement {
           )}
           ${this.renderTextField(
             'explanation',
-            html`Required: If you checked any box above, explain why you
-            checked it. And, provide any other relevant context.`
+            html`Required: If you checked any box above, explain why you checked
+            it. And, provide any other relevant context.`
           )}
         </ol>
       </div>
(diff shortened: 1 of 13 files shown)


client-src/elements/chromedash-survey-questions.ts
```diff
@@ -151,6 +168,11 @@ export class ChromedashSurveyQuestions extends LitElement {
             discussed with someone on the privacy team, enter their email
             address. Or, enter "None".`
           )}
+          ${this.renderTextField(
+            'explanation',
+            html`Required: If you checked any box above, explain why you checked
+            it. And, provide any other relevant context.`
```
KyleJu, line 174:
```suggestion
            it, and provide any other relevant context.`
```
maybe this?

gen/js/chromestatus-openapi/package.json
```diff
@@ -16,6 +16,6 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "typescript": "^5.7"
+    "typescript": "^4.0 || ^5.0"
```
KyleJu, line 19: unrelated?

KyleJu (approved):

PR #4811, opened 2025-02-26
Search web feature names in link

KyleJu:
A part of https://github.com/GoogleChrome/chromium-dashboard/issues/4799. In #4807, it is using a list of available feature names from [enums.xml](https://source.corp.google.com/piper///depot/google3/analysis/uma/ukm/aggregation/webdx_features/enums.xml;bpv=1), however it does not provide the exact Web feature ID. Update the webstatus.dev link to use search instead.

The downside of this change is that some feature names cannot be found, e.g.  https://webstatus.dev/?q=OverflowShorthand

### DIFF
--- a/client-src/elements/chromedash-form-field.ts
+++ b/client-src/elements/chromedash-form-field.ts
@@ -479,7 +479,7 @@ export class ChromedashFormField extends LitElement {
               <td colspan="2" class="webdx">
                 See web feature
                 <a
-                  href="https://webstatus.dev/features/${this.value}"
+                  href="https://webstatus.dev/?q=${this.value}"
                   target="_blank"
                 >
                   ${this.value}
--- a/client-src/elements/chromedash-form-field.ts
+++ b/client-src/elements/chromedash-form-field.ts
@@ -479,7 +479,9 @@ export class ChromedashFormField extends LitElement {
               <td colspan="2" class="webdx">
                 See web feature
                 <a
-                  href="https://webstatus.dev/features/${this.value.toLowerCase()}"
+                  href="https://webstatus.dev/features/${enumLabelToFeatureKey(
+                    this.value
+                  )}"
                   target="_blank"
                 >
                   ${this.value}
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-form-field.ts
```diff
@@ -479,7 +479,7 @@ export class ChromedashFormField extends LitElement {
               <td colspan="2" class="webdx">
                 See web feature
                 <a
-                  href="https://webstatus.dev/features/${this.value}"
+                  href="https://webstatus.dev/?q=${this.value}"
```
jcscottiii, line 482: ## Prior behavior:

A user clicks on the link for a feature id that does not exist.
They get the 404 page

## New behavior:

A user clicks on the link for a feature id that does not exist.
They either get 1) an empty table if the name is truly unique or 2) maybe related features but still not their feature.

---

I'm not sure if the new behavior would help the user understand what is going on more than the previous behavior. Showing related features might mislead the user into thinking their feature exists under a different name when it doesn't, or if the related features are too dissimilar, it might add noise and confusion rather than clarity.

## You could:

1. Do a fetch to webstatus.dev and see if the feature exists. If you get a 404 back from the API for that feature ID, direct them to start the process of adding their feature to the source of truth (https://github.com/web-platform-dx/web-features). Or,
2. Add an information icon to warn them on hover that it is okay if it does not exist. And that they just need to add it to the source of truth.


Thoughts?

jrobbins: In the (hopefully) most common case where the web feature does exist, I think it is best to keep the direct link as-is.  To cope with the non-existing web feature case, how about making the webstatus.dev 404 page for that case more helpful by replacing the "Go back home" button with a "Search for similar names" button or "Did you mean ___?" links.  That would also help with users who type in webstatus.dev URLs or who click links in other documents or websites that contain mistakes.

jcscottiii: > To cope with the non-existing web feature case, how about making the webstatus.dev 404 page for that case more helpful by replacing the "Go back home" button with a "Search for similar names" button or "Did you mean ___?" links? That would also help with users who type in webstatus.dev URLs or who click links in other documents or websites that contain mistakes.

I like `"Did you mean ___?" links?` idea! I created this issue for it https://github.com/GoogleChrome/webstatus.dev/issues/1214

jcscottiii: Instead of doing straight to lowercase, the mojom file has [instructions](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/use_counter/metrics/webdx_feature.mojom;l=35-43;drc=004786a4bca997d71618c8fc0b3f0c98ea660c27) on how to convert the enum label to the web feature IDs in the [source of truth](https://github.com/web-platform-dx/web-features/tree/main/packages/web-features)(You can ignore the leading `k` character part)

For example:
- Take `FocusVisible` from the enums.xml [file](https://source.chromium.org/chromium/chromium/src/+/main:tools/metrics/histograms/metadata/blink/enums.xml;l=7589;drc=ad07ffe6771c2897dd8b99cb2fb726617cf56d9f)
- Instead of converting to `focusvisible`, it should be converted to [focus-visible](https://github.com/web-platform-dx/web-features/blob/main/features/focus-visible.yml) 

webstatus.dev has this conversion logic that you can look at too.  ([Code](https://github.com/GoogleChrome/webstatus.dev/blob/6ae7ecf7c26e2563a6e1685cc6d92fc12bcc7941/lib/gcpspanner/spanneradapters/chromium_historgram_enum_consumer.go#L102-L125))([test](https://github.com/GoogleChrome/webstatus.dev/blob/6ae7ecf7c26e2563a6e1685cc6d92fc12bcc7941/lib/gcpspanner/spanneradapters/chromium_historgram_enum_consumer_test.go#L198))

jcscottiii (approved):

client-src/elements/chromedash-form-field.ts
KyleJu: Implemented! Thanks!

PR #4719, opened 2025-01-21
[Draft] WebDX feature IDs API

KyleJu:
#4666

PR #4815, opened 2025-02-27
Revert original web_feature field implementation

### LINKED ISSUE #4799: Use WebDXFeature use counter instead of Web Feature ID

KyleJu:
Fix #4799. Revert changes in #4666 and mark `web_feature = ndb.StringProperty()` as deprecated.


KyleJu:
FYI @jcscottiii, since we are deleting all changes related to the webstatus API, the smaller file generation mentioned in https://github.com/GoogleChrome/chromium-dashboard/pull/4719#pullrequestreview-2570064013 is no longer needed

jrobbins (approved):

PR #4816, opened 2025-02-27
Fix accessing confidential features

### LINKED ISSUE #4778: 500 when accessing confidential feature entries
When there is a confidential feature entry, and the signed in user is not using a @google.com or @chromium.org account, and they are not a site admin, then they should get a 403, but instead they get a 500 response.

The problem is that the check in permissions.py is looking for feature_dict['owners'], but that field is only there for features converted to dicts via `feature_entry_to_json_basic()`, not `feature_entry_to_json_verbose()`.  The verbose conversion uses `'owner_emails'`.  Also, for search results, we sometimes produce dicts with just the name and ID.

jrobbins:
Fixes #4778.

Basically, we have a couple of different JSON formats that we can't update without breaking clients, so code that works on the JSON dicts needs to handle all possibilities.


KyleJu (approved):

shal44m:
> Fixes #4778.
> 
> Basically, we have a couple of different JSON formats that we can't update without breaking clients, so code that works on the JSON dicts needs to handle all possibilities.

shal44m:
> Fixes #4778.
> 
> Basically, we have a couple of different JSON formats that we can't update without breaking clients, so code that works on the JSON dicts needs to handle all possibilities.

shal44m:
Hak

shal44m:
Whwhh

PR #4821, opened 2025-02-28
Count NA_SELF votes as both a review request and a response.

jrobbins:
One of the main points of offering self-certification is to speed up review times, so to measure the impact of this process change we should count self-certified gates as having a zero review latency. We do that by setting both `gate.reviewed_on` and `gate.responded_on` at the time that the NA_SELF vote is recorded.

In this PR:
* Classify a NA_SELF vote as both a review request vote and a review response vote.  Existing logic in slo.py sets `gate.reviewed_on` and `gate.responded_on` if the vote state is in those lists.
* Related topic: Also, bold the word "required" for the explanation field to make it stand out more.


KyleJu (approved):

Varryme:
🔑

PR #4834, opened 2025-03-05
Upgrade to python3.12

jrobbins:
Our cloudtop machines seem to have removed python3.11 so we need to upgrade to 3.12.
I am also upgrading the GAE runtime to match the python version we will use when running locally.

This change seems pretty simple, and I didn't see anything urgent in the python release notes, but it is possible that there is something that I missed and that we will hit some problems with some part of the app.

All developers will need to be sure to have python3.12 on their local machines, and they will need to run `npm run clean-setup`.

Also, add an `r` to a regex to make it a raw string.  The need for this was caught by the 3.12 regex library.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -208,7 +208,7 @@ LOGICAL_OPERATORS_PATTERN = r'OR\s+|-'
 # Structured terms look like: FIELD OPERATOR VALUE.
 # Full-text terms look like: SINGLE_WORD, or like: "QUOTED STRING".
 TERM_RE = re.compile(
-    '(?P<logical>%s)?(?:(?P<field>%s)(?P<op>%s)(?P<val>%s)|(?P<textterm>%s))\s+' % (
+    r'(?P<logical>%s)?(?:(?P<field>%s)(?P<op>%s)(?P<val>%s)|(?P<textterm>%s))\s+' % (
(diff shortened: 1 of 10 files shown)


internals/search.py
```diff
@@ -208,7 +208,7 @@ def parse_query_value_list(
 # Structured terms look like: FIELD OPERATOR VALUE.
 # Full-text terms look like: SINGLE_WORD, or like: "QUOTED STRING".
 TERM_RE = re.compile(
-    '(?P<logical>%s)?(?:(?P<field>%s)(?P<op>%s)(?P<val>%s)|(?P<textterm>%s))\s+' % (
+    r'(?P<logical>%s)?(?:(?P<field>%s)(?P<op>%s)(?P<val>%s)|(?P<textterm>%s))\s+' % (
```
jcscottiii, line 211: Just checking: did you mean to change this line?

jcscottiii (commented):
LGTM

internals/search.py
jrobbins: Yes.  Thanks for pointing that out.  I just added it to the PR description.

jcscottiii: I just checked and the other patterns above in the file are also raw strings so this makes sense.

PR #4835, opened 2025-03-06
Optional redis instructions

siddhiguptas:


abhayporwals:
hi, @jrobbins! Given that the Redis client fails to connect at startup, if we set redis_client = None, the server will never use Redis for the rest of its uptime. Would you prefer an alternative approach, such as:

- Retry Mechanism – Implementing an automatic retry with exponential backoff for reconnection attempts.
- Periodic Health Checks – Running a background task to attempt reconnection at intervals.
- Lazy Initialization – Retrying Redis initialization whenever a cache operation is attempted.

Let me know your thoughts on the preferred approach!

siddhiguptas:
hi, @jrobbins ! Given that the Redis client fails to connect at startup, if we set redis_client = None, the server will never use Redis for the rest of its uptime. Would you prefer an alternative approach, such as:

Retry Mechanism – Implementing an automatic retry with exponential backoff for reconnection attempts.
Periodic Health Checks – Running a background task to attempt reconnection at intervals.
Lazy Initialization – Retrying Redis initialization whenever a cache operation is attempted.
Let me know your thoughts on the preferred approach!

DanielRyanSmith:
I could be wrong here, but I think we're not taking fixes for this issue for now. I've removed the "goodfirstbug" label from this issue. Thank you for your initial attempt at contribution.

Issue #4790, opened 2025-02-19
Change the roadmap page to consider approvals rather than implementation status


jrobbins:
Our current roadmap page was designed before we implemented approval tracking in chromestatus.  It decides whether to include a feature and where to place it based on `FeatureEntry.impl_status_chrome` and the relevant milestone fields.  This is problematic because feature owners sometimes don't update `impl_status_chrome`, or they update it without regard for approvals.  Because feature owners often forgot to set `impl_status_chrome`, we include new incubation and deprecation features in the roadmap regardless of `impl_status_chrome`. That can be a problem because we sometimes show features that are no longer being pursued or that have not gotten needed approvals.

The current milestone card sections are:
* Deprecations
* Enabled by default
* In developer trial
* Origin trial

In the new approach, we will

- [x] Use `impl_status_chrome` only to avoid displaying features that are no longer being pursued.
- [x] Otherwise, display all features according to their type and milestones
- [x] Label all features with a small badge that summarizes that gate's review state
- [x] Include rollout stages
- [ ] Label features that have experiments

PR #4838, opened 2025-03-08
Add gate status badge to features on the roadmap page

jrobbins:
Resovles #4790.

We already changed the roadmap to display features based on milestone, regardless of `impl_status_chrome` (unless the feature owner explicitly set a status that indicates that the features is not shipping).  That leaves the other half of that issue, which is to add a review status badge.

In this PR:
* Add an `earliest_milestone` field to the Gates API responses.  This includes the needed NDB queries, converter logic, and openapi changes.  It is needed so that we can consider only the feature gates relevant to a given roadmap feature item.
* Create an element `chromedash-review-status-icon` that fetches gates and checks their state to chooses which icon symbol to display and which gate to navigate to if the user clicks it.
* Add the new icon to the front of each roadmap feature item.

One tricky bit of this change was that `chromedash-roadmap` needed to change to use `repeat()` to render `chromedash-roadmap-milestone-card` elements rather than using `map()`.  That is needed because Lit reuses DOM elements by default, so when the user scrolls sideways the same `chromedash-roadmap-milestone-card` element is used over and over.  That was a problem because the gates are only fetched once for each `chromedash-review-status-icon`, so they could link to the wrong gates after scrolling.


jrobbins:
While testing this on staging I found that the earliest_milestone approach is not good enough.  Specifically, a feature could ship in milestone 140 on android and 142 on desktop.  earliest_milestone would be 140, so the icon would appear correctly there, but the same feature would also be listed for 142 with no icon.

I am going to try a different API change and see if that works better.

jrobbins:
I have the improved designed implemented.  I just need to update the tests.

jrobbins:
Tests are updated and the PR is ready for re-review.

KyleJu (approved):
LGTM again

PR #4855, opened 2025-03-12
Set longer web test timeout and reduce concurrency

jrobbins:
Yesterday it was quite difficult to get a PR through CI (and when running locally) because of repeated failures of Firefox to start within 30 seconds.  The error message says to increasing the timeout, which can only be done in a config file.  However, that did not seem to completely resolve the problem.  So, I also tried specifying the level of concurrency.  Concurrency previously defaulted to half the number of CPUs available, which could be kind of a lot, and I imagine could trigger concurrent browsers to deadlock on some resource.

In this PR:
* Add a config file for running web tests one at a time and allow longer for the browser to start up.


KyleJu:
This is not ready?

jrobbins:
> This is not ready?

It is ready now.  I was running the CI checks multiple times to validate that it would consistently pass.  Using concurrency 4 did not help enough, so I lowered it to concurrency 1.

KyleJu (approved):

PR #4856, opened 2025-03-12
Limit num= to 1000.

jrobbins:
Now that we have over 3000 `FeatureEntry`s in our NDB, it is possible that a user could request all of them, which will take a long time and could slow the site for other users.  This adds a limit to the number of search results in a given response.  A user could still send multiple requests in parallel, but that would at least spin up more GAE instances to handle those requests.


KyleJu (commented):
Would this affect the pagination of the API?

jrobbins:
> Would this affect the pagination of the API?

The client would still get an accurate `total_count`.  The response can always have a subset of the total list of items, so that is no different.  Of course, a client that asked for num=10000 would only get back 1000 items so it would need to know to make its next request be start=1000 rather than start=10000.  I added logic for that in our UI pagination element.  Other clients will need to figure it out when they see results capped at 1000.

KyleJu (approved):

PR #4860, opened 2025-03-13
update devcontainer setup

jcscottiii:
When rebuilding my devcontainer environment for python 3.12, I saw that it failed to start because the previous location for the python image has been archived and did not receive anymore updates after python 3.11
https://github.com/microsoft/vscode-dev-containers

Updated the image to the one in the updated location. README: https://github.com/devcontainers/images/blob/main/src/python/README.md

Also, updated the docker-compose setup in the config to get that working. (Maybe it was needed when moving from bullseye to bookworm?)


jrobbins (approved):
Thanks. I was just relying on CI to check my previous change to this.

PR #4874, opened 2025-03-18
Add label for experiments on roadmap.

### LINKED ISSUE #4790: Change the roadmap page to consider approvals rather than implementation status

jrobbins:
This should resolve #4790.  It adds a label on features on the roadmap page to indicate ones that have a finch experiment.  Most WP features do not deploy as experiments, so the ones that do are highlighted.  The user can click the link to go to the page that the feature owner provided for the experiment.

In this PR:
* Simplify the layout CSS for roadmap feature items by using `.hbox` and adding `.align-top`.  Also, allow long feature names to wrap rather than end with `...`.
* Add a `<a>` link for each experiment listed for each feature in the server's response.  Group it with a the feature name so that they can both wrap together.
* Simplify the server-side logic for generating the roadmap responses.  Rather than using one python variable for each reason that a feature could be in the roadmap, use one variable for each section of the roadmap.
* Refactor the server-side logic that added the `roadmap_stage_ids` and now also add `finch_urls`.


jrobbins:
Screenshot:
image

KyleJu (approved):

PR #4875, opened 2025-03-18
Offer NA_VERIFIED for review audits.

jrobbins:
As part of our support for self-certification, we need to support reviewers doing audits to double-check that a self-certified N/A is actually valid.  This PR adds support for a new Vote value NA_VERIFIED.

In this PR:
* In review_models.py: Define the new vote value.
* In reviews_api.py: check that NA_VERIFIED is being cast only if the gate state was NA_SELF (or NA_VERIFIED, in case two reviewers both verify).
* In chromestatus-gate-chip.ts: Display NA_VERIFIED as "N/a (ver)" with a green background and checkmark icon.
* In chromestatus-gate-column.ts: Display an explanation at the top of the column for gates in the NA_VERIFIED state, and offer the option to vote NA_VERIFIED only if there was some NA_SELF or NA_VERIFIED vote already.
* In approvals_defs.py: Simply the vote counting logic by using `collections.Counter` and add logic to make the gate verified if there are any relevant NA_VERIFIED votes.

### DIFF
--- a/api/reviews_api.py
+++ b/api/reviews_api.py
@@ -70,6 +70,7 @@ class VotesAPI(basehandlers.APIHandler):
         feature_id=feature_id, gate_id=gate_id, set_by=user.email())
     old_state = old_votes[0].state if old_votes else Vote.NO_RESPONSE
     self.require_permissions(user, fe, gate, new_state)
+    self.check_voting_rules(gate, new_state)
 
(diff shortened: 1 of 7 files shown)


jrobbins:
Screenshot of a verified gate:
image

KyleJu (approved):

PR #4885, opened 2025-03-22
Fix: Add 'No issues found' option to review status dropdown

tiwaryfied:
This PR adds a new option, "No issues found", to  client-src/elements/form-field-enums.ts to improve clarity in security and privacy review statuses.

This ensures that reviews where no issues are detected can be clearly marked, avoiding ambiguity between "Pending" and "Not applicable"


jrobbins:
I'm glad that you are getting familiar with our code.  If you were to continue with this change, you would need to change both the client-side and the server-side.

However, I don't think that the review teams desire this change.  They have been successfully using "Approved" for the situation in which a review was done and now issues were found.  So, we are unlikely to merge this change.

PR #4901, opened 2025-03-27
Add multiple spec link feature

suprith-hub:


suprith-hub:
@jrobbins Could I get a quick go-ahead on the draft PR? I’d like to ensure I’m on the right track before finalizing the remaining changes for this PR.

DanielRyanSmith:
This would require multiple PRs and incremental changes to fix the issue. We can't simply replace `spec_link` with `spec_links`.

We would need to:
1. Create a new field in FeatureEntry, `spec_links`. Leave the existing `spec_link` field.
2. Add new logic to save the `spec_link` field changes from the UI to both `spec_link` and `spec_links`.
3. Create a migration script to move all existing `spec_link` values into the new `spec_links`.
4. (Deploy the above changes so that there is no user interruption).
5. Fix all references in the UI and backend to point to `spec_links` rather than `spec_link`.
6. (Deploy the above changes so that there is no user interruption).
7. Deprecate the old `spec_link` field.

Thank you for the submission here. I'm going to close it due to the long period of time it has been open.

PR #4921, opened 2025-04-04
Clarify the updated timestamp on the releasenotes page

jrobbins:
The Enterprise team asked for the timestamp for editing a feature to be shown on the release notes page.  It was actually already there, but not in a very clear format.

In this PR:
* Instead of making that whole paragraph bold, just bold the labels so that the values and labels are more differentiated.
* Replace the timestamp string with `<sl-relative-time>` and link to the feature activity page for more details.
* When the user saves an edit, update the timestamp locally so that it will show `(this minute)` and they can see which ones they have already edited


jrobbins:
Old screenshot:
image

New screenshot:
image

KyleJu (approved):

PR #4923, opened 2025-04-04
Add rollout_plan field and include it in I2S emails

jrobbins:
This was a suggestion from Dharani to help align ChromeStatus's fields with those used by the internal launch tracking tool.  We will now ask feature owners to specify what their rollout plans are by choosing from a small set of enums.

In this PR:
* Add a new field to FeatureEntry for rollout_plan and include it in the features_api responses
* Define 4 enum values for it
* Add a form field for it in the shipping stage
* Display its string value on the feature detail page and in amendments
* Include it in intent-to-ship emails


internals/core_enums.py
```diff
@@ -150,6 +150,19 @@
...
+ROLLOUT_0_THEN_100 = 1
+ROLLOUT_GRADUAL = 2
+ROLLOUT_OTHER = 3
+
+ROLLOUT_PLAN_DISPLAYNAMES = collections.OrderedDict([
+  (ROLLOUT_100, 'Will ship enabled for all users'),
+  (ROLLOUT_0_THEN_100, '(RARE) Ships disabled, then flips on for all users'),
+  (ROLLOUT_GRADUAL, '(RARE) Experiment users ramp up over time'),
+  (ROLLOUT_OTHER, "(RARE) It's complicated or unusual"),
+])
```
DanielRyanSmith, line 164: Why is an OrderedDict used here? I don't see anything that references the order here

internals/core_models.py
```diff
@@ -116,6 +116,7 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
   finch_name = ndb.StringProperty()
   non_finch_justification = ndb.TextProperty()
   ongoing_constraints = ndb.TextProperty()
+  rollout_plan = ndb.IntegerProperty(required=True, default=ROLLOUT_100)
```
DanielRyanSmith, line 119: I imagine this is the case, but hopefully this will now assume that all previous FeatureEntry entities have a value here and this will not require backfilling.

KyleJu (approved):

internals/core_enums.py
jrobbins: Good point.  I was just copying the existing pattern.  However, in python these days, all dicts are ordered.  I'll try changing it.

internals/core_models.py
jrobbins: Correct, the default value is used when an existing stored lacks that field.

Issue #4736, opened 2025-01-27
Add new security review API client


DanielRyanSmith:
This is part of the work for integrating ChromeStatus with IssueTracker API in order to create and reference security review bugs. A number of requests to the OT API will need to be implemented as a ChromeStatus client.

- Check if a given continuity ID is valid, and to check if a launch bug exists for that continuity ID.
- Create a new launch bug.
- Get the status of some number of bug IDs to check for updates.

PR #4925, opened 2025-04-05
New Security Review API

DanielRyanSmith:
Part of #4736

This change adds the new SecurityReviewsAPI that handles requests related to the security review process, as well as adds a new (not yet available) endpoint to create a new security launch issue through the OT API.


gen/py/webstatus_openapi/webstatus_openapi.egg-info/requires.txt
```diff
@@ -0,0 +1,4 @@
+urllib3<2.1.0,>=1.25.3
```
jrobbins, line 1: I suppose that we are generating egg-info for the first time now because we upgraded openapitools/openapi-generator-cli in #5037 and this is the first API change since then.

PR #4991, opened 2025-05-01
Allow feature editors to archive.

### LINKED ISSUE #4633: Allow any feature owner to delete a feature entry
Right now it is limited to the feature creator and to site admins.  The we should allow feature owners to do it.

jrobbins:
Resolves #4633.  This aligns the server-side permission check with what was already being done on the client-side.  Any user who can edit a feature can also archive it (which is a soft-delete).

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import flask
 from datetime import datetime
 import re
 from typing import Any
@@ -338,7 +339,7 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         stage_ids)
     feature.put()
 
-  def do_patch(self, **kwargs):
+  def do_patch(self, **kwargs) -> flask.Response | dict[str, str]:
     """Handle PATCH requests to update fields in a single feature."""
     body = self.get_json_param_dict()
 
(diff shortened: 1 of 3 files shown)


api/features_api.py
```diff
@@ -375,17 +376,19 @@ def do_patch(self, **kwargs):
...
 
-  @permissions.require_create_feature
-  def do_delete(self, **kwargs) -> dict[str, str]:
+  def do_delete(self, **kwargs) -> flask.Response | dict[str, str]:
     """Delete the specified feature."""
     # TODO(jrobbins): implement undelete UI.  For now, use cloud console.
-    feature_id = kwargs.get('feature_id', None)
+    if 'feature_id' not in kwargs:
+      self.abort(404, msg='Feature ID not specified')
+    feature_id: int = int(kwargs.get('feature_id', 0))
```
DanielRyanSmith, line 384: Nit: We might want to also check to confirm here that feature ID is an integer before trying to cast it.

```py
    if not isinstance(kwargs['feature_id'], int):
      self.abort(400, msg='Invalid feature ID.')
```

Not required though, and unlikely to cause much problems. 🙂

api/features_api.py
```diff
@@ -375,17 +376,19 @@ def do_patch(self, **kwargs):
...
+    # Validate the user has edit permissions and redirect if needed.
+    redirect_resp = permissions.validate_feature_edit_permission(
+        self, feature_id)
+    if redirect_resp:
+      return redirect_resp
```
DanielRyanSmith, line 389: This will mean that site editors have permission to archive any feature. I don't have any issue with that, but it might be something to consider.

DanielRyanSmith, line 389: I'm realizing after seeing this that there is no redirect that can occur with an API request like this, so this will not behave the expected way (i.e. redirecting the user somewhere else). Instead we should just have an abort with a reason.

```suggestion
    # Validate the user has edit permissions and redirect if needed.
    redirect_resp = permissions.validate_feature_edit_permission(
        self, feature_id)
    if redirect_resp:
      self.abort(403, msg='User does not have permission to edit this feature.')
```

I know many of the endpoints we have are doing this incorrectly, so that's something to take care of in the future. 😅

DanielRyanSmith (approved):
A couple of nits, but otherwise LGTM 🙂

api/features_api.py
jrobbins: Good point.  Actually, it should always be an int because it is declared as an int path parameter in main.py.  The cast to int should probably never have been there.

api/features_api.py
jrobbins: Yeah, that seems fine.  Worst case, we can always unarchive.

PR #5029, opened 2025-05-16
Make the form field capture feature IDs.

jrobbins:
Change the form field to capture Web Feature IDs instead of UseCounters.

In this PR:
* Define a new API endpoint to return the list of web feature IDs.  Unlike the previous implementation that was UI-centric, this one just gives a list of strings for the web feature IDs.  Also, add it to the openapi spec for documentation.
* In the UI, add "Missing feature" and format it into a JS object as expected by the rendering logic.
* The regular editing form processing was handled in a previous PR, but this one includes a needed update to guide.py for the feature creation form.


api/webdx_feature_api.py
```diff
@@ -16,12 +16,24 @@
...
+
+  def do_get(self, **kwargs):
+    """Returns a sorted list of WebDX feature IDs."""
+    singleton: WebdxFeatures | None = WebdxFeatures.get_webdx_feature_id_list()
+    web_feature_ids = sorted(singleton.feature_ids if singleton else [])
+    return web_feature_ids
+
+
+# TODO(jrobbins): Move the ot_webfeature_use_counter form field a shipping
+# stage and change have it use datalist populated by this API.
```
DanielRyanSmith, line 34: Nit: wording on this might need to be clarified.

pages/guide.py
```diff
@@ -74,9 +74,7 @@ def process_post_data(self, **kwargs):
       }):
       enterprise_notification_milestone = get_default_first_notice_milestone_for_feature()
 
-    webdx_usecounter_enum = WebDXFeatureObserver.get_enum_by_web_feature(
-      self.form.get('web_feature')
-    )
+    web_feature = self.form.get('web_feature', '')
```
DanielRyanSmith, line 77: I had almost forgotten that this file still exists - this process should feasibly be handled by a POST to the features API, but I suppose that cleanup is not a high priority. 🙂

DanielRyanSmith (approved):

api/webdx_feature_api.py
jrobbins: Fixed

PR #5065, opened 2025-05-28
Remove Dev API

DanielRyanSmith:
I'm pretty sure this is not being used by anyone (I don't use it and I implemented it 😅). It's best for this to be removed since it would be misleading to new contributors, as this code is not maintained. If we want to populate data for a local environment, then pulling data from staging or programmatically generating data similar to webstatus.dev should be the path forward.

### DIFF
--- a/main.py
+++ b/main.py
@@ -370,11 +370,16 @@ internals_routes: list[Route] = [
         maintenance_scripts.SendManualOTActivatedEmail),
 ]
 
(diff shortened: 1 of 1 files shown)


jrobbins:
I agree with this, but isn't the mock login used by the playwright tests?

jrobbins (approved):

PR #5082, opened 2025-06-03
Add cron job to generate Review Activities CSV file

DanielRyanSmith:
This change adds a new cron job that generates a new CSV file, `chromestatus-review-activity.csv` containing all review activity that has occurred in ChromeStatus. The file is uploaded to Google Cloud Storage, where it can be ingested by CDA feeds to make the data available for the Skyhook dashboard.

My original implementation exposed a new endpoint that generates the data on-demand, but after investigating the ingestion process further, this approach is a bit more feasible and streamlined. The file is regenerated every day, but we can adjust this cadence as needed.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -946,10 +946,11 @@ class GenerateReviewActivityFile(FlaskHandler):
     return csv_rows
 
   def _get_last_run_timestamp(self, bucket):
-    """Get the starting timestamp for """
+    """Get the timestamp for the starting interval of querying for new
+    activities."""
     blob = bucket.blob('review-activity-last-timestamp.txt')
     if blob.exists():
-      with blob.open('r')as f:
+      with blob.open('r') as f:
         timestamp_str = f.read()
       return datetime.strptime(timestamp_str, self.DATE_FORMAT)
     # If no previous timestamp exists, start from the beginning.
@@ -960,9 +961,11 @@ class GenerateReviewActivityFile(FlaskHandler):
     does not exist."""
     blob = bucket.blob('chromestatus-review-activity.csv')
     csv_contents = ''
+    logging.info(f'adding {len(csv_rows)} new rows to CSV.')
     if blob.exists():
       with blob.open('r') as f:
         existing_csv = f.read()
+        logging.info(f'Existing csv is {existing_csv.count("\n")} lines long')
         csv_contents = existing_csv + '\n' + '\n'.join(csv_rows)
     else:
       csv_contents = (
@@ -988,4 +991,5 @@ class GenerateReviewActivityFile(FlaskHandler):
     self._write_csv(bucket, csv_rows)
     self._write_last_run_timestamp(bucket, now)
 
-    return f"File chromestatus-review-activity.csv uploaded."
+    return (f'{len(csv_rows)} '
+            'new rows added to chromestatus-review-activity.csv uploaded.')
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -963,10 +963,10 @@ class GenerateReviewActivityFile(FlaskHandler):
       with blob.open('a') as f:
         f.write('\n'.join(csv_rows))
     else:
-      csv_rows = [
+      csv_rows.insert(
+        0,
         'FeatureID,TeamName,EventType,EventDate,ReviewStatus,ReviewAssignee,'
-        'Author,Content'
-      ].extend(csv_rows)
+        'Author,Content')
       blob.upload_from_string('\n'.join(csv_rows))
 
   def _write_last_run_timestamp(self, bucket, timestamp: datetime) -> None:
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -16,7 +16,7 @@ import collections
 from datetime import date, datetime
 import logging
 from typing import Any
-from google.cloud import ndb  # type: ignore
+from google.cloud import ndb, storage  # type: ignore
 import requests
 
 from api import converters, channels_api
@@ -32,6 +32,7 @@ from internals.feature_links import batch_index_feature_entries
 from internals import stage_helpers
 from internals.webdx_feature_models import WebdxFeatures
 from webstatus_openapi import ApiClient, DefaultApi, Configuration, ApiException, Feature
+from chromestatus_openapi.models import ReviewActivity as ReviewActivityModel
 import settings
 
 class EvaluateGateStatus(FlaskHandler):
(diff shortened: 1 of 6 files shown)


DanielRyanSmith:
I've tested this on staging and permissions seem to work as expected. The file is successfully uploaded to a bucket in the `cr-status-staging` project.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,69 @@ def get_template_data(self, **kwargs):
...
+    activities: list[Activity] = Activity.query(
+      ).order(Activity.created).fetch()
+
+    # Filter deleted activities the user can't see, and activities that have
+    # no gate ID, meaning they do not represent review activity.
+    activities = list(filter(
+      lambda a: (a.deleted_by is None
+                 and a.gate_id is not None),
+      activities))
+    gate_ids = set([a.gate_id for a in activities])
```
jrobbins: Nit: This is slightly more efficient without the square brackets.  A set can be constructed from an iterator.  So it could be: `set(a.gate_id for a in activities)`.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,69 @@ def get_template_data(self, **kwargs):
...
+  DATE_FORMAT = '%Y-%m-%dT%H:%M:%S'
+  
+  def _generate_csv(self):
+    # Note: We assume that anyone may view approval comments.
+    activities: list[Activity] = Activity.query(
+      ).order(Activity.created).fetch()
+
+    # Filter deleted activities the user can't see, and activities that have
+    # no gate ID, meaning they do not represent review activity.
+    activities = list(filter(
```
jrobbins, line 909: We should probably filter out activities on deleted features.  Maybe just add a TODO for that and we can check with Brandon about the desired behavior.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,69 @@ def get_template_data(self, **kwargs):
...
     return 'Email task enqueued'
+
+
+class GenerateReviewActivityFile(FlaskHandler):
+  """Generate a CSV file with all review activity in ChromeStatus."""
+  DATE_FORMAT = '%Y-%m-%dT%H:%M:%S'
+  
+  def _generate_csv(self):
+    # Note: We assume that anyone may view approval comments.
+    activities: list[Activity] = Activity.query(
```
jrobbins, line 900: This is getting all activities ever.  Can we limit it to a start date and a maximum number of results like with did for the API approach?  Or does the ingestion expect a complete CSV file?  There are currently 25,000 Activity entities on prod, and 12000 of those were added in the past year.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,69 @@ def get_template_data(self, **kwargs):
...
+          review_status,
+          review_assignee,
+          a.author or '',
+          comment,
+        ]
+      ))
+    return csv_rows
+
+  def get_template_data(self, **kwargs):
+    self.require_cron_header()
```
jrobbins, line 983: There are some concerns about the security of serving CSV files and the potential for tricking a browser's content sniffing.  I think that we are OK here because we never send the CSV in the response.

jrobbins (commented):
Looks pretty good, but I am wondering about scalability of the number of rows generated.

internals/maintenance_scripts.py
DanielRyanSmith: Yeah, it does take some time to generate this CSV, and as of now, the approach is to regenerate the entire file.

The good news is that the CDA feed can support CSV sizes of [up to 5-10 GB](http://go/cda-feeds-transform-formats#supported-file-sizes), which I don't think we'll come close to. The computational time is likely the bottleneck.

A possible approach would be to pull the CSV file down from GCS and then append additional rows to it, rather than generating it from scratch. Do you think this is a better approach?

There is also the issue of deleted comments, and that we do not update any timestamp on the entity when it is flagged for deletion, but that is a problem that would occur relatively rarely.

internals/maintenance_scripts.py
DanielRyanSmith: Good catch - I've updated this. 🙂

internals/maintenance_scripts.py
DanielRyanSmith: TODO added!

internals/maintenance_scripts.py
jrobbins: OK, let's move ahead with the current approach of regenerating the entire file and then measure how well that works on prod.  The request timeout is 10 minutes and 32MB.

DanielRyanSmith: I'm not sure how much this will help, but I added timestamp compatibility so that we'll only look at entries for the last day. However, GCS does not support "append" actions, so we have to pull the existing CSV down and add the new rows to it. However, this is still an improvement, since we're avoiding querying for every activity in the DB each time we run this. 🙂

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,108 @@ def get_template_data(self, **kwargs):
...
+          review_status,
+          review_assignee,
+          a.author or '',
+          comment,
+        ]
+      ))
+    return csv_rows
+
+  def _get_last_run_timestamp(self, bucket):
+    """Get the starting timestamp for """
```
jrobbins: Your docstring is incompl

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,108 @@ def get_template_data(self, **kwargs):
...
+    blob = bucket.blob('review-activity-last-timestamp.txt')
+    if blob.exists():
+      with blob.open('r')as f:
```
jrobbins: nit: space before `as`.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,108 @@ def get_template_data(self, **kwargs):
...
+          approval_defs.APPROVAL_FIELDS_BY_ID[gate_type].team_name,
+          a.amendments[0].field_name if len(a.amendments) else 'comment',
+        str(datetime.strftime(a.created, self.DATE_FORMAT)),
+          review_status,
+          review_assignee,
+          a.author or '',
+          comment,
+        ]
+      ))
+    return csv_rows
```
jrobbins, line 946: Let's also log the number of csv rows generated.  I don't have a specific use for it in mind, but it could be useful if we ever need to debug.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,108 @@ def get_template_data(self, **kwargs):
...
+
+  def _write_csv(self, bucket, csv_rows: list[str]) -> None:
+    """Append the rows to the review activity CSV, or create a new CSV if it
+    does not exist."""
+    blob = bucket.blob('chromestatus-review-activity.csv')
+    csv_contents = ''
+    if blob.exists():
+      with blob.open('r') as f:
+        existing_csv = f.read()
+        csv_contents = existing_csv + '\n' + '\n'.join(csv_rows)
```
jrobbins, line 969: Let's log the size of existing_csv.

internals/maintenance_scripts.py
```diff
@@ -883,3 +884,108 @@ def get_template_data(self, **kwargs):
...
+    else:
+      csv_contents = (
+        'FeatureID,TeamName,EventType,EventDate,ReviewStatus,ReviewAssignee,'
+        'Author,Content'
+      ) + '\n'.join(csv_rows)
+    blob.upload_from_string(csv_contents)
```
jrobbins, line 975: And, log the size of the new csv_contents.

jrobbins (approved):
Still looks good.  I had a couple more small suggestions.

internals/maintenance_scripts.py
DanielRyanSmith: Agreed. I've added the number of rows to the return value, as well as a log statement.

internals/maintenance_scripts.py
DanielRyanSmith: Added!

Fabiansatachiathayde:
What about receipts and money when is that coming back not willing to collaborate

PR #5083, opened 2025-06-03
try array-grouping

jcscottiii:


client-src/elements/chromedash-feature-table.ts
```diff
@@ -93,14 +93,7 @@ export class ChromedashFeatureTable extends LitElement {
...
       .then(res => {
-        const gatesByFID: Record<number, GateDict[]> = {};
-        for (const g of res.gates) {
-          if (!gatesByFID.hasOwnProperty(g.feature_id)) {
-            gatesByFID[g.feature_id] = [];
-          }
-          gatesByFID[g.feature_id].push(g);
-        }
-        this.gates = gatesByFID;
+        this.gates = Object.groupBy(res.gates, ({feature_id}) => feature_id);
```
jrobbins, line 96: This one is a clear upgrade.  Jules also detected this one.

Personally, I have mixed feelings about using destructuring in the arrow function parameters.  So, there's basically two changes happening even in this small part of the PR: a library usage upgrade, and a coding style change.

client-src/elements/chromedash-report-review-latency-page.ts
```diff
@@ -86,13 +86,15 @@ export class ChromedashReportReviewLatencyPage extends LitElement {
   }
 
   renderFeatureReviewsRow(reviewLatency) {
+    const groupedByGateType = Object.groupBy(reviewLatency.gate_reviews, ({gate_type}) => gate_type);
     const gateLatency = {};
-    for (const gr of reviewLatency.gate_reviews) {
+    for (const gateType in groupedByGateType) {
+      const gr = groupedByGateType[gateType][0]; // Assuming one review per gate_type for a feature
```
jrobbins, line 92: I wonder where this assumption comes from, and I don't think that it is true.  At least it added a comment to mention it.

client-src/elements/chromedash-typeahead.ts
```diff
@@ -142,24 +142,20 @@ export class ChromedashTypeahead extends LitElement {
...
+    const groupedByGroup = Object.groupBy(candidates, ({ group }) => group);
     const result: Candidate[] = [];
-    for (const c of candidates) {
-      if (!groupsSeenTwice.has(c.group)) {
-        result.push(c);
-      } else if (!groupsSeenTwiceProcessed.has(c.group)) {
-        result.push({group: c.group, name: c.group + ':', doc: c.doc});
-        groupsSeenTwiceProcessed.add(c.group);
+
+    for (const [groupName, groupCandidatesArray] of Object.entries(groupedByGroup)) {
```
jrobbins, line 148: This might work and would be a nice upgrade that simplifies the code.  However, I think that order is preserved, but that is now implicit so I am not 100% sure.

client-src/elements/utils.ts
```diff
@@ -83,17 +83,9 @@ export function shouldShowDisplayNameField(feStages, stageType) {
...
-    if (feStages[i].stage_type === stageType) {
-      matchingStageCount++;
-      // If we find two of the same stage type, then display the display name field.
-      if (matchingStageCount > 1) {
-        return true;
-      }
-    }
-  }
-  return false;
+  const stagesGroupedByType = Object.groupBy(feStages, ({stage_type}) => stage_type);
```
jrobbins, line 86: Since we only care about one group, this could actually be a call to `filter()`.  This might be an example of a shortcoming of asking AI to perform one upgrade at a time.

client-src/elements/utils.ts
```diff
@@ -130,13 +122,11 @@ export function unambiguousStageName(
 
   // Count the stages of the same type that appear before this one. This is O(n^2) when it's used to
   // number every stage, but the total number of stages is generally <20.
-  const index = feature.stages
-    .filter(s => s.stage_type === stage.stage_type)
-    .findIndex(s => s.id === stage.id);
+  const stagesOfThisType = Object.groupBy(feature.stages, (s: StageDict) => s.stage_type)[stage.stage_type] || [];
```
jrobbins, line 125: I think the chained method calls and use of .filter() was actually more clear.  This seems like an example of AI finding a place to do an upgrade that we asked for, but it doesn't really improve over the existing code.  Maybe we could provide an additional criteria as part of the prompt, e.g., "only use groupBy() when more than one of the resulting groups are read."

client-src/elements/utils.ts
```diff
@@ -586,61 +576,75 @@ interface FeatureUpdateInfo {
...
       continue;
     }
+    hasChanges = true;
 
-    // Arrays should be submitted as comma-separated strings.
-    let formattedValue = value;
-    if (Array.isArray(formattedValue)) {
-      formattedValue = formattedValue.join(',');
-    }
+    const formattedValue = Array.isArray(value) ? value.join(',') : value;
```
jrobbins, line 600: These off-topic changes make it harder to see the actual upgrade that was requested.

client-src/elements/chromedash-report-review-latency-page.ts
```diff
@@ -86,13 +86,15 @@ export class ChromedashReportReviewLatencyPage extends LitElement {
   }
 
   renderFeatureReviewsRow(reviewLatency) {
+    const groupedByGateType = Object.groupBy(reviewLatency.gate_reviews, ({gate_type}) => gate_type);
```
jrobbins, line 89: gate_type is an int and the docs for Object.groupBy() say to only use it with strings.  I have a feeling that it might work in this case, but Map.groupBy() might be better.

client-src/elements/utils.ts
```diff
@@ -586,61 +576,75 @@ interface FeatureUpdateInfo {
...
+      pendingStageChanges.push({
+        stageId,
+        originalName: name,
         value: formattedValue,
-      };
+      });
+    }
+  }
+
+  for (const change of directFeatureChanges) {
```
jrobbins, line 617: I'm not sure why this for-loop is needed.  The existing code was kind of messy, but this proposed change adds a lot of new lines that don't make it simpler.

jrobbins (commented):
I realize this is not a regular code review. I just put in some comments to capture my thoughts on the proposed upgrade.

PR #5084, opened 2025-06-03
Jules PR to identify opportunities for Object.groupBy()

jrobbins:
Note: At first it complained that groupBy() was "not available in the execution environment".  I eventually figured out that our eslint was configured to generate code that would work for browsers in 2022, whereas groupBy() became newly available in 2024.

Here's Jules's PR description:

Specifically, I set `ecmaVersion` to "latest" in eslint.config.js.

I also identified and marked several locations in TypeScript files within `client-src/elements/` with TODO comments. These are places where I think `Object.groupBy()` could potentially be used to simplify code for grouping arrays of objects.

Here are the files I modified with TODOs:
- client-src/elements/chromedash-enterprise-release-notes-page.ts
- client-src/elements/chromedash-feature-row.ts
- client-src/elements/chromedash-feature-table.ts
- client-src/elements/chromedash-gantt.ts
- client-src/elements/chromedash-report-external-reviews-page.ts
- client-src/elements/chromedash-review-status-icon.ts
- client-src/elements/chromedash-typeahead.ts

### DIFF
--- a/client-src/elements/chromedash-enterprise-release-notes-page.ts
+++ b/client-src/elements/chromedash-enterprise-release-notes-page.ts
@@ -153,7 +153,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
   }
 
   convertShippingStageToRolloutStages(stage) {
-// TODO: Consider using Object.groupBy() here to group items by milestone.
+    // TODO: Consider using Object.groupBy() here to group items by milestone.
     const milestones = [
       stage.desktop_first,
       stage.android_first,
--- a/client-src/elements/chromedash-enterprise-release-notes-page.ts
+++ b/client-src/elements/chromedash-enterprise-release-notes-page.ts
@@ -153,6 +153,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
   }
 
   convertShippingStageToRolloutStages(stage) {
+// TODO: Consider using Object.groupBy() here to group items by milestone.
     const milestones = [
       stage.desktop_first,
       stage.android_first,
@@ -287,6 +288,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
         return impactB - impactA;
       });
 
+    // TODO: Consider using Object.groupBy() here to group items by enterprise_product_category.
     this.currentChromeBrowserUpdates = currentFeatures.filter(
       f =>
         f.enterprise_product_category ===
@@ -330,6 +332,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
           ) || 0;
         return minA - minB;
       });
+    // TODO: Consider using Object.groupBy() here to group items by enterprise_product_category.
     this.upcomingChromeBrowserUpdates = upcomingFeatures.filter(
       f =>
         f.enterprise_product_category ===
--- a/client-src/elements/chromedash-gantt.ts
+++ b/client-src/elements/chromedash-gantt.ts
@@ -139,6 +139,7 @@ export class ChromedashGantt extends LitElement {
 
   // Get lists of all dev trial, origin trial, and shipping stages
   // associated with the feature.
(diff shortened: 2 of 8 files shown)


client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -153,6 +153,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
   }
 
   convertShippingStageToRolloutStages(stage) {
+// TODO: Consider using Object.groupBy() here to group items by milestone.
```
jrobbins: If I saw this TODO, it would not be enough to make the upgrade myself.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -287,6 +288,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
         return impactB - impactA;
       });
 
+    // TODO: Consider using Object.groupBy() here to group items by enterprise_product_category.
```
jrobbins, line 291: This one makes sense, but it a bit hard to see.

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -330,6 +332,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
           ) || 0;
         return minA - minB;
       });
+    // TODO: Consider using Object.groupBy() here to group items by enterprise_product_category.
```
jrobbins, line 335: Same.

client-src/elements/chromedash-gantt.ts
```diff
@@ -139,6 +139,7 @@ export class ChromedashGantt extends LitElement {
 
   // Get lists of all dev trial, origin trial, and shipping stages
   // associated with the feature.
+  // TODO: Consider using Object.groupBy() here to group stages by type (dev_trial, origin_trial, shipping).
```
jrobbins, line 142: Good identification.

client-src/elements/chromedash-report-external-reviews-page.ts
```diff
@@ -132,6 +132,7 @@ export class ChromedashReportExternalReviewsPage extends LitElement {
     args: () => [this.reviewer],
   });
 
+  // TODO: Consider using Object.groupBy() here to group reviews by current_stage.
```
jrobbins, line 135: Good suggestion.  Slightly complicated by the need to sort.

client-src/elements/chromedash-review-status-icon.ts
```diff
@@ -71,6 +71,7 @@ export class ChromedashReviewStatusIcon extends LitElement {
     });
   }
 
+  // TODO: Consider using Object.groupBy() here to group gates by state, then determine overall status.
```
jrobbins, line 74: I can see how groupBy() could be applied here, but using filter() and every() would probably be more clear.  And, given that we want the target gate to be the last one, using a for-loop is understandable, but I suppose that we could reverse the list first.

client-src/elements/chromedash-typeahead.ts
```diff
@@ -141,6 +141,9 @@ export class ChromedashTypeahead extends LitElement {
     return !COMPARE_OPS.some(op => s.includes(op));
   }
 
+  // TODO: Consider using Object.groupBy() here to group candidates by `c.group`.
```
jrobbins, line 144: This is a pretty good suggestion to use groupBy().  But, one thing that becomes less clear is that the order of the groups corresponds to the order of the original items.

eslint.config.js
```diff
@@ -11,7 +11,7 @@ export default [
   },
   {
     languageOptions: {
-      ecmaVersion: 2022,
+      ecmaVersion: "latest",
```
jrobbins, line 14: I had to specifically add this to the prompt. 
It was set to 2022 as part of my upgrading the config file from an older format.
This is probably an example of something that could hold back upgrades in other repos if they also have older config files that seemed to be working well enough so far.

PR #5091, opened 2025-06-06
One more review activity CSV fix

DanielRyanSmith:
One last fix that I missed yesterday. 😞 A new line is not added for the header row with the current iteration. This fixes that.


jcscottiii (approved):
LGTM as-is. But we should consider using the python standard library csv package.

https://docs.python.org/3.10/library/csv.html

That would automatically handle this for us

You can use io.stringsio as the handle

PR #5131, opened 2025-06-16
Fix OT requests admin page

DanielRyanSmith:
This change updates the OT requests admin page to be more visually useful for long-term use. The data is partitioned in a more effective way and is more actionable.

The "Copy to Spreadsheet" functionality has been completely removed, since the OT Pipeline spreadsheet has been completely deprecated.

Additionally, only extension stages that have been approved but not initiated are now displayed on the page, rather than all OT extension requests.

This was an attempt to use Gemini to upgrade this page. Everything seems to function as expected here, but there could be something I've overlooked. The risk is low here since this is a page visible only to admins.

## New View Examples
Screenshot from 2025-06-16 21-43-04

Screenshot from 2025-06-16 21-39-18

### DIFF
--- a/pages/ot_requests.py
+++ b/pages/ot_requests.py
@@ -22,6 +22,7 @@ from internals.core_enums import (
     OT_CREATION_FAILED,
     OT_ACTIVATION_FAILED)
 from internals.core_models import Stage
+from internals.review_models import Gate, Vote
 
 from framework import basehandlers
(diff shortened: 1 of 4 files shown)


DanielRyanSmith:
I've added some unit tests for how the page is being populated, but some screenshot tests might also be useful here. Hopefully this will do for now. 🙂

PR #5136, opened 2025-06-16
New helper method for fetching entities

DanielRyanSmith:
We do a lot of fetching specific entities based on IDs, especially in our API handlers. The checks to ensure the ID inputs are good gets redundant (Make sure ID is provided, ID is the correct format and type, and that the entity exists in the DB). This change adds a new helper method (get_validated_entity) to streamline these checks a bit more.

Now, the potential entity ID, as well as the entity type, are provided to the method and the validation checks are all handled as needed, and the process is aborted if needed.

I've gone ahead and updated the Stages API methods as an example of the simplification this method can add to the codebase.

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -64,8 +64,8 @@ URL_RE = re.compile(r'\b%s%s%s\b' % (
     SCHEME_PATTERN, DOMAIN_PATTERN, PATH_PARAMS_ANCHOR_PATTERN))
 ALLOWED_SCHEMES = [None, 'http', 'https']
 
-
-T = TypeVar('T')
+# Generic type variable for our model entities.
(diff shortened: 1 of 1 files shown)


framework/basehandlers.py
```diff
@@ -64,6 +64,8 @@
     SCHEME_PATTERN, DOMAIN_PATTERN, PATH_PARAMS_ANCHOR_PATTERN))
 ALLOWED_SCHEMES = [None, 'http', 'https']
 
+# Generic type variable for our model entities.
+M = TypeVar('M', bound=ndb.Model)
```
jrobbins, line 68: Cool, our first TypeVar.

jrobbins (approved):

Issue #4738, opened 2025-01-27
Add UI components that allows feature owners to supply a continuity ID for security review


DanielRyanSmith:
This is part of the work for integrating ChromeStatus with IssueTracker API in order to create and reference security review bugs. Feature owners should be able to supply a existing continuity ID when requesting a security review.

A new “Continuity ID” feature-level field will exist in ChromeStatus (with one ID associated at maximum with each feature). When the feature owner selects “Request Review” or “Request N/A” in ChromeStatus for the “WP Security” gate in an origin trial or shipping stage, a new modal will appear asking the user to optionally provide a continuity ID if it exists, along with a detailed description of which bug represents the continuity bug.
If the user provides a continuity ID, ChromeStatus will run a query against other existing features to determine if the given continuity ID is already in use. The ID will be rejected if it is found to be associated with another feature.

PR #5149, opened 2025-06-19
New dialog component for Continuity IDs

DanielRyanSmith:
Part of #4738 

This change adds a new dialog component (that is not integrated yet with the application). This dialog component will be displayed when a user attempts to request a security review on a feature that does not already have an associated security "Continuity ID". The dialog allows the user to supply a security Continuity ID that will link their existing security review to this feature. The dialog will handle the cases when the given Continuity ID is invalid, improperly formatted, or is in use by another feature. Additionally, the user can select the button for "I don't have a Continuity ID" to instead create a new security review issue and be delegated a new Continuity ID.

The text and instructions displayed in the dialog were taken from the [security review questionnaire](https://source.corp.google.com/piper///depot/google3/security/spur/config/questionnaires/chrome_security_launch.pi;l=100-117;rcl=721104173).

#### Default view
Screenshot from 2025-06-19 18-47-10

#### Verification and submission in progress
Screenshot from 2025-06-19 18-47-56

#### Verification of existing continuity issue succeeded
Screenshot from 2025-06-19 18-50-36

#### Verification failure
Screenshot from 2025-06-19 18-51-07


client-src/elements/chromedash-continuity-id-dialog.ts
```diff
@@ -0,0 +1,313 @@
...
+        sl-dialog::part(panel) {
+          max-width: 720px;
+          width: 90vw;
+        }
+        sl-dialog::part(body) {
+          padding: var(--sl-spacing-large);
+        }
+
+        p {
+          color: var(--sl-color-neutral-700);
```
jrobbins: Let's avoid these color, line-height, and text-decoration properties.  Instead we can use the defaults that apply to the entire application already.

client-src/elements/chromedash-continuity-id-dialog.ts
```diff
@@ -0,0 +1,313 @@
...
+    const body: {feature_id: number; gate_id: number; continuity_id?: number} =
+      {
+        feature_id: this.featureId,
+        gate_id: this.gateId,
+      };
+    if (usingContinuityId) {
+      body.continuity_id = parseInt(this._idValue);
+    }
+    try {
+      const resp = await window.csClient.createSecurityLaunchIssue(body);
```
jrobbins: It would be closer to the pattern of existing csClient methods if you passed all the needed info to createSecurityLaunchIssue() as parameters and had the logic to form the body inside that method.

client-src/elements/chromedash-continuity-id-dialog.ts
```diff
@@ -0,0 +1,313 @@
...
+          >
+            ${verificationButtonText}
+          </sl-button>
+        </div>
+
+        <div class="footer-actions">
+          <sl-button
+            class="no-id-btn"
+            variant="default"
+            @click=${this._handleSubmitClickNoContinuityId}
```
jrobbins, line 300: I can foresee some users entering the continuity ID and then mistekenly hitting this button simply because it is near the bottom of the dialog.  How about making it disabled whenever `this._canCheck` is true?

jrobbins (changes requested):
This looks pretty good, but I have a few suggestions.

client-src/elements/chromedash-continuity-id-dialog.ts
DanielRyanSmith: I agree here. I've updated them to use app-defined vars. :slightly_smiling_face:

client-src/elements/chromedash-continuity-id-dialog.ts
DanielRyanSmith: Changed!

client-src/elements/chromedash-continuity-id-dialog.ts
DanielRyanSmith: Good idea - I implemented this :slightly_smiling_face:

PR #5155, opened 2025-06-22
fix: accessibility bug in the header logo

FranciscoImanolSuarez:
I add the alt attribute to the img tag

BUG: https://github.com/GoogleChrome/chromium-dashboard/pull/5155


FranciscoImanolSuarez:
Hi @jrobbins , I have a question, why would we set Chrome Logo? Having "Chrome Platform Status" the screen reader would read 
Chrome logo
Chrome Platform Status

In my case I thought it would be better to say, Chrome Platform Status and the logo is decorative for the screen reader.

I also noticed that platforms like gmail do something similar to what I was telling you.

https://www.w3.org/WAI/tutorials/images/decorative/

Sorry for the question, it's just to understand and learn.

jrobbins (approved):

Issue #5154, opened 2025-06-22
Accessibility bug in the header logo


FranciscoImanolSuarez:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
The Chrome logo in the header, being descriptive, should have the alt empty, so that the screen reader understands that it is a decorative image.

**To Reproduce**
1. Go to /
2. Run axe developer tool
3. Locate the error: “Images must have alternative text”.

**Expected behavior**
Adding an empty tag to the img tag would solve the problem.

**Additional context**
https://www.w3.org/TR/WCAG20-TECHS/H37.html
https://dequeuniversity.com/rules/axe/4.1/image-alt

FIX PR: https://github.com/GoogleChrome/chromium-dashboard/pull/5155

jrobbins:
The alt attribute was added.

PR #5195, opened 2025-07-06
Update chromedash-header.ts

love12yadav:
fix(a11y): mark Chrome logo as decorative with empty alt and presentation role

The header logo was interpreted as meaningful content by screen readers. Setting alt="" and role="presentation" ensures it's treated as decorative, fixing the accessibility issue flagged by tools like axe and Lighthouse.

Related Issue: Accessibility bug in the header logo https://github.com/GoogleChrome/chromium-dashboard/issues/5154

I have signed the CLA


jrobbins:
Thanks for the explanation.  I was not thinking about that before.

PR #5199, opened 2025-07-07
Fix occasional missing login button bug

### LINKED ISSUE #4540: On desktop, the Sign In button sometimes does not render
Sign out and then reload the page a bunch of times.  Every once in a while, the Sign In button does not show up.  The console shows an error trying to access `id` of `undefined`.

My initial guess is that it is a race condition.

Panos noticed this happening after adding a stage.  It's not clear what connection there could be, and it is reproducible without that action.

DanielRyanSmith:
Fixes #4540 

This issue was due to the Google Identity Services library not being loaded in before the page is rendered.


jrobbins (approved):
Good catch!

PR #5320, opened 2025-08-13
Add playwright check to make tests more robust.

jrobbins:
One of our most common playwright test failures is after creating a feature it might capture a screenshot with the skeletons still showing.  So, hopefully a longer delay will avoid that.

### DIFF
--- a/client-src/elements/chromedash-feature-page.ts
+++ b/client-src/elements/chromedash-feature-page.ts
@@ -738,7 +738,6 @@ export class ChromedashFeaturePage extends LitElement {
     return html`
       <chromedash-feature-detail
         appTitle=${this.appTitle}
-        .loading=${this.loading}
         .user=${this.user}
         ?canEdit=${this.userCanEdit()}
         .feature=${this.feature}
@@ -754,8 +753,6 @@ export class ChromedashFeaturePage extends LitElement {
   }
 
   render() {
-    // TODO: create another element - chromedash-feature-highlights
-    // for all the content of the <div id="feature"> part of the page
     // If loading, only render the skeletons.
     if (this.loading) {
       return this.renderSkeletons();


packages/playwright/tests/test_utils.js
```diff
@@ -364,7 +364,7 @@ export async function createNewFeature(page) {
 
   // Wait until we are on the Feature page.
   await page.waitForURL('**/feature/*');
-  await delay(1500);
```
jcscottiii, line 367: Since you use sl-skeleton as the loading indicator, thoughts on waiting for the skeleton component to actually be gone?

```ts
const skeletonLocator = page.locator('sl-skeleton');
await expect(skeletonLocator).toHaveCount(0);
```

or 

```ts
const skeletonLocator = page.locator('sl-skeleton');
await skeletonLocator.waitFor({ state: 'hidden' });
```

We started replacing some places where we used `page.waitForTimeout` with something similar in [webstatus](https://github.com/GoogleChrome/webstatus.dev/blob/c764df32411c5c0db46e327119d8c1f93cb0a67e/e2e/tests/utils.ts#L61-L65) and it helped out a lot.

On the notes for [waitForTimeout](https://playwright.dev/docs/api/class-page#page-wait-for-timeout), they actually advise against using that.

jrobbins: Looking for the sl-skeletons to be missing was not working because the gate column also has skeletons that somehow count as visible.  Instead, I changed it to wait for the feature details to appear.

jcscottiii (approved):
I see that you wait for the feature-detail element now. LGTM on these changes.

I have a question out the scope of this PR.

It seems like that element only renders when `this.loading` equals false.

https://github.com/GoogleChrome/chromium-dashboard/blob/f4a66c2c39f036ef667f9a5a30cae18137bff3af/client-src/elements/chromedash-feature-page.ts#L737-L782

But I don't see a loading property for that element

https://github.com/GoogleChrome/chromium-dashboard/blob/f4a66c2c39f036ef667f9a5a30cae18137bff3af/client-src/elements/chromedash-feature-detail.ts#L79-L105

I think we can remove this line:

https://github.com/GoogleChrome/chromium-dashboard/blob/f4a66c2c39f036ef667f9a5a30cae18137bff3af/client-src/elements/chromedash-feature-page.ts#L741

Thoughts?

jrobbins:
> I think we can remove this line:

Good point.  Gone.

PR #5322, opened 2025-08-13
Add web feature id search term

### LINKED ISSUE #5242: Add query term for web_feature_id
We added a `web_feature` field to `FeatureEntry`, but we did not define a query term for it in search_query.py.  It should be a simple addition.

jrobbins:
This should resolve #5242.
When I added the field for the web feature ID, I missed adding it to the list of searchable terms.

### DIFF
--- a/internals/search_queries.py
+++ b/internals/search_queries.py
@@ -268,7 +268,7 @@ QUERIABLE_FIELDS: dict[str, Property] = {
 
     'browsers.chrome.blink_component': FeatureEntry.blink_components,
     'star_count': FeatureEntry.star_count,
-    'tags': FeatureEntry.search_tags,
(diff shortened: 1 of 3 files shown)


internals/search_queries.py
```diff
@@ -270,6 +270,7 @@ def sorted_by_review_date(descending: bool) -> Future:
     'star_count': FeatureEntry.star_count,
     'tag': FeatureEntry.search_tags,
     'feature_notes': FeatureEntry.feature_notes,
+    'web_feature_id': FeatureEntry.web_feature,
```
jrobbins, line 273: This is the line that maps from search term name to NDB field.

jcscottiii: Thanks!

Issue #5234, opened 2025-07-15
Implement a CSV import page for admins


jrobbins:
Sometimes we want to do bulk edits using data from spreadsheets.  E.g., we would like to import the web feature IDs for each chromestatus entry linked in this spreadsheet: http://bit.ly/webdxuc.

I am thinking of:
* A form where the admin can upload a CSV file
* Some `<select>` form fields where they can choose the column that holds the chromestatus entry URL, and the column that holds the web feature ID.
* A button to preview that the changes would be.
* After the preview is viewed, there would be a button to make the changes.

I am thinking that the changes would be made via our existing feature_api and they would cause all the normal permission checks and notification emails individually rather than have a new bulk edit API.

Later we might generalize the form to update some other chromestatus fields.

jrobbins:
This is live now and has been used.
I didn't implement the <select> but the naming convention for the needed columns seems good enough.

PR #5324, opened 2025-08-14
Implement a basic version of csv bulk edit for setting a lot of web feature IDs

jrobbins:
This should resolve issue #5234.

This PR adds a new admin-only page that allows the user to open a local CSV file.  It then parses certain columns that have the chromestatus feature ID and the desired web_feature_id.  Each row of info is displayed in a preview table on the web page.  Then it loads the chromestatus data for each referenced feature and updates the display to show the feature entry name and its current web feature ID.  Those rows that have different values for the existing and desired web feature ID will be highlighted with background colors.  If the user presses submit, then the page will make a series of calls to update each feature to set its web feature ID and update the preview table.  Eventually all existing values with match the desired value and no rows will be highlighted.


jrobbins:
Here's what I have so far.  This is using an old CSV copy of a spreadsheet that was shared in July:
bulk-edit-so-far

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+        showToastMessage(
+          'Some errors occurred. Please refresh the page or try again later.'
+        );
+      });
+  }
+
+  async handleSubmit(e) {
+    e.preventDefault();
+    this.submitting = true;
+    console.log('submitting');
```
jcscottiii:
```suggestion
```

client-src/elements/chromedash-bulk-edit-page.ts
jcscottiii, line 1: Could you add a comment (either on the page itself, or in the code as a comment) about the format of the CSV?

This would be helpful for @foolip. He showed me this [project](https://github.com/foolip/chromestatus-mapper) yesterday. I could see that being helpful when constructing the appropriate format

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+  parseFileContent(fileContent: string) {
+    this.cells = parse(fileContent);
+    this.detectColumns();
+    this.selectItems();
+  }
+
+  getFileAndParse() {
+    var fileField: HTMLInputElement = this.shadowRoot!.querySelector(
+      '#id_file_form_field'
+    ) as HTMLInputElement;
```
jcscottiii:
```suggestion
    const fileField = this.shadowRoot!.querySelector<HTMLInputElement>(
      '#id_file_form_field');
```

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+    if (fileField && fileField.files) {
```
jcscottiii: Question: In the else case, do you want to reset this.parsing to false?

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
+import {LitElement, css, html, nothing} from 'lit';
+import {customElement, state} from 'lit/decorators.js';
+import {FORM_STYLES} from '../css/forms-css.js';
+import {SHARED_STYLES} from '../css/shared-css.js';
+import {handleSaveChangesResponse, showToastMessage} from './utils';
```
jcscottiii:
```suggestion
import {showToastMessage} from './utils';
```

Don't see handleSaveChangesResponse being used

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+        if (e.target && e.target.result) {
+          const fileContent: string = e.target.result as string;
+          this.parseFileContent(fileContent);
+          const promises = this.items.map(item => this.fetchFeature(item));
+          Promise.all(promises).then(() => {
+            this.parsing = false;
+          });
+        }
+      };
+      reader.readAsText(file);
```
jcscottiii: We should use  `await file.text()` instead.

Something like this:

```ts
    if (!fileField?.files?.length) {
      this.parsing = false;
      // Could show a toast too?
      return;
    }

    try {
      const file = fileField.files[0];
      const fileContent = await file.text();
      this.parseFileContent(fileContent);
      await Promise.all(this.items.map(item => this.fetchFeature(item)));
    } catch (error) {
      showToastMessage(
        `An error occurred while processing the file. ${JSON.stringify(error)}`
      );
    } finally {
      this.parsing = false;
    }
```

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+    for (const item of this.items) {
+      if (item.existing != item.desired) {
+        // Save one at a time to avoid a load spike.
+        await this.saveFeature(item);
+      }
+    }
+    this.submitting = false;
+  }
+
+  handleChange(e) {
```
jcscottiii:
```suggestion
  handleChange() {
```

client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -0,0 +1,292 @@
...
+        this.fetchFeature(item);
+      })
+      .catch(() => {
+        showToastMessage(
+          'Some errors occurred. Please refresh the page or try again later.'
+        );
+      });
+  }
+
+  async handleSubmit(e) {
```
jcscottiii: I think you could do this:

```suggestion
  async handleSubmit(e: PointerEvent) {
```

https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event

jcscottiii (approved):
LGTM with consideration of nits and questions

client-src/elements/chromedash-bulk-edit-page.ts
jrobbins: Gone.

client-src/elements/chromedash-bulk-edit-page.ts
jrobbins: Good idea.  Done.

client-src/elements/chromedash-bulk-edit-page.ts
jrobbins: I followed your code example below, which handles this case.

client-src/elements/chromedash-bulk-edit-page.ts
jrobbins: That is much cleaner.  I used that.

PR #5346, opened 2025-08-22
Don't automatically select 0th autocomplete candidate.

jrobbins:
This is the chromestatus change to the typeahead component to make it less aggressive at selecting an autocomplete item.

The root cause is that the shoelace component `sl-menu` automatically selects the 0th menu item whenever the menu items are set.  It is not clear how we could differentiate that from a desired user selection without further complicating a UI element that already has a lot of state.  And, it is not possible to subclass `SlMenu` to override just that behavior because that class uses many `private` methods.

In this PR, I nerf the undesired behavior by prepending an invisible 0th item to the menu.  It is automatically selected, but if the user hits `Enter` when it is selected we do not treat it as an autocomplete completion and instead assume that the user wants to submit the search to the server.  This achieves the desired behavior of only performing a completion after the user hits the down arrow (to move the selection to the 1st or later item) or clicks an item in the menu.

### DIFF
--- a/client-src/elements/chromedash-typeahead.ts
+++ b/client-src/elements/chromedash-typeahead.ts
@@ -26,6 +26,8 @@ export interface Candidate {
   doc: string;
 }
 
+const NONSELECTABLE_ITEM_VALUE = 'NONSELECTABLE_ITEM_VALUE';
+
 @customElement('chromedash-typeahead')
 export class ChromedashTypeahead extends LitElement {
   slDropdownRef: Ref<SlDropdown> = createRef();
@@ -201,14 +203,19 @@ export class ChromedashTypeahead extends LitElement {
 
   // Check if the user is pressing Enter to send a query.  This is detected
   // on keyDown so that the handler is run before the dropdown keyDown is run.
+  // If stopPropagation() is not called, the event will cause a completion.
   handleInputFieldKeyDown(event) {
     if (event.key === 'Enter') {
-      const slDropdown = this.slDropdownRef.value;
+      const slDropdown = this.slDropdownRef
+        .value as ChromedashTypeaheadDropdown;
       if (!slDropdown) return;
+      const currentItem = slDropdown.getCurrentItem();
       if (
         !slDropdown.open ||
-        !(slDropdown as ChromedashTypeaheadDropdown).getCurrentItem()
+        !currentItem ||
+        currentItem.value === NONSELECTABLE_ITEM_VALUE
       ) {
+        // User wants to send the query to the server.
         this._fireEvent('sl-change', this);
         event.stopPropagation();
       }
@@ -279,11 +286,19 @@ export class ChromedashTypeahead extends LitElement {
   }
 
   renderAutocompleteMenu() {
+    // sl-menu automatically selects the first menu item.  This nerfs
+    // that undesired behavior by adding an invisible 0th menu item.
+    const nonselectableItem = html`
+      <chromedash-typeahead-item
+        value="${NONSELECTABLE_ITEM_VALUE}"
+      ></chromedash-typeahead-item>
+    `;
     return html`
       <sl-menu
         @click=${e => e.preventDefault()}
         @sl-select=${this.handleCandidateSelected}
       >
+        ${nonselectableItem}
         ${this.candidates.map(
           c => html`
             <chromedash-typeahead-item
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-typeahead.ts
```diff
@@ -201,14 +203,19 @@ export class ChromedashTypeahead extends LitElement {
 
   // Check if the user is pressing Enter to send a query.  This is detected
   // on keyDown so that the handler is run before the dropdown keyDown is run.
+  // If stopPropagation() is not called, the event will cause a completion.
   handleInputFieldKeyDown(event) {
     if (event.key === 'Enter') {
-      const slDropdown = this.slDropdownRef.value;
+      const slDropdown = this.slDropdownRef
+        .value as ChromedashTypeaheadDropdown;
```
jcscottiii: You can change  the line above from:

```ts
 slDropdownRef: Ref<SlDropdown> = createRef();
```

to 
```ts
  slDropdownRef: Ref<ChromedashTypeaheadDropdown> = createRef();
```

Then here, you can keep the line as `const slDropdown = this.slDropdownRef.value;`

client-src/elements/chromedash-typeahead.ts
```diff
@@ -450,6 +469,9 @@ export class ChromedashTypeaheadItem extends LitElement {
   }
 
   render() {
+    if (this.value === NONSELECTABLE_ITEM_VALUE) {
+      return nothing;
```
jcscottiii: I think this has to be like this:

```suggestion
      return html`${nothing}`;
```

client-src/elements/chromedash-typeahead.ts
```diff
@@ -363,12 +378,16 @@ export class ChromedashTypeaheadDropdown extends SlDropdown {
         if (event.key === 'ArrowDown' && menuItems[currentItemIndex + 1]) {
           this.setCurrentItem(menuItems[currentItemIndex + 1]);
         }
-        if (event.key === 'ArrowUp' && menuItems[currentItemIndex - 1]) {
+        if (
+          event.key === 'ArrowUp' &&
```
jcscottiii, line 386: Instead of doing this additional check with currentItemIndex here and the change to the index below, you could filter out the invisible item before. Then the code here could stay as-is.

jcscottiii (approved):
LGTM! Thanks for doing this! I tried a [version](https://github.com/GoogleChrome/webstatus.dev/compare/jcscottiii/port-fix-for-autocomplete?expand=1) of this on webstatus.dev and it worked!

Other comments are things to consider but not hard requirements. The slDropdownRef change will prevent the need to use `as` for type conversion (since that is unsafe)

client-src/elements/chromedash-typeahead.ts
```diff
@@ -450,6 +469,9 @@ export class ChromedashTypeaheadItem extends LitElement {
   }
 
   render() {
+    if (this.value === NONSELECTABLE_ITEM_VALUE) {
```
jcscottiii, line 480: Also, thoughts of having a subclass of the item type? Then it could have its own render method

client-src/elements/chromedash-typeahead.ts
jrobbins: Done.

client-src/elements/chromedash-typeahead.ts
jrobbins: It was not initially needed, but since I have now added render()'s return type, it became needed.  I added some more type info in a few places to make type checking more precise and to align this code more closely to the webstatus.dev version.  I think the difference came from the fact that webstatus.dev's version started in typescript whereas this one started in javascript and was converted to typescript.

client-src/elements/chromedash-typeahead.ts
jrobbins: I feel the current comparison seems simpler than filtering.

client-src/elements/chromedash-typeahead.ts
jrobbins: If it got more complicated than a single sentinel value and an if-statement, then I might favor subclassing.

jrobbins:
On second thought, using a subclass also looks pretty good.  I'll send a followup PR to switch to that and to keep the two implementations closer together.

PR #5348, opened 2025-08-22
Fix row numbering in bulk edit page

foolip:


foolip:
This is what I'm seeing and attempting to fix:

image

The numbering starts at 2, so maybe just not adding 1 will fix the problem. Untested.

jrobbins:
Oh, I was trying to make it line up with the spreadsheet row numbers.  Don't the headers occupy spreadsheet row 1?

foolip:
That's true, but I wanted to see what the total number of features was, and before scrolling to the top I couldn't understand why it was off by one.

If this was intentional it's OK, I just assumed it was accidental.

jrobbins:
> That's true, but I wanted to see what the total number of features was, and before scrolling to the top I couldn't understand why it was off by one.
> 
> If this was intentional it's OK, I just assumed it was accidental.

OK, to support that use case, I'll display a count of total rows, matching rows, and rows to update.

foolip:
That sounds great, thanks @jrobbins!

jrobbins:
OK, see my new PR #5350.

jrobbins:
I'll close this PR since mine merged.

foolip:
Thank you @jrobbins, that's a nice solution!

PR #5350, opened 2025-08-22
Display counts of rows in the bulk edit preview.

jrobbins:
This is intended to address the use case identified in #5348.  The bulk edit page will now show a line like
`123 rows: 100 already matching and 23 to be updated.`
at the bottom of the preview table.

The purpose is to help confirm that the CSV file was parsed correctly and that it includes the expected changes.

### DIFF
--- a/client-src/elements/chromedash-bulk-edit-page.ts
+++ b/client-src/elements/chromedash-bulk-edit-page.ts
@@ -247,9 +247,12 @@ export class ChromedashBulkEditPage extends LitElement {
     `;
   }
 
+  isDifferent(item: Item): boolean {
+    return item.existing != 'Loading...' && item.existing != item.desired;
+  }
+
   renderItemRow(item: Item) {
-    const different =
-      item.existing != 'Loading...' && item.existing != item.desired;
+    const different = this.isDifferent(item);
     const differentClass = different ? 'diff' : '';
     return html`
       <tr>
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-bulk-edit-page.ts
```diff
@@ -277,6 +280,13 @@ export class ChromedashBulkEditPage extends LitElement {
         </p>
       `;
     }
+    const totalCount = this.items.length;
+    const matchCount = this.items.filter(
+      item => !this.isDifferent(item)
+    ).length;
+    const updateCount = this.items.filter(item =>
+      this.isDifferent(item)
+    ).length;
```
jcscottiii, line 289: You could calculate the last count based on the other information.
```suggestion
    const updateCount = totalCount - matchCount;
```

jcscottiii: I also asked gemini. It said you could do something like this instead too:

```ts
const { matchCount, updateCount } = this.items.reduce(
  (counts, item) => {
    if (this.isDifferent(item)) {
      counts.updateCount++;
    } else {
      counts.matchCount++;
    }
    return counts;
  },
  { matchCount: 0, updateCount: 0 }
);
```

jcscottiii (approved):
LGTM with one thought to consider.

client-src/elements/chromedash-bulk-edit-page.ts
jrobbins: The reduce looks pretty complicated compared to individual lines to set each value.
I considered using minus, but I wanted to count each independently just in case isDifferent() ever gets replaced by something more complicated and there are more than two subtotals to count.

PR #5366, opened 2025-08-28
Add web_feature as a prerequisite to pre-flight and pre-vote dialogs

### LINKED ISSUE #5365: Add logic to flag web_feature as a prerequisite in the pre-flight and pre-vote dialogs
We have existing logic to detect when fields needed for a given action are missing.  It will need to be enhanced to handle fields that are in the feature's metadata rather than in any stage.

jrobbins:
This should resolve #5365.  Basically, this adds support for flagging a missing web_feature field in the pre-flight and pre-vote dialog.  Additional changes were needed because the web_feature field is located on the metadata form rather than being part of any stage.

Background: In our old Guide UX there was a strong concept of a progress item that got checked off as the user filled in certain fields, and there was an explicit mapping from any given action to its prerequisite progress items.  It was also supposed to be possible to check that certain enum fields had progressed to a certain value, or that feature-related CLs had landed, but that was never implemented.  Since the only kind of progress we ever actually use is looking at field values, it probably makes more sense to phase out progress items and instead mark some fields as some flavor of required.

In this PR:
* Allow a ProgressItem to have no stage.  And, when rendering such a progress item, use the word "Metadata" and link to the metadata editing form.
* Special-case the detection of the "Web feature" progress item because it does not reside on the stage where that progress item is defined, but is instead on the metadata editing form.
* Add missing field detection to the prevote dialog.  Start with this as a special case of a general detection of missing fields rather than using the progress item logic because I would like to phase out progress items.


client-src/elements/chromedash-gate-column.ts
```diff
@@ -360,12 +360,13 @@ export class ChromedashGateColumn extends LitElement {
...
         this.featureGates = gatesRes.gates;
-        const vote = this.voteSelectRef.value?.value;
+        const vote = this.voteSelectRef.value!.value;
         maybeOpenPrevoteDialog(
+          this.feature,
           this.featureGates,
           this.stage,
           this.gate,
-          vote
+          Number(vote)
```
jcscottiii: This is outside the scope of this PR. But something to consider. If the structure of `VOTE_OPTIONS` ever changed so that the "value" was no longer an int, we might not catch it here. Instead, we should use the selectedIndex or something to get the value directly from VOTE_OPTIONS.

client-src/js-src/cs-client.js
```diff
@@ -195,6 +195,7 @@
  * @property {string} summary
  * @property {string} category
  * @property {number} category_int
+ * @property {string=} web_feature
```
jcscottiii, line 198:
```suggestion
 * @property {string} web_feature
```

client-src/elements/chromedash-preflight-dialog.ts
```diff
@@ -162,7 +162,11 @@ export class ChromedashPreflightDialog extends LitElement {
     this.hide();
   }
 
-  renderEditLink(stage: ProcessStage, feStage: StageDict, pi: ProgressItem) {
+  renderEditLink(
```
jcscottiii, line 165: Thoughts on something like this?

```ts
  renderEditLink(
    stage: ProcessStage | null,
    feStage: StageDict,
    pi: ProgressItem
  ) {
    // This function only renders links for progress items that have a field.
    if (!pi.field) {
      return nothing;
    }

    const pathSegment = stage
      ? `${stage.outgoing_stage}/${feStage.id}`
      : 'metadata';

    return html`
      <a
        class="edit-progress-item"
        href="/guide/stage/${this._feature.id}/${pathSegment}#id_${pi.field}"
        @click=${this.hide}
      >
        Edit
      </a>
    `;
  }
```

Or you could also create a separate renderEditMetadataLink since you're already check if the stage exists in renderDialogContent. (But that may duplicate code)

jcscottiii (approved):
LGTM

client-src/elements/chromedash-gate-column.ts
jrobbins: I changed this code to make it more explicit and consistent with the function above it, so if we ever do make a change to VOTE_OPTIONS it will be harder to miss that this code needs to be updated.

client-src/elements/chromedash-preflight-dialog.ts
jrobbins: Good suggestion.  Done.

client-src/js-src/cs-client.js
jrobbins: Actually, the `=` in that notation means that the property is optional.

jcscottiii: Makes sense. Missed that. Thanks!

Issue #5263, opened 2025-07-22
Fall back to class usecounters when WebDX usecounter info is not available


jrobbins:
Our web feature charts only go back to around June 2024, that is because the backend that we fetch that data from only has data for them starting in June 2024.

The WebDX usecounters themselves were established around June 2024, but they are based on classic usecounters that go back much further.  There is a map from classic to WebDX in a CL, it could be used to make the same mapping in reverse in chromestatus so that whenever a WebDX usecounter is missing data, we could use the classic one instead.

We should add a note to the charts page explaining that classic data is used prior to 2024.

jstenback:
We did have discussions about doing this when we first introduced WebDXFeature use counters in Chromium, but at least back then decided against doing that. And I'd question whether doing this reverse mapping in Chromestatus is the right approach as we show the exact same metrics data on webstatus.dev as well, so we'd need to replicate the reverse mapping there too, or have a different amount of data available on the two sites. I'm not opposed to doing this work, but we should think about where it'd be best done, and whether it's worth it given the fact that this mapping isn't static and it'll add more things to maintain over time.

PR #5370, opened 2025-08-29
Add WebDX to classic usecounter fallback mechanism

MorganaFuture:
Implement automatic fallback from WebDX usecounter data to classic FeatureObserver data when WebDX data is insufficient (< 5 data points after June 2024). This provides continuous historical metrics for web features even when WebDX data is missing.

Fixes: [#5263](https://github.com/GoogleChrome/chromium-dashboard/issues/5263)


DanielRyanSmith:
I could be wrong here, but I think we're not taking fixes for this issue for now. Thank you for your initial attempt at contribution.

PR #5371, opened 2025-08-29
Filter confidential features from cached results in release notes and re-order filters to improve performance

yanndago:
- Filter confidential features from cached results in release notes
- Re-order filters for release note features to improve performance by putting the faster filters first


jcscottiii (approved):
LGTM.

There might be some use to add a test case that calls a mock cache and assert that filter_confidential is called and its results are used instead

PR #5374, opened 2025-08-29
Speed up get_features_in_release_notes and add logging.

jrobbins:
This should resolve b/441927391, or at least give more info needed to debug the performance of that page.

In this PR:
* Add logging statements so that we can see where time is spent in this function.
* Simplify some python code syntax to avoid creation of temporary objects
* Prefetch stages needed to convert a list of feature entries all at once rather than allowing sequential queries to happen

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -95,6 +95,7 @@ def get_features_in_release_notes(milestone: int):
   logging.info('Fetched %r relevent stages', len(stages))
 
   feature_ids = list({s.feature_id for s in stages})
+  features_future = get_entries_by_id_async(feature_ids)
   # Prefetch all stages for all those features, not just the stages that
   # qualified the features to be considered.
   prefetched_stages = organize_all_stages_by_feature(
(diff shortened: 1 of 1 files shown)


internals/feature_helpers.py
```diff
@@ -90,21 +91,35 @@ def get_features_in_release_notes(milestone: int):
           Stage.milestones.webview_last >= milestone,
           Stage.rollout_milestone >= milestone),
       Stage.stage_type.IN([STAGE_BLINK_SHIPPING, STAGE_PSA_SHIPPING,
-          STAGE_FAST_SHIPPING, STAGE_DEP_SHIPPING, STAGE_ENT_ROLLOUT])).filter().fetch()
+          STAGE_FAST_SHIPPING, STAGE_DEP_SHIPPING, STAGE_ENT_ROLLOUT])).fetch()
+  logging.info('Fetched %r relevent stages', len(stages))
+
+  feature_ids = list({s.feature_id for s in stages})
```
jcscottiii, line 97: Is there a chance for duplicates?

I see before you used a set:

```
  feature_ids = list(set({
      *[s.feature_id for s in stages]}))
 ```

jcscottiii (approved):
LGTM with one question.

Also, there's a similar PR that is trying to improve performance: https://github.com/GoogleChrome/chromium-dashboard/pull/5371

internals/feature_helpers.py
jrobbins: The `{...}` does create a set.  The old code was actually creating two sets and two lists.

jrobbins:
> Also, there's a similar PR that is trying to improve performance: #5371

They both look good.  I will merge both.

PR #5376, opened 2025-08-31
Improve formatting of intent template

DanielRyanSmith:
This pull request introduces formatting improvements to the intent templates to reduce unnecessary whitespace and enhance readability.

Changes include:
- Using Jinja's whitespace control (`{%- ... -%}`) to remove extraneous whitespace from the rendered HTML output.
- Some reformatting of the template structures for better organization and consistency.

### Per file:

`intent_to_implement.html`:

   * The conditional statements for the "Contact emails" and "Explainer" sections now use if/else blocks. This change makes the logic more explicit, clearly defining the output for cases where content exists versus when it is absent ("None").
   * For the "Non-finch justification" section, a nested if statement is converted to an elif. This flattens the control flow, which makes the conditions for rendering the content easier to read and maintain.
   * A non-breaking space (`&nbsp;`) now appears before the parenthesized links and notes for browser compatibility signals (Gecko, WebKit, etc.). This ensures a clean visual separation in the rendered output, even if the line wraps.
   * The logic for adding a comma between contact emails is refined by moving the closing `</a>` tag outside the conditional, which reduces code duplication.

`estimated-milestones-table.html`

   * The for loops and the if conditions within them are separated onto individual lines. This improves the readability of the template's source code by making the control structure more apparent.
   * In the template source, the `<td>` elements for each table row are grouped onto a single line. This makes the template code more compact and visually reinforces the structure of each row.


DanielRyanSmith:
It seems that the Playwright tests track screenshots that change shipping year based on the current year, so we'll have issues with the tests at yearly intervals unless some updates to the tests are made.

Fabiansatachiathayde:
The issue with test lasted a long time for at least 6 to 8 years that's a
lil to long don't you think

jrobbins:
> It seems that the Playwright tests track screenshots that change shipping year based on the current year, so we'll have issues with the tests at yearly intervals unless some updates to the tests are made.

Whoops, it's September already.  I have a fix in mind, but for the short-term I will just submit a no-op PR today that updates all the screenshots.  You can rebase onto that.

jrobbins:
This does seem to really help reduce the blank lines in the generated HTML source code.

I deployed this PR to staging and generated this email:
https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/lLg-XZPmdOE?e=48417069

It still has a lot of vertical space because of the margin on `<h4>`s.  Maybe we should use `<div><b>` instead.

DanielRyanSmith:
> Maybe we should use <div><b> instead.

Updated! Also, removing the `<p>` tags seemed to help a lot in this regard, especially since they were only sporadically used. [Here is an example](https://groups.google.com/a/google.com/g/cr-status-staging-emails/c/FfB7AUl8xT0?e=48417069) I generated on staging with the current changes.

templates/blink/intent_to_implement.html
```diff
@@ -1,36 +1,53 @@
...
+<br>
+<br>
+<div><b>Blink component</b></div>
+{% for c in feature.browsers.chrome.blink_components -%}
+  <a href="https://issues.chromium.org/issues?q=customfield1222907:%22{{c}}%22" target="_blank" rel="noopener">{{c}}</a>
+{%- endfor -%}
+
+<br>
+<br>
+<div><b>Web Feature ID</b></div>
```
jrobbins, line 50: It looks like you merged in my changes, but you did not make those lines use `{%-` and `-%}`.

PR #5377, opened 2025-09-02
Make web_features field required for some feature types

jrobbins:
With this PR, the web_feature field will be required on the feature creation page when the feature type is a new feature or catching up with an existing standard, but not otherwise.


client-src/elements/chromedash-guide-new-page.ts
```diff
@@ -86,6 +89,24 @@ export class ChromedashGuideNewPage extends LitElement {
...
+    const webFeatureField = this.shadowRoot?.querySelector(
+      'chromedash-form-field[name="web_feature"]'
+    ) as ChromedashFormField;
+    if (!webFeatureField) {
+      return;
+    }
+    let featureType = FEATURE_TYPES.FEATURE_TYPE_INCUBATE_ID[0];
+    for (const fv of this.fieldValues) {
+      if (fv.name == 'feature_type' && fv.value !== undefined) {
+        featureType = fv.value;
```
jcscottiii, line 102: Question: Are you trying to get the last instance of field value with name == 'feature_type'? If so, this is good. If not, you may want to consider using a `break` after you get the feature type to prevent extra looping.

client-src/elements/chromedash-guide-new-page.ts
```diff
@@ -86,6 +89,24 @@ export class ChromedashGuideNewPage extends LitElement {
     });
   }
 
+  maybeMakeWebFeatureRequired() {
+    const webFeatureField = this.shadowRoot?.querySelector(
+      'chromedash-form-field[name="web_feature"]'
+    ) as ChromedashFormField;
```
jcscottiii: Nit: We should use the generic parameter instead of type casting.

```suggestion
    const webFeatureField = this.shadowRoot?.querySelector<ChromedashFormField>(
      'chromedash-form-field[name="web_feature"]'
    );
```

client-src/elements/chromedash-form-field.ts
jcscottiii, line 1: I see there are five places where you modified the conditions for the `required` property to be `this.fieldProps.required || this.forceRequired`

Thoughts on creating a helper like this:

```ts
  isRequired(): boolean { return this.fieldProps.required || this.forceRequired }
```

Then you could change all of those lines to be: `?required=${this.isRequired()}`

client-src/elements/chromedash-guide-new-page.ts
jrobbins: Done.  It shouldn't matter but exiting after the first is fine.

client-src/elements/chromedash-form-field.ts
jrobbins: Done.  It was just used inside that one function, so I just extracted it out as a constant.

PR #5383, opened 2025-09-04
Add web_feature to intents.

jrobbins:
Add web_feature to the intent email generation.  If it has a value other than "Missing feature" we link to webstatus.dev.

### DIFF
--- a/api/converters.py
+++ b/api/converters.py
@@ -30,6 +30,7 @@ from internals.core_models import (
 )
 from internals.data_types import FeatureDictInnerViewInfo, StageDict, VerboseFeatureDict
 from internals.review_models import Gate, Vote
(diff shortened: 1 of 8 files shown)


jrobbins:
@DanielRyanSmith FYI, this may conflict with your #5376.

jrobbins:
> * While building the [template_data](https://github.com/GoogleChrome/chromium-dashboard/blob/aeb0cd9ceeb16412312ea42754cffed71849b7f1/api/intents_api.py#L83-L93), add a new boolean (`is_missing_feature`) that equals `feature.web_feature == MISSING_FEATURE_ID`. Then your template can check that boolean instead.

I went with a variation of this.

PR #5419, opened 2025-09-16
Allow editing of milestones on the enterprise release notes page

jrobbins:


client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -239,39 +251,45 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
...
-          stages.some(s => STAGE_TYPES_SHIPPING.has(s.stage_type))
+  convertFeatureShippingStagesToRolloutStages(f: Feature): Feature {
+    const rollouts: StageDict[] = f.stages.filter(
+      s => s.stage_type === STAGE_ENT_ROLLOUT
+    );
+    const converted: StageDict[] = f.stages
+      .filter(s => STAGE_TYPES_SHIPPING.has(s.stage_type))
+      .map(
+        s =>
+          this.convertShippingStageToRolloutStages(s) as unknown as StageDict[]
```
jcscottiii:
```suggestion
          this.convertShippingStageToRolloutStages(s) as StageDict[]
```

This `as unknown as StageDict[]` is a very forceful type assertion. It looks like it's necessary because the convertShippingStageToRolloutStages function (called inside this map) creates rollout_platforms as a number[]. The StageDict type definition, however, requires rollout_platforms to be string[]. Is this on purpose? If so, we could probably create a different type of StageDict with rollout_platforms being of type number[]. If not, this type mismatch could lead to a runtime error if any downstream code consumes this stages array and expects rollout_platforms to actually contain strings.

jcscottiii: To do this, you would need to change `convertShippingStageToRolloutStages` to have the following diff

```diff
@@ -170,7 +170,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
     ];
   }
 
-  convertShippingStageToRolloutStages(stage) {
+  convertShippingStageToRolloutStages(stage): Partial<StageDict>[] {
     const milestones = [
       stage.desktop_first,
       stage.android_first,
@@ -245,7 +245,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
       ([milestone, platforms]) => ({
         stage_type: STAGE_ENT_ROLLOUT,
         rollout_milestone: Number(milestone),
-        rollout_platforms: Array.from(platforms),
+        rollout_platforms: Array.from(platforms).map(String),
         rollout_impact: 1,
       })
     );
```

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -562,27 +588,48 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
   }
 
   save(f: Feature) {
-    let textarea: SlTextarea = this.shadowRoot?.querySelector(
+    const fieldValues: FieldInfo[] = [];
+    const addFieldValue = (name, el, originalValue, stage?) => {
```
jcscottiii:
```suggestion
        const addFieldValue = (name: string, el: ValueElement, originalValue: string| string[] | number| undefined, stage?: StageDict) => {
```


We should add some types to this.

You would add this type of interface to the top too:

```ts
// A simple interface for any Shoelace element that has a .value
interface ValueElement {
  value: string | string[] | undefined;
}
```

client-src/elements/chromedash-enterprise-release-notes-page.ts
jrobbins: Excellent suggestion.  Done.

Issue #4988, opened 2025-04-30
accurate_as_of reminder emails say "No milestone specified" for enterprise features


jrobbins:
Our template accuracy_notice_email.html includes template estimated_milestone_table.html to generate a list of milestones like
`Shipping on desktop: 137`
That works for web platform features that have a shipping stage, but it shows `No milestone specified` for enterprise features that have rollout stages rather than shipping stages.

DanielRyanSmith:
This is unfortunately  a lingering side effect of #3767

DanielRyanSmith:
This is now fixed after landing #5430

Issue #4498, opened 2024-10-30
Accurate-as-of email confusing for enterprise features


jrobbins:
An accurate-as-of email was recently sent for
https://chromestatus.com/feature/6138627554541568

It included the text
```
Your feature is slated to launch soon for m131:

No milestones specified
```

That text caused some confusion because it seems to contradict itself: if it is launching in m131, then why does it say there are no milestones specified.

The feature has a percentage rollout in 131, but no shipping stages because it is an enterprise feature.

DanielRyanSmith:
This will be fixed after an update to `estimated-milestones-table.html`

PR #5430, opened 2025-09-22
Migrate the `rollout_milestone` field

DanielRyanSmith:
Part of #3767, #4498, and #4988

A number of issues are being caused by the `rollout_milestone` field being stored in its own field, rather than in the `milestones` stage field. A lot of Chromestatus logic relies on checking for milestones in the `milestones` field, so this separate storage of a milestone has unintended consequences (see issues above).

This change adds a migration script that copies all `rollout_milestone` stage values into `milestones.desktop_first`. Additionally, new logic is added that saves any `rollout_milestone` changes to `milestones.desktop_first`. This will allow for the field to be completely migrated, so that references to `rollout_milestone` can be removed in a future PR.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -188,6 +188,8 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
       changed_fields: CHANGED_FIELDS_LIST_TYPE
     ) -> list[Stage]:
     """Update stage fields with changes provided in the PATCH request."""
+    # TODO(DanielRyanSmith): This method should be updated to use the logic
+    # for basehandlers.update_stage(). This logic is mostly duplicated otherwise.
     stages_to_store: list[Stage] = []
     for change_info in stage_changes_list:
       stage_was_updated = False
@@ -223,6 +225,12 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         form_field_name = change_info[field]['form_field_name']
         old_value = getattr(milestones, field)
         new_value = change_info[field]['value']
+        # If this is the rollout milestone, also save it to the old field.
+        # TODO(DanielRyanSmith): Remove this double-storage once the
+        # rollout_milestone field is deprecated.
+        if form_field_name == 'rollout_milestone':
+          self.update_field_value(stage, 'rollout_milestone', 'int', new_value)
+
         self.update_field_value(milestones, field, field_type, new_value)
         changed_fields.append((form_field_name, old_value, new_value))
         stage_was_updated = True
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -225,6 +225,9 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         form_field_name = change_info[field]['form_field_name']
         old_value = getattr(milestones, field)
         new_value = change_info[field]['value']
+        # desktop_first will be the new default field for "start" milestone,
+        # like rollout_milestone.
+
         # If this is the rollout milestone, also save it to the old field.
         # TODO(DanielRyanSmith): Remove this double-storage once the
         # rollout_milestone field is deprecated.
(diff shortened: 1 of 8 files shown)


internals/maintenance_scripts.py
```diff
@@ -1004,3 +1004,21 @@ def get_template_data(self, **kwargs):
...
+    changed_stages: list[Stage] = []
+    for stage in stages:
+      if stage.milestones and stage.milestones.desktop_first:
+        continue
+      if not stage.milestones:
+        stage.milestones = MilestoneSet()
+      stage.milestones.desktop_first = stage.rollout_milestone
+      changed_stages.append(stage)
+    if changed_stages:
+      ndb.put_multi(changed_stages)
```
jrobbins, line 1028: Are we going to hit a limit on the number of entities that can be set at one time?  There is other code in this file that writes in batches.

api/features_api.py
```diff
@@ -223,6 +225,12 @@ def _patch_update_stages(
         form_field_name = change_info[field]['form_field_name']
         old_value = getattr(milestones, field)
         new_value = change_info[field]['value']
+        # If this is the rollout milestone, also save it to the old field.
```
jrobbins, line 231: Is this code needed?  The rollout_milestones are already being stored because rollout_milestone is listed in STAGE_FIELD_DATA_TYPES that is processed on lines 205-216.

DanielRyanSmith:
> Why put this value into milestones.desktop_first, rather than add a new field like milestones.rollout? It could be confusing to use a field with desktop in the name because rollouts specify their own list of platforms, which might be iOS or Android.

I do think that this wording is confusing, but many of our timeline and notification checks use the existing milestone fields to display information, so `desktop_first` has been used as a default single milestone field when there does not need a platform differentiation (OT extensions use only `desktop_last` for example).

I think the better approach might be to create a new non-differentiated field (e.g. `Stage.milestones.first` `Stage.milestones.last`) that does not unnecessarily categorize the values if there is a 1:1 relationship between the stage and a start/end milestone, but that would also require a lot of refactoring in our codebase. Let me know what you think. 🙂

internals/maintenance_scripts.py
DanielRyanSmith: Good catch - I'll add the batch processing like the other scripts. 🙂

api/features_api.py
DanielRyanSmith: This is needed because the change to form-field-enums.ts will cause the request to send this change over as the new field name instead, and this logic will also save it to the old `rollout_milestone` field while that field is still used.

jrobbins: OK, got it.

However, I have recently added the ability to edit the rollout milestone in chromedash-enterprise-release-notes-page.ts line 620.  So, I think that would need to change too.  Maybe that should use STAGE_FIELD_NAME_MAPPING too.

jrobbins:
> > Why put this value into milestones.desktop_first, rather than add a new field like milestones.rollout? It could be confusing to use a field with desktop in the name because rollouts specify their own list of platforms, which might be iOS or Android.
> 
> I do think that this wording is confusing, but many of our timeline and notification checks use the existing milestone fields to display information, so `desktop_first` has been used as a default single milestone field when there does not need a platform differentiation (OT extensions use only `desktop_last` for example).
> 
> I think the better approach might be to create a new non-differentiated field (e.g. `Stage.milestones.first` `Stage.milestones.last`) that does not unnecessarily categorize the values if there is a 1:1 relationship between the stage and a start/end milestone, but that would also require a lot of refactoring in our codebase. Let me know what you think. 🙂

Let's stick with desktop_first and explain it in a comment rather than make major changes.

jrobbins (approved):

api/features_api.py
DanielRyanSmith: Actually, the changes you made should still work in that context, because you format the request using `utils.formatFeatureChanges()` [here](https://github.com/GoogleChrome/chromium-dashboard/blob/3da4bc568c9329ac82ce7eb9def5036450ff2daf/client-src/elements/chromedash-enterprise-release-notes-page.ts#L632). This function takes STAGE_FIELD_NAME_MAPPING into account when preparing the request to be sent to the API. 🙂

PR #5447, opened 2025-09-24
Make enterprise_feature_categories conditionally required.

jrobbins:
This should resolve b/446658199.

In this PR:
* Add the field `enterprise_feature_category` to the feature entry creation page, and give it new non-enterprise help text.
* Add logic to make that field required based on the value of `enterprise_impact`.

### DIFF
--- a/client-src/elements/chromedash-guide-new-page.ts
+++ b/client-src/elements/chromedash-guide-new-page.ts
@@ -7,7 +7,7 @@ import {
   NEW_FEATURE_FORM_FIELDS,
   ENTERPRISE_NEW_FEATURE_FORM_FIELDS,
 } from './form-definition';
-import {FEATURE_TYPES} from './form-field-enums';
+import {FEATURE_TYPES, ENTERPRISE_IMPACT} from './form-field-enums';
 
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-guide-new-page.ts
```diff
@@ -108,6 +108,25 @@ export class ChromedashGuideNewPage extends LitElement {
...
+    const enterpriseFeatureCategoriesField =
+      this.shadowRoot?.querySelector<ChromedashFormField>(
+        'chromedash-form-field[name="enterprise_feature_categories"]'
+      );
+    if (!enterpriseFeatureCategoriesField) {
+      return;
+    }
+    let enterpriseImpact = ENTERPRISE_IMPACT.IMPACT_NONE[0];
+    for (const fv of this.fieldValues) {
+      if (fv.name == 'enterprise_impact' && fv.value !== undefined) {
```
DanielRyanSmith: Optional nit
```suggestion
      if (fv.name === 'enterprise_impact' && fv.value !== undefined) {
```

PR #5450, opened 2025-09-24
Add test delay after feature creation, and use bookworm

jrobbins:
Fairly often our tests fail on this step.  Looking at the actual screenshots, I see that the skeletons are still showing, which should not be possible given that the test code awaits for the element the replaces the skeletons.  However, maybe there is a slight delay between updating the DOM and updating the rendered page.  So, I am adding a delay.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG PYTHON_VARIANT=3.13-bullseye
+ARG PYTHON_VARIANT=3.13-bookworm
 FROM python:${PYTHON_VARIANT} as app
 
 RUN groupadd -g 1001 appuser && \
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):

jcscottiii (approved):

jcscottiii:
@jrobbins I'm not sure if the added delay is working as intended.

I saw that the job was failing a few times.

image

It kept detecting pixel differences.

But I will leave it to you on whether you want to merge this.

I wonder if there is some other component we can wait explicitly for instead.

jrobbins:
Yeah, I am also not sure that this will solve our playwright problems.  I want to merge this now to get the bookworm change, and I will try to work more on playwright.

PR #5451, opened 2025-09-25
Query for features missing in cache all at once

DanielRyanSmith:
This change updates some of the asynchronous logic in feature_helpers.py to send a single request for all features that are not cached. Behavior should be identical.

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -396,7 +396,6 @@ def get_feature_names_by_ids(feature_ids: list[int],
   """Return a list of JSON dicts for the specified feature names.
   """
   result_dict = {}
-  futures_by_id = {}
 
   if update_cache:
     lookup_keys = [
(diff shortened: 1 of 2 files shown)


internals/feature_helpers.py
```diff
@@ -409,24 +408,31 @@ def get_feature_names_by_ids(feature_ids: list[int],
...
-    if fe and not fe.deleted:
-      feature_id = fe.key.integer_id()
-      result_dict[feature_id] = converters.feature_entry_to_json_tiny(fe)
+  if needed_ids:
+    # Create ndb.Key objects for the batch operation
+    needed_keys = [ndb.Key(FeatureEntry, fid) for fid in needed_ids]
 
-  if update_cache:
+    features_future = FeatureEntry.query(
+        FeatureEntry.key.IN(needed_keys)
```
jcscottiii, line 418: Do we anticipate there ever being more than 30 keys? There's a limit to how many keys you can use. You would need to batch them.

> Use the IN operator to combine up to 30 equality (==) clauses on the same property with a logical OR. An IN query returns entities where the given property matches any of the comparison values.

https://cloud.google.com/datastore/docs/concepts/queries#in

internals/feature_helpers.py
```diff
@@ -456,37 +461,43 @@ def get_by_ids(feature_ids: list[int],
...
-      futures_by_id[feature_id] = FeatureEntry.get_by_id_async(feature_id)
+  needed_ids = [fid for fid in feature_ids if fid not in result_dict]
+
+  if needed_ids:
+    # Create ndb.Key objects for the batch operation
+    needed_keys = [ndb.Key(FeatureEntry, fid) for fid in needed_ids]
+
+    feature_futures = ndb.get_multi_async(needed_keys)
+    stages_future = Stage.query(
+      Stage.feature_id.IN(needed_ids), Stage.archived == False).fetch_async()
```
jcscottiii, line 472: Same here.

internals/feature_helpers.py
DanielRyanSmith: You're right about this. some of this implementation will have issues on large datasets. I'm going to drop this PR for now, and potentially come back to it later.

PR #5462, opened 2025-09-25
Fix the new usage of enterprise_feature_categories.

jrobbins:
This is a follow-up to #5447.  I had focused too much on making that field conditionally required and failed to think through how it would be saved and later edited.

In this PR:
* Also make the enterprise_feature_categories field required when creating an enterprise feature entry.  This is done through TS logic rather than simply marking the field always required because it is not always required for non-enterprise features.
* Add enterprise_feature_categories to the WP feature metadata editing form.  Without this, there would be no way so edit it after feature entry creation.
* Update the help text for enterprise feature creation to indicate that the user must choose at lease one value.
* Actually process the form field when submitted to the server.  I had though that this would go through features_api.py but this is still a full form submission.

### DIFF
--- a/pages/guide.py
+++ b/pages/guide.py
@@ -79,6 +79,8 @@ class FeatureCreateHandler(basehandlers.FlaskHandler):
     # Write for new FeatureEntry entity.
     feature_entry = FeatureEntry(
         category=int(self.form.get('category')),
+        enterprise_feature_categories=self.form.getlist(
+            'enterprise_feature_categories'),
         name=self.form.get('name'),
         feature_type=feature_type,
         summary=self.form.get('summary'),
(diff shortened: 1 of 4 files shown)


pages/guide.py
```diff
@@ -79,6 +79,8 @@ def process_post_data(self, **kwargs):
     # Write for new FeatureEntry entity.
     feature_entry = FeatureEntry(
         category=int(self.form.get('category')),
+        enterprise_feature_categories=self.form.getlist(
+            'enterprise_feature_categories'),
```
DanielRyanSmith, line 83: I had almost forgotten that we still pass through guide.py when creating new features. I think the features API would work just fine, but it would take some time to make sure all the behavior matches the existing behavior in the POST request. I'll make an issue for it.

PR #5465, opened 2025-09-25
For roadmap dates, use early_stable going forward

jrobbins:
The chromiumdash team advised me to use this field of their responses instead of the field that we were previously using.  The values are the same for past and current milestones and will be the correct one to use going forward.

(I had thought that we needed to keep the reference to final_beta for backward compatibility with copies of the chromiumdash JSON that we cache in chromestatus, however each new chromestatus version uses its own redis namespace.  So, this deployed version of chromestatus will never see data in the old format.)


jrobbins:
CC @harrysouders@google.com, @benmason@google.com

PR #5469, opened 2025-09-28
Stop over-using list-style.

jrobbins:
Back before 2020 when ChromeStatus.com was implemented in as a MPA, it had the same over-arching CSS approach where a lot of the user agent's styles for things are overridden to make things start off very plain. In particular, HTML lists had their bullets hidden and indentation reduced to nothing.  Since then, every time we actually needed a bullet list, we wrote new CSS rules to make it look like bullets again.  That's a hassle, and it will make future markdown lists look wrong.

In this PR:
* Remove the rule that set `<li>` list-style to none site-wide.  And, revert the `<ol>` and `<ul>` padding to the default.
* Remove CSS rules that were needed to make lists look like lists.
* Change from ordered to unordered in one place where that was the way that it was styled anyway.

There are some visual differences in indentation and bullet style, but I think those are OK.


DanielRyanSmith (approved):
Even though not intentional, I think the visual change is actually pretty nice 🙂

PR #5502, opened 2025-10-04
Create features via the API rather than form POST.

### LINKED ISSUE #5463: Send feature creation requests to the Features API
Chromestatus still uses the code in `guide.py` to create new features. Most of the functionality (editing features, deleting features, etc.)  has been migrated to the Features API. This last piece should also be moved to the Features API, and guide.py should be fully deprecated.

jrobbins:
This should resolve #5463.

Basically, this PR finishes the job of making feature creation use an API handler rather than a form POST handler. 

In this PR:
* In the features_api do_post handler: 
    * Set some defaults, set some server-determined values, add the feature to the search index, and flush redis.  
    * These are all steps that are one-to-one with code in guide.py.
    * Loop over a much smaller set of required fields from core_models.py because most were not strictly required.
* On the client side: 
    * Use the same utility function that we use to format the JSON request body for changes, but only send the feature changes part of it.
    * Fill in some defaults based on the enterprise or non-enterprise mode.
    * When initializing `fieldValues`, take into account the initial values listed on the field specs.  Also, consider every field touched so that it is all sent to the server, even if the user does not type in that particular field.
    * Remove the logic to set attributes on the `<form>` since they are no longer needed.
 * Leave TODO comments to remove guide.py after we can assume that most users no longer have tabs open on the old page.


DanielRyanSmith (approved):
Thanks for handling this!

PR #5504, opened 2025-10-05
New shipping features endpoint

### LINKED ISSUE #5498: Create a separate endpoint that surfaces features that are very likely to ship in a given milestone
It is very common for features to be marked as shipping in a specific milestone, but then those features get delayed and are no longer part of that release. This scenario seems to be true for 20-30% of features listed as shipping each milestone in Chromestatus.

Features are much more likely to be accurately marked as shipping in a milestone if they have certain requirements met. Namely:

- I2S posted
- LGTMs obtained by API owners
- Finch name or non-finch justification provided

(note that this requirement does not apply to PSA features)

To make it easier to aggregate which features will actually ship in a given release, it would be useful to have an endpoint that obtains only the features that meet these requirements.

CC @nairnandu

DanielRyanSmith:
Fixes #5498 

This change adds a new endpoint that aggregates features that are shipping in a certain milestone and partitions them into "complete features" and "incomplete features". Incomplete features are features that are missing certain criteria needed to ship, like LGTMs, Intent to Ship threads, or Finch names (or non-finch justifications).

This should make it easier for our users to see which features are truly shipping in a milestone (specifically for blog posts) .


api/shipping_features_api.py
```diff
@@ -0,0 +1,95 @@
...
+        Stage.milestones.ios_first == milestone,
+        Stage.milestones.webview_first  == milestone,
+      )
+    ).fetch()
+
+    return shipping_stages
+
+  def do_get(self, **kwargs) -> GetShippingFeaturesResponse:
+    """Get all features that have met all conditions to ship for a given milestone"""
+    milestone_arg = kwargs.get('mstone', None)
```
jrobbins: Can you call `self.get_int_arg('mstone') to replace some of these lines?

api/shipping_features_api.py
```diff
@@ -0,0 +1,95 @@
...
+        logging.warning(f'Feature {stage.feature_id} not found.')
+        continue
+
+      if feature.feature_type == FEATURE_TYPE_CODE_CHANGE_ID:
+        # PSA features do not require intents or approvals.
+        feature_dict = converters.feature_entry_to_json_verbose(feature)
+        complete_features.append(feature_dict)
+        continue
+
+      api_owner_gate: Gate | None = Gate.query(
```
jrobbins, line 77: Shipping stages have 6 gates, how do you know the one you are getting is the API Owners gate?

api/shipping_features_api.py
DanielRyanSmith: Yes, good idea. 🙂 Added!

api/shipping_features_api.py
DanielRyanSmith: Yeah, I definitely missed querying by gate type. I've added the constraint to the query, along with a small test change to cover the case. 😅

jrobbins (approved):

PR #5518, opened 2025-10-08
Make feature POST and PATCH set markdown_fields.

jrobbins:
This adds server-side handling of feature changes in the form of `'summary_is_markdown': true`.

Also, when creating a new feature, do more work on the feature object rather than on the dictionary used to create it because that has a cleaner syntax and any errors will be identified on the line where they happen.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -126,14 +126,61 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
       return self.get_one_feature(feature_id)
     return self.do_search()
 
+  def _shared_update_special_fields(
+      self,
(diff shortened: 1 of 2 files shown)


api/features_api.py
```diff
@@ -299,6 +326,7 @@ def _patch_update_special_fields(
       attachments.delete_orphan_attachments(
           feature_id, feature_changes['screenshot_links'])
 
+    has_updated |= self._shared_update_special_fields(feature, feature_changes)
```
DanielRyanSmith, line 329: I like the clean use of the in-place "or" operator 🙂

PR #5523, opened 2025-10-08
Show PSA features on the roadmap with a checkmark and hide API owners ship gate

### LINKED ISSUE #3403: The PSA features seem to imply that they still need an Intent to Ship
I've filed a chromestatus entry that is a "change to existing code". My understanding is that this requires a PSA, which is what the tool is letting me draft. However, the bullet points also list an "Intent to Ship email", which I think is incorrect. See screenshot below:

Screenshot 2023-10-12 at 12 20 45 PM

DanielRyanSmith:
Fixes #3403 

PSA features do not require approvals to ship, so most PSA features on the roadmap state that their review process has not started. This change adds a conditional to display PSA features shipping features with the approved checkmark when no reviews exist stating the feature needs work or is denied. Also, the "API owners" gate will no longer display in the UI for PSA features, which will avoid user confusion on whether or not approvals are required.

This conditional workaround does not completely fix the issue from a technical perspective, and all API owner Gate entities associated with PSA ship stages should be deleted. However, that solution has potential regressions that would take additional work to avoid (for example, it is assumed in the codebase that all shipping stages have an API owners gate).


DanielRyanSmith:
> Would it make sense to check the stage intent_thread_url field and somehow communicate to the user that it should be filled in?

I do think this is a good idea. It might take a bit of an overhaul in implementation, since we do not fetch feature stages for the roadmap, and the intent thread exists in the stage. It could cause some additional latency on the roadmap, but perhaps the difference will be negligible. 🙂 

I'll add an issue to capture this!

PR #5524, opened 2025-10-08
Add use markdown checkbox to chromedash textarea

jrobbins:
This is progress toward markdown support for the summary field of enterprise features.

In this PR:
* Fix some CSS and further revert heavy-handed CSS rules that spoil the look of rendered markdown.
* In chromedash-form-field, look for `offer_markdown` and `enterprise_offer_markdown` in the field spec and pass them into the chromedash-textarea component.
* In chromedash-textarea, when markdown should be offered, show a checkbox, and check it if isMarkdown is specified.
* In the places where summary can be edited, handle changes to the markdown checkbox and store them in fieldValues.
* When formatting fieldValues to send to the server, look for isMarkdown booleans and add that to the JSON data.

### DIFF
--- a/client-src/elements/utils_test.ts
+++ b/client-src/elements/utils_test.ts
@@ -304,6 +304,11 @@ go/this-is-a-test
           touched: true,
           isMarkdown: false,
         },
+        {
+          name: 'motivation',
+          value: 'Makes no change to markdown',
+          touched: true,
+        },
       ];
       const expected = {
         feature_changes: {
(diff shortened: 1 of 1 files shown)


client-src/elements/utils_test.ts
```diff
@@ -290,6 +290,34 @@ go/this-is-a-test
...
+          value: 'Now with **Markdown**',
+          touched: true,
+          isMarkdown: true,
+        },
+        {
+          name: 'description',
+          value: 'No longer desires markdown',
+          touched: true,
+          isMarkdown: false,
+        },
```
DanielRyanSmith, line 306: Maybe add a test here to make sure `{field}_is_markdown` is not present in the result if `isMarkdown` is undefined in fieldValues

jrobbins: Done.  It was already covered by the fact that such fields were not in the expected results for existing test cases, but I agree that it is helpful to add it here explicitly.

PR #5528, opened 2025-10-09
Implement markdown preview

jrobbins:
This is the final PR for markdown support for now.  It implements a preview mode so that people can see how their markdown will render before submitting.  To avoid taking up additional vertical space, the new "Preview" checkbox is positioned on the same row as the "Use markdown" checkbox.

In this PR:
* On the enterprise release notes page: 
    * Keep track of which items are in preview mode
    * Display a "Preview" checkmark when editing that controls which items are being previewed
* For chromedash-textarea fields:
    * Same behavior, but different code since only one thing is being edited.  Also, the "Preview" checkbox display is conditional, which would have been complicated on the release notes page.

### DIFF
--- a/client-src/elements/chromedash-enterprise-release-notes-page.ts
+++ b/client-src/elements/chromedash-enterprise-release-notes-page.ts
@@ -762,15 +762,14 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
   renderEditableFeatureSummary(f: Feature): TemplateResult {
     const isMarkdown = (f.markdown_fields || []).includes('summary');
     const isPreviewing = this.previewingFeatureIds.has(f.id);
-    const editor = html`
-    <sl-textarea
-    class="feature-summary"
-        id="edit-summary-${f.id}"
-        value=${f.summary}
-        size="small"
-        resize="auto"
-      >
-        </sl-textarea>`;
+    const editor = html` <sl-textarea
+      class="feature-summary"
+      id="edit-summary-${f.id}"
+      value=${f.summary}
+      size="small"
+      resize="auto"
+    >
+    </sl-textarea>`;
     const preview = html`
       <div
         id="preview"


client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -747,26 +749,60 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
...
+      id="edit-summary-${f.id}"
+      value=${f.summary}
+      size="small"
+      resize="auto"
+    >
+    </sl-textarea>`;
+    const preview = html`
+      <div
+        id="preview"
+        style="border:var(--card-border); padding:0 var(--content-padding); min-height:14em; background:var(--table-alternate-background)"
```
DanielRyanSmith, line 776: Would it be better to move the CSS to a CSS file or the component level?

client-src/elements/chromedash-textarea.ts
```diff
@@ -88,20 +91,43 @@ export class ChromedashTextarea extends SlTextarea {
...
+
+    // Since this class subclasses SlTextarea, we need to render (but
+    // not display) the content from that class, otherwise there will
+    // be errors in validation code.
+    const preview = html`
+      <div style="display:none">${editor}</div>
+      <div
+        id="preview"
+        style="border:var(--card-border); padding:0 var(--content-padding); min-height:14em; background:var(--table-alternate-background)"
+      >
```
DanielRyanSmith, line 119: Same question here about the CSS

client-src/elements/chromedash-textarea.ts
```diff
@@ -88,20 +91,43 @@ export class ChromedashTextarea extends SlTextarea {
...
+        ${autolink(this.value, [], true)}
+      </div>
+    `;
+
     return html`
-      ${editor}
+      ${this.showPreview ? preview : editor}
```
DanielRyanSmith, line 125: What do you think about showing the markdown preview at all times, similar to how we do with Buganizer? I know that GitHub doesn't take this approach, so this may just be preference.

client-src/elements/chromedash-textarea.ts
jrobbins: I don't think that it is possible in this case because we are subclassing.  I could try it in a future clean-up PR that might avoid subclassing.

client-src/elements/chromedash-textarea.ts
jrobbins: Buganizer does that for one editing field, so even though it take up extra vertical space, it is only in one spot.  We have long forms that could have many markdown fields.  Displaying them all could almost double the length of the page, which would make the user scroll a lot more to find fields that they were looking for.

jrobbins:
Here's a screenshot of the summary field in an enterprise entry:
image


And here it is in preview mode:
image

PR #5529, opened 2025-10-09
Clearer default value for fields with no value provided in intent templates

### LINKED ISSUE #4898: Consider whether "None" is the right default value for certain fields when generating Intent mails
When generating Intent mails like Intent to Prototype, it is the case for some of the ChromeStatus fields that if they're left blank, the corresponding field in the generated intent mail will be populated with the value "None".

For a lot of the fields, that is a reasonable default. But for some of the "Risk" fields like `Interoperability and Compatibility Risks`, it can be ambiguous. If the submitter leaves the field blank the Intent will generate with:

> Interoperability and Compatibility
None

Which is ambiguous. Does the submitter mean to say that they think the feature has no Interop or Compat risk, or did they just leave the Chromestatus field blank?

Other defaults to consider would be to just put no default text there, or to put something like "No response provided" or "To be determined".

**To Reproduce**
1. Create a new ChromeStatus entry
2. Leave the `Interoperability and Compatibility Risks` field blank.
3. Generate an Intent to Prototype mail.
4. Observe that the mail generates with:
> Interoperability and Compatibility
None

DanielRyanSmith:
Fixes #4898 

Adds new macros to the Jinja template so that null or empty strings are not directly printed into the intent template.

### DIFF
--- a/templates/blink/intent_to_implement.html
+++ b/templates/blink/intent_to_implement.html
@@ -1,18 +1,18 @@
 {%- macro show_value(value) -%}
-  {{ value | default('<i>No value provided</i>', true) | safe }}
+  {{ value | default('<i>No information provided</i>', true) | safe }}
 {%- endmacro -%}
 
 {%- macro urlize_with_default(value) -%}
   {%- if value -%}
     {{ value | urlize | safe }}
   {%- else -%}
-    <i>No value provided</i>
+    <i>No information provided</i>
   {%- endif -%}
 {%- endmacro -%}
 
 <div><b>Contact emails</b></div>
 {%- if not feature.browsers.chrome.owners -%}
-  <i>No value provided</i>
+  <i>No information provided</i>
 {%- else -%}
   {%- for owner in feature.browsers.chrome.owners -%}
     <a href="mailto:{{owner}}">{{owner}}</a>{%- if not loop.last -%}, {% endif %}
@@ -24,7 +24,7 @@
 <br>
 <div><b>Explainer</b></div>
   {%- if not feature.explainer_links -%}
-    <i>No value provided</i>
+    <i>No information provided</i>
   {%- else -%}
     {%- for link in feature.explainer_links -%}
       {% if loop.index0 %}<br>{% endif %}<a href="{{link}}">{{link}}</a>
@@ -42,7 +42,7 @@
 <br>
 <div><b>Design docs</b></div>
   {%- if not feature.explainer_links -%}
-    <i>No value provided</i>
+    <i>No information provided</i>
   {%- endif -%}
   {%- for link in feature.resources.docs -%}
     <br><a href="{{link}}">{{link}}</a>
@@ -68,7 +68,7 @@
 {%- elif feature.web_feature -%}
   {{ show_value(feature.web_feature) }}
 {%- else -%}
-  <i>No value provided</i>
+  <i>No information provided</i>
 {%- endif %}
 
 {% if 'motivation' in sections_to_show -%}
@@ -297,7 +297,7 @@
 <br>
 <br>
 <div><b>Non-finch justification</b></div>
-  <i>No value provided</i>
+  <i>No information provided</i>
 {%- endif -%}
 {%- endif -%}
 
--- a/internals/testdata/notifier_test/test_make_intent_email.html
+++ b/internals/testdata/notifier_test/test_make_intent_email.html
@@ -2,7 +2,7 @@
 <br>
 
 <div><b>Specification</b></div>
-<i>No value provided</i>
+<i>No information provided</i>
 
 <br>
 <br>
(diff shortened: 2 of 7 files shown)


templates/blink/intent_to_implement.html
```diff
@@ -1,9 +1,21 @@
...
+{%- endmacro -%}
+
 <div><b>Contact emails</b></div>
 {%- if not feature.browsers.chrome.owners -%}
-  None
+  <i>No value provided</i>
 {%- else -%}
   {%- for owner in feature.browsers.chrome.owners -%}
-    <a href="mailto:{{owner}}">{{owner}}</a>{%- if not loop.last -%}, {% endif %}
+    <a href="mailto:{{ owner | default('', true) }}">{{ show_value(owner) }}</a>{%- if not loop.last -%}, {% endif %}
```
jrobbins: I think that `owner` would always have a value because this is inside a for-loop of values.

templates/blink/intent_to_implement.html
```diff
@@ -12,81 +24,81 @@
 <br>
 <div><b>Explainer</b></div>
   {%- if not feature.explainer_links -%}
-    None
+    <i>No value provided</i>
   {%- else -%}
     {%- for link in feature.explainer_links -%}
-      {% if loop.index0 %}<br>{% endif %}<a href="{{link}}">{{link}}</a>
+      {% if loop.index0 %}<br>{% endif %}<a href="{{ link | default('#', true) }}">{{ show_value(link) }}</a>
```
jrobbins: Same

templates/blink/intent_to_implement.html
```diff
@@ -12,81 +24,81 @@
...
+{{ urlize_with_default(feature.standards.spec) }}
 
 {% if feature.resources and feature.resources.docs -%}
 <br>
 <br>
-  <div><b>Design docs</b></div>
+<div><b>Design docs</b></div>
   {%- for link in feature.resources.docs -%}
-<br><a href="{{link}}">{{link}}</a>
+<br><a href="{{ link | default('#', true) }}">{{ show_value(link) }}</a>
```
jrobbins: Same.  To get a `No value provided` result, we should check if the entire list is empty.

templates/blink/intent_to_implement.html
```diff
@@ -12,81 +24,81 @@
...
+  {{ urlize_with_default(feature.initial_public_proposal_url) }}
 {%- endif -%}
 
 {% if feature.tags -%}
 <br>
 <br>
 <div><b>Search tags</b></div>
   {%- for tag in feature.tags -%}
-    <a href="/features#tags:{{tag}}">{{tag}}</a>{%- if not loop.last -%}, {% endif %}
+    <a href="/features#tags:{{ tag | default('', true) }}">{{ show_value(tag) }}</a>{%- if not loop.last -%}, {% endif %}
```
jrobbins: Same.

templates/blink/intent_to_implement.html
```diff
@@ -95,31 +107,31 @@
...
 <div><b>Origin Trial Name</b></div>
-    {{stage.ot_display_name}}
+    {{ show_value(stage.ot_display_name) }}
   {%- endif -%}
   {%- if stage.ot_chromium_trial_name -%}
 <br>
 <br>
 <div><b>Chromium Trial Name</b></div>
-    {{stage.ot_chromium_trial_name}}
+    {{ show_value(stage.ot_chromium_trial_name) }}
```
jrobbins: This is already inside an if-statement that checks that value.

templates/blink/intent_to_implement.html
DanielRyanSmith: True - I've removed the redundant check

templates/blink/intent_to_implement.html
DanielRyanSmith: removed redundancy

templates/blink/intent_to_implement.html
DanielRyanSmith: Added a conditional to display the _No value provided_ text only if the entire list is empty. 🙂

templates/blink/intent_to_implement.html
DanielRyanSmith: Fixed

templates/blink/intent_to_implement.html
DanielRyanSmith: Removed the redundancy on all OT fields here. 🙂

templates/blink/intent_to_implement.html
```diff
@@ -1,6 +1,18 @@
+{%- macro show_value(value) -%}
+  {{ value | default('<i>No value provided</i>', true) | safe }}
```
jrobbins: I think this should be "No information provided".  We use that phrase in the UI because humans deal with "information", whereas a "value" is more of an implementation concept.

templates/blink/intent_to_implement.html
```diff
@@ -1,6 +1,18 @@
...
+{%- endmacro -%}
+
+{%- macro urlize_with_default(value) -%}
+  {%- if value -%}
+    {{ value | urlize | safe }}
+  {%- else -%}
+    <i>No value provided</i>
```
jrobbins: And here.

templates/blink/intent_to_implement.html
DanielRyanSmith: I do like this working better - updated!

templates/blink/intent_to_implement.html
DanielRyanSmith: Updated

Issue #5493, opened 2025-10-03
Make a list of stale features available to admins


DanielRyanSmith:
In order to maintain data accuracy in Chromestatus, it is important that feature entries do not become stale, especially as an important milestone is approaching, such as a shipping milestone. Although we send notifications to users as a reminder to confirm the accuracy of their feature's data, this is sometimes not enough to maintain the freshness of each feature's data.

It would be useful to have a view or dashboard that displayed which features are currently stale and nearing an important milestone, so that further action can be taken if needed.

jstenback:
Agreed. And I wonder if it'd be useful a bit more broadly. Maybe to all site editors as well, or maybe even all google + Chromium + explicit users?

DanielRyanSmith:
This is complete with the new `/features/stale` view. Separately, we'll need to determine if we want to increase visibility on this through making it more available. In the meantime, it should be a helpful tool for tracking down feature owners that need to verify their data.

PR #5532, opened 2025-10-10
New Stale Features endpoint

DanielRyanSmith:
Part of #5493 

This change adds a new endpoint that surfaces stale features in a similar way that stale features are notified via weekly emails. This will allow for site admins easier access to this data. I did not add site admin requirements to this endpoint, because the data is essentially public.

Additionally, the `get_current_milestone_info` function in reminders.py has been moved to the utils.py file for reusability.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,25 +1,25 @@
-# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/python-3/.devcontainer/base.Dockerfile
+# Copyright 2025 Google LLC
 
-# [Choice] Python version (use -bullseye variants on local arm64/Apple Silicon): 3, 3.10, 3.9, 3.8, 3.7, 3.6, 3-bullseye, 3.10-bullseye, 3.9-bullseye, 3.8-bullseye, 3.7-bullseye, 3.6-bullseye, 3-buster, 3.10-buster, 3.9-buster, 3.8-buster, 3.7-buster, 3.6-buster
-# Update 'VARIANT' to pick a version of Python: 3, 3.10, 3.9, 3.8, 3.7, 3.6
-# Append -bullseye or -buster to pin to an OS version.
-# Use -bullseye variants on local arm64/Apple Silicon.
-ARG VARIANT="3.13-bookworm"
-FROM mcr.microsoft.com/devcontainers/python:${VARIANT} as ide
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
 
-# [Choice] Node.js version: none, lts/*, 16, 14, 12, 10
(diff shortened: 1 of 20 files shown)


internals/feature_helpers.py
```diff
@@ -544,3 +547,62 @@ def get_features_by_impl_status(limit: int | None=None, update_cache: bool=False
     rediscache.set(cache_key, feature_list)
 
   return filter_confidential(feature_list)
+
+
+def get_stale_features() -> list[tuple[FeatureEntry, int, str]]:
+  features: list[FeatureEntry] = FeatureEntry.query(
```
jrobbins: I think this might be slow on prod because it is fetching almost all feature entries and then looping to get the stages of each.  Could more of the python conditions be implemented as NDB conditions.  E.g., `accurate_as_of` should not be None for any recently created feature entires, so you could use an NDB inequality on that.  E.g., the range from min_mstone to max_mstone is only a few values, so you could try fetching the stages first via an ndb.OR() and several ndb.IN() operators for the milestone fields, then fetch the feature entries that own those stages afterwards.  The roadmap query and release notes query might be starting points.  Not many features are marked deleted, so that might be better to filter in python if you would otherwise not need to add an index.

DanielRyanSmith: Thanks for the suggestion here - I think I've implemented it as you described, to be a bit more efficient. 🙂

internals/feature_helpers.py
```diff
@@ -544,3 +547,131 @@ def get_features_by_impl_status(limit: int | None=None, update_cache: bool=False
...
+    mstones_by_feature_id,
+    relevant_ent_stages,
+    ent_milestone_fields,
+    min_mstone,
+  )
+
+  relevant_features: list[FeatureEntry] = ndb.get_multi(feature_keys)
+  now = datetime.now()
+
+  return [
```
jrobbins, line 693: Given that this API is not limited to admins or site admins, then you need to filter out unlisted and confidential feature entries.   IIRC, there are helper functions for that.

DanielRyanSmith: Good call - the helper functions were only for formatted features, but I went ahead and created similar helper methods for FeatureEntry entities as well. 🙂

internals/feature_helpers.py
```diff
@@ -31,7 +31,22 @@
...
+  """Filters a FeatureEntry list to display only features the user should see."""
+  user = users.get_current_user()
+  email = None
+  if user:
+    email = user.email()
+  return [
+    f for f in feature_list
+    if (not f.unlisted
+        or (email in f.owner_emails or email in f.editor_emails
+        or (email is not None and f.creator_email == email))
```
jrobbins, line 44: The indentation of this line looks off.  I think you meant to have one of the closing parens at the end of the line above.

internals/feature_helpers.py
jrobbins: Thanks for fixing that.

PR #5533, opened 2025-10-10
New Stale Features UI view

### LINKED ISSUE #5493: Make a list of stale features available to admins

DanielRyanSmith:
Fixes #5493 

This change adds a new view in Chromestatus to view current stale features. The view contains a sortable table for easier reference.

This view is not surfaced in the UI for navigation to normal users. Instead, there is an admin link for signed in admin users. However, the URL (`/features/stale`) is not gated by admin access.

### Example Screenshot

Screenshot From 2025-10-10 08-12-37

### DIFF
--- a/client-src/elements/chromedash-drawer.ts
+++ b/client-src/elements/chromedash-drawer.ts
@@ -367,7 +367,7 @@ export class ChromedashDrawer extends LitElement {
       ${this.renderNavItem('/reports/review-latency', 'Review latency')}
       ${this.renderNavItem('/admin/blink', 'Subscriptions')}
       ${this.renderNavItem('/admin/find_stop_words', 'Find stop words JSON')}
-      ${this.renderNavItem('/features/stale', 'Stale Features')}
+      ${this.renderNavItem('/features/stale', 'Stale features')}
     `;
   }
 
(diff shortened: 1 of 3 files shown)


client-src/elements/chromedash-drawer.ts
```diff
@@ -367,6 +367,7 @@ export class ChromedashDrawer extends LitElement {
       ${this.renderNavItem('/reports/review-latency', 'Review latency')}
       ${this.renderNavItem('/admin/blink', 'Subscriptions')}
       ${this.renderNavItem('/admin/find_stop_words', 'Find stop words JSON')}
+      ${this.renderNavItem('/features/stale', 'Stale Features')}
```
jrobbins: Use sentence capitalization: Stale features.

DanielRyanSmith: Fixed!

PR #5534, opened 2025-10-10
Refactor: Improve Playwright and Dev Container DX

jcscottiii:
This commit introduces several changes to improve the developer experience, particularly for those using VS Code and dev containers.

The Gemini CLI was also configured in the dev container. Something carried over from webstatus.dev

The `playwright.config.ts` file has been moved to the repository root. This allows VS Code's Playwright extension to automatically discover and run tests through its UI, which works for developers both locally and inside a dev container.

To support this, the Playwright Dockerfile now copies the root config and patches the `testDir` path to match the container's file structure. The `.devcontainer/Dockerfile` was also updated to leverage built-in dev container features for installing tools like python and java, rather than using manual installation steps. A minor code quality improvement was made in `test_utils.js` to use JSDoc type annotations as suggested by the IDE.

This change enables the following next steps for our testing workflow:

- **One-off Screenshot Regeneration:** Developers can now regenerate screenshots for a single test directly from the VS Code UI, avoiding the need to run the full `pwtests-update` script.

- **CI Parity for Dev Container Users:** We use containers in GitHub CI to guarantee a consistent browser environment, as recommended by Playwright. By aligning the dev container to use the same base image as CI in the future, developers can run tests within VS Code and get results that perfectly mirror the CI environment. This simplifies debugging CI-specific failures in a local, IDE-integrated workflow.

**NOTE** These changes enhance the developer experience WITHOUT making dev containers a requirement; existing workflows for all developers remain unaffected.

### DIFF
--- a/packages/playwright/tests/test_utils.js
+++ b/packages/playwright/tests/test_utils.js
@@ -1,6 +1,5 @@
 // @ts-check
 import { expect } from '@playwright/test';
-import { Page } from "playwright-core";
 
 
 /**
@@ -14,12 +13,12 @@ export async function delay(ms) {
 /**
  * Call this, say in your test.beforeEach() method, to capture all
  * console messages and copy them to the playwright console.
- * @param {Page} page
+ * @param {import("playwright-core").Page} page
  */
 export function captureConsoleMessages(page) {
   page.on('console', async msg => {
     // ignore warnings for now.  There are tons of them.
-    if (msg.type() === 'warn') {
+    if (msg.type() === 'warning') {
       return;
     }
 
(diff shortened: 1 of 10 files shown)


packages/playwright/tests/test_utils.js
jcscottiii, line 1: The error I was seeing before I made any changes image

Initially, I thought about separating this into a separate PR. But if we want to have the ability to run the Playwright tests individually in the IDE, it prints out an error and does not run the  tests

jcscottiii:
Talking about the future experience, once we pick a common image between the devcontainer and playwright containers, you can do things like this:

Here's an example of debugging an individual test live

image

jcscottiii:
This should help with your work as you want to expand the playwright tests in the future @jrobbins @DanielRyanSmith

PR #5538, opened 2025-10-13
Shipping Features endpoint fix

DanielRyanSmith:
Part of #5498

This change adds two changes to the shipping features endpoint: 

- Changes the URL to send requests to the endpoint so the milestone argument is not redundantly provided twice.
- Reduces the feature information provided in the response to only include the URL to the feature. After trying this endpoint on prod, sending all of the fields of the feature for each feature is just too much unneeded information and makes the response hard to read.

### DIFF
--- a/api/shipping_features_api.py
+++ b/api/shipping_features_api.py
@@ -72,7 +72,6 @@ class ShippingFeaturesAPI(basehandlers.EntitiesAPIHandler):
 
       if feature.feature_type == FEATURE_TYPE_CODE_CHANGE_ID:
         # PSA features do not require intents or approvals.
-        feature_dict = converters.feature_entry_to_json_verbose(feature)
         complete_features.append(chromestatus_url)
         continue
 


api/shipping_features_api.py
```diff
@@ -59,19 +58,22 @@ def do_get(self, **kwargs) -> GetShippingFeaturesResponse:
...
       if feature is None:
         logging.warning(f'Feature {stage.feature_id} not found.')
         continue
 
+      chromestatus_url = (f'{self.request.scheme}://{self.request.host}'
+                          f'/feature/{feature.key.integer_id()}')
+
       if feature.feature_type == FEATURE_TYPE_CODE_CHANGE_ID:
         # PSA features do not require intents or approvals.
         feature_dict = converters.feature_entry_to_json_verbose(feature)
```
jrobbins: Do you still need feature_dict?  Avoiding this call would avoid more NDB operations.

DanielRyanSmith: Ah, good catch - I did not remove this line

PR #5557, opened 2025-10-14
Default value from channels endpoint for unit tests

### LINKED ISSUE #5556: The channels endpoint uses live requests to external APIs, even for unit tests
I did some investigating of why I would occasionally see long feature load times in local mode. The occurrence was very sporadic, but I was able to check the network requests while it happened once:

Image

Requests to our channels endpoint can sometimes take 15+ seconds. This is likely due to it being reliant on multiple calls to other APIs to collect the data. We cache this data for 24 hours, but locally this does not help. This might be what is causing flakiness in our Playwright tests.

DanielRyanSmith:
Fixes #5556, at least when running unit tests.

Hopefully this will help with playwright test flakiness. :slightly_smiling_face:

### DIFF
--- a/packages/playwright/Dockerfile
+++ b/packages/playwright/Dockerfile
@@ -21,7 +21,7 @@ USER pwuser
 WORKDIR /work
 
 # Set environment variable so we can tell we're performing Playwright tests.
-ENV PLAYWRIGHT_MODE=1
+ENV PLAYWRIGHT_MODE=true
 
 # Install a consistent version of Playwright. The duplicate ARG line is needed to copy the value
 # into this build stage.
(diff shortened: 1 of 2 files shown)


jrobbins:
I think this will help with the python unit tests, so let's try it.

However, I don't think that it will affect the playwright tests.  The `npm test` command works by basically executing the *_test.py files which then import testing_config.py which sets os.environ["SERVER_SOFTWARE"] to have "test" in the value before importing settings.py.  The playwright tests run `npm run start` which does not do that.

jrobbins (approved):

DanielRyanSmith:
> However, I don't think that it will affect the playwright tests. The npm test command works by basically executing the *_test.py files which then import testing_config.py which sets os.environ["SERVER_SOFTWARE"] to have "test" in the value before importing settings.py. The playwright tests run npm run start which does not do that.

Hmm, I see. Well, we need to make this change for playwright tests. Maybe we can add an environment variable so that we can discern when we're in pwtest mode

DanielRyanSmith:
This seems to work - I'm going to merge it and see if it affects the PW tests at all

PR #5539, opened 2025-10-13
New identifiers for denoting intent-specific fields in intent template

DanielRyanSmith:
Fixes #5540 

(This change got a lot larger than I intended. 😅  I'll add descriptions of changes per-file)

This PR updates the `sections_to_show` argument that is used to discern which fields to generate in the intent template. The `INTENT_EMAIL_SECTIONS` dictionary that contained intent-specific field information seemed like an artifact of an older implementation, specifically adding different fields based on `intent_stage` (and also including values that no longer changed anything within the intent, like `i2p_thread`).

Now, the intent template takes an enum value `IntentDraftType`, and displays certain fields in the intent based on this.

Additionally, the old MPA URL for the intent templates, `admin/features/launch/...` has since been removed, and so this PR removes those addresses. HOWEVER, the flask handler used to generate the old page (intentpreview.py) has been a very useful tool in thoroughly testing the intent template generation. I've updated the implementation to simply generate intent templates, and have updated `intentpreview_test.py` to thoroughly test the templates.

### Description of changes by file:

- api/intents_api.py: Moved the `compute_subject_prefix()` functionality to this file, rather than living in intentpreview.py, since that is now only used for testing. Also, adjusted the implentation to use the new `IntentDraftType` enum, rather than determining intent stages from the gate type.
- internals/core_enums.py: Add the new `IntentDraftType` enums, as well as a mapping of stage types to intent draft types.
- internals/notifier.py: Small change to ensure email is generated using new enums.
- internals/processes.py: Removal of the old mapping of fields that should be shown by intent stage.
- main.py: Removal of old MPA addresses.
- pages/intentpreview.py: Adjust old logic to instead generate intent templates purely for testing purposes.
- pages/intentpreview_test.py: Went a little overboard here. I generated a fully filled out feature so that we can see what every field would look like in a fully filled out intent template.
- pages/testdata/intentpreview_test/test_html_rendering.html: Completely removed, along with the test, since we're no longer generating the MPA intent preview page.
- templates/blink/intent_to_implement.html: Intent template is updated so that all fields using `sections_to_show` are now instead using `intent_type`. The fields that are displayed for each intent draft now should be the same as before.

PR #5560, opened 2025-10-15
Remove intent fields from I2P that are not asked for by Chromestatus

### LINKED ISSUE #1691: Generated Intent to Prototype emails are missing fields because I was never asked for them
**Describe the bug**

When creating a new feature, a Chromium developer's goal is to generate an Intent to Prototype email that meets the standards of the API owners and blink-dev.

However, when using the wizard, the generated email is missing a bunch of fields. This generates emails that look empty, as if the developer was intentionally leaving important information out. This is not good and causes extra traffic between the implementer and their spec mentor (if we're lucky) or between the implementer and the API owners/rest of blink-dev (if it gets sent without a spec-mentor check).

**To Reproduce**
1. Go to https://chromestatus.com/features
2. Click on "Add new feature"
3. Fill name, summary. Choose "new feature incubation".
4. Press "Submit".
5. Press "Start" in the "Start incubating" phase.
6. Fill out all the fields, and check "Set to: Start incubating".
7. Press "Submit".
8. Press "Draft Intent to Prototype email".

The resulting email contains the following fields which are empty:

- Specification
- TAG review
- Interoperability and Compatibility Risks text
- Debuggability (@mathiasbynens gets sad when this is empty)
- Flag name

and the following fields which have bad or useless default values:

- TAG review status
- Signals, notably including "Web developers: no signals". (Why am I prototyping an API if no developers have given me signals that they want to use it?)
- Is this feature fully tested by web-platform-tests?
- Estimated milestones

**Expected behavior**

The resulting Intent to Prototype should either not contain empty/left-as-default fields, because they're not appropriate for Intent to Prototype. Or, I should have been prompted to fill them out before reaching this stage.

DanielRyanSmith:
Fixes #1691

Built on top of #5539

Some fields were populating all intent templates, even when they are not part of the prototype templates. This led to the intent not looking complete, even when the user could not fill out those fields without moving to a further stage.

Additionally, some fields that are asked for in the "dev trials" stage are added to the "Ready for Developer Testing" email template. Previously, some of these were only visible in the origin trial or ship intents.

### DIFF
--- a/templates/blink/intent_to_implement.html
+++ b/templates/blink/intent_to_implement.html
@@ -251,7 +251,7 @@
 <div><b>Debuggability</b></div>
 {{ urlize_with_default(feature.debuggability) }}
 {% endif -%}
-{% if intent_type in ['developer_testing', 'experiment', 'extend_experiment', 'ship'] -%}
+{% if intent_type in ['developer_testing', 'experiment', 'extend_experiment', 'psa', 'ship'] -%}
 <br>
 <br>
 <div><b>Will this feature be supported on all six Blink platforms
--- a/pages/testdata/intentpreview_test/test_html_psa_ship_rendering.html
+++ b/pages/testdata/intentpreview_test/test_html_psa_ship_rendering.html
@@ -49,6 +49,14 @@ sum
 <div><b>Debuggability</b></div>
 <i>No information provided</i>
 <br>
+<br>
+<div><b>Will this feature be supported on all six Blink platforms
+      (Windows, Mac, Linux, ChromeOS, Android, and Android WebView)?</b></div>No<br>
+<br>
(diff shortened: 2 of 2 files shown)


templates/blink/intent_to_implement.html
```diff
@@ -243,22 +245,22 @@
...
+{% endif -%}
+{% if intent_type in ['developer_testing', 'experiment', 'extend_experiment', 'psa', 'ship'] -%}
 <br>
 <br>
 <div><b>Will this feature be supported on all six Blink platforms
       (Windows, Mac, Linux, ChromeOS, Android, and Android WebView)?</b></div>
   {%- if feature.all_platforms -%}Yes{%- else -%}No{%- endif -%}
-  {%- if feature.all_platforms_descr -%}
+  {%- if feature.all_platforms_descr %}
+    <br>
```
DanielRyanSmith, line 261: This is the fix for the spacing you mentioned in another PR @jrobbins

PR #5564, opened 2025-10-15
Check Chromium code to ensure features are shipping

### LINKED ISSUE #5517: Adding a check against Chromium code to ensure features are shipping could solve some data accuracy issues
Features are enabled in either [runtime_enabled_features.json5](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5) or [content_features.cc](https://source.chromium.org/chromium/chromium/src/+/main:content/public/common/content_features.cc). It may be possible to reference these files in Chromium to check that the current milestone data is accurate and features are indeed shipping in a specific milestone. This may allow us to only consider a feature as shipped if it exists in these files, or to ask feature owners to confirm their shipping milestone if the feature's name is not detected in these files.

I would imagine corner cases exist here, and more exploration on that would be useful. For example, it could be possible that a feature is enabled, but finch is overriding it.

DanielRyanSmith:
Fixes #5517 

This updates the Shipping features endpoint with additional checks against current Chromium code to check if the feature is actually shipping as expected. If anything missing or incorrect is found with the Chromium code, then those missing criteria are surfaced in the "incomplete_features" property of the response.

Some constants for Chromium file URLs have been moved to core_enums.py, and the helper `get_chromium_file(url)` has been moved to the `utils.py` file for reusability.

### DIFF
--- a/framework/utils_test.py
+++ b/framework/utils_test.py
@@ -2,7 +2,6 @@
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 # you may not use this file except in compliance with the License.
-# You may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
 #     http://www.apache.org/licenses/LICENSE-2.0


api/shipping_features_api.py
```diff
@@ -49,6 +65,45 @@ def _get_shipping_stages(self, milestone: int) -> list[Stage]:
...
+  def _validate_feature_in_chromium(
+    self,
+    name: str,
+    enabled_features_json: dict,
+    content_features_file: str
+  ) -> list[Criteria]:
+    """Verify required info exists in Chromium files. Return a list of missing criteria"""
+    criteria_missing = []
+    feature_found = False
+    feature_info = next(
```
jrobbins, line 77: Clever.  I have not seen this pattern used in our code before.

framework/utils_test.py
```diff
@@ -2,6 +2,7 @@
 #
 # Licensed under the Apache License, Version 2.0 (the "License")
 # you may not use this file except in compliance with the License.
+# You may not use this file except in compliance with the License.
```
jrobbins: Redundant with the line above it.

jrobbins (approved):

api/shipping_features_api.py
DanielRyanSmith: I tend to forget the best functional way to do these things for each language, and this is what the AI summary gave me when posing the question in Google search - reads pretty intuitively!

PR #5565, opened 2025-10-15
Make settings link be keyboard navigable.

jrobbins:
Gemini suggested using`<a><sl-menu-item>Settings</sl-menu-item></a>`, but it turns out that that is not keyboard navigable.  I also tried `<sl-menu-item><a>Settings</a></sl-menu-item>` and that could be selected via the keyboard, but the link could not be followed using the keyboard.  So, we are back to `@click`.

### DIFF
--- a/client-src/elements/chromedash-header.ts
+++ b/client-src/elements/chromedash-header.ts
@@ -244,6 +244,10 @@ export class ChromedashHeader extends LitElement {
       });
   }
 
+  gotoSettings() {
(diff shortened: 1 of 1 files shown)


DanielRyanSmith (approved):
It seems like the suggest was a dud. 😕

jrobbins:
> Was this the AI suggestion? If so, unfortunately, it seems like it was a dud. 😕

The problem was that I followed it without fully testing it.

PR #5566, opened 2025-10-16
Supress dev mode warnings while testing

DanielRyanSmith:
In our webtest logs, we get spammed with this warning:

```
Lit is in dev mode. Not recommended for production! See https://lit.dev/msg/dev-mode for more information.
```

This change suppresses those warnings so our logs are less busy with non-actionable info.

### DIFF
--- a/web-test-runner.config.mjs
+++ b/web-test-runner.config.mjs
@@ -1,3 +1,7 @@
+const filteredLogs = [
+  'Lit is in dev mode',
+];
+
 export default {
   files: 'build/**/*_test.{js,ts}',
   concurrentBrowsers: 1,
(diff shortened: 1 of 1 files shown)


web-test-runner.config.mjs
```diff
@@ -4,4 +4,13 @@ export default {
   concurrency: 1,
   nodeResolve: true,
   browserStartTimeout: 60000,
+  filterBrowserLogs(log) {
```
jcscottiii, line 11: For now, we only have one line, but you may want to move all of these into an array for future lines.

https://github.com/GoogleChrome/webstatus.dev/blob/7a6ffd51e418cceb4e3486e6d5865074715e1f9a/frontend/web-test-runner.config.mjs#L17-L27

webstatus.dev does this and then passes it in to filterBrowserLogs

DanielRyanSmith: Excellent! I'll do that

PR #5567, opened 2025-10-16
Add browser-side feature file to Finch name description

DanielRyanSmith:
Related to #5517

Not all Finch features are referenced in runtime_enabled_features.json5 - browser content features are instead referenced in another file. This change updates our description to clarify that.

Additionally, this field is more strict, and requires only alphanumeric characters.


client-src/elements/form-field-specs.ts
```diff
@@ -2567,7 +2567,11 @@ export const ALL_FIELDS: Record<string, Field> = {
 
   finch_name: {
     type: 'input',
-    attrs: TEXT_FIELD_ATTRS,
+    attrs: {
+      ...TEXT_FIELD_ATTRS,
+      placeholder: 'e.g. "StorageBuckets"',
+      pattern: String.raw`[a-zA-Z0-9]*`,
```
jcscottiii: Nit: Could you make this a const and move it to the top with the other regular expressions?

https://github.com/GoogleChrome/chromium-dashboard/blob/9730ee2a84a6d6cb7c500d5595e2de13ffb8e644/client-src/elements/form-field-specs.ts#L126-L142

DanielRyanSmith: Good call - updated!

PR #5568, opened 2025-10-16
Update accuracy verification page to be more succinct

### LINKED ISSUE #5561: Refine feature accuracy notifications to increase user interaction
There are some possible changes that can be made to the accuracy verification emails that are sent out based on reported milestones.

- Currently, accuracy verification emails are sent out at least twice for each milestone - at the time any dev trial, origin trial, or shipping milestone approaches. Origin trial milestones should be kept fairly accurate, since Chromestatus milestones around origin trials are required to be accurate to activate an origin trial. Additionally, dev trial milestones may not be a high accuracy priority until the feature is later in the launching process. This means that it may simplify things to only send verification emails for shipping milestones.
- When the user accesses the accuracy verification page, they are given a longer list of "important" milestones or fields. There might be some adjustments to the form to make it more intuitive to the user what they need to change, and that they need to submit the form regardless of making any changes.

DanielRyanSmith:
Fixes #5561 

This change updates the view of the feature accuracy form.

- A new alert is added to the top of the page to notify users that they need to submit the form to verify accuracy, even if no changes are made.
- Many milestone fields were part of the form that were not a high priority. Users will only receive verify accuracy notifications when their feature is shipping, so dev trial milestones are not as useful here. Also, origin trial milestones are kept very accurate in Chromestatus now that the origin trial process is coordinated in the tool. Now, only shipping milestones are part of the form.
- Some very important fields are now part of the metadata subsection, named as "Important fields". These are fields that are typically required for shipping.
- The submit button is now disabled when the form is submitted.
- More thorough testing.

### Screenshot
Screenshot From 2025-10-16 06-47-10

### DIFF
--- a/client-src/elements/chromedash-guide-verify-accuracy-page.ts
+++ b/client-src/elements/chromedash-guide-verify-accuracy-page.ts
@@ -331,7 +331,9 @@ export class ChromedashGuideVerifyAccuracyPage extends LitElement {
           <sl-icon slot="icon" name="info-circle"></sl-icon>
           <strong
             >Please review your information below and click 'Submit' to confirm
-            it is accurate, <p id="alertUnderline">even if no changes are made</p>!</strong
+            it is accurate,
+            <p id="alertUnderline">even if no changes are made</p>
+            !</strong
           >
         </sl-alert>
         <chromedash-form-table ${ref(this.registerFormSubmitHandler)}>


client-src/elements/chromedash-guide-verify-accuracy-page.ts
```diff
@@ -339,12 +323,24 @@ export class ChromedashGuideVerifyAccuracyPage extends LitElement {
           value=${allFormFields.join(',')}
         />
         <h3>${title}</h3>
+        <sl-alert variant="warning" open class="verify-banner">
+          <sl-icon slot="icon" name="info-circle"></sl-icon>
+          <strong
+            >Please submit this form to verify accuracy, even if no changes are
```
jcscottiii: Another phrasing of this: Please review your information below and click 'Submit' to confirm it is accurate, even if no changes are made.

We could also underline the "even if no changes are made." part too

jcscottiii:
> Many milestone fields were part of the form that were not a high priority. Users will only receive verify accuracy notifications when their feature is shipping, so dev trial milestones are not as useful here. Also, origin trial milestones are kept very accurate in Chromestatus now that the origin trial process is coordinated in the tool. Now, only shipping milestones are part of the form

This makes sense! And I like it

jcscottiii:
If this doesn't help, another iteration of this could have a checkbox under each form to "Certify" that the data is correct for each form before the "Submit" form becomes active. 

But that will slow productivity, so I don't think we need that. But just in case in the future.

client-src/elements/chromedash-guide-verify-accuracy-page.ts
DanielRyanSmith: I like it! I've updated the wording.

client-src/elements/form-definition.ts
```diff
@@ -643,55 +643,16 @@ const DEPRECATION_PREPARE_TO_SHIP_FIELDS: MetadataFields = {
...
-      name: 'Trial extension milestones',
-      fields: [
-        'extension_android_last',
-        'extension_desktop_last',
-        'extension_webview_last',
+        'summary',
+        'owner',
+        'finch_name',
+        'non_finch_justification',
+        'bug_url',
```
jcscottiii, line 655: We really need to make enums /const strings of the fields. That's outside the scope of this PR. But when I was reviewing this (not Gemini lol), I was trying to make sure there were no typos in this. I finally made found it in https://github.com/GoogleChrome/chromium-dashboard/blob/ebb4ce4ac325d88ffb3047c5fd4564752fd0650c/client-src/elements/form-field-specs.ts

DanielRyanSmith: Yes, that's a good call-out. Definitely worth creating enums for these field names. I've created #5569 to track this

PR #5570, opened 2025-10-16
Consider only shipping milestones for staleness

DanielRyanSmith:
This change updates the features that are surfaced as stale on the dashboard to be features only that are about to ship and have had at least one outstanding notification sent out.

### DIFF
--- a/internals/feature_helpers_test.py
+++ b/internals/feature_helpers_test.py
@@ -824,7 +824,7 @@ class FeatureHelpersTest(testing_config.CustomTestCase):
 
   @mock.patch('internals.feature_helpers.get_current_milestone_info')
   @mock.patch('internals.feature_helpers.datetime')
-  def test_get_stale_features__no_outstanding_notifications(
+  def test_get_stale_features__non_shipping_stages(
       self, mock_dt, mock_mstone_info):
     """Ignores features with relevant milestones in non-shipping stages."""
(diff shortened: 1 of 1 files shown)


jcscottiii:
This looks like a great simplification. I had a couple of questions and ideas, mostly around test coverage to make sure we're fully validating the new logic. Let me know what you think.

**`internals/feature_helpers.py`**

*   **L597:** Good catch removing the dev trial and origin trial logic. I noticed this didn't cause any test failures, which made me wonder if we have a small gap in our test coverage. What do you think about adding a test for a feature that's *only* in a dev trial stage to ensure it's correctly ignored now? That might help protect against future regressions.
*   **L699:** Minor thought: would `> 0` be slightly more idiomatic here for checking a positive integer? Happy to defer to your preference.

**`internals/feature_helpers_test.py`**

*   **L794:** To be extra sure about the new `outstanding_notifications` logic, do you think it would be worth adding a quick test for a feature where it's `0`? Just to confirm it's filtered out as we expect.
*   **L810:** Looks like this test name was updated but the docstring might have been missed. Could we update it to match the new logic?

DanielRyanSmith:
> Also another question: any reason this endpoint didn't make it into the openapi.yaml?

This was an oversight - I can refactor it and add it in a separate PR. 🙂

PR #5571, opened 2025-10-16
Only features that have been sent notifications are displayed as stale

DanielRyanSmith:
This change adjusts the aggregation of stale features to be migrated to internal tools.

Occasionally, a feature will become stale mid-week and the dashboard will display this, but no notification has been sent and feature owners do not yet know to take action. Instead, we now only assume features whose feature owners have been contacted are considered potentially stale.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -1029,7 +1029,7 @@ class GenerateStaleFeaturesFile(FlaskHandler):
     for f in stale_features:
       # We should only surface features we have sent notifications about.
       # (This cannot be added to the query, since only 1 inequality is allowed per query.)
-      if f.outstanding_notifications > 0:
+      if f.outstanding_notifications == 0:
         continue
       shipping_stage_type = STAGE_TYPES_SHIPPING[f.feature_type]
       upcoming_ship_stages = Stage.query(


jcscottiii:
Thanks for this change, @DanielRyanSmith! This is a great improvement to how we identify stale features. The logic is clear, and I really appreciate the thorough test updates.

I had a couple of minor, non-blocking thoughts based on some great feedback from a teammate:

**1. Consider a constant for the notification threshold:**

A teammate made a great suggestion for future maintainability. Right now, in `internals/maintenance_scripts.py`, we're checking `f.outstanding_notifications >= 1`. A search of the codebase shows that other parts of the system check this field against different values (e.g., `internals/reminders.py` checks against `2`).

To make the intent of this number clear and avoid confusion with other thresholds, what do you think about defining a constant for it, maybe in `internals/core_enums.py` or `framework/constants.py`?

For example:
`MIN_NOTIFICATIONS_FOR_STALE = 1`

And then the code would read:
`stale_features = [f for f in stale_features if f.outstanding_notifications >= MIN_NOTIFICATIONS_FOR_STALE]`

This is just a thought for making our codebase even more readable. No pressure to add it in this PR, but it seemed like a great idea worth sharing.

**2. A potential micro-optimization:**

This logic is perfectly clear! One small thought for a potential micro-optimization: right now, we iterate through all the initial `stale_features` results to build a new list in memory, and then iterate over that new list. We could combine these into a single loop to avoid creating the intermediate list.

Something like this:
```python
stale_features_query = FeatureEntry.query(...)
stale_features_with_upcoming_ship_stages: list[FeatureEntry] = []
for f in stale_features_query:
  if f.outstanding_notifications < 1:
    continue

  shipping_stage_type = STAGE_TYPES_SHIPPING[f.feature_type]
  # ... rest of the logic
```

No strong opinion, and what you have is very readable, but thought I'd mention it as a possibility.

Overall, this is a solid change. Great work!

DanielRyanSmith:
I decided to add the micro-optimization from number 2 of your comment, but I did not add the constant, since I did not do this with the other PRs, and I've now changed to working to be `f.outstanding_notifications > 0`. 🙂

PR #5572, opened 2025-10-16
Reset outstanding notifications counter

DanielRyanSmith:
This change adds a script that resets all `FeatureEntry.outstanding_notifications` values to 0, if they were incremented.

Our plan is to add more proactive corrective measures for stale features (e.g. unsetting the shipping milestone values if an owner fails to take action verifying data accuracy over multiple weeks). If this is the case, we want to avoid the chance that our logic preemptively takes this action for features that have not had the chance to verify their feature with the new system. Resetting the notifications counter here seems like the right choice to avoid adverse side effects.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -1140,7 +1140,7 @@ class ResetOutstandingNotifications(FlaskHandler):
   def get_template_data(self, **kwargs) -> str:
     self.require_cron_header()
     notified_features: list[FeatureEntry] = FeatureEntry.query(
-      FeatureEntry.outstanding_notifications >= 1
+      FeatureEntry.outstanding_notifications > 0
     ).fetch()
     for f in notified_features:
       logging.info(f'Setting outstanding notifications for feature {f.key.integer_id()} '


jcscottiii:
Question: Are there cases when we don't want to blindly reset all offending features?

internals/maintenance_scripts.py
```diff
@@ -1132,3 +1132,17 @@ def get_template_data(self, **kwargs):
...
+
+class ResetOutstandingNotifications(FlaskHandler):
+  """Reset the FeatureEntry.outstanding_notifications counter for all features"""
+
+  def get_template_data(self, **kwargs) -> str:
+    self.require_cron_header()
+    notified_features = FeatureEntry.query(
+      FeatureEntry.outstanding_notifications >= 1
+    ).fetch()
+    for f in notified_features:
```
jcscottiii, line 1145: You may want to add a log statement here like `Resetting outstanding_notifications for feature %d from %d to 0.` as well for auditing reasons

DanielRyanSmith:
> Question: Are there cases when we don't want to blindly reset all offending features?

This is definitely a one-time thing. We're only wanting to do this now that we've drastically changed the process and requirements around accuracy verification notifications. Otherwise, I don't think there's any risk here, outside of some people who are currently receiving escalated notifications now instead receiving normal notifications again, but that is fine since we're micro-managing the shipping features for the upcoming milestones, since we're in the process of trying to increase accuracy.

internals/maintenance_scripts.py
DanielRyanSmith: Log statement added 🙂

internals/maintenance_scripts.py
```diff
@@ -1132,3 +1132,19 @@ def get_template_data(self, **kwargs):
...
     return f'{count} rollout_milestone fields migrated'
+
+
+class ResetOutstandingNotifications(FlaskHandler):
+  """Reset the FeatureEntry.outstanding_notifications counter for all features"""
+
+  def get_template_data(self, **kwargs) -> str:
+    self.require_cron_header()
+    notified_features: list[FeatureEntry] = FeatureEntry.query(
+      FeatureEntry.outstanding_notifications >= 1
```
jcscottiii: Change this to be >0 like the other PRs

PR #5573, opened 2025-10-16
Update accuracy emails to be explicitly required

DanielRyanSmith:
This change updates the accuracy verification emails to be explicit in the fact that they are now required, as the plan is to unset the shipping milestones for features whose accuracy has not been verified over an extended period.

Additionally, a caveat about shipping milestones being unset is added to escalated emails.

### DIFF
--- a/internals/reminders.py
+++ b/internals/reminders.py
@@ -41,6 +41,7 @@ WEBSTATUS_EMAIL = 'webstatus@google.com'
 CBE_ESCLATION_EMAIL = 'cbe-releasenotes@google.com'
 STAGING_EMAIL = 'jrobbins-test@googlegroups.com'
 EMAIL_SUBJECT_PREFIX = '[Action required]'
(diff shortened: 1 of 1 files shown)


internals/reminders.py
```diff
@@ -100,7 +100,7 @@ def build_email_tasks(
     subject = subject_format % fe.name
     if is_escalated:
       if EMAIL_SUBJECT_PREFIX in subject:
-        subject = subject.replace(EMAIL_SUBJECT_PREFIX, 'Escalation request')
+        subject = subject.replace(EMAIL_SUBJECT_PREFIX, '[Escalated Request]')
```
jcscottiii: Nit: Since we are changing this, it would be a good time to change this to a constant. ESCALATION_EMAIL_SUBJECT_PREFIX

jcscottiii (approved):
LGTM with one nit

Issue #5496, opened 2025-10-03
Add "stale feature" banner to the "Edit all fields" page


DanielRyanSmith:
After interacting with prominent users, it seems that the "Edit all fields" functionality is very commonly used. We might get even more traction on users verifying the accuracy of their data if this banner is also surfaced on that page.

PR #5574, opened 2025-10-17
Refactor stale feature banner to be reusable

DanielRyanSmith:
Part of #5496 

(This ended up being more work than I realized)

This change updates the existing logic for displaying the outdated feature banner and moves it to a utility function so that it can be used on any page.

### DIFF
--- a/client-src/elements/utils.ts
+++ b/client-src/elements/utils.ts
@@ -4,7 +4,12 @@ import {html, nothing, TemplateResult} from 'lit';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
 import DOMPurify from 'dompurify';
 import {marked} from 'marked';
-import {Feature, FeatureLink, StageDict} from '../js-src/cs-client.js';
+import {
+  Channels,
+  Feature,
+  FeatureLink,
+  StageDict,
+} from '../js-src/cs-client.js';
 import {markupAutolinks} from './autolink.js';
 import {FORMS_BY_STAGE_TYPE, FormattedFeature} from './form-definition.js';
 import {
@@ -30,6 +35,9 @@ const NARROW_WINDOW_MAX_WIDTH = 700;
 // Represent a 4-week period in milliseconds. This grace period needs
 // to be consistent with ACCURACY_GRACE_PERIOD in internals/reminders.py.
 const ACCURACY_GRACE_PERIOD = 4 * 7 * 24 * 60 * 60 * 1000;
+
+// A 9-week grace period used to approximate 2 months for shipped features.
+const SHIPPED_FEATURE_OUTDATED_GRACE_PERIOD = 9 * 7 * 24 * 60 * 60 * 1000;
 
 export const IS_MOBILE = (() => {
   const width =
@@ -748,15 +756,11 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
 /**
  * A feature is outdated if it has shipped, and its
  * accurate_as_of is before its latest shipping date before today.
- * @param {Feature} feature The feature object.
- * @param {boolean} hasShipped Whether the feature has any shipping stages.
- * @param {string | undefined} closestShippingDate The ISO string of the latest shipping date.
- * @returns {boolean} True if the shipped feature is outdated, false otherwise.
  */
 function isShippedFeatureOutdated(
-  feature,
+  feature: Feature,
   hasShipped: boolean,
-  closestShippingDate: string | undefined
+  closestShippingDate: string
 ): boolean {
   // Check if a feature has shipped.
   if (!hasShipped) {
@@ -775,17 +779,12 @@ function isShippedFeatureOutdated(
 /**
  * Determine if it should show warnings to a feature author, if
  * a shipped feature is outdated, and it has edit access.
- * @param {Feature} feature The feature object.
- * @param {boolean} userCanEdit Whether the current user can edit the feature.
- * @param {boolean} hasShipped Whether the feature has any shipping stages.
- * @param {string | undefined} closestShippingDate The ISO string of the latest shipping date.
- * @returns {boolean} True if the banner should be shown for an author.
  */
 function isShippedFeatureOutdatedForAuthor(
-  feature,
+  feature: Feature,
   userCanEdit: boolean,
   hasShipped: boolean,
-  closestShippingDate
+  closestShippingDate: string
 ): boolean {
   return (
     userCanEdit &&
@@ -796,36 +795,27 @@ function isShippedFeatureOutdatedForAuthor(
 /**
  * Determine if it should show warnings to all readers, if
  * a shipped feature is outdated, and last update was > 9 weeks.
- * @param {Feature} feature The feature object.
- * @param {boolean} hasShipped Whether the feature has any shipping stages.
- * @param {number} currentDate The current date as a timestamp.
- * @param {string | undefined} closestShippingDate The ISO string of the latest shipping date.
- * @returns {boolean} True if the banner should be shown for all users.
  */
 function isShippedFeatureOutdatedForAll(
-  feature,
+  feature: Feature,
   hasShipped: boolean,
   currentDate: number,
-  closestShippingDate
+  closestShippingDate: string
 ): boolean {
   if (!isShippedFeatureOutdated(feature, hasShipped, closestShippingDate)) {
     return false;
   }
 
-  // Represent two months grace period.
-  const nineWeekPeriod = 9 * 7 * 24 * 60 * 60 * 1000;
   const isVerified = isVerifiedWithinGracePeriod(
     feature.accurate_as_of,
     currentDate,
-    nineWeekPeriod
+    SHIPPED_FEATURE_OUTDATED_GRACE_PERIOD
   );
   return !isVerified;
 }
 
 /**
  * Fetches the shipping date (final_beta) for a specific milestone.
- * @param {number} milestone The milestone number to fetch data for.
- * @returns {Promise<string>} A promise that resolves to the shipping date ISO string, or empty string.
  */
 async function fetchClosestShippingDate(milestone: number): Promise<string> {
   if (milestone === 0) {
@@ -847,7 +837,7 @@ async function fetchClosestShippingDate(milestone: number): Promise<string> {
 
 export interface closestShippingDateInfo {
   // The closest milestone shipping date as an ISO string.
-  closestShippingDate: string | undefined;
+  closestShippingDate: string;
   isUpcoming: boolean;
   hasShipped: boolean;
 }
@@ -856,14 +846,10 @@ export interface closestShippingDateInfo {
  * Determine if this feature is upcoming - scheduled to ship
  * within two milestones, then find the closest shipping date
  * for that upcoming milestone or an already shipped milestone.
- * @param {object} channels Channel information object.
- * @param stages Array of feature stages.
- * @returns {Promise<closestShippingDateInfo>} An object containing the closest shipping date,
- *   and booleans for isUpcoming and hasShipped.
  */
 export async function findClosestShippingDate(
-  channels,
-  stages
+  channels: Channels,
+  stages: StageDict[]
 ): Promise<closestShippingDateInfo> {
   const latestStableVersion = channels?.stable?.version;
   if (!latestStableVersion || !stages) {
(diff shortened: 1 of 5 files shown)


client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+    isUpcoming &&
+    !isVerifiedWithinGracePeriod(feature.accurate_as_of, currentDate)
+  );
+}
+
+/**
+ * Returns a warning banner TemplateResult if a feature is considered outdated,
+ * otherwise returns null.
+ * @param {Feature} feature The feature object.
+ * @param {closestShippingDateInfo} shippingInfo The shipping information from findClosestShippingDate.
```
jcscottiii, line 950: Suggestion: This is a `.ts` file, so we should use TypeScript's explicit type annotations for function parameters instead of relying on JSDoc. Could we add the `: Feature` type to the `feature` parameter here? This would improve type safety and editor support. This comment applies to all the new functions added in this PR (`isShippedFeatureOutdatedForAuthor`, `isShippedFeatureOutdatedForAll`, etc.), which are also missing explicit types for their `feature` and other parameters.

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+    currentDate = Date.now();
+  }
+  const {closestShippingDate, hasShipped, isUpcoming} = shippingInfo;
+  if (isUpcomingFeatureOutdated(feature, isUpcoming, currentDate)) {
+    if (userCanEdit) {
+      return html`
+        <div class="warning layout horizontal center">
+          <span class="tooltip" id="outdated-icon" title="Feature outdated ">
+            <sl-icon name="exclamation-circle-fill" data-tooltip></sl-icon>
+          </span>
```
jcscottiii, line 971: Suggestion: Along the lines of the previous comment, could we add the missing types here? It looks like `feature` should be `Feature` and `closestShippingDate` should be `string | undefined`.

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+              >verify that your feature is accurate</a
+            >.
+          </span>
+        </div>
+      `;
+    } else {
+      return html`
+        <div class="warning layout horizontal center">
+          <span class="tooltip" id="outdated-icon" title="Feature outdated ">
+            <sl-icon name="exclamation-circle-fill" data-tooltip></sl-icon>
```
jcscottiii, line 990: Suggestion: Same as above, could we type the `feature` and `closestShippingDate` parameters here?

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+          <span>
+            Your feature hasn't been verified as accurate since&nbsp;
+            <sl-relative-time
+              date=${feature.accurate_as_of ?? ''}
+            ></sl-relative-time
+            >, but it claims to have shipped&nbsp;
+            <sl-relative-time date=${closestShippingDate}></sl-relative-time>.
+            Please
+            <a href="/guide/verify_accuracy/${feature.id}"
+              >verify that your feature is accurate</a
```
jcscottiii, line 1032: Suggestion: Could we add types for `channels` and `stages`? Based on the code, it looks like `stages` should be `StageDict[]`. For `channels`, even typing it as `any` for now would be an improvement, though a more specific `Channels` type would be ideal if one exists.

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+              date=${feature.accurate_as_of ?? ''}
+            ></sl-relative-time
+            >, but it is scheduled to ship&nbsp;
+            <sl-relative-time date=${closestShippingDate}></sl-relative-time>.
+            Please
+            <a href="/guide/verify_accuracy/${feature.id}"
+              >verify that your feature is accurate</a
+            >.
+          </span>
+        </div>
```
jcscottiii, line 984: Suggestion: This is a great use of a fixed week count to approximate a "two-month" period. To make the code a little more self-documenting, what do you think about extracting the magic number into a named constant? We could also clarify the comment slightly.

For example, we could define this at the top of the file:
```typescript
// A 9-week grace period used to approximate 2 months for shipped features.
const SHIPPED_FEATURE_OUTDATED_GRACE_PERIOD = 9 * 7 * 24 * 60 * 60 * 1000;
```
And then use the constant here. This would make the purpose of the value immediately clear at the point of use.

client-src/elements/utils_test.ts
```diff
@@ -346,4 +364,307 @@ go/this-is-a-test
...
+    const CURRENT_DATE_MS = Date.parse('2025-03-01T00:00:00Z');
+    // Verified: Feb 15, 2025 (within 4 weeks)
+    const VERIFIED_DATE = new Date(CURRENT_DATE_MS - FOUR_WEEKS / 2).toISOString();
+    // Outdated: Jan 15, 2025 (outside 4 weeks)
+    const OUTDATED_DATE = new Date(
+      CURRENT_DATE_MS - FOUR_WEEKS - 1000
+    ).toISOString();
+    // Very Outdated: Dec 1, 2024 (outside 9 weeks)
+    const VERY_OUTDATED_DATE = new Date(
+      CURRENT_DATE_MS - NINE_WEEKS - 1000
```
jcscottiii, line 589: Suggestion: This test case name, `'should return a banner if accurate_as_of is missing for upcoming'`, is a little ambiguous. It correctly tests that a banner *is* shown, but it's not immediately obvious *why*. What do you think about renaming it to something like `'should return upcoming banner if accurate_as_of is missing'` and adding a brief comment inside the test explaining that a missing `accurate_as_of` date is considered outdated for an upcoming feature? It would make the intent clearer for future maintainers.

client-src/elements/utils.ts
jcscottiii: wrong location gemini

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+ * Determine if it should show warnings to all readers, if
+ * a shipped feature is outdated, and last update was > 9 weeks.
+ * @param {Feature} feature The feature object.
+ * @param {boolean} hasShipped Whether the feature has any shipping stages.
+ * @param {number} currentDate The current date as a timestamp.
+ * @param {string | undefined} closestShippingDate The ISO string of the latest shipping date.
+ * @returns {boolean} True if the banner should be shown for all users.
+ */
+function isShippedFeatureOutdatedForAll(
+  feature,
```
jcscottiii: Same thing. Add the types.

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+ * Determine if it should show warnings to a feature author, if
+ * a shipped feature is outdated, and it has edit access.
+ * @param {Feature} feature The feature object.
+ * @param {boolean} userCanEdit Whether the current user can edit the feature.
+ * @param {boolean} hasShipped Whether the feature has any shipping stages.
+ * @param {string | undefined} closestShippingDate The ISO string of the latest shipping date.
+ * @returns {boolean} True if the banner should be shown for an author.
+ */
+function isShippedFeatureOutdatedForAuthor(
+  feature,
```
jcscottiii: Same thing around adding the types instead of relying on jsdocs

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+ * Determine if this feature is upcoming - scheduled to ship
+ * within two milestones, then find the closest shipping date
+ * for that upcoming milestone or an already shipped milestone.
+ * @param {object} channels Channel information object.
+ * @param stages Array of feature stages.
+ * @returns {Promise<closestShippingDateInfo>} An object containing the closest shipping date,
+ *   and booleans for isUpcoming and hasShipped.
+ */
+export async function findClosestShippingDate(
+  channels,
```
jcscottiii: We should import the Channels and StageDict from cs-client.js and use those

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+/**
+ * A feature is outdated if it is scheduled to ship in the next 2 milestones,
+ * and its accurate_as_of date is at least 4 weeks ago.
+ * @param {Feature} feature The feature object.
+ * @param {boolean} isUpcoming Whether the feature is shipping in the next 2 milestones.
+ * @param {number} currentDate The current date as a timestamp.
+ * @returns {boolean} True if the upcoming feature is outdated, false otherwise.
+ */
+function isUpcomingFeatureOutdated(
+  feature,
```
jcscottiii: Same thing for adding the types

client-src/elements/utils.ts
jcscottiii: wrong place gemini

client-src/elements/utils.ts
jcscottiii: Wrong place gemini

client-src/elements/utils.ts
```diff
@@ -742,3 +744,343 @@ export const METRICS_TYPE_AND_VIEW_TO_SUBTITLE = {
...
+  hasShipped: boolean,
+  currentDate: number,
+  closestShippingDate
+): boolean {
+  if (!isShippedFeatureOutdated(feature, hasShipped, closestShippingDate)) {
+    return false;
+  }
+
+  // Represent two months grace period.
+  const nineWeekPeriod = 9 * 7 * 24 * 60 * 60 * 1000;
```
jcscottiii: Suggestion: This is a great use of a fixed week count to approximate a "two-month" period. To make the code a little more self-documenting, what do you think about extracting the magic number into a named constant? We could also clarify the comment slightly.

For example, we could define this at the top of the file:
```typescript
// A 9-week grace period used to approximate 2 months for shipped features.
const SHIPPED_FEATURE_OUTDATED_GRACE_PERIOD = 9 * 7 * 24 * 60 * 60 * 1000;
```
And then use the constant here. This would make the purpose of the value immediately clear at the point of use.

client-src/elements/utils.ts
jcscottiii: wrong place gemini

client-src/elements/utils_test.ts
jcscottiii: wrong place gemini

client-src/elements/utils_test.ts
```diff
@@ -346,4 +364,307 @@ go/this-is-a-test
...
+      const result = getFeatureOutdatedBanner(
+        feature,
+        shippingInfoNeither,
+        CURRENT_DATE_MS,
+        true
+      );
+      assert.isNull(result);
+    });
+
+    it('should return a banner if accurate_as_of is missing for upcoming', () => {
```
jcscottiii: Suggestion: This test case name, `'should return a banner if accurate_as_of is missing for upcoming'`, is a little ambiguous. It correctly tests that a banner *is* shown, but it's not immediately obvious *why*. What do you think about renaming it to something like `'should return upcoming banner if accurate_as_of is missing'` and adding a brief comment inside the test explaining that a missing `accurate_as_of` date is considered outdated for an upcoming feature? It would make the intent clearer for future maintainers.

client-src/elements/utils.ts
DanielRyanSmith: I've updated this to now use the types effectively. This did, however, chain into a lot of other changes, since the tests required more thorough values to make sure the variables were properly typed. Then, other components that previously used some of these existing functions also needed updating. This even uncovered that there were some data type discrepancies in the Feature type definition, so there's a lot of updates coming with this change. 🙂

client-src/elements/utils.ts
DanielRyanSmith: Updated!

PR #5576, opened 2025-10-17
Display enterprise milestones on the estimated milestones table

### LINKED ISSUE #4498: Accurate-as-of email confusing for enterprise features

DanielRyanSmith:
Fixes #4498 

Enterprise stages with milestones now display in the estimated milestones table, in emails and intents.

### DIFF
--- a/templates/estimated-milestones-table.html
+++ b/templates/estimated-milestones-table.html
@@ -30,7 +30,7 @@
     {%- endfor -%}
     {%- for stage in stage_info.ot_stages -%}
       {%- if stage.milestones.android_first %}
-      <tr><td>Origin trial Android first</td><td>{{stage.milestones.android_first}}</td></tr>
+        <tr><td>Origin trial Android first</td><td>{{stage.milestones.android_first}}</td></tr>
       {%- endif -%}
       {%- if stage.milestones.android_last %}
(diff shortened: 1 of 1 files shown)


templates/estimated-milestones-table.html
```diff
@@ -30,7 +30,7 @@
     {%- endfor -%}
     {%- for stage in stage_info.ot_stages -%}
       {%- if stage.milestones.android_first %}
-        <tr><td>Origin trial Android first</td><td>{{stage.milestones.android_first}}</td></tr>
+      <tr><td>Origin trial Android first</td><td>{{stage.milestones.android_first}}</td></tr>
```
jrobbins: Why not have this indented under the if-statement?

DanielRyanSmith: This was an accidental spacing fix - thanks for catching!

PR #5578, opened 2025-10-20
Cron job to reset stale shipping milestones after 4+ notifications

DanielRyanSmith:
This adds a new cron job (not yet activated), that will reset all shipping milestones of a feature that has not verified its accuracy after being notified 4+ times during an upcoming shipping milestone cycle.

### DIFF
--- a/internals/maintenance_scripts.py
+++ b/internals/maintenance_scripts.py
@@ -1187,16 +1187,19 @@ class ResetStaleShippingMilestones(FlaskHandler):
         # Create an activity that shows all the shipping milestones have been set to null.
         activity = Activity(
           feature_id=f.key.integer_id(),
+          amendments=[],
           content='Shipping milestones were unset due to failure to verify accuracy.')
         if s.milestones:
           self._reset_milestone(s, 'desktop_first', activity)
           self._reset_milestone(s, 'android_first', activity)
           self._reset_milestone(s, 'ios_first', activity)
           self._reset_milestone(s, 'webview_first', activity)
-          entities_to_update.append(s)
-          entities_to_update.append(activity)
-      entities_to_update.append(f)
+          # Only update the stage and save the activity if some milestones were unset.
+          if activity.amendments:
+            entities_to_update.append(s)
+            entities_to_update.append(activity)
       f.outstanding_notifications = 0
+      entities_to_update.append(f)
       num_features_reset += 1
     if entities_to_update:
       ndb.put_multi(entities_to_update)


internals/maintenance_scripts.py
```diff
@@ -1152,3 +1152,38 @@ def get_template_data(self, **kwargs) -> str:
...
+
+      # Get all the shipping stages of the stale feature and reset any
+      # milestones that are set.
+      stages: list[Stage] = Stage.query(
+        Stage.feature_id == f.key.integer_id(),
+        Stage.stage_type == STAGE_TYPES_SHIPPING[f.feature_type]
+      ).fetch()
+      for s in stages:
+        if s.milestones:
+          s.milestones.desktop_first = None
```
jrobbins: Do we have feature entries that shipped several milestones ago that would be reset?  Maybe we should apply this from, e.g., 140 onwards.

internals/maintenance_scripts.py
```diff
@@ -1152,3 +1152,38 @@ def get_template_data(self, **kwargs) -> str:
...
+          s.milestones.android_first = None
+          s.milestones.ios_first = None
+          s.milestones.webview_first = None
+          entities_to_update.append(s)
+        entities_to_update.append(f)
+      f.outstanding_notifications = 0
```
jrobbins, line 1201: I realize that the feature owners have already ignored 4 reminders, but I think we should queue up an outgoing email to inform them of the change and the reason for it.

Also, we need to add an Activity record for this change.  That activity could also include comment text to explain what happened.

DanielRyanSmith: I do agree that there should be a notification sent to owners as part of this process. However, I can add that in a separate PR, since it will require wording drafting separately.

I've also added the activity generation to this PR. 🙂

internals/maintenance_scripts.py
DanielRyanSmith: Agreed - I've added a conditional for this.

internals/maintenance_scripts.py
```diff
@@ -1152,3 +1152,53 @@ def get_template_data(self, **kwargs) -> str:
...
+        activity = Activity(
+          feature_id=f.key.integer_id(),
+          content='Shipping milestones were unset due to failure to verify accuracy.')
+        if s.milestones:
+          self._reset_milestone(s, 'desktop_first', activity)
+          self._reset_milestone(s, 'android_first', activity)
+          self._reset_milestone(s, 'ios_first', activity)
+          self._reset_milestone(s, 'webview_first', activity)
+          entities_to_update.append(s)
+          entities_to_update.append(activity)
```
jrobbins: I think you need to only add the Activity if it has some Amendments.  Otherwise, when time this Cron job runs, a feature with milestones before 140 will get an empty Activity.

jrobbins (approved):
LGTM after you address one potential problem.

PR #5606, opened 2025-10-25
Query for features and stages missing in cache all at once

DanielRyanSmith:
This change updates some of the asynchronous logic in feature_helpers.py to fetch all required features and stages that are not cached at the same time. The behavior should be identical.

This was originally #5451, but I closed that PR to address it later.


internals/feature_helpers.py
```diff
@@ -480,37 +482,48 @@ def get_by_ids(feature_ids: list[int],
...
-      futures_by_id[feature_id] = FeatureEntry.get_by_id_async(feature_id)
+  needed_ids = [fid for fid in feature_ids if fid not in result_dict]
+
+  if needed_ids:
+    # Create ndb.Key objects for the batch operation
+    needed_keys = [ndb.Key(FeatureEntry, fid) for fid in needed_ids]
+
+    # Get all features and stages asynchronously.
+    unformatted_features_future: list[ndb.Future] = ndb.get_multi_async(needed_keys)
+    stages_futures: list[ndb.Future] = []
```
jcscottiii, line 492: Question:

Looking at the current database, what's the worst case scenario of how many stages will need to be fetched?

If it's a lot, we might want to do batches of what you were kinda doing in the previous PR:

```
BATCH_SIZE = 30
...

stages_futures = []
for i in range(0, len(ids), BATCH_SIZE):
  batch_ids = ids[i:i + BATCH_SIZE]
  # fetch_async returns a Future object immediately
  stages_futures.append(
     Stage.query(Stage.feature_id.IN(batch_ids), Stage.archived == False)
      .fetch_async(None))
```

If the number is low, keeping it as-is is good. But if we are talking a high amount, this could help a lot (e.g. 6k individual feature requests to 200 batched requests)


But if you do this, you will probably want to add a case to feature_helpers_test.py to assert that the batching works.

Let me know what you think.

jcscottiii: Actually I just saw that these methods are used in search.py which look already paginated. We should be okay. But please verify.

jcscottiii (approved):
One question about potential batching.

Another observation: I noticed that get_by_ids and get_feature_names_by_ids share a lot of similar logic. There might be an opportunity to refactor the common code into a shared helper function / generics to reduce duplication. Not a hard requirement though.

LGTM

internals/feature_helpers.py
DanielRyanSmith: I do think that the suggestion might make a slightly more optimal set of queries, but I don't think the number of entities that we query for is high enough to worry too much about it. If we find the site is starting to feel bogged-down, this might be the right move.

Do you think I should create an issue regarding this possible optimization?

jcscottiii: Could you do me a favor: time some of the requests in production currently and time it again after this is deployed. If it's not bad with the new code, we won't need to file that issue.

jcscottiii: We can still merge this in the meantime.

DanielRyanSmith:
I still have some metrics to take for this PR, so I'll leave it open and merge it in next week. 🙂

DanielRyanSmith:
I'm going to finally merge this. 😅 

I did some quick timing of the requests for populating uncached pages of the ["All features" list](https://chromestatus.com/features). The server processes seem to typically take 2-3 seconds. I'll do some additional testing for comparison once this is available on prod. 

Screenshot 2025-12-11 1 57 48 PM
Screenshot 2025-12-11 1 57 33 PM
Screenshot 2025-12-11 1 57 21 PM
Screenshot 2025-12-11 1 56 45 PM

PR #5619, opened 2025-10-28
Remove mentioned of Browser Interventions

### LINKED ISSUE #3527: "Interventions" terminology is discouraged
See https://github.com/WICG/interventions/.

This is leading to beta blog posts being written that propagate the usage of this terminology, which is counter to the efforts outlined above to sunset it.

DanielRyanSmith:
Fixes #3527 

This change removes all mention of Browser Interventions in the Chromestatus UI and API. The enum for the implementation status still exists, but as far as I can tell in `form-definition.ts`, we do not surface this field anywhere.


jrobbins (approved):
code looks good, but what's up with those playwright image changes?

DanielRyanSmith:
> what's up with those playwright image changes?

The shipping milestone field descriptions have very slightly changed - I think that's the only difference for the most part

PR #5620, opened 2025-10-28
Add expectation for TypeScript error

DanielRyanSmith:
This is a fix for the issue that is blocking #5611

I originally thought that we would need to update the component to instead extend LitElement, but this also works as long as the behavior is what we expect it to.

### DIFF
--- a/client-src/elements/chromedash-textarea.ts
+++ b/client-src/elements/chromedash-textarea.ts
@@ -1,8 +1,9 @@
-import {TemplateResult, html, nothing} from 'lit';
+import {html, nothing} from 'lit';
 import SlTextarea from '@shoelace-style/shoelace/dist/components/textarea/textarea.js';
 import {customElement, property, state} from 'lit/decorators.js';
 import {autolink} from './utils.js';
 
+// @ts-expect-error ts(1238)
 @customElement('chromedash-textarea')
 export class ChromedashTextarea extends SlTextarea {
   @property({type: Boolean})


DanielRyanSmith:
# Gemini's explanation of the problem to me:

This is a classic type collision problem that often happens when extending components that interact with the standard DOM element interfaces.

The error message is telling you everything you need to know:
`The types of 'prototype.autocorrect' are incompatible... Type 'string' is not assignable to type 'boolean'.`

-----

### 🧬 The Origin of the Error

Here is the breakdown of the conflict:

1.  **Your Base Class:** You are extending `SlTextarea`. The Shoelace `SlTextarea` component defines its own `autocorrect` property as a `string` (to match the HTML attribute, which accepts values like `"on"` and `"off"`).
2.  **The Decorator's Expectation:** The `@customElement` decorator validates that your class signature is compatible with the standard `HTMLElement` interface provided by TypeScript's built-in DOM libraries (`lib.dom.d.ts`).
3.  **The Type Conflict:** In many versions of the TypeScript DOM library, the `HTMLElement` interface *itself* (or a related interface it uses) defines an `autocorrect` property as a `boolean`.

Therefore, when the `@customElement` decorator "inspects" your `ChromedashTextarea` class, it sees:

  * An inherited `autocorrect: string` (from `SlTextarea`).
  * An expected `autocorrect: boolean` (from `HTMLElement`).

TypeScript flags this as an error because `string` is not assignable to `boolean`, and it blocks the decorator.

-----

### 🔧 How to Fix the Error

You have two simple and effective ways to resolve this.

#### Option 1: Suppress the Error (Recommended)

Since you know this is a type-only issue and your code works functionally, the cleanest and most targeted fix is to tell TypeScript to ignore this specific error.

```typescript
import {TemplateResult, html, nothing} from 'lit';
import SlTextarea from '@shoelace-style/shoelace/dist/components/textarea/textarea.js';
import {customElement, property, state} from 'lit/decorators.js';
import {autolink} from './utils.js';

// V-- FIX: Add this comment to suppress the known type error --V
// @ts-expect-error ts(1238)
@customElement('chromedash-textarea')
export class ChromedashTextarea extends SlTextarea {
  // ... rest of your component implementation
}
```

Using `@ts-expect-error` is generally preferred over `@ts-ignore` as it will cause a TypeScript error *if* the underlying error (1238) ever goes away, letting you know you can remove the suppression comment.

#### Option 2: Re-declare the Conflicting Property

You can also satisfy the type-checker by explicitly re-declaring the conflicting property in your subclass with a type of `any`. This overrides the inherited type and resolves the conflict.

```typescript
import {TemplateResult, html, nothing} from 'lit';
import SlTextarea from '@shoelace-style/shoelace/dist/components/textarea/textarea.js';
import {customElement, property, state} from 'lit/decorators.js';
import {autolink} from './utils.js';

@customElement('chromedash-textarea')
export class ChromedashTextarea extends SlTextarea {
  // V-- FIX: Re-declare the conflicting property as 'any' --V
  // This satisfies the type conflict between SlTextarea's
  // 'string' and HTMLElement's 'boolean' type.
  @property()
  autocorrect: any;

  @property({type: Boolean})
  multiple;
  // ... rest of your properties
}
```

This fix is more explicit but adds a bit of boilerplate to your class. Both methods are perfectly valid.

DanielRyanSmith:
Hmm, it seems the comment for the error expectation is causing errors itself because TypeScript hasn't been updated. I'm going to try adding this change to the actual PR that updates TypeScript instead.

jrobbins (approved):

DanielRyanSmith:
Checking the other PR, it seems like we came to the same conclusion @jrobbins 🙂 - the wording change seems to have made a difference. I'll close this PR, since the other PR now contains this change.

PR #5622, opened 2025-10-29
Exclude unreviewed WP features from enterprise release notes.

jrobbins:
This should resolve b/454934046.

The enterprise release notes page consists of enterprise features and WP features that have some enterprise impact.  Of those WP features, only ones that have approval on the enterprise gate should be included.

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -104,11 +104,14 @@ def _filter_out_wp_features_lacking_enterprise_approval(
   approved_enterprise_gates: list[Gate] = []
   if wp_feature_ids:
     approved_enterprise_gates = Gate.query(
-        Gate.feature_id.IN(wp_feature_ids),
-        Gate.gate_type == GATE_ENTERPRISE_SHIP).fetch()
+        # Use a range because the IN operator is limited to 30.  This gets
+        # extra Gates, but that doesn't matter.
+        Gate.feature_id >= min(wp_feature_ids),
+        Gate.feature_id <= max(wp_feature_ids),
+        Gate.gate_type == GATE_ENTERPRISE_SHIP,
+        Gate.state.IN(Gate.APPROVED_STATES)).fetch()
   approved_wp_feature_ids = {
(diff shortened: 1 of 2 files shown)


internals/feature_helpers.py
```diff
@@ -94,6 +95,27 @@ def get_future_results(async_features: Future | None) -> list[FeatureEntry]:
...
+
+def _filter_out_wp_features_lacking_enterprise_approval(
+    features: list[FeatureEntry]) -> list[FeatureEntry]:
+  """Take out any WP features that did not yet earn enterprise approval."""
+  wp_feature_ids = {fe.key.integer_id() for fe in features
+                    if fe.feature_type != FEATURE_TYPE_ENTERPRISE_ID}
+  approved_enterprise_gates: list[Gate] = []
+  if wp_feature_ids:
+    approved_enterprise_gates = Gate.query(
+        Gate.feature_id.IN(wp_feature_ids),
```
DanielRyanSmith: @jcscottiii identified [here](https://github.com/GoogleChrome/chromium-dashboard/pull/5451#discussion_r2379941894) that the IN operator might come with some constraints, requiring 30 or less equalities. I handled a similar query requirement [in this change](https://github.com/GoogleChrome/chromium-dashboard/pull/5606/files) regarding fetching multiple groups of stages.

jrobbins: OK, I handled this by using `>=` and `<=` operators instead.  It causes a few extra Gates to be retrieved, but I was able to add an `IN()` operator for the state field, which should help. Performance on staging seems good, and no new index was needed.

I also switched the release notes cache key from DEFAULT_CACHE_KEY to SEARCH_CACHE_KEY because generating this page is a kind of search.  Specifically, that is needed because when we save a Vote, we update the FeatureEntry's updated field and call `put()` which invalidates just one DEFAULT_CACHE_KEY entry but all of the SEARCH_CACHE_KEY entries, which will now include these release notes entries.

PR #5628, opened 2025-10-29
Remove @polymer deps

jrobbins:
I recently removed all the usages of polymer elements, so removing the whole dependency is pretty straightforward.


DanielRyanSmith (approved):
Excellent!

PR #5634, opened 2025-11-01
Implement release notes reviewers.

jrobbins:
This should resolve b/446217660.  Basically, the enterprise team requested the ability to check off which release notes items have been reviewed by their tech writer, and to have those two checkmarks reset if anyone else edits the name or summary fields.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -303,6 +303,11 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         stage_was_updated = True
       stage.milestones = milestones
 
+      # Changing other rollout details also counts.
+      if ('rollout_platforms' in change_info or
+          'rollout_details' in change_info):
+        ship_milestones_were_updated = True
+
       if stage_was_updated:
         stages_to_store.append(stage)
 
@@ -313,6 +318,18 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
     # Return the list of modified stages, and whether any shipping/rollout
     # milestones were updated.
     return stages_to_store, ship_milestones_were_updated
+
+  def _maybe_reset_releasenotes_flags(
+      self, feature: FeatureEntry,
+      changed_fields: CHANGED_FIELDS_LIST_TYPE) -> None:
+    if permissions.can_review_release_notes(self.get_current_user()):
+      return
+    if feature.is_releasenotes_content_reviewed:
+      feature.is_releasenotes_content_reviewed = False
+      changed_fields.append(('is_releasenotes_content_reviewed', True, False))
+    if feature.is_releasenotes_publish_ready:
+      feature.is_releasenotes_publish_ready = False
+      changed_fields.append(('is_releasenotes_publish_ready', True, False))
 
   def _patch_update_special_fields(
       self,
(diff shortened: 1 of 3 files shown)


jrobbins:
Screenshot:
image

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -702,27 +737,71 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
...
   }
 
-  renderConfidential(f: Feature): TemplateResult {
+  renderLabels(f: Feature): TemplateResult {
+    let confidentialLabel = html``;
     if (f.confidential) {
-      return html`<span class="confidential"
-        ><strong>CONFIDENTIAL</strong></span
+      confidentialLabel = html`<span class="confidential">CONFIDENTIAL</span>`;
+    }
```
DanielRyanSmith, line 741: These labels would be a good option for a Shoelace tag component. We use these for the badges [on webstatus.dev](https://github.com/search?q=repo%3AGoogleChrome%2Fwebstatus.dev%20sl-tag&type=code)

https://shoelace.style/components/tag

gen/js/chromestatus-openapi/package.json
```diff
@@ -16,6 +16,6 @@
     "prepare": "npm run build"
   },
   "devDependencies": {
-    "typescript": "^5.9"
+    "typescript": "^4.0 || ^5.0"
```
DanielRyanSmith: I know this is an unintended consequence of generating the new models. 😅 This should be reverted

package-lock.json
```diff
@@ -82,7 +82,7 @@
     "gen/js/chromestatus-openapi": {
       "version": "1.0.0",
       "devDependencies": {
-        "typescript": "^5.9"
+        "typescript": "^4.0 || ^5.0"
```
DanielRyanSmith: This will also need to be reverted

gen/js/chromestatus-openapi/package.json
jcscottiii: Broader question / Outside the scope of this PR: I know we checked in the generated code to help others learn more about openapi. How do we feel about removing it from the source tree now / do we feel comfortable enough to remove it? We would then generate it locally before every deployment. Which means every developer would need to have Java on their machine. 

Just an option if this is getting in the way. Open to either way. We would add something [similar](https://github.com/GoogleChrome/webstatus.dev/blob/4e494f69a0fb9c11a85d4b9b5f21a50ca7362f7f/.gitignore#L42-L46) to this repository's gitignore file too.

jrobbins: James, I prefer to avoid having generated code checked in.  I think we could require all developers to be able to run `npm run openapi`.  However, what about this generated package.json file that will trigger a security advisory (IIRC)?

jcscottiii: When we remove the full `gen` directory, it will remove this file and it should go away I thought (because it no longer tracks it)

jrobbins: > When we remove the full `gen` directory, it will remove this file and it should go away I thought (because it no longer tracks it)

Ah, and that is good enough because we also don't actually use those files.  OK.

PR #5643, opened 2025-11-04
Fetch WPT test file contents based on wpt.fyi URLs

DanielRyanSmith:
This change adds new functions to interact with the GitHub API to fetch the contents of test files within web-platform-tests. The capacity is available to fetch either the contents of a single test, or the contents of every file in a single directory.

To support this, a new GitHub token secret is added to the Chromestatus.

These functions will be used later to populate prompts to have Gemini evaluate feature test coverage in WPT.

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -210,7 +210,7 @@ def _parse_wpt_fyi_url(url: str) -> str:
   return path
 
 
-def get_file_contents(file_url: str) -> str:
+def get_wpt_file_contents(file_url: str) -> str:
   """
   Fetches the raw text contents of a single file from a wpt.fyi URL.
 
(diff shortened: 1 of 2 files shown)


framework/secrets.py
```diff
@@ -152,6 +151,33 @@ def get_ot_api_key() -> str|None:
...
+    try:
+      with open(f'{settings.ROOT_DIR}/github_token.txt', 'r') as f:
+        settings.GITHUB_TOKEN = f.read().strip()
+        return settings.GITHUB_TOKEN
+    except:
+      logging.info('No key found locally for the GitHub API.')
+      return None
+  else:
+    # If in staging or prod, pull the token from the project secrets.
+    from google.cloud.secretmanager import SecretManagerServiceClient
```
jcscottiii, line 170: I see there's some repetition between this secret and the existing OT API Key secret.

https://github.com/GoogleChrome/chromium-dashboard/blob/4d31bb987f7b6f19d2bd67fbd9c9aa8b30fcd096/framework/secrets.py#L127-L152

We may want to create a secret_manager class or some other common abstraction that handles the dynamic importing depending on the environment only in one place instead of in each secret. Same thing for reading the secret locally.

DanielRyanSmith: Good call! I've added #5648 to track this

PR #5650, opened 2025-11-05
Implement testing gate self certification

jrobbins:
This is progress toward resolving: b/448868070.  This PR defines a new testing gate questionnaire that contains questions that will allow for self-certification.

### DIFF
--- a/client-src/elements/chromedash-survey-questions.ts
+++ b/client-src/elements/chromedash-survey-questions.ts
@@ -194,7 +194,8 @@ export class ChromedashSurveyQuestions extends LitElement {
   renderTestingForm(): TemplateResult {
     return html`
       <div id="questionnaire">
-        Do your tests for this feature cover the following?
+        Does your feature have WPT or other automated tests that cover the
+        following?
         <ol>
           ${this.renderBooleanField(
             'covers_existence',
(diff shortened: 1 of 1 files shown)


api/converters_test.py
```diff
@@ -606,10 +606,16 @@ def test_maxmimal(self, mock_now):
           'is_api_polyfill': False,
           'is_language_polyfill': True,
           'is_same_origin_css': False,
+          'covers_existance': False,
```
DanielRyanSmith: small error, but will need to be changed in a few places:
```suggestion
          'covers_existence': False,
```

DanielRyanSmith (approved):

api/converters_test.py
jrobbins: Fixed

jrobbins:
Thanks for the review on this PR.  I'm going to hang onto it until after I demo to Kadir and Philip again next Wednesday.

jrobbins:
FYI, I got feedback from Kadir and Johnny that:
1. The checkbox text should make it clear that items should be checked if the concern is not relevant.  I will make those changes in this PR. 
1. The old Testing gate questions should still be presented in the case where a full review is needed.  I will create a follow-up PR that branches off of this one to add a dialog that generates comments.

PR #5652, opened 2025-11-05
Fix Chromium file mapping for WebDXFeatures

DanielRyanSmith:
Unfortunately, the OT creation form was using the wrong file contents to check for WebDXFeature use counters due to a minor typo.

This was brought to our attention in a support request, and I went ahead and landed a tainted version on prod to make sure the user could submit the form properly.


DanielRyanSmith:
Notably, this is the first time we know of that a user has attempted to provide a WebDXFeature use counter to track the usage of an origin trial. 🙂

PR #5653, opened 2025-11-05
Add prompts for WPT coverage evaluation

DanielRyanSmith:
This change adds the prompts defined in the WPT coverage evaluation design doc to Chromestatus. They are in Jinja2 format to be populated effectively with the data relevant to the prompt.

### DIFF
--- a/templates/prompts/gap-analysis.html
+++ b/templates/prompts/gap-analysis.html
@@ -45,7 +45,7 @@ Your response must be a formal coverage analysis structured as follows:
 
 For each category, compare the spec requirements to the test coverage and provide a verdict.
 
-### 1. Feature Existence
+#### 1. Feature Existence
 
   * **Spec Requirement:** [Briefly summarize required API surface from `FEATURE_SPEC_SUMMARY`]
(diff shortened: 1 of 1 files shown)


templates/prompts/gap-analysis.html
```diff
@@ -0,0 +1,114 @@
...
+1.  **Overall Verdict:** Start with a brief, direct conclusion (e.g., "VERDICT: Minimum satisfactory coverage has **NOT** been met.")
+2.  **Summary of Analysis:** In 1-2 sentences, explain the *primary reason* for your verdict (e.g., "While feature existence and common use cases are well-tested, the suite fails to cover any error scenarios or the critical invalidation behavior defined in the spec.")
+
+-----
+
+### Detailed Coverage Analysis
+
+For each category, compare the spec requirements to the test coverage and provide a verdict.
+
+### 1. Feature Existence
```
jrobbins: Should this have `###`? That puts it at the same level as the section heading.

jrobbins (approved):
LGTM after considering one comment

templates/prompts/gap-analysis.html
DanielRyanSmith: You're right! I updated these subheaders to use `####`. 🙂

Issue #5648, opened 2025-11-04
Generalize and abstract the app's logic for importing secrets


DanielRyanSmith:
From @jcscottiii 

> I see there's some repetition between this secret and the existing OT API Key secret.
> 
> [chromium-dashboard/framework/secrets.py](https://github.com/GoogleChrome/chromium-dashboard/blob/4d31bb987f7b6f19d2bd67fbd9c9aa8b30fcd096/framework/secrets.py#L127-L152)
> 
> We may want to create a secret_manager class or some other common abstraction that handles the dynamic importing depending on the environment only in one place instead of in each secret. Same thing for reading the secret locally.

PR #5654, opened 2025-11-05
Add new Gemini client

DanielRyanSmith:
This PR adds the [google-genai](https://ai.google.dev/gemini-api/docs) library to Chromestatus, as well as a basic client to initialize the Gemini client, submit prompts, and receive responses from Gemini.

Additionally, a new secret is added to Chromestatus (GEMINI_API_KEY).

### DIFF
--- a/framework/gemini_client.py
+++ b/framework/gemini_client.py
@@ -66,7 +66,7 @@ class GeminiClient:
                  '---------------------------------')
     try:
       response = self.client.models.generate_content(
-        model=self.GEMINI_MODEL,
+        model=GeminiClient.GEMINI_MODEL,
         contents=prompt
       )
 


DanielRyanSmith:
@jcscottiii This is more evidence that we need to act on issue #5648. I'll try and put it in my work queue. 😅

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+
+  Returns:
+      The text response from the model, or an error message if the request failed.
+  """
+  logging.info('--- Sending Prompt to Gemini --- \n'
+               f'{prompt}\n'
+               '---------------------------------')
+  try:
+    response = response = client.models.generate_content(
+      model='gemini-2.5-pro',
```
jcscottiii: Nit: move to constant

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+    if response.text:
+      return response.text
+    else:
+      return 'Error: No text response received from the API.'
+  except Exception as e:
+    logging.error(f'An error occurred while obtaining Gemini response: {e}')
+    raise RuntimeError(f'Failed to get response from Gemini API: {e}') from e
+
+
+def initialize_gemini():
```
jcscottiii: Nit: Typing for return type

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from google import genai
+
+import settings
+
+
+def get_gemini_response(prompt: str, client) -> str:
```
jcscottiii: Nit: type hint for client.

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+  """
+  Sends a prompt to the Gemini 2.5 Pro model and returns the text response.
+
+  Args:
+      prompt: The input prompt string.
+
```
jcscottiii, line 29: Nit: Add a `raises` section to this python docstring for the RuntimeException

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+  Returns:
+      The text response from the model, or an error message if the request failed.
+  """
+  logging.info('--- Sending Prompt to Gemini --- \n'
+               f'{prompt}\n'
+               '---------------------------------')
+  try:
+    response = response = client.models.generate_content(
```
jcscottiii: Question: Do we need to possibly implement any retry mechanism (either here or have the caller handle it)?

framework/gemini_client.py
```diff
@@ -0,0 +1,55 @@
...
+      return response.text
+    else:
+      return 'Error: No text response received from the API.'
+  except Exception as e:
+    logging.error(f'An error occurred while obtaining Gemini response: {e}')
+    raise RuntimeError(f'Failed to get response from Gemini API: {e}') from e
+
+
+def initialize_gemini():
+  """Configure and initialize Gemini."""
```
jcscottiii: Question: Are you planning to initialize the client per request or some other matter? I ask because the client needs to be [closed](https://github.com/googleapis/python-genai/blob/7370d24899f8dfe9babcab247253d15bdd0b1c87/README.md#close-a-client) (if you're not using context managers). 

Also, maybe helpful to add a comment that the client needs to be closed.

framework/gemini_client.py
jcscottiii: Thoughts on creating your own GeminiClient class? And that class could maintain the client instance and `get_gemini_response` could be an instance method (def get_gemini_response(self, prompt:str))

Then you could also handle the destruction of this client in the `__del__` destructor method (see comment below about closing the client)

framework/gemini_client.py
DanielRyanSmith: We probably do - I haven't come to a final conclusion on how to best do that. I will address it in a subsequent PR.

framework/gemini_client.py
DanielRyanSmith: Added!

framework/gemini_client.py
DanielRyanSmith: This is a great idea, and I've changed to implementation to reflect it 🙂

framework/gemini_client.py
DanielRyanSmith: Updated

framework/gemini_client.py
DanielRyanSmith: Added

framework/gemini_client.py
DanielRyanSmith: Added this behavior to the `__del__` destructor as suggested 🙂

DanielRyanSmith:
@jcscottiii this PR has changed a lot, so I'll wait for a re-review before merging 🙂

requirements.txt
```diff
@@ -18,6 +18,7 @@ google-cloud-logging==3.12.1
 google-cloud-secret-manager==2.25.0
 google-cloud-storage==3.4.1
 google-auth==2.42.1
+google-genai==1.48.0
```
jcscottiii, line 21: 1.49.0 was just released! Can update here or let dependabot update it if you want.

framework/gemini_client.py
```diff
@@ -0,0 +1,80 @@
...
+    Raises:
+      RuntimeError: If the API request fails, times out, or if the API
+        succeeds but returns no text content.
+    """
+    logging.info('--- Sending Prompt to Gemini --- \n'
+                 f'{prompt}\n'
+                 '---------------------------------')
+    try:
+      response = self.client.models.generate_content(
+        model=self.GEMINI_MODEL,
```
jcscottiii: nit: self is referring to the instance of the class. Since we are using this as a class constant, you can do this:

```suggestion
        model=GeminiClient.GEMINI_MODEL,
```

jcscottiii (approved):
Thanks for making those updates.

LGTM. With one nit and an optional note about the package version.

requirements.txt
DanielRyanSmith: I guess it's fine to leave this as is, and we'll see if Dependabot picks it up as expected.

PR #5655, opened 2025-11-05
New FeatureEntry fields for AI WPT coverage evaluation

DanielRyanSmith:
Adds the basic required fields, as described in [the design doc](https://docs.google.com/document/d/17ZU8dTKMRIiOme2MGdnyRCoJGx5A4_SgFWmy-eDEccQ/edit?resourcekey=0-Nj7Mx73Sat0-CTjTHU17lA&tab=t.0#heading=h.nkshwck5hr8y). None of these new fields will be mutable by the user, and only one will be surfaced in the UI, so required changes here are pretty sparse.

### DIFF
--- a/client-src/elements/form-field-specs.ts
+++ b/client-src/elements/form-field-specs.ts
@@ -128,7 +128,7 @@ const DOMAIN_NAME_REGEX: string = String.raw`(([A-Za-z0-9\-]+\.)+[A-Za-z]{2,6})`
 const LOCALHOST_REGEX: string = String.raw`(localhost|127\.0\.0\.1)`;
 const DOMAIN_REGEX: string =
   '(' + DOMAIN_NAME_REGEX + '|' + LOCALHOST_REGEX + ')';
-const ALPHANUMERIC_REGEX = String.raw`[-_a-zA-Z0-9,]*`;
+const FINCH_NAMES_REGEX = String.raw`[\-_a-zA-Z0-9,]*`;
 
 const EMAIL_ADDRESS_REGEX: string = USER_REGEX + '@' + DOMAIN_NAME_REGEX;
 const GOOGLE_EMAIL_ADDRESS_REGEX: string = `${USER_REGEX}@google.com`;
@@ -2571,7 +2571,7 @@ export const ALL_FIELDS: Record<string, Field> = {
     attrs: {
       ...TEXT_FIELD_ATTRS,
       placeholder: 'e.g. "StorageBuckets"',
-      pattern: ALPHANUMERIC_REGEX,
+      pattern: FINCH_NAMES_REGEX,
     },
     required: false,
     label: 'Finch feature name',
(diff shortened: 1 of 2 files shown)


internals/core_models.py
```diff
@@ -154,6 +157,11 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
   web_dev_views_notes = ndb.TextProperty()
   other_views_notes = ndb.TextProperty()
 
+  # AI evaluation fields.
+  ai_test_eval_report = ndb.TextProperty(indexed=False)
+  ai_test_eval_in_progress = ndb.BooleanProperty(default=False)
```
jcscottiii: Do we want to use an enumeration instead? COMPLETE | IN_PROGRESS | FAILED

jcscottiii (approved):
LGTM with one question

internals/core_models.py
DanielRyanSmith: Yes! I think that would be useful. I've added a new enum to reflect this.

internals/core_enums.py
```diff
@@ -809,6 +809,12 @@ class SkyhookDashStatus(str, Enum):
     'rollout_plan': ROLLOUT_PLAN_DISPLAYNAMES,
   }
 
+# Valid values for FeatureEntry.ai_test_eval_run_status
+class AITestEvaluationStatus(int, Enum):
+  COMPLETE = 1
```
jcscottiii, line 814: Question for you: Do you plan on backfilling all feature entries with the enum? If so, you may want to have a `NONE = 0` state too.

I think the bigger question is: What different states are you going to check before doing something? Maybe the NONE state can help guide users in a different way, or simplify the difference between showing `Generate` vs `Regenerate`. On the other hand, one could also argue that the lack of `last_request_time` could mean the None state too. Just wondering if you gave it any thought.

internals/core_models.py
```diff
@@ -156,6 +159,11 @@ class FeatureEntry(ndb.Model):  # Copy from Feature
   web_dev_views_notes = ndb.TextProperty()
   other_views_notes = ndb.TextProperty()
 
+  # AI evaluation fields.
+  ai_test_eval_report = ndb.TextProperty(indexed=False)
+  ai_test_eval_run_status = ndb.IntegerProperty()
+  ai_test_eval_last_request_time = ndb.DateTimeProperty()
```
jcscottiii: One more question. Given this is asynchronous, do we want a `ai_test_eval_last_complete_time` field. Or rename this to be like `ai_test_eval_status_timestamp`. Where it corresponds to the timestamp for the current status

internals/core_enums.py
DanielRyanSmith: Yeah, I was thinking of leaving it null by default, with a similar use case you mentioned (Generate vs. Regenerate). True that `last_request_time` could also mean that. The last request time was mostly in place to clear the "in progress" status if the pipeline was terminated inelegantly

internals/core_models.py
DanielRyanSmith: > Or rename this to be like ai_test_eval_status_timestamp. Where it corresponds to the timestamp for the current status

I think this sounds like a really good idea 🙂 - I'll make the adjustment!

PR #5660, opened 2025-11-07
Update WPT test content fetching for better concurrency

DanielRyanSmith:
This PR updates the logic for fetching the contents of WPT test files and files within directories in WPT. The new implementation uses the asyncio library for better concurrency, and fetches all content in two phases. The first phase obtains test file download URLs within directories, as well as download URLs for individual test files concurrently. The second phase takes all test file download URLs and fetches the contents of those files concurrently.

Note: We probably want better retry logic here, but I will address this in a follow-up PR.

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -303,10 +303,6 @@ async def get_mixed_wpt_contents_async(
   """
   Orchestrates concurrent fetching of complete directories and individual WPT file URLs.
 
-  Phase 1: concurrently resolves ALL directory listings and individual file URLs
-           to their raw GitHub download links.
-  Phase 2: concurrently fetches the actual content for every unique download link found.
-
   Args:
     dir_urls: List of WPT directory URLs to scan.
     additional_file_urls: list of individual WPT file URLs.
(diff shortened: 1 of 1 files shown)


framework/utils.py
```diff
@@ -288,66 +239,133 @@ def _fetch_file_content(url: str) -> str | None:
...
+    additional_file_urls: list[str]
+) -> dict[str, str]:
+  """
+  Orchestrates concurrent fetching of complete directories and individual WPT file URLs.
 
-    # Use a ThreadPoolExecutor to fetch files concurrently.
-    with concurrent.futures.ThreadPoolExecutor() as executor:
-      # 'map' maintains the order of the results corresponding to 'files_to_fetch'
-      contents = executor.map(_fetch_file_content, files_to_fetch)
+  Phase 1: concurrently resolves ALL directory listings and individual file URLs
```
jrobbins: I think the comments in the body of the code explain what is happening, so you don't need it here.  Also, a docstring should explain the purpose, not the implementation.

PR #5661, opened 2025-11-07
Add major components for the WPT coverage eval pipeline

DanielRyanSmith:
This PR implements the overall flow of the WPT coverage evaluation pipeline.

- GeminiClient now has methods to support obtaining multiple responses from Gemini concurrently.
- A new file, `gemini_helpers`, contains the rough implementation of the AI evaluation process described in the design.

Similar to #5660, this process needs better handling for errors and retries. I will address this in a subsequent PR alongside #5660.

### DIFF
--- a/framework/gemini_client_test.py
+++ b/framework/gemini_client_test.py
@@ -214,7 +214,7 @@ class GeminiClientTest(testing_config.CustomTestCase):
     """Test that the outer async timeout works correctly."""
     client = gemini_client.GeminiClient()
     # Set a very short timeout for this specific test to make it run fast.
-    client.ASYNC_TIMEOUT_SECONDS = 0.1
+    gemini_client.GeminiClient.ASYNC_TIMEOUT_SECONDS = 0.1
 
     # Mock get_response to sleep longer than the timeout.
     def slow_response(prompt):
--- a/framework/gemini_helpers_test.py
+++ b/framework/gemini_helpers_test.py
@@ -198,7 +198,7 @@ class GeminiHelpersTest(testing_config.CustomTestCase):
         asyncio.run(gemini_helpers.run_wpt_test_eval_pipeline(self.feature))
 
       # Should have logged warnings for the failures
-      self.assertEqual(self.mock_logging.warning.call_count, 2)
+      self.assertEqual(self.mock_logging.error.call_count, 2)
 
   def test_run_pipeline__partial_test_analysis_success(self):
     """Pipeline should continue if at least ONE test analysis succeeds."""
@@ -225,7 +225,7 @@ class GeminiHelpersTest(testing_config.CustomTestCase):
 
       self.assertEqual(self.feature.ai_test_eval_report, 'Final Report')
       # Verify one warning logged for the failure
-      self.mock_logging.warning.assert_called_once()
+      self.mock_logging.error.assert_called_once()
       # Verify the gap analysis prompt only contains the successful one
       # We can check the 'test_analysis' kwarg passed to render_template for the gap analysis
       call_args_list = self.mock_render_template.call_args_list


framework/gemini_client.py
```diff
@@ -78,3 +82,43 @@ def get_response(self, prompt: str) -> str:
...
+    it does not block the main event loop when called.
+
+    Args:
+      prompt: The input prompt string.
+
+    Returns:
+      The text response from the model.
+    """
+    # Use a semaphore to limit how many threads are hitting the API at once.
+    async with self._semaphore:
```
jcscottiii: We should have a timeout so that the process does not indefinitely hold the semaphore (In the event the Gemini API hangs)

jcscottiii (commented):
Higher level question: I saw in the design doc that we plan to use Cloud Tasks.

Instead of managing this locally, we should configure `max_concurrent_dispatches` [1] on the Cloud Task queue itself.

Given our current App Engine infrastructure, this local semaphore will result in a much higher concurrency than intended because it is per-process, not global:
Example: 5 App Engine instances * 8 Gunicorn workers [2] * 20 limit = **800 potential concurrent requests** hitting the Gemini API. 

If you really want to have control of the concurrency in the code, we would need to use something like distributed locking. Which could happen with redis since we already do that. [3]

- [1] https://docs.cloud.google.com/tasks/docs/configuring-queues
- [2] https://cloudlogging.app.goo.gl/x5exnt1yAFgXapgn8
- [3] https://redis.io/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-3-counting-semaphores/

DanielRyanSmith:
> Higher level question: I saw in the design doc that we plan to use Cloud Tasks.
> 
> Instead of managing this locally, we should configure `max_concurrent_dispatches` [1] on the Cloud Task queue itself.
> 
> Given our current App Engine infrastructure, this local semaphore will result in a much higher concurrency than intended because it is per-process, not global: Example: 5 App Engine instances * 8 Gunicorn workers [2] * 20 limit = **800 potential concurrent requests** hitting the Gemini API.
> 
> If you really want to have control of the concurrency in the code, we would need to use something like distributed locking. Which could happen with redis since we already do that. [3]
> 
> * [1] https://docs.cloud.google.com/tasks/docs/configuring-queues
> * [2] https://cloudlogging.app.goo.gl/x5exnt1yAFgXapgn8
> * [3] https://redis.io/ebook/part-2-core-concepts/chapter-6-application-components-in-redis/6-3-counting-semaphores/

Thanks for finding this - that was definitely not the intended outcome of this implementation 🙃 Very good catch! I'm not super familiar with handling this kind of concurrency, so I'm glad you are knowledgeable about this 😊 

I've removed the semaphore here, and went into the GCP project and [updated the max concurrent dispatches in the UI directly](https://pantheon.corp.google.com/cloudtasks?project=cr-status-staging).

framework/gemini_client.py
DanielRyanSmith: I added a timeout for both the async request and the and in the actual request as well. I arbitrarily chose 3 minutes, assuming that we should have a response from the API by that time.

framework/gemini_client.py
```diff
@@ -78,3 +90,46 @@ def get_response(self, prompt: str) -> str:
...
+      TimeoutError: If the request exceeds ASYNC_TIMEOUT_SECONDS.
+    """
+    try:
+      return await asyncio.wait_for(
+        asyncio.to_thread(self.get_response, prompt),
+        timeout=self.ASYNC_TIMEOUT_SECONDS
+      )
+    except asyncio.TimeoutError:
+      logging.error(f'Gemini request timed out after {self.ASYNC_TIMEOUT_SECONDS} seconds.')
+      raise TimeoutError(f'Gemini request timed out after {self.ASYNC_TIMEOUT_SECONDS}s')
```
jcscottiii:
```suggestion
    try:
      return await asyncio.wait_for(
        asyncio.to_thread(self.get_response, prompt),
        timeout=GeminiClient.ASYNC_TIMEOUT_SECONDS
      )
    except asyncio.TimeoutError:
      logging.error(f'Gemini request timed out after {GeminiClient.ASYNC_TIMEOUT_SECONDS} seconds.')
      raise TimeoutError(f'Gemini request timed out after {GeminiClient.ASYNC_TIMEOUT_SECONDS}s')
```

framework/gemini_helpers.py
```diff
@@ -0,0 +1,129 @@
...
+
+  spec_synthesis_response = all_responses.pop()
+  if not isinstance(spec_synthesis_response, str):
+    raise PipelineError(f'Spec synthesis prompt failure: {spec_synthesis_response}')
+
+  test_analysis_responses_formatted = []
+  counter = 1
+  for resp in all_responses:
+    if not isinstance(resp, str):
+      logging.warning(f'Test analysis prompt failure: {resp}')
```
jcscottiii: Any reason not to make this `logging.error`? I'm just wondering about alerting in case we need to slow things down. I think if we log an error, we will get notified, right?

DanielRyanSmith: I'll update this to log an error. I still don't think we'll get notified in the event of a failure, but I plan to update our error handling and retry logic in an upcoming PR.

framework/gemini_client.py
DanielRyanSmith: Whoops, yeah, should have caught this 😅

PR #5668, opened 2025-11-10
New endpoint and Cloud Task to run WPT coverage evaluation pipeline

DanielRyanSmith:
This PR adds a new endpoint, `/api/v0/features/generate-wpt-coverage-evaluation`, that enqueues a new Cloud Task to run the WPT coverage evaluation pipeline.

### DIFF
--- a/framework/gemini_helpers_test.py
+++ b/framework/gemini_helpers_test.py
@@ -12,11 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from internals import core_enums
 import testing_config  # Must be imported before the module under test.
 
 import asyncio
 from unittest import mock
+from internals import core_enums
 from internals.core_models import FeatureEntry
 from framework import gemini_helpers
 


api/wpt_coverage_api.py
```diff
@@ -0,0 +1,48 @@
...
+from internals.core_models import FeatureEntry
+
+
+class WPTCoverageAPI(basehandlers.EntitiesAPIHandler):
+  """Accepts requests related to WPT AI coverage evaluations."""
+
+  def do_post(self, **kwargs):
+    """Enqueue a Cloud Task for generating a WPT coverage evaluation report."""
+    feature_id = self.get_int_param('feature_id')
+    feature = self.get_validated_entity(feature_id, FeatureEntry)
```
jcscottiii, line 29: If two people are on the same feature entry page at the same time. They will both see the generate button. However, we should only allow one to submit the request successfully and return an error status to the other one to let them know that it is in progress already. This will prevent duplicate work.

framework/gemini_helpers.py
```diff
@@ -127,3 +130,36 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
...
+      asyncio.run(run_wpt_test_eval_pipeline(feature))
+    except Exception as e:
+      feature.ai_test_eval_run_status = core_enums.AITestEvaluationStatus.FAILED
+      feature.ai_test_eval_status_timestamp = datetime.now()
+      feature.ai_test_eval_report = (
+        'Web Platform Tests coverage evaluation report failed to generate. '
+        'Try again later.'
+      )
+      feature.put()
+      error_message = f'WPT coverage evaluation report failure: {e}'
```
jcscottiii, line 158: you may want to include the feature Id so it's easy to find in the logs

api/wpt_coverage_api.py
jcscottiii: Actually, I see you just implemented this! Thanks!

DanielRyanSmith: Lol, I realized it as soon as I requested your review (naturally)

framework/gemini_helpers.py
DanielRyanSmith: Added!

api/wpt_coverage_api.py
jcscottiii, line 29: Question: Do we need to check for None here before passing into can_edit_feature?

framework/gemini_helpers.py
```diff
@@ -127,3 +130,36 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
...
+class GenerateWPTCoverageEvalReportHandler(basehandlers.FlaskHandler):
+  """Cloud Task handler for running the AI-powered WPT coverage evaluation."""
+
+  IS_INTERNAL_HANDLER = True
+
+  def process_post_data(self, **kwargs):
+    self.require_task_header()
+
+    feature_id = self.get_int_param('feature_id')
+    feature = self.get_validated_entity(feature_id, FeatureEntry)
```
jcscottiii, line 144: Question: We don't delete feature entries, right? I'm thinking about the time from an entry being enqueued to when it finally gets processed. Is there a possibility we might delete the entry? If so, you may want to check for None

framework/gemini_helpers.py
```diff
@@ -127,3 +130,36 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
...
+
+    logging.info(f'Starting WPT coverage evaluation pipeline for feature {feature_id}')
+
+    try:
+      asyncio.run(run_wpt_test_eval_pipeline(feature))
```
jcscottiii, line 149: This could come as a follow up. But maybe we should handle the updating of the FeatureEntry in datatstore here instead of in run_wpt_test_eval_pipeline? Then it keeps all of the updates of `feature.ai_test_eval*` (success or failure) in this method.



```py
    try:
      result = asyncio.run(run_wpt_test_eval_pipeline(feature))
      feature.ai_test_eval_run_status = COMPLETED
      message = {'message': 'WPT coverage evaluation report generated.'}
      ...
    except Exception as e:
      feature.ai_test_eval_run_status = FAILED
      message = {'message': error_message}
      ...
    feature.put()
    return message
```

api/wpt_coverage_api.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Validate the user has edit permissions.
+    can_edit = permissions.can_edit_feature(
+      self.get_current_user(), feature_id)
+    if not can_edit:
+      self.abort(403, f'User does not have dit access to feature {feature_id}')
+
+    if feature.ai_test_eval_run_status == core_enums.AITestEvaluationStatus.IN_PROGRESS:
```
jcscottiii: You might also want to check the timestamp. I'm concerned about a potential edge case where a feature gets marked IN_PROGRESS, but GenerateWPTCoverageEvalReportHandler fails to clear it for some reason. This would leave the AI information in a stuck state, preventing future regeneration.

While I know the current code likely handles cleanup on failure, adding a timeout would make it more robust against future changes. For example:

```
if feature.ai_test_eval_run_status == core_enums.AITestEvaluationStatus.IN_PROGRESS && (time.now() - feature.ai_test_eval_run_status_timestamp) < 1 day
```

Alternatively (but only if don't have time for the first suggestion), we could add a comment in GenerateWPTCoverageEvalReportHandler warning that failure to properly clear the status can permanently halt AI generation for the feature.

framework/gemini_helpers.py
DanielRyanSmith: We don't delete entities _often_ at least. However, this logic still does check if the feature exists and aborts if it doesn't 🙂

api/wpt_coverage_api.py
DanielRyanSmith: `APIHandler.get_validated_entity` checks that the feature exists and is pulled correctly from the DB, and aborts with a 404 if the entity isn't found.

framework/gemini_helpers.py
DanielRyanSmith: This sounds good to me. Created #5677

api/wpt_coverage_api.py
DanielRyanSmith: This is a great idea, and I instead assumed that if a run has been in progress for over an hour (rather than a day), it's safe to assume that it's hanging and we allow retries.

PR #5679, opened 2025-11-11
Add logic for retries to Gemini API requests [1/4]

DanielRyanSmith:
This change updates the GeminiClient logic to allow 3 tries for both "hard" and "soft" failures.

- If the Gemini API fails to send a response, we will retry the request after a short wait (2 seconds).
- If the Gemini API sends a response but the response denotes a failure, then we will also retry.

The logic for soft failures is prepended to all of our prompts:
```
### **Critical Error Handling Rule**

**Before all other instructions, apply this rule:** If the provided `SPECIFICATION_DOCUMENT` content is empty, nonsensical, or clearly not a valid specification (e.g., an HTML 'Not Found' page, boilerplate text), your **entire** response must be *only* the following string, followed by a brief explanation:
`RESPONSE FAILED: <short explanation of the issue>`

**Examples:**
* `RESPONSE FAILED: Provided specification document was empty.`
* `RESPONSE FAILED: Content is an HTML '404 Not Found' page, not a specification.`
* `RESPONSE FAILED: Content does not appear to be a valid specification document.`
```

This now updates the maximum wait time for async requests to 9 minutes, to allow the maximum (3 minute timeout * 3 tries).

### DIFF
--- a/framework/gemini_client.py
+++ b/framework/gemini_client.py
@@ -18,6 +18,7 @@ import time
 from google import genai
 from google.genai import types
 
+from framework import utils
 import settings
 
 
@@ -58,6 +59,7 @@ class GeminiClient:
     if hasattr(self, 'client') and self.client:
       self.client.close()
 
+  @utils.retry(MAX_RETRIES, delay=RETRY_BACKOFF_SECONDS)
   def get_response(self, prompt: str) -> str:
     """Sends a prompt to the Gemini model and returns the text response.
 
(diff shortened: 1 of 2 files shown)


framework/gemini_client.py
```diff
@@ -66,42 +68,68 @@ def get_response(self, prompt: str) -> str:
...
         )
-      )
 
-      if response.text:
+        if not response.text:
+          raise RuntimeError('No text response received from the API.')
+
+        # Check for the specific failure sentinel in the response.
+        # Using lstrip().upper() for robustness against minor formatting variations.
+        if response.text.lstrip().upper().startswith("RESPONSE FAILED"):
```
jcscottiii: Nit: Is there documentation about "RESPONSE FAILED" being an error to check against? A link would be helpful so others know where to look for that and future errors that are not thrown.

framework/gemini_client.py
```diff
@@ -66,42 +68,68 @@ def get_response(self, prompt: str) -> str:
...
-    try:
-      response = self.client.models.generate_content(
-        model=GeminiClient.GEMINI_MODEL,
-        contents=prompt,
-        config=types.GenerateContentConfig(
-          # timeout is passed to the config using milliseconds.
-          http_options=types.HttpOptions(timeout=GeminiClient.API_TIMEOUT_SECONDS * 1000)
+    last_exception = None
+
+    for attempt in range(1, GeminiClient.MAX_RETRIES + 1):
```
jcscottiii: Question: Any reason not to use the generic utils.retry decorator?

https://github.com/GoogleChrome/chromium-dashboard/blob/98d6e2f4b43b6af2f18f44c578a6046a1add13fa/framework/utils.py#L48-L86

Example of it being used:

https://github.com/GoogleChrome/chromium-dashboard/blob/98d6e2f4b43b6af2f18f44c578a6046a1add13fa/internals/fetchmetrics.py#L46-L60

jcscottiii (approved):
LGTM with two questions

framework/gemini_client.py
DanielRyanSmith: Created #5687 to track this - unfortunately, this is only part of the design doc at the moment, although the prompts have clear guidelines on when this should occur.

framework/gemini_client.py
DanielRyanSmith: I didn't know this was a thing tbh - I've updated the implementation to use this 🙂

PR #5681, opened 2025-11-11
Use test files names when populating Gap Analysis prompt [2/4]

DanielRyanSmith:
I noticed that the responses were referencing "Test 1", "Test 2", etc. This is because the test summaries were populated in the prompt with a counter (e.g. "Test 1 Summary:). This change removes the counter and instead uses the file names so that the gap analysis report is clearer about which tests it is talking about.

### DIFF
--- a/framework/gemini_helpers.py
+++ b/framework/gemini_helpers.py
@@ -113,7 +113,6 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
     if not isinstance(resp, str):
       logging.error(f'Test analysis prompt failure: {resp}')
       continue
-    print(f'\n\n\n\n TEST NAME: {fname}')
     test_analysis_responses_formatted.append(f'Test {fname} summary:\n')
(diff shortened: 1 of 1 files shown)


framework/gemini_helpers.py
```diff
@@ -106,16 +108,15 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
...
-  test_analysis_responses_formatted = []
-  counter = 1
-  for resp in all_responses:
+  test_analysis_responses_formatted: list[str] = []
+  for fname, resp in zip(file_names, all_responses):
     if not isinstance(resp, str):
       logging.error(f'Test analysis prompt failure: {resp}')
       continue
-    test_analysis_responses_formatted.append(f'Test {counter} summary:\n')
+    print(f'\n\n\n\n TEST NAME: {fname}')
```
DanielRyanSmith: Oops, I'll fix this when I get home 😬

jcscottiii (approved):
LGTM

PR #5682, opened 2025-11-11
Make WPT coverage analysis visible for Googlers and Chromium users

DanielRyanSmith:
This will allow us to get some feedback on this feature without initially making it visible to non-Googlers.


DanielRyanSmith:
I think it's best to not land this change until December, when I can babysit things a little more. 🙂

jrobbins:
Did you mean to have this merging into your other branch instead of main?

DanielRyanSmith:
> Did you mean to have this merging into your other branch instead of main?

These changes are built off of other changes that haven't yet landed, so updating the merge target makes it so the PR view is isolated to only changes in this PR (for hopefully easier review) 🙂 

Once the other PRs get merged in, the merge target here will change to main as it should

DanielRyanSmith:
Looking at the number of Googlers are on the roadmap page for m145 using their Chromium accounts, I'm wondering if it would be a better choice to make this available to Chromium email accounts as well. We have items in place to avoid  overuse here (e.g. the 30 minute cooldown to regenerate the report, etc.). It might increase the velocity of feedback we get for this new feature 🤔 

Edit: It is probably best to address #5922 before doing this.

DanielRyanSmith:
This should be okay to review and land. Some polishing PRs are in flight, but will also be deployed at the same time.

DanielRyanSmith:
(This needs Playwright snapshots update, which are being generated now)

client-src/elements/chromedash-feature-detail.ts
```diff
@@ -322,20 +322,26 @@ export class ChromedashFeatureDetail extends LitElement {
...
     `;
-    // TODO(DanielRyanSmith): Add this to the main view when all aspects of
-    // the WPT coverage evaluation pipeline have been implemented
-    // const wptEvalButton = html` <chromedash-wpt-eval-button
-    //   .featureId=${this.feature.id}
-    // ></chromedash-wpt-eval-button>`;
+    let wptEvalButton = html`${nothing}`;
+    // For now, the WPT coverage evaluation page is only visible to Googlers and Chromium users.
+    if (
+      this.user?.email?.endsWith('@google.com') ||
```
jcscottiii, line 328: Question: We also enforce this in the backend too?

client-src/elements/chromedash-feature-detail.ts
```diff
@@ -322,20 +322,26 @@ export class ChromedashFeatureDetail extends LitElement {
...
+      this.user?.email?.endsWith('@chromium.org')
+    ) {
+      if (
+        this.canEdit &&
+        // Enterprise features don't provide spec URLs or Web Platform Tests info.
+        this.feature.feature_type_int !==
+          FEATURE_TYPES.FEATURE_TYPE_ENTERPRISE_ID[0]
+      ) {
+        wptEvalButton = html` <chromedash-wpt-eval-button
```
jcscottiii, line 337: Completely optional and does not block this PR: This is a cool new feature. Maybe add something like a tooltip is already open to highlight this.

client-src/elements/chromedash-feature-detail.ts
DanielRyanSmith: No we do not. I can add a PR enforcing that. 🙂

client-src/elements/chromedash-feature-detail.ts
jcscottiii: Or you could add like the Gemini logo next to it or part of the eval button component.

DanielRyanSmith: > Or you could add like the Gemini logo next to it or part of the eval button component.

The Gemini logo is added to the button already 🙂 I'm glad to hear we had the same idea!

PR #5683, opened 2025-11-11
Enforce a 30 minute cooldown on WPT pipeline requests [3/4]

DanielRyanSmith:
This is to avoid any risk of hitting rate limits for the GitHub and Gemini APIs.

Also, a caveat is added to the "Before you begin" section stating that a maximum of 50 individual tests are accepted for WPT coverage evaluation runs. This is not yet enforced, but will be added in a subsequent PR.

### DIFF
--- a/api/wpt_coverage_api.py
+++ b/api/wpt_coverage_api.py
@@ -64,6 +64,8 @@ class WPTCoverageAPI(basehandlers.EntitiesAPIHandler):
       # (which can happen if the condition evaluated true milliseconds ago but time passed)
       retry_after_seconds = int(max(0, retry_after.total_seconds()))
       error_resp = {'error': msg}
+      # TODO(#5688): Create a BaseHandler.abort_with_headers helper method
+      # and refactor these lines to use that new method.
       resp = flask.make_response(flask.jsonify(error_resp), 409)
       resp.headers['Retry-After'] = retry_after_seconds
       flask.abort(resp)


client-src/elements/chromedash-wpt-eval-page_test.ts
```diff
@@ -318,4 +318,83 @@ describe('chromedash-wpt-eval-page', () => {
       expect(el.shadowRoot!.querySelector('sl-skeleton')).to.not.exist;
     });
   });
+
+  // NEW: Cooldown Logic Tests
```
jcscottiii, line 322:
```suggestion
```

:)

client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -10,6 +10,9 @@ import {AITestEvaluationStatus} from './form-field-enums.js';
 // Matches http or https, followed by wpt.fyi/results, followed by any non-whitespace and non-question mark characters.
 const WPT_RESULTS_REGEX = /(https?:\/\/wpt\.fyi\/results[^\s?]+)/g;
 
+// 30 minute cooldown for regenerating the evaluation report.
+const COOLDOWN_MS = 30 * 60 * 1000;
```
jcscottiii, line 14: This decouples the cool down used in the backend from the one in the frontend. For example:
- User loads version A of the frontend
- We do a deployment that changes the cooldown
- User gets the 409, and expects 30 mins still.

jcscottiii: Instead, maybe we should send the timestamp back. Example: GitHub sets the [retry-after header](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit).

jrobbins:
Will a per-feature cooldown work? What are our site-wide quotas for these Github and Gemini APIs?

Brainstorm: What if we combined the prompts and URLs into one big prompt that the user could copy from our site and paste into Gemini or another LLM?  That would use their personal quota rather than our site quota, and it would allow them to tweak the prompt and ask follow-up questions.  Maybe keep the UI you have, but also add a "Copy prompt" button as an alternative.

client-src/elements/chromedash-wpt-eval-page_test.ts
DanielRyanSmith: 😳

DanielRyanSmith:
> Will a per-feature cooldown work? What are our site-wide quotas for these Github and Gemini APIs?
>
> Brainstorm: What if we combined the prompts and URLs into one big prompt that the user could copy from our site and paste into Gemini or another LLM? That would use their personal quota rather than our site quota, and it would allow them to tweak the prompt and ask follow-up questions. Maybe keep the UI you have, but also add a "Copy prompt" button as an alternative.

Our app's rate limits are in line with a [tier 1 project](https://ai.google.dev/gemini-api/docs/rate-limits#current-rate-limits). I do believe a per-feature cool down will keep us in line here, especially if we consider [analyzing multiple tests in a single prompt](https://github.com/GoogleChrome/chromium-dashboard/issues/5685). 🙂 

It's difficult to try and give the user something to copy, because it's a 3-prompt pipeline, and the test-analysis prompt can be split into [N number of prompts](https://docs.google.com/document/d/17ZU8dTKMRIiOme2MGdnyRCoJGx5A4_SgFWmy-eDEccQ/edit?resourcekey=0-Nj7Mx73Sat0-CTjTHU17lA&tab=t.0#heading=h.tphg2uux54fk). This is done to avoid overflowing a single prompt with too much context.

If we were to generate a single prompt for the user, it would be structured differently from the feedback that is generated now. It's possible, but it would need a fundamental overhaul in how we're obtaining these analyses. This might be an approach worth looking at after we've gotten deeper feedback from users. That being said, if that is something we want to pursue, I think it could be worth exploring.

client-src/elements/chromedash-wpt-eval-page.ts
DanielRyanSmith: > Instead, maybe we should send the timestamp back. Example: GitHub sets the [retry-after header](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit).

What if the user reloads the page in this scenario? The cooldown isn't displayed to the user unless they refresh the page. Maybe we should go off of `feature.ai_test_eval_status_timestamp` instead?

DanielRyanSmith: Actually, I want to clarify this more - should we be sending a separate timestamp to the user for when they can retry their request? For example, should the feature JSON have a new field similar to `feature.ai_test_eval_retry_after`, which is populated with `feature.ai_test_eval_status_timestamp + timedelta(COOLDOWN_CONSTANT)`?

api/wpt_coverage_api.py
```diff
@@ -35,14 +35,29 @@ def do_post(self, **kwargs):
...
+      # Assume that a request that is in progress for over an hour is hanging.
+      and last_status_time + one_hour > datetime.now())
+
+    # Be more generous with the cooldown so we don't block the UI sending a
+    # request accidentally.
+    twenty_nine_minutes = timedelta(minutes=29)
+    on_cooldown = (
+      feature.ai_test_eval_run_status == core_enums.AITestEvaluationStatus.COMPLETE
+      and last_status_time
+      # Assume that a request that is in progress for over an hour is hanging.
```
jcscottiii: This comment needs to be corrected/removed. It's a copy and paste from above.

client-src/elements/chromedash-wpt-eval-page.ts
jcscottiii: In this case, no. If we ever change the timedelta of 29mins in the future, we could be potentially telling clients to wait too long or wait too soon. 


If we really wanted to, we would probably want to add a new COOLDOWN state to the enum to prevent different mini conditions of COMPLETE.

I think the simpler thing would be to just add a header before you abort with the 409: self.response.headers['Retry-After'] = duration

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After

client-src/elements/chromedash-wpt-eval-page_test.ts
jcscottiii: No worries. I smile because the same thing has happened to me before lol.

PR #5689, opened 2025-11-13
Pull GEMINI_API_KEY and GITHUB_TOKEN from settings

DanielRyanSmith:
This PR ensures that the GEMINI_API_KEY and GITHUB_TOKEN are populated by the GCP Secret Manager the first time they are referenced. (I missed this after exclusively testing locally 😅)

### DIFF
--- a/framework/gemini_client.py
+++ b/framework/gemini_client.py
@@ -50,6 +50,9 @@ class GeminiClient:
         API key issue or other unexpected error.
     """
     api_key = settings.GEMINI_API_KEY
+    # This error should only be raised locally. The application will already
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
I deployed a tainted version to staging and confirmed that this works as intended. 🙂

jcscottiii:
> If these secrets are mandatory, we should probably populate them at start up time.
> 
> Maybe around here:
> 
> https://github.com/GoogleChrome/chromium-dashboard/blob/7b1e6e014e99240e079c947df76884b1d979fe36/main.py#L78-L89

Also, open to other files that run at start up time. Just so that we fail fast if we are missing any dependencies.

Issue #5690, opened 2025-11-13
Some wpt.fyi results URLs don't map to WPT repo contents properly


DanielRyanSmith:
This needs to be figured out to properly fetch test data correctly.

This feature: https://chromestatus.com/feature/6295810820145152
lists a wpt.fyi result URL that is not mapped directly to a WPT file: https://wpt.fyi/results/url/IdnaTestV2.window.html

This test is actually a JS file located at: https://github.com/web-platform-tests/wpt/blob/master/url/IdnaTestV2.window.js

Tests like these will cause the WPT coverage evaluation pipeline to fail because we're not able to properly pull the file contents of the test.

DanielRyanSmith:
I haven't found an example of this problem that is not solved by the (somewhat ungraceful) fix landed in #5693.

The long-term fix for this is [described here](https://github.com/GoogleChrome/chromium-dashboard/pull/5693#pullrequestreview-3460057614)

PR #5693, opened 2025-11-13
Add ".js" fallback when failing to fetch ".html" test file contents

DanielRyanSmith:
Partly addresses #5690 

Most of the results with the behavior described in the issue have a counterpart file that is the same name and directory as the test file results, but with a ".js" file extension. This change updates the logic to have a fallback case to attempt fetching the same file with a js extension when fetching an html file fails.

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -262,7 +262,19 @@ def _fetch_file_content(url: str) -> str | None:
     return response.text
   except requests.exceptions.RequestException as e:
     logging.error(f'Warning: Failed to fetch {url}: {e}')
-    return None
+  # Some tests results are represented with ".html" file extensions, but
+  # their test contents are located in an equivalent ".js" file.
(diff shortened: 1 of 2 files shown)


jcscottiii (approved):
LGTM. But long term I think we should actually invest in reading the MANIFEST.json file.

Looking at your linked bug, I started to wonder that `./wpt serve` knows how to go from the .js file to .HTML. So it has to be able to go backwards too.


I used grep on the MANIFEST file and it indeed has that information.
image

DanielRyanSmith:
> LGTM. But long term I think we should actually invest in reading the MANIFEST.json file.

Good call. I'll update my description to make sure the issue stays open and we can track this later. @foolip pointed out this has been a longstanding bug in wpt.fyi: https://github.com/web-platform-tests/wpt.fyi/issues/1476

PR #5706, opened 2025-11-20
Add more enterprise release notes reviewers.

jrobbins:
This should address comment 10 on b/446217660.  Basically, the enterprise team just wants more release notes reviewers.  This list has some overlap with their review gate reviewers, but it is not the same list.


jcscottiii:
Long term question: Does this list update frequently? If so, maybe we should have it in a datastore table? If not, I can see keeping it here

jrobbins:
> Long term question: Does this list update frequently? If so, maybe we should have it in a datastore table? If not, I can see keeping it here

It has not been frequent.

PR #5722, opened 2025-12-01
Update WPT coverage evaluation prerequisites checklist

DanielRyanSmith:
This change updates the "Prerequisite checklist" section of the WPT coverage evaluation page to contain more information.

- All fields now display their values if they have been provided for easier referencing.
- "Feature name" and "Feature summary" requirements have been added. Although these are already required for a feature, adding these makes it clear to the user that these aspects of the feature will be used in evaluating test coverage.

### Before

Screenshot 2025-12-01 5 59 01 PM

### After

Screenshot 2025-12-01 5 57 28 PM

### DIFF
--- a/client-src/elements/chromedash-wpt-eval-page.ts
+++ b/client-src/elements/chromedash-wpt-eval-page.ts
@@ -458,6 +458,23 @@ export class ChromedashWPTEvalPage extends LitElement {
     `;
   }
 
+  /**
+   * Renders pre-formatted text within a div, applying `prettier-ignore`
+   * to prevent unwanted whitespace from being introduced by Prettier's
+   * formatting when `white-space: pre-wrap` is used.
+   * @param content The text content to render.
+   * @returns A TemplateResult containing the pre-formatted text, or nothing if content is undefined.
+   */
+  private _renderPreformattedText(
+    content: string | undefined
+  ): TemplateResult | typeof nothing {
+    if (!content) {
+      return nothing;
+    }
+    // prettier-ignore
+    return html`<div class="url-list prewrap">${content}</div>`;
+  }
+
   renderRequirementsChecks(): TemplateResult {
     if (!this.feature) {
       return html`${nothing}`;
@@ -480,8 +497,7 @@ export class ChromedashWPTEvalPage extends LitElement {
           </div>
           ${this.renderRequirementItem(true, 'Feature summary', 'id_summary')}
           <div class="url-list-container">
-            <!-- prettier-ignore -->
-            <div class="url-list prewrap">${this.feature.summary}</div>
+            ${this._renderPreformattedText(this.feature.summary)}
           </div>
           ${this.renderRequirementItem(hasSpecLink, 'Spec URL', 'id_spec_link')}
           ${hasSpecLink
@@ -505,8 +521,7 @@ export class ChromedashWPTEvalPage extends LitElement {
           ${hasWptDescr
             ? html`
                 <div class="url-list-container">
-                  <!-- prettier-ignore -->
-                  <div class="url-list prewrap">${this.feature.wpt_descr}</div>
+                  ${this._renderPreformattedText(this.feature.wpt_descr)}
                 </div>
               `
             : nothing}


client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -466,12 +469,45 @@ export class ChromedashWPTEvalPage extends LitElement {
...
         <h2>Prerequisites Checklist</h2>
         <div class="requirements-list">
+          <!-- Name and summary are assumed to always be filled -->
+          ${this.renderRequirementItem(true, 'Feature name', 'id_name')}
+          <div class="url-list-container">
+            <div class="url-list">${this.feature.name}</div>
+          </div>
+          ${this.renderRequirementItem(true, 'Feature summary', 'id_summary')}
+          <div class="url-list-container">
+            <!-- prettier-ignore -->
```
jcscottiii: Do we need this ignore?
```suggestion
```

client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -466,12 +469,45 @@ export class ChromedashWPTEvalPage extends LitElement {
...
+            : nothing}
           ${this.renderRequirementItem(
             hasWptDescr,
             'WPT description',
             'id_wpt_descr'
           )}
+          ${hasWptDescr
+            ? html`
+                <div class="url-list-container">
+                  <!-- prettier-ignore -->
```
jcscottiii: Do we need this ignore?
```suggestion
```

client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -466,12 +469,45 @@ export class ChromedashWPTEvalPage extends LitElement {
...
+                  <div class="url-list" style="white-space: pre-wrap;"
```
jcscottiii: Nit: Move inline style to the styles method.

client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -466,12 +469,45 @@ export class ChromedashWPTEvalPage extends LitElement {
...
         <div class="requirements-list">
+          <!-- Name and summary are assumed to always be filled -->
+          ${this.renderRequirementItem(true, 'Feature name', 'id_name')}
+          <div class="url-list-container">
+            <div class="url-list">${this.feature.name}</div>
+          </div>
+          ${this.renderRequirementItem(true, 'Feature summary', 'id_summary')}
+          <div class="url-list-container">
+            <!-- prettier-ignore -->
+            <div class="url-list" style="white-space: pre-wrap;"
```
jcscottiii: Nit: Move inline style to the styles method.

client-src/elements/chromedash-wpt-eval-page.ts
DanielRyanSmith: This ignore is required because the linter turns this:
```html
<div class="url-list" style="white-space: pre-wrap;"
  >${this.feature.wpt_descr}
</div>
```

Into this:
```html
  <div class="url-list" style="white-space: pre-wrap;">
    ${this.feature.wpt_descr}
  </div>
```

Which is subtle, but adds unintended whitespace to the beginning of each section

client-src/elements/chromedash-wpt-eval-page.ts
DanielRyanSmith: Done!

client-src/elements/chromedash-wpt-eval-page.ts
DanielRyanSmith: Done!

client-src/elements/chromedash-wpt-eval-page.ts
jcscottiii: What about creating a helper method that handles this then:

```js
  /**
   * Renders pre-formatted text within a div, applying `prettier-ignore`
   * to prevent unwanted whitespace from being introduced by Prettier's
   * formatting when `white-space: pre-wrap` is used.
   * @param content The text content to render.
   * @returns A TemplateResult containing the pre-formatted text, or nothing if content is undefined.
   */
  private _renderPreformattedText(
    content: string | undefined
  ): TemplateResult | typeof nothing {
    if (!content) {
      return nothing;
    }
    // prettier-ignore
    return html`<div class="url-list prewrap">${content}</div>`;
  }
```

Then the two places that have `prettier-ignore` can use this instead.

DanielRyanSmith: This seems fine, but I don't think it fixes the problem, since we'd still have the "prettier-ignore" comment. Is the intention to avoid the comment as an HTML comment?

jcscottiii: The idea was to centralize the use of prettier-ignore comment. 

For me, it was non obvious as to why we needed it at first glance. (Then I understood it when I read up on the white-space CSS rule you were using). You could add additional comments to each usage of the prettier-ignore. But a central helper method does that for you.

Another idea was to have the helper method not use the white-space CSS rule and do something like this in the helper:

```ts
private _renderPreformattedText(content:string)
...
const lines = content.split('\n'); 
return html`${lines.map((line, i) =>
  i === lines.length - 1 ? line : html`${line}<br />`
)}`;
```

But that seemed like overkill. Especially when the browser can handle this with the CSS rule. So having a central helper that has the ignore line with more context helps out with future people reading the code and its intention.

DanielRyanSmith: > The idea was to centralize the use of prettier-ignore comment.
> 
> For me, it was non obvious as to why we needed it at first glance. 

Ah! That makes sense now. Yes, I think this is a good idea to make it clear why we need the prettier-ignore comment. Thanks for clarifying for me 😊

jcscottiii: Also, with this type of helper, we could eventually move this into https://github.com/GoogleChrome/chromium-dashboard/blob/main/client-src/elements/utils.ts if we find ourselves having to do this often.

PR #5724, opened 2025-12-01
Change button location for generating WPT coverage reports

DanielRyanSmith:
This change moves the location of the button for generating reports under the WPT coverage report display, since it's less useful once the report has been generated. Also, the button now changes text and color if a report has already been generated to properly convey that the user will be overwriting the existing report if they generate a new one.

New button appearance:
Screenshot 2025-12-02 1 13 54 AM

### DIFF
--- a/client-src/elements/chromedash-wpt-eval-page.ts
+++ b/client-src/elements/chromedash-wpt-eval-page.ts
@@ -541,13 +541,15 @@ export class ChromedashWPTEvalPage extends LitElement {
           : nothing}
 
         <sl-button
-          variant="${(this.feature.ai_test_eval_report) ? 'danger' : 'primary'}"
+          variant="${this.feature.ai_test_eval_report ? 'danger' : 'primary'}"
           size="large"
           class="generate-button"
           ?disabled=${!this.isRequirementsFulfilled || isCooldownActive}
           @click=${this.handleGenerateClick}
         >
-          ${(this.feature.ai_test_eval_report) ? 'Discard this report and reevaluate test coverage' : 'Evaluate test coverage'}
+          ${this.feature.ai_test_eval_report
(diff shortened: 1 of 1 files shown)


client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -541,13 +541,15 @@ export class ChromedashWPTEvalPage extends LitElement {
           : nothing}
 
         <sl-button
-          variant="primary"
+          variant="${this.feature.ai_test_eval_report ? 'danger' : 'primary'}"
           size="large"
           class="generate-button"
           ?disabled=${!this.isRequirementsFulfilled || isCooldownActive}
```
jcscottiii, line 547: Question: Do we want to disable the delete button if the report is present but have changed the artifacts to where  isRequirementsFulfilled is now false. (I don't think isCooldownActive will ever be true if ai_test_eval_report is present). If so, we can leave this as-is. But otherwise, you may want to do something like this and have a helper to render different types of buttons.

```ts
${this.feature.ai_test_eval_report === undefined ? this.renderGenerateEvalButton(): this.renderDeleteEvalButton()}
```

I ask this question because:
1. I'm wondering if we are unintentionally merging conditions on the button between when the report is present vs the report is not present. (Example: Should we consider the cooldown if the report is present)
2. You are modifying two properties on the same element based on if the report is present.

DanielRyanSmith: We won't have separate functionality to delete the existing report, only to generate a report, which essentially overwrites the previous report once the process is finished. So this change is just cosmetic - we just want to make it clear to users that when they press the button to generate a report a second time, the previous report will be gone after that.

The cooldown will still be important in this case, because we don't want users to be able to spam report generation repeatedly, as that could be costly. 😅

PR #5725, opened 2025-12-01
Increase Gunicorn timeout for local dev

DanielRyanSmith:
We were using the default 30 second timeouts here, which causes failures in WPT coverage evaluation pipeline. This increase to 2 minutes fixes this.

### DIFF
--- a/scripts/start_server.sh
+++ b/scripts/start_server.sh
@@ -11,7 +11,7 @@ export GAE_ENV='localdev'
 export DATASTORE_EMULATOR_HOST=${DATASTORE_EMULATOR_HOST:-'localhost:15606'}
 
 
-gunicorn --bind :7777 --workers 4 main:app
+gunicorn --bind :7777 --workers 4 main:app --timeout 120
 
 
 # TODO(jrobbins): Consider switching back to dev_appserver when


DanielRyanSmith:
> LGTM. But one question: Have you checked the performance on staging/prod?
> 
> gunicorn's default of 30 seconds may be used [there too](https://pantheon.corp.google.com/logs/query;cursorTimestamp=2025-12-01T09:00:21.645922Z;duration=P1D;query=resource.type%3D%22gae_app%22%0Aresource.labels.module_id%3D%22default%22%0Aresource.labels.version_id%3D%2228ee68a%22%0ASEARCH%2528%22gunicorn%22%2529%0Atimestamp%3D%222025-12-01T04:00:45.235525Z%22%0AinsertId%3D%22692d12ed0003980553b48e54%22?project=cr-status-staging) You may also need to change the timeouts in app engine too.

Hmm, I have only noticed this being a problem locally, and have had no issues on staging or prod. I'll have to look into where to make the change for this, since it's better that we be explicit about this everywhere.

On a similar note, this was only intended to deal with local issues, but if we're fixing things to correspond with staging and prod, it should be noted that [the maximum timeout with Gemini requests after retries is 9 minutes currently](https://github.com/GoogleChrome/chromium-dashboard/blob/bfb76b0323eaa77f30b8569c63944390b20203cd/framework/gemini_client.py#L42-L43). I guess this means it would be prudent to increase the timeout for app engine to same number?

jrobbins:
This page says that the appengine requests timeout is 10 minutes by defauilt
https://docs.cloud.google.com/appengine/docs/standard/how-instances-are-managed#timeout

jrobbins:
And, I think the gunicorn on appengine defaults to a long timeout
Code search for file:gunicorn.py includes the line
`timeout = 86400`
which is 24 hours.

DanielRyanSmith:
> This page says that the appengine requests timeout is 10 minutes by defauilt
https://docs.cloud.google.com/appengine/docs/standard/how-instances-are-managed#timeout

> And, I think the gunicorn on appengine defaults to a long timeout
Code search for file:gunicorn.py includes the line
timeout = 86400
which is 24 hours.

Excellent! Thanks for finding that information. I set the timeout at 2 minutes to be conservative, but if we have problems with this timeout number, it seems safe to increase it further in the future 🙂

PR #5729, opened 2025-12-01
Avoid errors when JSON serializing a datetime.

### LINKED ISSUE #5727: Exception "Object of type DatetimeWithNanoseconds is not JSON serializable" on prod
We got an alert on the Error Reporting page
https://pantheon.corp.google.com/errors/detail/CIjulsmiz8i-_QE;locations=global;time=P30D?project=cr-status&utm_source=error-reporting-notification&utm_medium=email&utm_content=new-error

jrobbins:
This should resolve #5727.

Passing a `datetime.datetime` object to `json.dump()` will cause an exception because "DateTime objects are not JSON serializable".  That is worked around easily by passing `default=str`.  We were already doing this in basehandlers.py, but here I have added it to cloud_task_helpers.py.

### DIFF
--- a/framework/cloud_tasks_helpers.py
+++ b/framework/cloud_tasks_helpers.py
@@ -76,6 +76,9 @@ def _get_client():
 
 
 def _make_task(handler_path, task_params):
+  # Note: `default=str` is used to handle datetimes. Be cautious, as
+  # this will also stringify any other non-serializable objects
+  # passed in task_params.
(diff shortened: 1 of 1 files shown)


jrobbins:
I added the suggested comment.  I checked all callers and didn't see anything passing in any python class instances.  If there were such a case, I believe that it would have already caused exceptions.

PR #5735, opened 2025-12-02
Test analysis prompt advises to fetch dependency files

### LINKED ISSUE #5691: Adjust the test analysis prompts to fetch resource/support files
I had some luck in initial testing adding this additional text to the test analysis prompt:

"If the test file lists any resource files that it uses for testing, fetch them from raw.githubusercontent.com and use their contents in your analysis as well."

I also added a clear URL to the prompt when presenting the test file contents, so that Gemini has an idea what the resource file URLs should look like.
"Here is the test file, fetched from https://raw.githubusercontent.com/web-platform-tests/wpt/refs/heads/master/url/IdnaTestV2.window.js:"

This could increase the accuracy of our test analysis summaries, which would increase the utility of the WPT coverage evaluation reports.

DanielRyanSmith:
Fixes #5691 

This change updates the test analysis prompt to advise Gemini to fetch any dependency files that are in the same repository to obtain a better understanding of the test contents.

Additionally, some functions have been updated to pass Path objects so that the test analysis prompt can be populated with the full path of the test, so that it is easier for Gemini to deduce where the dependency files are located in the repo in the event of relative paths.

### DIFF
--- a/templates/prompts/test-analysis.html
+++ b/templates/prompts/test-analysis.html
@@ -22,6 +22,7 @@ Do not output *any* other text or formatting if this rule is triggered.
 Scan the test file for references to other files within the same repository (e.g., `<script src="resources/test-helper.js">`, `import ... from './utils.js'`, or reference files listed in metadata).
 
 * **Action:** If such files exist, you must **fetch** them from `raw.githubusercontent.com` (using the context of the current file's location).
+* **Constraint:** * If a dependency file cannot be fetched, **do not** hallucinate or guess the file contents based on the file name. Instead, disregard the contents of that dependency file.
 * **Integration:** You must read the contents of these fetched files and incorporate their logic into your final analysis. Do not analyze the main test file in isolation if the assertions rely on external helper scripts.
 
 ---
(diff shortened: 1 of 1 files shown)


templates/prompts/test-analysis.html
```diff
@@ -16,10 +16,20 @@
 
 ---
 
+### **Dependency Resolution**
+
+**Before analyzing the code, perform this step:**
+Scan the test file for references to other files within the same repository (e.g., `<script src="resources/test-helper.js">`, `import ... from './utils.js'`, or reference files listed in metadata).
+
+* **Action:** If such files exist, you must **fetch** them from `raw.githubusercontent.com` (using the context of the current file's location).
```
jcscottiii, line 24: Have you tried this multiple times? Not with GitHub specifically, I tried to have AI do something like this in the prompt and if it cannot find it, it would imagine something instead.

DanielRyanSmith: I have tried this with a few features, but unfortunately, I cannot 100% confirm that this will not cause any unintended side-effects. Do you have a suggestion for how to best rule out this possibility?

jcscottiii: Let's try your new commit and iterate on that.

PR #5740, opened 2025-12-03
Add a dialog for the old testing gate questions

jrobbins:
When the feature owner requests a full review (either because no all the min-bar checkboxes are set or because the user opted for a full review), then show an "Additional questions" dialog box that helps the user compose a comment to be posted to the gate.  When the feature owner presses "Generate comment and request review" it will post a comment as their account to the gate and request the review.

### DIFF
--- a/client-src/elements/chromedash-secondary-survey-dialog.ts
+++ b/client-src/elements/chromedash-secondary-survey-dialog.ts
@@ -5,7 +5,8 @@ import {customElement, property} from 'lit/decorators.js';
 import {GateDict} from './chromedash-gate-chip.js';
 import {GATE_TYPES} from './form-field-enums.js';
 
-let secondarySurveyDialogEl;
+let secondarySurveyDialogEl: ChromedashSecondarySurveyDialog | undefined =
+  undefined;
 
 export function isTestingGate(gate: {gate_type: number}): boolean {
   return (
@@ -34,7 +35,7 @@ async function openSecondarySurveyDialog(gate, resolve) {
   if (!secondarySurveyDialogEl) {
     secondarySurveyDialogEl = document.createElement(
       'chromedash-secondary-survey-dialog'
-    );
+    ) as ChromedashSecondarySurveyDialog;
     document.body.appendChild(secondarySurveyDialogEl);
   }
   secondarySurveyDialogEl.gate = gate;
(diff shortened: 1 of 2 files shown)


client-src/elements/chromedash-secondary-survey-dialog.ts
```diff
@@ -0,0 +1,214 @@
...
+    document.body.appendChild(secondarySurveyDialogEl);
+  }
+  secondarySurveyDialogEl.gate = gate;
+  secondarySurveyDialogEl.resolve = resolve;
+  await secondarySurveyDialogEl.updateComplete;
+  secondarySurveyDialogEl.show();
+}
+
+@customElement('chromedash-secondary-survey-dialog')
+export class ChromedashSecondaySurveyDialog extends LitElement {
```
jcscottiii: Nit: typo
```suggestion
export class ChromedashSecondarySurveyDialog extends LitElement {
```

client-src/elements/chromedash-secondary-survey-dialog.ts
```diff
@@ -0,0 +1,214 @@
+import {LitElement, css, html, nothing} from 'lit';
+import {SHARED_STYLES} from '../css/shared-css.js';
+import {SlRadioGroup, SlTextarea} from '@shoelace-style/shoelace';
+import {customElement, property} from 'lit/decorators.js';
+import {GateDict} from './chromedash-gate-chip.js';
+import {GATE_TYPES} from './form-field-enums.js';
+
+let secondarySurveyDialogEl;
```
jcscottiii: Quick question regarding the lifecycle of the chromedash-secondary-survey-dialog. I noticed that once it's created and appended to the <body>, it's hidden but not explicitly removed from the DOM, and the global secondarySurveyDialogEl isn't reset.

Was this intentional? For example, was it to preserve user input if the dialog were to be reopened? My concern is that keeping it in the DOM could lead to a minor memory leak over long sessions.

jcscottiii: Also, could you add typing to the variable? Like `let secondarySurveyDialogEl: ChromedaySecondarySurveyDialog | undefined = undefined;`

jrobbins: I added the type info.

Regarding resetting the dialog, I mainly just followed existing examples that did not account for that.  However, I did notice that the dialog retains the values that the user entered.  I think it is unlikely to matter in practice because most feature owners will not be requesting review on the testing gate of multiple features in the same session.  However, in cases where they do, it seems potentially handy.  I don't think that it can be much of a memory leak because there can be at most one instance of each type of dialog.

PR #5741, opened 2025-12-03
Various fixes for the spec synthesis prompt

DanielRyanSmith:
There were a number of issues with the spec synthesis prompt and, while it was effective, had some fundamental flaws that needed addressing.

1. The prompt stated that Gemini will receive a SPECIFICATION_DOCUMENT value that contained the contents of the feature's spec, but instead the prompt was populated with a URL that linked to the specification contents. While this seemed to not cause too much issue with the results, we obviously want to be clear with the data we're providing to Gemini. Now, the prompt is clear that Gemini will receive URL to explore.
2. Removed some artifacts during prompt generation that were unfortunately overlooked. (e.g. "(Maps to # 1)"). This was in reference to the "Minimum bar" requirements for WPT coverage, but the prompt does not actually reference the minimum bar outside of this.
3. The prompt now advises Gemini completeness over brevity. This is to avoid Gemini's summary omitting important spec details.
4. The `spec_url` and `feature_definition` inputs are no longer encased in a code block to make it clearer that the input is not code.

### DIFF
--- a/framework/gemini_helpers.py
+++ b/framework/gemini_helpers.py
@@ -83,7 +83,7 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> None:
     raise PipelineError('No valid wpt.fyi results URLs found in WPT description.')
 
   template_data = {
-    'spec_contents': feature.spec_link,
+    'spec_url': feature.spec_link,
     'feature_definition': _create_feature_definition(feature)
   }
   spec_synthesis_prompt = render_template(SPEC_SYNTHESIS_TEMPLATE_PATH, **template_data)


templates/prompts/spec-synthesis.html
```diff
@@ -1,55 +1,53 @@
...
 
-Core Directive: I will provide you with a full `SPECIFICATION_DOCUMENT` and a `FEATURE_DEFINITION`. Your task is to:
+Core Directive: I will provide you with a `SPECIFICATION_URL` and a `FEATURE_DEFINITION`. Your task is to:
 
 1.  Use the `FEATURE_DEFINITION` to identify the precise *scope* of the feature.
-2.  Scour the `SPECIFICATION_DOCUMENT` to find all technical details that fall within that scope.
-3.  **Synthesize** all of this information into a *new, single, coherent, and detailed summary* that describes *only* the feature, organized by the structure below.
+2.  **Action:** Access the `SPECIFICATION_URL` using your Search and Retrieval tools.
+    * **Constraint:** You must read the actual content of the page. **Do not** rely on a search snippet or a high-level summary.
+    * **Constraint:** If the specific URL cannot be read, **do not** hallucinate or guess the spec contents based on the URL text.
```
DanielRyanSmith, line 8: @jcscottiii This is a shot in the dark, but maybe this last constraint would help avoid hallucinations? 😅

jcscottiii: Let's try it out and see.

PR #5743, opened 2025-12-03
Make the AI gap analysis comprehensive

DanielRyanSmith:
Previously, the gap analysis response listed a summary of what tests might be missing to achieve minimum satisfactory test coverage, as well as having a "Prioritized Recommendations" section for giving suggestions on some tests to add to increase test coverage. However, this is not sufficient, and ideally the gap analysis would identify every single test that is required to meet the minimum bar in test coverage.

This PR updates the gap analysis prompt to adjust the structure of the response to include every test that will be required to meet the minimum bar. If a feature owner adds all the suggested tests, it should be assumed that the subsequent gap analysis will pass with no suggestions.

### DIFF
--- a/templates/prompts/gap-analysis.html
+++ b/templates/prompts/gap-analysis.html
@@ -92,16 +92,17 @@ For each category, compare the spec requirements to the test coverage and provid
 
 -----
 
-### Required Tests to Achieve Minimum Satisfactory Coverage
+### Test Suggestions to Achieve Minimum Satisfactory Coverage
 
-This is the most critical section. Based on the "Gaps & Notes" identified above, provide a checklist of every new test that must be written.
+This is the most critical section. Based on the "Gaps & Notes" identified above, provide a checklist of every new test suggestion.
 
-**Format:** Group the tests by the 5 definition categories. If a category has no gaps, omit it.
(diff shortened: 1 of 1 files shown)


templates/prompts/gap-analysis.html
```diff
@@ -86,16 +88,20 @@
...
 
 -----
 
-### Prioritized Recommendation
+### Required Tests to Achieve Minimum Satisfactory Coverage
+
+This is the most critical section. Based on the "Gaps & Notes" identified above, provide a checklist of every new test that must be written.
 
-Based on the gaps, provide a 1-3 bullet-point list of the *most critical* tests that must be written to achieve a minimum satisfactory level of coverage.
+**Format:** Group the tests by the 5 definition categories. If a category has no gaps, omit it.
```
jcscottiii: Do you want to print the category but say "no gaps" in that case?

I'm wondering about the case of the API owner reviewing the AI report for the first time. Would they know that a category was evaluated if it was omitted?

jcscottiii (approved):
LGTM with one question.

templates/prompts/gap-analysis.html
DanielRyanSmith: That's a good idea - I've reworded this to keep the category name with a default of "No gaps." when no test suggestions exist. 🙂

PR #5750, opened 2025-12-07
Allow WPT coverage report pipeline to retry after 1 hour in progress

### LINKED ISSUE #5726: Allow for retry of WPT coverage evaluation pipeline if request hangs for 1 hour
The user should be able to retry generating a new WPT coverage evaluation report if the request has been hanging for over an hour. This is already in [the backend logic](https://github.com/GoogleChrome/chromium-dashboard/blob/bfb76b0323eaa77f30b8569c63944390b20203cd/api/wpt_coverage_api.py#L41-L45), but needs to be added to [the UI page](https://github.com/GoogleChrome/chromium-dashboard/blob/bfb76b0323eaa77f30b8569c63944390b20203cd/client-src/elements/chromedash-wpt-eval-page.ts).

DanielRyanSmith:
Fixes #5726

This change allows the user to submit another request to run the WPT coverage evaluation pipeline if the process has been in progress for over an hour, which likely means the process is hanging.

The retry button will display with an explanation of the failure once the process is available for retry:
Screenshot 2025-12-08 1 31 55 AM

### DIFF
--- a/api/wpt_coverage_api.py
+++ b/api/wpt_coverage_api.py
@@ -20,6 +20,13 @@ from framework import permissions
 from internals import core_enums
 from internals.core_models import FeatureEntry
 
+# 29 minute cooldown for regenerating the evaluation report.
+# 29 minutes instead of 30 so we don't block the UI sending a request
+# accidentally.
+COOLDOWN_THRESHOLD = timedelta(minutes=29)
+
+# 59 minute timeout to allow retrying if the process hangs.
+HANGING_TIMEOUT_THRESHOLD = timedelta(minutes=59)
 
 class WPTCoverageAPI(basehandlers.EntitiesAPIHandler):
   """Accepts requests related to WPT AI coverage evaluations."""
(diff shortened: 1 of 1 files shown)


api/wpt_coverage_api.py
```diff
@@ -37,12 +37,13 @@ def do_post(self, **kwargs):
 
     last_status_time = feature.ai_test_eval_status_timestamp
 
-    one_hour = timedelta(hours=1)
+    fifty_nine_minutes = timedelta(minutes=59)
```
jcscottiii: Nit: Let's make this a constant at the top. Named like `HANGING_PROCESS_THRESHOLD`. 

Then if you change it again, you won't have to change all places to match the value.

DanielRyanSmith: Updated!

DanielRyanSmith:
> Outside the scope of this PR, but it made me think about the scenario where two people try to try/retry at roughly the same time. I wonder if there's a small chance of a race condition where the pipeline could be triggered twice. It might be worth investigating if adding @ndb.transactional to the do_post method would be a good safeguard here. Just a thought for the future!

Filed #5760 to track this 🙂

PR #5764, opened 2025-12-10
Add rollout plan to rollout stages

yanndago:


yanndago:
Screenshot 2025-12-10 at 13 57 02
Screenshot 2025-12-10 at 13 57 16

yanndago:
Hey jrobbins, I cc'ed you on the initial bug http://b/445987210.
IWe need to allow custom strings for the rollout plan here. I thought about adding another field custom_stage_rollout_plan that would be a string and be shown if the user selects custom.
Do you think that is a good approach ?

internals/core_enums.py
```diff
@@ -173,6 +173,17 @@
   ROLLOUT_OTHER: "(RARE) It's complicated or unusual",
 }
 
+ROLLOUT_FULL = 0,
```
jrobbins: Please name these constants with the full field name to distinguish them from the constants on the lines above.  E.g., ROLLOUT_FULL -> ROLLOUT_STAGE_PLAN_FULL.

jrobbins:
> Hey jrobbins, I cc'ed you on the initial bug http://b/445987210. IWe need to allow custom strings for the rollout plan here. I thought about adding another field custom_stage_rollout_plan that would be a string and be shown if the user selects custom. Do you think that is a good approach ?

Yes, I think that is a good approach.

internals/core_enums.py
yanndago: Done

yanndago:
I updated the implementation and I would love some feedback on the best way to make this work on all forms.
Screenshot 2025-12-12 at 11 12 56
Screenshot 2025-12-12 at 11 13 02
Screenshot 2025-12-12 at 11 13 12
Screenshot 2025-12-12 at 11 13 22

jrobbins:
I realize this is getting complicated.  But, it also seems a bit complicated from the user's perspective.   Maybe it could be simpler if the rollout plan menu had the "Custom rollout" option as "Custom rollout (explain in Rollout details field)".  That way there would be no new text field for the other value. Also, don't try to make the details required because the fact that it is expected is clear from the name of that option.  WDYT?

yanndago:
That seems like a good idea, let me forward that to the PM

client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -959,6 +975,17 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
             )}
           </sl-select>
         </div>
+        <sl-input
```
jrobbins: You don't need an input for the rollout details because it is already in `${editor}`

jrobbins (commented):
Generally looks good, but I think you left in an sl-input from a previous iteration of coding.

Also, the link to the screenshot in the PR discussion is not working.

yanndago:
Screenshot 2025-12-16 at 15 28 24

MayLove31:
Why do I keep getting these emails even when I unsubscribe from it?

LuckyGemimi

jrobbins (approved):

Issue #5766, opened 2025-12-10
Suppress logs while running Python tests


DanielRyanSmith:
When running the Python tests, our logs are quite noisy from all of the arbitrary logging going on. We can clean this up by patching our logs in the tests.

e.g.
https://github.com/GoogleChrome/chromium-dashboard/blob/dcc4eb1bb43e2c660e50666772f1a72fc5a140cd/framework/csp_test.py#L99

PR #5767, opened 2025-12-10
Allow for retries for obtaining GCP secrets

DanielRyanSmith:
This should make it less likely to see failures on service start.

### DIFF
--- a/framework/secrets.py
+++ b/framework/secrets.py
@@ -129,6 +129,7 @@ class ApiCredential(ndb.Model):
     self.put()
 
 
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
It's odd that github diff view adds comment cards about these being CI failures.
https://github.com/GoogleChrome/chromium-dashboard/pull/5767/files

DanielRyanSmith:
> It's odd that github diff view adds comment cards about these being CI failures.
https://github.com/GoogleChrome/chromium-dashboard/pull/5767/files

It might be pulling the errors out of the logs, which means the problem is related to #5766 (but this is just a guess)

PR #5768, opened 2025-12-11
Always use markdown for rollout_details.

jrobbins:
This should resolve b/429239394 comment #9.  The enterprise team is requesting that we support markdown in the rollout details text field.  

For the summary field, we implemented a list of fields to treat as markdown vs plain text because many existing feature entries had text with unescaped HTML that would not render correctly if treated as markdown.  A review of production data for the rollout_details field did not find any instance of that problem.  So, we can simply treat rollout_details as always being markdown.
In the long term, I'd like to eventually treat all fields as always being markdown.  We are not ready for that yet because of the existing data, but for this particular field we can jump ahead to the desired end-state of always treating it as markdown.

In this PR:
* On the release notes page: Track the preview checkbox for both features and individual stages.
* On the release notes page and in chromestatus-textarea: handle fields that are always markdown by showing an always-checked checkbox.
* Fix some margin problems that were visible when the user had a numbered list in rendered markdown.

### DIFF
--- a/client-src/elements/chromedash-enterprise-release-notes-page.ts
+++ b/client-src/elements/chromedash-enterprise-release-notes-page.ts
@@ -915,11 +915,7 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
       </div>
     `;
     const controls = html`
-      <sl-checkbox
-        class="markdown-checkbox"
-        size="small"
-        checked
-        disabled
+      <sl-checkbox class="markdown-checkbox" size="small" checked disabled
         >Use markdown</sl-checkbox
       >
       <sl-icon-button
@@ -962,8 +958,8 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
               `
             )}
           </sl-select>
-          </div>
-       ${isPreviewing ? preview : editor} ${controls}
+        </div>
+        ${isPreviewing ? preview : editor} ${controls}
       </li>
     `;
   }


client-src/elements/chromedash-enterprise-release-notes-page.ts
```diff
@@ -890,15 +890,53 @@ export class ChromedashEnterpriseReleaseNotesPage extends LitElement {
...
+      <div
+        id="preview"
+        style="border:var(--card-border); padding:0 var(--content-padding); min-height:14em; background:var(--table-alternate-background)"
+      >
+        ${autolink(s.rollout_details, [], true)}
+      </div>
+    `;
+    const controls = html`
+      <sl-checkbox class="markdown-checkbox" size="small" checked disabled
+        >Use markdown</sl-checkbox
```
DanielRyanSmith, line 919: Since `chromedash-textarea` is updated to support `alwaysMarkdown`, could you use `<chromedash-textarea>` here instead of `<sl-textarea>`?

jrobbins: Right now there is a lot of overlap and I could probably reuse chromedash-textarea. However the enterprise release notes page has had a fairly rapid pace of requirements changes. So, I'd like to keep the implementations separate for now so that I am have a freer hand to make changes.  It could be that we consolidate as part of building a v2.0 release notes page some time in the future that would also help with the WP beta blog post.

PR #5788, opened 2025-12-17
Don't email bounced users

### LINKED ISSUE #5739: email bounce detection is not working
We correctly detect email bounces and send a notification email to our bounces mailing list.  However, the bounces keep coming, so we must be failing to set the bouncing flag in the user account, or failing to check it when sending some future emails.

jrobbins:
This should resolve #5739.

Instead of trying to check user_pref.bounced in each place where we construct an email task, we now have one place where that is checked as the tasks are added to the task queue.

### DIFF
--- a/util/deployment/deploy-env
+++ b/util/deployment/deploy-env
@@ -28,7 +28,7 @@ npm run clean-setup
 npm run build
 
 # 2. Deploy the application services.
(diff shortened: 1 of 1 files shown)


jrobbins:
I tested this on staging, but since any change to email processing is a little risky, I'll hold on to this PR until after the holidays.

DanielRyanSmith (approved):

Issue #5828, opened 2026-01-07
Fetch resource/dependency files of Web Platform Tests for direct integration into evaluation report prompts


DanielRyanSmith:
Right now, we're relying on the LLM to identify and fetch required resource/dependency files that are listed in the test file contents. Instead, we should populate the prompt with the resource files and provide them directly in the prompt.

For example:
```
<test_suite>
  <file name="test-1.html">
    <script src="resources/test-helper.js"></script>
  </file>

  <file name="resources/test-helper.js">
    function assert_true(condition) { ... }
  </file>
</test_suite>
```

DanielRyanSmith:
This is fixed with #5830

PR #5830, opened 2026-01-07
Fetch resource/dependency files of Web Platform Tests for evaluation reports

DanielRyanSmith:
Part of #5828

This change adds recursive fetching of dependency file contents from WPTs. These file contents will be used to populate test analysis prompts, which will come in a subsequent PR.

A list of files we have already fetched is kept to avoid infinite recursion, or fetching the same file twice.

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -439,7 +439,7 @@ async def get_mixed_wpt_contents_async(
     initial_paths.add(fpath)
     visited_paths.add(fpath)
 
-  # PHASE 2: Recursive Content Fetching
+  # PHASE 2: Iterative Content Fetching
   while processing_queue:
     # Create fetch tasks for the current batch
     file_tasks = [


framework/utils.py
```diff
@@ -36,6 +37,13 @@
 WPT_GITHUB_API_URL = 'https://api.github.com/repos/web-platform-tests/wpt/contents/'
 WPT_GITHUB_RAW_CONTENTS_URL = 'https://raw.githubusercontent.com/web-platform-tests/wpt/master/'
 
+# Match <script src="...">
+SCRIPT_DEPENDENCY_REGEX = r'<script\s+[^>]*src=["\']([^"\']+)["\']'
+
+# Match import/export ... from "..." or import "..."
+# Matches: import { x } from './foo.js' | import './bar.js' | export * from './baz.js'
```
jrobbins, line 44: Ah, good point that that form of `export` is actually importing first.

framework/utils.py
```diff
@@ -36,6 +37,13 @@
 WPT_GITHUB_API_URL = 'https://api.github.com/repos/web-platform-tests/wpt/contents/'
 WPT_GITHUB_RAW_CONTENTS_URL = 'https://raw.githubusercontent.com/web-platform-tests/wpt/master/'
 
+# Match <script src="...">
+SCRIPT_DEPENDENCY_REGEX = r'<script\s+[^>]*src=["\']([^"\']+)["\']'
```
jrobbins: It would probably be better to create regular expression objects here so that any Regex syntax errors are detected immediately.  E.g.,
SCRIPT_DEPENDENCY_RE = re.compile(r'...').
findall() accepts a regex object for the pattern.

framework/utils.py
```diff
@@ -313,68 +321,157 @@ async def _fetch_and_pair(fpath: Path, furl: str) -> tuple[Path, str | None]:
...
+
+      # Analyze for dependencies
+      dependencies = extract_dependencies(content)
+
+      for dep in dependencies:
+        resolved_path = resolve_dependency_path(fpath, dep)
+
+        # If valid path and we haven't seen it yet
+        if resolved_path and resolved_path not in visited_paths:
+          visited_paths.add(resolved_path)
```
jrobbins, line 474: For safety against the possibility of a very long run, maybe bail out if len(visited_paths) > some limit like 50.

framework/utils.py
```diff
@@ -313,68 +321,157 @@ async def _fetch_and_pair(fpath: Path, furl: str) -> tuple[Path, str | None]:
...
-  file_results = await asyncio.gather(*file_tasks)
+  # Initialize processing queue
+  processing_queue = list(files_to_fetch_map.items())
+
+  # Record the initial set of paths so we know they are "Tests"
+  for fpath, _ in processing_queue:
+    initial_paths.add(fpath)
+    visited_paths.add(fpath)
+
+  # PHASE 2: Recursive Content Fetching
```
jrobbins: This is an iterative BFS, right? Not actually recursive.

jrobbins (approved):

framework/utils.py
DanielRyanSmith: You're right - the outcome is a recursive tree of dependencies, but the method uses BFS. 🙂

framework/utils.py
DanielRyanSmith: This is a good idea - I've added a constant that enforced an upper bound of 100 dependency files for now

framework/utils.py
DanielRyanSmith: This is correct, and is the way we do it in other parts of the codebase anyway 🙂

Issue #5829, opened 2026-01-07
Fetch spec contents for direct integration into evaluation report prompts


DanielRyanSmith:
Right now, we're relying on the LLM to fetch the contents of a spec from a URL. Instead, we should populate the prompt with the contents and provide it directly in the prompt.

PR #5831, opened 2026-01-07
Add function to fetch contents of spec URLs

DanielRyanSmith:
Part of #5829

This PR adds a new function that takes the `spec_link` value of a feature and fetches its contents in text form. Use of this new function will be implemented in a subsequent PR.

- If the URL is a GitHub PR, the text diff of that PR is fetched.
- If the URL is a GitHub file, the text contents of that file are fetched.
- If the URL is any other webpage, the text contents are cleaned and provided in markdown format.

This process uses the [Trafilatura](https://trafilatura.readthedocs.io/en/latest/) library to obtain the text contents of webpages.

### DIFF
--- a/client-src/css/shared-css.js
+++ b/client-src/css/shared-css.js
@@ -11,7 +11,7 @@ export const SHARED_STYLES = [
 
   * {
     box-sizing: border-box;
-     /* font: inherit; */
+    /* font: inherit; */
     text-decoration: inherit;
     -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
   }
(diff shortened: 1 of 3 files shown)


framework/gemini_helpers.py
```diff
@@ -42,6 +45,58 @@ def _create_feature_definition(feature: FeatureEntry) -> str:
...
+    # include_tables=True is critical for Specs which often use tables for definitions.
+    result = trafilatura.extract(
+      downloaded,
+      include_comments=False,
+      include_tables=True,
+      output_format="markdown"
+    )
+
+    if not result:
+      return f"Error: No main content found at {url}"
```
jrobbins, line 92: So, do these error messages actually get passed to Gemini along with successfully-fetched content? How might that affect Gemini?

jrobbins (approved):
LGTM with one question.

framework/gemini_helpers.py
DanielRyanSmith: Yes, they are. There could be a situation where a test doesn't exist anymore with the same name, or a dependency file name isn't correct, but I didn't want it to derail the entire evaluation. However, now that we're just fetching all the test file contents, we can just omit the file contents we don't fetch correctly.

I've created #5833 to describe and track this

PR #5836, opened 2026-01-08
New Unified Gap Analysis prompt and adjustments

DanielRyanSmith:
This change adds a new "Unified Gap Analysis" prompt, which is a new prompt used to combine the entire 3-prompt flow into one single prompt using Chain of Thought prompting. Initial tests with this prompt have been positive for small test sets.

Additionally, some modifications to the existing 3-prompt flow are made to ensure that wording is consistent across all of the prompts, and the output format between the two gap analysis prompts is the same.

### DIFF
--- a/client-src/css/shared-css.js
+++ b/client-src/css/shared-css.js
@@ -11,7 +11,7 @@ export const SHARED_STYLES = [
 
   * {
     box-sizing: border-box;
-     /* font: inherit; */
+    /* font: inherit; */
     text-decoration: inherit;
     -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
   }
@@ -70,6 +70,23 @@ export const SHARED_STYLES = [
     text-decoration: underline;
     color: var(--link-hover-color);
     cursor: pointer;
+  }
+
+
+  .roadmap-sidebar nav {
+    display: flex;
+    flex-direction: column;
+    gap: var(--content-padding-quarter);
+  }
+
+  .roadmap-sidebar nav a {
+    display: block;
+    line-height: 1.4;
+  }
+
+  .roadmap-sidebar nav a[aria-current="page"],
+  .roadmap-sidebar nav a.active {
+    margin-bottom: 0; /* spacing handled by gap */
   }
 
   input:not([type="submit"]),
@@ -139,7 +156,7 @@ export const SHARED_STYLES = [
   }
 
   code {
-   font-family: monospace;
+    font-family: monospace;
   }
 
   .description {
@@ -239,7 +256,7 @@ export const SHARED_STYLES = [
     background: var(--accordion-background);
     font-weight: 300;
     font-size: 20px;
-    padding: var(--content-padding-quarter) var(--content-padding-half)
+    padding: var(--content-padding-quarter) var(--content-padding-half);
   }
 
   sl-details::part(content) {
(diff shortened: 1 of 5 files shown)


templates/prompts/gap-analysis.html
```diff
@@ -34,75 +34,66 @@
...
+* **Commpn Use Cases:** ["No gaps." or "X Test suggestion(s)."]
+    * [Test suggestion]
+* **Error Scenarios:** ["No gaps." or "X Test suggestion(s)."]
+    * [Test suggestion]
+* **Invalidation:** ["No gaps." or "X Test suggestion(s)."]
+    * [Test suggestion]
+* **Integration:** ["No gaps." or "X Test suggestion(s)."]
+    * [Test suggestion]
+
+#### End Output Format
```
jrobbins, line 96: How did you decide to use `#### Output Format` and `#### End Output Format` as the block delimiters?  Is there some guide or examples that you are following?

templates/prompts/gap-analysis.html
```diff
@@ -34,75 +34,66 @@
...
+* **Gaps:** [If not covered, explicitly state what is missing]
+
+### 2. Common Use Cases
+* **Status:** [**Covered** / **Partially Covered** / **Not Covered**]
+* **Spec Requirement:** [Describe the core successful behaviors]
+* **Evidence:** [Cite specific test files and logic that verify these behaviors]
+* **Gaps:** [List any core scenarios missing from the tests]
+
+### 3. Likely Error Scenarios
+* **Status:** [**Covered** / **Partially Covered** / **Not Covered**]
```
jrobbins, line 65: How did you decide to make these options bold rather than the way that you had them?  Why not backquotes?

templates/prompts/unified-gap-analysis.html
```diff
@@ -0,0 +1,125 @@
...
+2.  `<feature_definition>`: The Name and Description of the specific feature you are auditing. **You must use this to define the scope of your analysis.**
+3.  `<test_suite>`: One or more raw test files (including dependencies) written to test this feature.
+
+Your task is to act as an auditor. You must compare the `<test_suite>` against the *scoped requirements* found in the `<spec_document>` to determine if a "minimum satisfactory level of test coverage" has been met.
+
+**Goal:** Identify every missing test scenario required to achieve minimum satisfactory coverage. The output must be a list identifying any tests needed to achieve this.
+
+---
+
+### Phase 1: Scope & Spec Extraction
```
jrobbins, line 17: Prompt engineering can seem so magical.

jrobbins (approved):
LGTM, but I am curious about a few details.

templates/prompts/gap-analysis.html
DanielRyanSmith: To be honest, this was just me trying to ensure that the output format was more properly sectioned off in a way that is understandable. I haven't had an issue with it yet, but this version of the output instructions does not contain examples like the previous one, so it seemed intuitive to specify exactly where the format instructions began and ended.

I need to look for a specific guide in cases like this 😅 - For things like this, I am making decisions based off of whether the LLM seems to understand what I'm asking after some number of content generations.

templates/prompts/gap-analysis.html
DanielRyanSmith: This was a personal preference that I added. I was thinking that the user might want to see this result at a glance, so I made these bold. 🙂

templates/prompts/unified-gap-analysis.html
DanielRyanSmith: I needed to read more on this, but I am impressed with its ability to essentially handle subtasks within a single prompt like this. We're in interesting times! 🙂

PR #5843, opened 2026-01-09
Don't include non-ready items in releasenotes.

jrobbins:
This should resolve b/444431984#comment17 item 2.  In addition to checking the `confidential` boolean field, we also check `is_releasenotes_publish_ready` when returning the list of results to display for the release notes page.  Users who are signed in with a google, chromium, or admin account still see the full list, e.g., to edit those items.  However, anon users will see the filtered list, e.g., as it will be displayed on the separate release notes site.


DanielRyanSmith:
> I'm not remembering when is_releasenotes_publish_ready is set to true, but I imagine it's within the UI and at the feature owner's discretion. 🙂

I confirmed this in the code

Issue #5834, opened 2026-01-08
"Post intent to blink-dev" button updated feature status but failed to send/post the email


yoshisatoyanagisawa:
Thank you for letting us know that something needs to be fixed. This page is for recording bugs on Chrome Status (https://www.chromestatus.com). If you need to report a problem with Chrome or Chromium, please use crbug (http://crbug.com/new). Thank you.

**Describe the bug**
When using the "Post intent to blink-dev" feature from an "Intent to Extend Experiment" draft page, the feature status on chromestatus.com was successfully updated. However, the actual email was never sent to the blink-dev mailing list, nor does it appear in the Google Groups archives (https://groups.google.com/a/chromium.org/g/blink-dev).

**To Reproduce**
1. Fill in the required fields for the intent.
2. Go to a feature's "Intent to Extend Experiment" (or supposing any Intent) stage on chromestatus.com, and push "Draft Intent to Extend Experiment email" button.
3. Click the "Post intent to blink-dev" button in the opened page.
4. Observe that the feature status changes on the dashboard, but no email is received by the mailing list.

**Expected behavior**
Clicking "Post intent to blink-dev" should both update the feature status on the dashboard and successfully post the drafted message to the blink-dev@chromium.org mailing list.

**Additional context**
- I waited for over 24 hours to account for any potential delivery delays.
- I checked the official blink-dev Google Groups archives (https://groups.google.com/a/chromium.org/g/blink-dev), but the post is missing.
- This resulted in a discrepancy where the dashboard shows the intent has been "Trial Extension Pending" but the community has not been notified.
- https://chromestatus.com/feature/6265472244514816 is the feature I tried to extend experiment.

jrobbins:
From the logs
```
POST /tasks/email-intent-to-blink-dev
INFO 2026-01-08T03:40:35.594103Z Submitting email task: To: blink-dev@chromium.org CC: ... Subject: Intent to Extend Experiment: SharedWorker on Android
INFO 2026-01-08T03:40:35.594617Z Processing 1 email tasks
INFO 2026-01-08T03:40:35.623455Z Skipping email task with no non-bounced To: addrs
```

The problem was the blink-dev was marked as a bouncing email address.  We don't sent emails to addresses that bounced previously, but that rule is apparently too strict.  I am tracking that general problem in #5853.  But, I will keep this issue open until I specifically track down why (or at least when) blink-dev@chromium.org was marked as bouncing.

[...]

PR #5852, opened 2026-01-12
Prevent intent stage advancement without blink-dev post

pranjalisr:
For issue #5834
I have another PR open (#5848 ) for a separate issue. This PR is independent.

Problem
When using the “Post intent to blink-dev” action, the feature’s intent stage could advance even if the email was never successfully posted to the blink-dev mailing list. This resulted in an inconsistent state where ChromeStatus showed the intent as pending or posted, but no corresponding message existed in the blink-dev archives.
Because ChromeStatus does not receive delivery confirmation or a thread URL from Google Groups, there is no reliable signal at email-send time that the intent was actually posted.

 Solution
This change prevents intent-related stage advancement unless a blink-dev thread URL is present. The presence of an archived thread URL is the only reliable indicator that an intent was successfully posted.
Specifically:
* Intent stage updates are blocked if intent_thread_url is not set.
* This prevents advancing intent state when the blink-dev email fails, is delayed, or has not yet been linked.
* No changes are made to the email-sending task itself.

 Why this works
* ChromeStatus cannot synchronously verify blink-dev email delivery.
* The thread URL is added asynchronously and is the source of truth.
* Enforcing this invariant avoids false “intent posted” states and keeps the dashboard consistent with the mailing list.

### DIFF
--- a/api/intents_api.py
+++ b/api/intents_api.py
@@ -184,4 +184,9 @@ class IntentsAPI(basehandlers.APIHandler):
 
     cloud_tasks_helpers.enqueue_task('/tasks/email-intent-to-blink-dev',
                                      params)
-    return {'message': 'Email task submitted successfully.'}
+    return {
+  'message': (
+    'Intent email queued for delivery. '
+    'Feature status will update after successful posting to blink-dev.'
+  )
+}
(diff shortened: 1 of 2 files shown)


jrobbins:
I prefer the current UX because developers like to be able to work ahead to fill in the information that they have when they have it rather than being forced to do everything in a strict order.

PR #5867, opened 2026-01-13
Bypass Chromium code checks for WebGPU features

DanielRyanSmith:
WebGPU features are hard to verify in the Chromium code, as there seems to be no single source of truth. This change lets WebGPU features bypass the code checks in order to avoid false positives on features missing shipping criteria.


6h5pfs5xrm-spec:
But a better strategy is to guarantee it is a true or false positive after being flagged as potentially positive.

PR #5869, opened 2026-01-13
Playwright test update playground

DanielRyanSmith:
This is an attempt to use Gemini to optimize the Playwright tests to avoid flakiness.
- Removed all `await delay(500)` calls; implemented `expect().toBeVisible()` and `waitForURL` to handle race conditions deterministically.
- Swapped implementation details (CSS classes, `value="150"`) for user-facing locators (`getByRole`, `hasText`).
- Wrapped logical flows in `test.step()` for clearer failure reports.
- Removed redundant `afterEach` logout (Playwright uses fresh `BrowserContexts` per test).


DanielRyanSmith:
This was just for testing changes in the CI. Closing this, as its job as been completed 🙂

PR #5872, opened 2026-01-14
Update external reviews page Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**

### DIFF
--- a/packages/playwright/tests/chromedash-report-external-reviews-page_pwtest.js
+++ b/packages/playwright/tests/chromedash-report-external-reviews-page_pwtest.js
@@ -16,7 +16,7 @@ test.beforeEach(async ({ page }) => {
  */
 function getSection(page, headingText) {
   // Finds a <section> that contains a heading with the specific name
-  return page.locator('chromedash-report-external-reviews-page section section').filter({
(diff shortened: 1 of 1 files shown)


packages/playwright/tests/chromedash-report-external-reviews-page_pwtest.js
```diff
@@ -1,165 +1,152 @@
...
- * @returns {import('@playwright/test').Locator}
+ * @param {string} headingText
  */
-function section(page, heading) {
-  return page
-    .locator('chromedash-report-external-reviews-page section')
-    .filter({has: page.getByRole('heading', {name: heading})});
+function getSection(page, headingText) {
+  // Finds any <section> that contains a heading with the specific name
+  return page.locator('section').filter({
```
jrobbins: We use `section` a fair amount in our code.  Shouldn't the test have the more specific locator?

DanielRyanSmith: I'll add back the locator `chromedash-report-external-reviews-page section section` from the original logic

PR #5873, opened 2026-01-14
Update new feature list Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**

### DIFF
--- a/packages/playwright/tests/chromedash-new-feature-list_pwtest.js
+++ b/packages/playwright/tests/chromedash-new-feature-list_pwtest.js
@@ -20,7 +20,7 @@ test('Typing slash focuses on searchbox', async ({ page }) => {
   // Use the utility to navigate (removes hardcoded localhost URL)
   await gotoNewFeatureList(page);
 
-  const searchbox = page.locator('#inputfield');
+  const searchbox = page.locator('#inputfield input');
 
   // Verify initial state
   await expect(searchbox).toBeVisible();


packages/playwright/tests/chromedash-new-feature-list_pwtest.js
```diff
@@ -1,32 +1,43 @@
...
+    // 2. "/"   -> Hotkey detected by app, moves focus to searchbox
+    // 3. "def" -> Typed into searchbox
+    // 4. "/"   -> Typed into searchbox (literal character now that it has focus)
+    // 5. "ghi" -> Typed into searchbox
+    await page.keyboard.type('abc/def/ghi');
+
+    // Assertion:
+    // We use toHaveValue() because user input changes the DOM property,
+    // not necessarily the HTML attribute.
+    await expect(searchbox).toHaveValue('def/ghi');
```
jrobbins, line 41: This looks right, but why did we have that `xx` before?  That looks like it should have always failed. I wrote it in #4325, but I don't recall the purpose of the `xx`.  Maybe I was testing having a value in the field initially and then typing should prepend to it, but I don't see that initial set up code.  

We should try to understand why this was not failing before.  Does it fail now if you add back the xx?

packages/playwright/tests/chromedash-new-feature-list_pwtest.js
```diff
@@ -1,32 +1,43 @@
...
 test.afterEach(async ({ page }) => {
     // Logout after running each test.
     await logout(page);
 });
 
-test('Typing slash focuses on searchbox', async ({page}) => {
-  await gotoNewFeatureList(page, 'http://localhost:5555/');
+test('Typing slash focuses on searchbox', async ({ page }) => {
+  // Use the utility to navigate (removes hardcoded localhost URL)
+  await gotoNewFeatureList(page);
```
jrobbins, line 21: The fact that we had an extra argument passed and that was never flagged as an error is a sign that we should upgrade this code from JS to TS.

packages/playwright/tests/chromedash-new-feature-list_pwtest.js
DanielRyanSmith: I'm now realizing the problem - the naming convention in this test is not correct, so the tests presumably don't run at all

jrobbins:
I think that this test was never run because the file name does not end with `_pwtest.js`.

jrobbins:
> I think that this test was never run because the file name does not end with `_pwtest.js`.

Sorry, I did not notice your earlier comment.  Yeah, that's the problem.

jrobbins (approved):

PR #5874, opened 2026-01-14
Update stage page Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**

### DIFF
--- a/packages/playwright/tests/chromedash-guide-stage-page_pwtest.js
+++ b/packages/playwright/tests/chromedash-guide-stage-page_pwtest.js
@@ -1,133 +1,140 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import {
-  captureConsoleMessages, login, logout, createNewFeature
+  captureConsoleMessages, login, logout,
+  createNewFeature
 } from './test_utils';
 
 test.beforeEach(async ({ page }, testInfo) => {
   captureConsoleMessages(page);
   testInfo.setTimeout(90000);
 
-  // Login is required to create and edit features
+  // Login before running each test.
   await login(page);
 });
 
 test.afterEach(async ({ page }) => {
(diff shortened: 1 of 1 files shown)


packages/playwright/tests/chromedash-guide-stage-page_pwtest.js
```diff
@@ -1,154 +1,140 @@
...
+  await page.getByText('Add stage').click();
+
+  // Select stage to create.
+  // We wait for the select element to be ready before interacting.
+  const stageSelect = page.locator('sl-select#stage_create_select');
+  await stageSelect.click();
+
+  // Click the origin trial stage option.
+  const originTrialStageOption = page.locator('sl-select sl-option[value="150"]');
+  await expect(originTrialStageOption).toBeVisible();
```
jrobbins, line 35: Doesn't click() wait for the item to be visible?
https://playwright.dev/docs/input#mouse-click

DanielRyanSmith: You're correct here - this is a redundant line!

However, Gemini gives a compelling reason for it to be kept:
> While strictly redundant, developers often add await expect(locator).toBeVisible() before a click for two reasons
>
> 1. Clearer Error Messages: If the dropdown fails to open, an explicit expectation fails with "Expected element to be visible," which can sometimes be easier to read in logs than a generic click timeout.
> 2. Animation Timing (The "Belt and Suspenders" approach): With custom UI components (like the Shoelace sl-select dropdown you are using), sometimes the element technically "exists" and is "visible" to the DOM before the animation (sliding down) is fully finished. Explicitly waiting for visibility can sometimes stabilize these specific interactions, though Playwright's built-in "Stability" check usually handles this.

PR #5875, opened 2026-01-14
Update new feature creation page Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**

### DIFF
--- a/packages/playwright/tests/chromedash-guide-new-page_pwtest.js
+++ b/packages/playwright/tests/chromedash-guide-new-page_pwtest.js
@@ -1,24 +1,27 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import {
-  captureConsoleMessages, login, logout, gotoNewFeaturePage, createNewFeature
+  captureConsoleMessages, login, logout,
+  gotoNewFeaturePage, enterBlinkComponent, createNewFeature, enterWebFeatureId
 } from './test_utils';
 
 test.beforeEach(async ({ page }, testInfo) => {
   captureConsoleMessages(page);
-  testInfo.setTimeout(90000);
+  testInfo.setTimeout(30000);
 
-  // Login is required for all tests in this file
+  // Login before running each test.
   await login(page);
 });
 
 test.afterEach(async ({ page }) => {
-    // Logout after running each test.
-    await logout(page);
+  // Logout after running each test.
+  await logout(page);
 });
 
 test('navigate to create feature page', async ({ page }) => {
   await gotoNewFeaturePage(page);
+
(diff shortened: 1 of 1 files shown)


packages/playwright/tests/chromedash-guide-new-page_pwtest.js
```diff
@@ -19,21 +18,18 @@ test.afterEach(async ({ page }) => {
...
 
-
-test('enter feature name', async ({page}) => {
+test('enter feature name', async ({ page }) => {
   await gotoNewFeaturePage(page);
 
-  test.setTimeout(90000);
   const featureNameInput = page.locator('input[name="name"]');
-  await expect(featureNameInput).toBeVisible({timeout: 60000});
+  await expect(featureNameInput).toBeVisible();
```
jrobbins, line 32: Does removing this timeout speed up the overall test run?  I had assumed that timeout: 60000 means that it would keep checking periodically as fast as it could, and give up after 60s, not wait a full 60s.

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
```diff
@@ -42,52 +38,43 @@ test('enter feature name', async ({page}) => {
 
   // Enter a feature name.
   await featureNameInput.fill('Test feature name');
-  await delay(500);
 
-  await expect(page).toHaveScreenshot('feature-name.png');
+  // Verify the input has the value before screenshotting (implicitly waits for fill to complete)
+  await expect(featureNameInput).toHaveValue('Test feature name');
```
jrobbins, line 43: Good insight.  But, doesn't toHaveScreenshot() also automatically retry comparing screenshots up until the default timeout?
https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
DanielRyanSmith: based on [the docs](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible), removing this timeout just has the method fall back to the default timeout (defined in TestConfig.expects). I'm not certain if it speeds up anything here, but I'm also not certain if there's a reason to have such a larger timeout for this input - is it something that requires being 1 minute long?

packages/playwright/tests/chromedash-guide-new-page_pwtest.js
DanielRyanSmith: This may be redundant based on what you mentioned in the docs

PR #5876, opened 2026-01-14
Update edit-all page Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**


packages/playwright/tests/chromedash-guide-editall-page_pwtest.js
```diff
@@ -1,75 +1,85 @@
...
-
 test('editall page', async ({ page }) => {
-    await createNewFeature(page);
-    await gotoEditAllPage(page);
+  await createNewFeature(page);
+  await gotoEditAllPage(page);
 
-  await expect(page).toHaveScreenshot('edit-all-fields.png', {timeout: 30000});
+  await expect(page).toHaveScreenshot('edit-all-fields.png', {
+    mask: [page.locator('#history')] // Mask volatile history if present
```
jrobbins: Huh? The only `#history` that I see in the source code is on the feature highlights on the feature detail page.  By this point in the test, we should have navigated away from the feature detail page.

packages/playwright/tests/chromedash-guide-editall-page_pwtest.js
```diff
@@ -1,75 +1,85 @@
...
+    // We check for visibility before screenshotting to ensure the UI has updated
+    const shippedField = page.locator('chromedash-form-field[name="shipped_milestone"]');
+    await expect(shippedField.locator('.check-warning')).toBeVisible();
+
+    // Scroll to next field to center the error for the screenshot
+    const nextField = page.locator('input[name="shipped_android_milestone"]');
+    await nextField.scrollIntoViewIfNeeded();
+
+    await expect(page).toHaveScreenshot('shipped-desktop-error.png', {
+      mask: [page.locator('#history')]
```
jrobbins: Same.

packages/playwright/tests/chromedash-guide-editall-page_pwtest.js
DanielRyanSmith: Agreed - this may be a remnant of Gemini masking history in another file and thinking it needs to do it here as well. Definitely unnecessary.

PR #5878, opened 2026-01-14
Update the feature detail page Playwright tests

DanielRyanSmith:
Daniel's summary:

Gemini's summary:

**Key Changes**


packages/playwright/tests/chromedash-guide-feature-page_pwtest.js
```diff
@@ -1,79 +1,85 @@
...
+  // Wait for the option to appear in the dropdown.
+  const originTrialStageOption = page.locator('sl-select sl-option[value="150"]');
+  await expect(originTrialStageOption).toBeVisible();
+
+  // Hover Origin trial stage option.
+  await originTrialStageOption.hover();
+
+  // Screenshot of this dialog.
+  await expect(page).toHaveScreenshot('create-origin-trial-stage-dialog.png', {
+    mask: [page.locator('section[id="history"]')]
```
jrobbins, line 44: If we mask `#history`, do we still need lines 27-31 to scroll away from it?

jrobbins (approved):

packages/playwright/tests/chromedash-guide-feature-page_pwtest.js
DanielRyanSmith: This is correct - I've removed the scroll logic and updated the screenshots to reflect this

PR #5879, opened 2026-01-14
Update utils functions for Playwright tests

DanielRyanSmith:
**This change also contains the removal of the use of setTimeout within our test sign-in functions in our Lit components**

Daniel's summary:

Gemini's summary:

**Key Changes**

### DIFF
--- a/packages/playwright/tests/test_utils.js
+++ b/packages/playwright/tests/test_utils.js
@@ -1,5 +1,5 @@
 // @ts-check
-import { expect } from '@playwright/test';
+import {expect} from '@playwright/test';
 
 /**
  * Call this, say in your test.beforeEach() method, to capture all
@@ -81,7 +81,7 @@ export async function decodeCookies(page) {
  */
 export async function isMobile(page) {
   const viewportSize = page.viewportSize();
-  return (viewportSize && viewportSize.width <= 700);
+  return viewportSize && viewportSize.width <= 700;
 }
 
 /**
@@ -133,8 +133,8 @@ export async function login(page) {
   // can occur in Chrome when not logged in.
   acceptAlertDialogs(page);
 
-  await page.goto('/', { timeout: 20000 });
-  await page.waitForURL('**/roadmap', { timeout: 20000 });
+  await page.goto('/', {timeout: 20000});
+  await page.waitForURL('**/roadmap', {timeout: 20000});
 
   await expect(page).toHaveTitle(/Chrome Status/);
 
@@ -156,7 +156,7 @@ export async function login(page) {
   // Waiting for 'domcontentloaded' ensures the reload has actually happened.
   await Promise.all([
     page.waitForEvent('domcontentloaded'),
-    loginButton.click()
+    loginButton.click(),
   ]);
 
   // Validate successful login.
@@ -250,7 +250,7 @@ export async function gotoNewFeaturePage(page) {
 
   // Expect "Add a feature" header to be present.
   const addAFeatureHeader = page.getByTestId('add-a-feature');
-  await expect(addAFeatureHeader).toBeVisible({ timeout: 10000 });
+  await expect(addAFeatureHeader).toBeVisible({timeout: 10000});
 }
 
 /**


client-src/elements/chromedash-drawer.ts
```diff
@@ -174,10 +174,8 @@ export class ChromedashDrawer extends LitElement {
           }
         })
         .then(() => {
-          setTimeout(() => {
-            const url = window.location.href.split('?')[0];
-            window.location.href = url;
-          }, 1000);
+          const url = window.location.href.split('?')[0];
+          window.location.href = url;
```
jrobbins, line 178: This setTimeout certainly complicates the timing of tests.  I'm trying to remember why Dan needed it in the first place.

packages/playwright/tests/test_utils.js
```diff
@@ -143,72 +116,56 @@ export function acceptAlertDialogs(page) {
...
   // But first accept alert dialogs, which
   // can occur in Chrome when not logged in.
   acceptAlertDialogs(page);
 
-  await page.pause();
-  // console.log('login: goto /');
-  await page.goto('/', {timeout: 20000});
-  await page.waitForURL('**/roadmap', {timeout: 20000});
+  await page.goto('/', { timeout: 20000 });
+  await page.waitForURL('**/roadmap', { timeout: 20000 });
```
jrobbins: Maybe we shouldn't go to the roadmap page for every single test that needs to log in.  It is a complex page that kicks off a lot of API fetches.  Maybe we should implement a /blank URL that is not navigable from the main menu and that would be used only for testing.  Or, maybe use /settings, but avoid the "please sign in" alert.

packages/playwright/tests/test_utils.js
```diff
@@ -143,72 +116,56 @@ export function acceptAlertDialogs(page) {
...
-
-    accountIndicator = page.getByTestId('account-indicator');
+  // We use the presence of the 'dev-mode-sign-in-button' as the definitive
+  // indicator of "Not Logged In" because the Account Indicator layout varies.
+  const loginButton = page.getByTestId('dev-mode-sign-in-button');
+
+  if (await loginButton.isHidden()) {
+    // If the login button is hidden, we assume we are logged in.
+    // We must logout first to ensure a clean state.
+    await logout(page);
```
jrobbins: For this case, are we sure that we need to log out and then log back in?  AFAIK, playwright can only be sighed in with just the one mock account, so if we are signed in already, maybe just return early.

packages/playwright/tests/test_utils.js
```diff
@@ -143,72 +116,56 @@ export function acceptAlertDialogs(page) {
...
-  // console.info('expect login button to be present and visible');
-  const loginButton = page.getByTestId('dev-mode-sign-in-button');
-  await expect(loginButton).toBeVisible({timeout: loginTimeout});
+  // Expect login button to be present now.
+  await expect(loginButton).toBeVisible();
 
-  await loginButton.click({timeout: 5000, delay: 1000});
-  await delay(loginTimeout / 3); // longer delay here, to allow for initial login.
+  // Click login and wait for the page reload that the button triggers.
+  // We cannot use waitForURL because the URL might be the same.
```
jrobbins, line 155: Good insight.

packages/playwright/tests/test_utils.js
```diff
@@ -222,73 +179,73 @@ export async function logout(page) {
   // accept leaving them unsaved.
   acceptBeforeUnloadDialogs(page);
 
-  // console.log('logout: goto /');
   await page.goto('/');
   await page.waitForURL('**/roadmap');
```
jrobbins, line 188: Again, it seems expensive to keep hitting the roadmap page.  Are we navigating there just so that we can check the URL later?

packages/playwright/tests/test_utils.js
```diff
@@ -143,72 +116,56 @@ export function acceptAlertDialogs(page) {
...
-    await signOutLink.click({timeout: 5000});
-
-    await delay(1000);
-    await page.waitForURL('**/roadmap');
-    await expect(page).toHaveTitle(/Chrome Status/);
-    page.mouse.move(0, 0); // Move away from content on page.
-    await delay(1000);
-
-    accountIndicator = page.getByTestId('account-indicator');
+  // We use the presence of the 'dev-mode-sign-in-button' as the definitive
```
jrobbins, line 142: It's good that you rewrote this.  The `while` loop in the old code is one potential root cause of our random 30-minute timeouts.  Although, I would expect individual tests to time out before the whole 30m is used up.

packages/playwright/tests/test_utils.js
```diff
@@ -222,73 +179,73 @@ export async function logout(page) {
...
-    await menuButton.click();
+    // If menu isn't visible, we might not be logged in or page is broken,
+    // but we check anyway.
+    if (await menuButton.isVisible()) {
+      await menuButton.click();
+
+      const signOutLink = page.getByTestId('sign-out-link');
+      if (await signOutLink.isVisible()) {
+        await signOutLink.click();
+        await page.waitForURL('**/roadmap');
```
jrobbins, line 201: I am not sure that the URL and title prove that we are signed out.  It seems better to check for dev-mode-sign-in-button to appear again.

packages/playwright/tests/test_utils.js
```diff
@@ -222,73 +179,73 @@ export async function logout(page) {
...
-  await page.waitForURL('**/roadmap');
-  await delay(500);
+      // Need to wait for the sign-out-link to be visible
+      const signOutLink = page.getByTestId('sign-out-link');
+      await expect(signOutLink).toBeVisible();
+      await signOutLink.click();
 
-  // console.log('logout: done');
+      // Confirm we are back on roadmap and signed out
+      await page.waitForURL('**/roadmap');
```
jrobbins, line 222: Same.

packages/playwright/tests/test_utils.js
```diff
@@ -299,13 +256,11 @@ export async function enterBlinkComponent(page) {
...
-  // Trying to show options, doesn't work yet.
-  await blinkComponentsInputWrapper.focus();
-  await delay(500);
-
   const blinkComponentsInput = blinkComponentsInputWrapper.locator('input');
   await blinkComponentsInput.fill('blink');
-  await delay(500);
+
+  // If you need to verify the value "stuck"
+  await expect(blinkComponentsInput).toHaveValue('blink');
```
jrobbins, line 272: Good insight.

packages/playwright/tests/test_utils.js
DanielRyanSmith: I am in agreement here. However, the current iteration of the app makes the test sign in button visible on ONLY the roadmap page. It would take a substantial change to adjust this, so if it's worth fixing (and I think it is), it's probably better to handle it in a separate change.

packages/playwright/tests/test_utils.js
DanielRyanSmith: I'm going to give that a shot, because I think you're right here

packages/playwright/tests/test_utils.js
DanielRyanSmith: See roadmap comment above

packages/playwright/tests/test_utils.js
DanielRyanSmith: Agreed - definitely the better approach. I made the change. 🙂

packages/playwright/tests/test_utils.js
DanielRyanSmith: See roadmap comment above

packages/playwright/tests/test_utils.js
jrobbins: That's fine.  I was thinking of it as future work.

jrobbins (approved):

PR #5881, opened 2026-01-15
Fix screenshot generation of feature creation page

DanielRyanSmith:
The extra help icon in the form field was causing occasional flaky screenshot tests. By removing the `setTimeout` function (which was not needed), and adding a step in the test to wait for the button to render before taking the screenshot, the test should no longer be flaky.

This may not be the only flaky screenshot that can be fixed this way! But we can fix them as we identify them.

### DIFF
--- a/packages/playwright/tests/chromedash-guide-new-page_pwtest.js
+++ b/packages/playwright/tests/chromedash-guide-new-page_pwtest.js
@@ -69,9 +69,10 @@ test('test semantic checks', async ({page}) => {
   const summaryLocator = page.locator('chromedash-form-field[name="summary"]');
   await expect(summaryLocator).toContainText('Feature summary should be');
 
-
   // This forces Playwright to wait for the icon to be fully painted
-  const helpIcon = page.locator('chromedash-form-field[name="name"] sl-icon-button');
+  const helpIcon = page.locator(
+    'chromedash-form-field[name="name"] sl-icon-button'
+  );
   await expect(helpIcon).toBeVisible();
 
   // Screenshot of warnings about feature name summary length


client-src/elements/chromedash-form-field.ts
```diff
@@ -176,14 +176,11 @@ export class ChromedashFormField extends LitElement {
...
-    // Do first semantic check after the document is ready.
+    // are available to do the semantic check.
     if (document.readyState === 'loading') {
-      document.addEventListener('DOMContentLoaded', () =>
-        setTimeout(() => {
-          this.doSemanticCheck();
-        })
-      );
+      document.addEventListener('DOMContentLoaded', () => {
+        this.doSemanticCheck();
```
jrobbins, line 182: Was that setTimeout() needed for any reason from the user's perspective rather than for tests?  I see that it was introduced in #3583, but that is a big PR so I have not dug through the discussion.

DanielRyanSmith: I had the same question! I asked Gemini before opening this

---

Me:
> I want to clarify one thing: You told me to remove the `setTimeout` function from the logic in `chromedash-form-field.ts`. Will that have any effect on the behavior of the Lit component? Or did it do nothing but delay the `doSemanticCheck` method call?

---

Gemini:
This is an excellent question. It is always wise to double-check "chesterton's fence"—why was that code there in the first place?

Based on the code analysis, **removing the `setTimeout` is safe and actually improves the component's behavior.**

Here is the breakdown of why it was likely added, why it isn't needed, and why removing it fixes your issue without regressions.

### 1. Why was it likely there? (The "Defensive Coding" Pattern)

In Web Components (and Lit), developers often use `setTimeout(() => ..., 0)` in `firstUpdated` as a "hack" to ensure that **sibling components** or **parent containers** have finished rendering.

The comment in the code says:

> *"We need to wait until the entire page is rendered, so later dependents are available to do the semantic check"*

This suggests the author was worried that `doSemanticCheck` might try to read data from other fields that hadn't attached to the DOM yet.

### 2. Why is it safe to remove? (The Data Flow)

I reviewed `doSemanticCheck` and its helper `getFieldValueWithStage`.

* **It is Data-Driven, not DOM-Driven:** The validation logic does **not** query the DOM (e.g., it doesn't use `document.querySelector('#other-field').value`).
* **It uses Properties:** It relies entirely on `this.fieldValues`, which is a property passed into the component by the parent.
* **Lifecycle Guarantee:** In Lit, `firstUpdated` is called only after the component's properties (like `fieldValues`) have been set and the initial template has been rendered. Therefore, the data required to run the check is guaranteed to be present in memory **synchronously**.

### 3. The "Visual Regression" is actually a fix

By removing `setTimeout`, you are changing the render order from **Scenario A** to **Scenario B**:

* **Scenario A (Old - With Timeout):**
1. Component renders.
2. Browser paints the component (without error message).
3. `setTimeout` fires (next event loop tick).
4. Semantic check runs -> Error found.
5. Component re-renders with error message.
6. **Result:** The UI "jumps" or "pops in" the error message milliseconds after loading. This is what caused your screenshot test to fail (it took the photo during the jump).


* **Scenario B (New - No Timeout):**
1. Component renders.
2. Semantic check runs immediately.
3. Browser paints the component (with error message already included).
4. **Result:** The UI is stable immediately. No layout shift.

DanielRyanSmith: Since this is an AI response, this of course does not confirm that there won't be a regression, but I was not able to produce one in my own manual check of creating or editing a feature

Issue #5865, opened 2026-01-13
Update "shipping features" logic checks to handle multiple shipping stages


DanielRyanSmith:
This problem came up recently with the [Trusted Types spec alignment feature](https://chromestatus.com/feature/5163792014245888). The feature entry has two shipping stages. One is essentially empty, and the other has all of the relevant information needed to complete the shipping process.

For now, when we are verifying the data for a shipping feature, it's best to use only the shipping stage that contains the actual shipping milestone relevant to the feature (in this case, only check the shipping stage that lists m145 as the shipping milestone).

DanielRyanSmith:
This is actually not the original problem I had thought it was. It wasn't that there was some information in one shipping stage, and the rest of the info in another. Instead, there were just two shipping stages, both set to M145, one with more info after that, and one with no info.

This is not really a solution that should be combated in the "shipping features" logic. It should be assumed that if a user adds a stage to their feature, it's one we should respect and should assume it will be populated accordingly.

PR #5882, opened 2026-01-15
Ignore empty placeholder shipping stages in shipping features API

pranjalisr:
Fixes issue #5865
Some features can define multiple shipping stages, where one stage may be
effectively empty while another contains the actual shipping milestone data.

The shipping features API previously treated all returned shipping stages as
equally relevant, which could cause validation to fail when placeholder stages
were present.

This change updates the API to filter out empty or placeholder shipping stages
and only consider stages that define milestone data when aggregating shipping
features. This ensures the correct shipping stage is used when multiple stages
exist for a feature.

### DIFF
--- a/api/shipping_features_api.py
+++ b/api/shipping_features_api.py
@@ -57,6 +57,16 @@ class ShippingFeaturesAPI(basehandlers.EntitiesAPIHandler):
 
     shipping_stages = self._get_shipping_stages(milestone)
 
+   shipping_stages = [
+    s for s in shipping_stages
+    if s.milestones and any([
+        s.milestones.desktop_first,
+        s.milestones.android_first,
+        s.milestones.ios_first,
+        s.milestones.webview_first,
+    ])
+]
+
     if len(shipping_stages) == 0:
       return {
         'complete_features': [],


api/intents_api.py
```diff
@@ -184,4 +184,9 @@ def do_post(self, **kwargs) -> Response|dict|MessageResponse:
 
     cloud_tasks_helpers.enqueue_task('/tasks/email-intent-to-blink-dev',
                                      params)
-    return {'message': 'Email task submitted successfully.'}
+    return {
+  'message': (
+    'Intent email queued for delivery. '
+    'Feature status will update after successful posting to blink-dev.'
+  )
+}
```
DanielRyanSmith, line 192: Why change this?

api/stages_api.py
```diff
@@ -113,21 +113,33 @@ def do_patch(self, **kwargs):
...
+
+    # Block intent advancement unless a blink-dev thread exists
+    if (
+        'intent_stage' in body and
+        not stage.intent_thread_url
+    ):
+        self.abort(
+            400,
+            msg='Intent stage cannot be advanced until the blink-dev email is successfully posted.'
+        )
```
DanielRyanSmith, line 128: Why change this?

client-src/elements/chromedash-feature-detail.ts
```diff
@@ -399,9 +399,21 @@ export class ChromedashFeatureDetail extends LitElement {
...
+
+   const reviewIsComplete =
+   typeof value === 'string' && value === 'Complete';
+
+   const showCheckmark =
+  isReviewStatusField ? reviewIsComplete : isDefined;
+
+   const icon = showCheckmark
+  ? html`<sl-icon library="material" name="check_circle_20px"></sl-icon>`
+  : html`<sl-icon library="material" name="blank_20px"></sl-icon>`;
```
DanielRyanSmith, line 415: What does this have to do with this change?

api/intents_api.py
pranjalisr: I updated the message to make the flow clearer to the user, since the stage update won’t happen immediately and only updates once the email is posted to blink-dev.

But I may be misunderstanding how this response is expected to behave.
If this text shouldn’t be changed or there’s a preferred wording, I’m happy to adjust or revert it.
Please let me know what you think.

api/stages_api.py
pranjalisr: I added this check because the intent stage was being advanced before the blink-dev email actually finished posting, which could leave the feature in an inconsistent state.
My goal was to make sure we only allow stage advancement after the email thread URL exists.

If this logic belongs somewhere else or should be implemented differently, please let me know.  I’m very open to changing it.

client-src/elements/chromedash-feature-detail.ts
pranjalisr: I added this because I thought we needed a way to visually reflect the review status as part of the intent-progress blocking flow. If this isn’t needed or should be handled elsewhere, please let me know.

DanielRyanSmith:
We're no longer taking fixes for this issue. See https://github.com/GoogleChrome/chromium-dashboard/issues/5865#issuecomment-3770240785 for details.

Issue #5644, opened 2025-11-04
incomplete steps have a green tick-mark


fergald:
I'm looking at [this feature](https://chromestatus.com/feature/4830298126417920) which currently has

> Security review status
> Pending

and beside that is a green tick-mark.

<img width="360" height="268" alt="Image" src="https://github.com/user-attachments/assets/376a79c0-fdf8-4ea3-9b83-7116d8501b14" />

Privacy also has a green tick-mark.

This seems wrong, I would expect these to only be green when they are complete. Maybe they should be orange when they are pending and perhaps red when they haven't even been started.

I see no other color tick-marks on the page. Things either have a green tick-mark or nothing at all. This makes it hard to spot important steps that are incomplete.

pranjalisr:
@fergald Hi! Is this issue still open? I’d like to take it up.

fergald:
@pranjalisr I'm just reporting the bug, I don't work on this project.

pranjalisr:
I'd like to work on this issue and send a CL if that's okay.
The plan is to remove the green checkmark for non-complete states
so pending steps are no longer shown as completed.

PR #5887, opened 2026-01-16
Fix review status checkmark rendering for security and privacy fields

pranjalisr:
**Summary**

For issue #5644
This PR fixes incorrect checkmark icon rendering for review status fields on the feature detail page.
Previously, the Security review and Privacy review fields displayed a checkmark as soon as the field was defined, even when the review was not complete. This change ensures that checkmarks are shown only when the review status is explicitly marked as “Complete.”

**Problem:**

On the feature detail page:

security_review_status

privacy_review_status

were treated like regular fields.
As a result, they showed a ✔️ icon whenever a value existed, which was misleading and inconsistent with how review completion is interpreted elsewhere in Chrome Status.

**Solution:**

Special-case security and privacy review fields.

Render a checkmark only when the value is exactly "Complete".

Preserve existing behavior for all other fields.

This keeps the UI accurate and avoids implying review completion prematurely.

**Implementation Details:**

Updated renderField() logic in chromedash-feature-detail.ts

Introduced explicit handling for review status fields

No backend changes

No behavioral changes to unrelated fields

Before

✔️ shown for security/privacy review fields even when incomplete

After

✔️ shown only when review status is “Complete”

Blank icon otherwise

### DIFF
--- a/client-src/elements/chromedash-feature-detail.ts
+++ b/client-src/elements/chromedash-feature-detail.ts
@@ -389,29 +389,43 @@ export class ChromedashFeatureDetail extends LitElement {
   }
 
   renderField(fieldDef, feStage) {
-    const [fieldId, fieldDisplayName, fieldType, deprecated, alwaysMarkdown] =
-      fieldDef;
-    const value = getFieldValueFromFeature(fieldId, feStage, this.feature);
-    const isDefined = isDefinedValue(value);
-    if (!isDefined && deprecated) {
-      return nothing;
-    }
-    const isMarkdown =
-      (this.feature.markdown_fields || []).includes(fieldId) || alwaysMarkdown;
+  const [fieldId, fieldDisplayName, fieldType, deprecated, alwaysMarkdown] =
+    fieldDef;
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
The biggest problem here is that the issue creator fundamentally misunderstood the intent of the checkmark in the UI in the first place. It's not to show that a review is complete. It's to show that the field has a value of any kind. This does mean that there's a problem with the UI causing interpretation, but it doesn't mean that the solution is as easy as changing it to how the user thinks it works.

Working in the wrong direction like this is something that can be remedied by asking clarifying questions on the issue, or wait until a repository maintainer adds a clear desired outcome for the issue. In this scenario, we have not come to a consensus on what the desired change would be, and so this issue is not a particularly useful one to work on.

Thank you for the attempt, but I think it's best we just close this for now.

pranjalisr:
@DanielRyanSmith

Thanks for the clarification! From now on, I’ll make sure to ask for confirmation and clearly understand the expected direction before starting work on any issue. Appreciate your guidance!

PR #5890, opened 2026-01-16
Some general fixes for flaky Playwright test behavior

DanielRyanSmith:
This change adds some fixes for various Playwright flaky tests.

1. Shoelace icons not visible. [Failure example](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21077031692/job/60621390000). This is due to the asynchronous rendering of the icons in the Shoelace icon button component. Ensuring that the SVG within the icon button has rendered should fix this issue.
The explanation for this is that Playwright takes the screenshot as soon as the element is "visible" and network activity is idle. If the icon SVG hasn't finished painting inside the Shadow DOM yet, you get a screenshot of an empty button.
2. Refactored logout function. [Failure example 1](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21157874856/job/60846635753), [example 2](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21161496346/job/60856858989). The logic in this function has been updated to remove flaky behavior, like just returning if we detect we're already logged out, and ensuring that the page has been reloaded before checking for the login button. Additionally, a recursive retry has been added in order to circumvent network lag or permissions call response failures.
3. Feature detail page not loading after creation. [Failure example](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21160626338/job/60854744115). This flake happened very infrequently in testing, so I can't be certain that it is fixed, but my assumption is that the feature detail is attempting to be navigated to before new feature data is eventually consistent. a reload has been added if the page fails to load after 5 seconds. I can only verify this fixes the problem by not seeing flakes. 🙃


jrobbins:
Doesn't the screenshot have a 5s timeout already?  If the menu icon is not appearing within 5s, then the new condition would fail at 5s as well.  I think this is a good attempt to fix the underlying problem and I hope that it works, but it does, I'd like to understand how exactly it made that difference.

DanielRyanSmith:
> Doesn't the screenshot have a 5s timeout already? If the menu icon is not appearing within 5s, then the new condition would fail at 5s as well. I think this is a good attempt to fix the underlying problem and I hope that it works, but it does, I'd like to understand how exactly it made that difference.



Well, based on the [merge queue result](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21079607626/job/60629875190), it does not fix the problem 😅 

I'll go back to the drawing board here 🙂 

Edit: I adjusted the code and found some fixes. I've updated the base description of this PR.

DanielRyanSmith:
> Thanks for executing this test in CI to prove it is working consistently.

My plan was to try 10 times, and we had a failure on the 10th attempt 😮‍💨 I think this has increased consistency enough to land, but we may experience some rogue flakiness that would need to be addressed later.

packages/playwright/tests/test_utils.js
jcscottiii, line 1: Great to see you fixed this. Gemini had one suggestion. But if it's working, we can leave it as-is. But I will paste the suggestion just in case for the future:


Currently, the test sometimes has to retry or wait for a timeout because it's checking for a specific element (like the sign-in button) before the app has fully settled its authentication state after a reload. This is a common challenge in E2E tests for dynamic web apps.

Instead of the current retry/reload logic, what if we made the `logout` test wait explicitly for *either* the sign-in button *or* the account indicator to become visible? This way, the test would know the app has finished its internal state checks, without relying on a hidden attribute or multiple retries.

Here's a conceptual snippet for `logout` that illustrates this idea:

```javascript
export async function logout(page) {
  // ... (existing code to trigger the logout click, e.g., signOutLink.click()) ...

  // Wait for the page to reload after logout action.
  await page.waitForEvent('load');

  const loginButton = page.getByTestId('dev-mode-sign-in-button');
  const accountIndicator = page.getByTestId('account-indicator');

  // Wait for either the login button OR the account indicator to be visible.
  // This ensures the app has settled into a known auth state.
  await expect(loginButton.or(accountIndicator)).toBeVisible({ timeout: 15000 });

  // Now, assert that the expected logged-out state is present.
  await expect(loginButton).toBeVisible();
  await expect(accountIndicator).not.toBeVisible();
}
```

This approach directly observes the user-facing UI, making the test less susceptible to intermittent backend/network delays while still ensuring the desired outcome. It would fail quickly if neither element appeared (indicating a truly broken state) but be stable otherwise.

It's a small tweak that could offer a nice improvement in how deterministically our tests reflect the actual user experience.

Let me know what you think!

---
*This review was generated by Gemini.*

DanielRyanSmith: This doesn't sound like a bad idea, but it doesn't address the root of the flakiness, which is that the page occasionally fails to load entirely, or at least for very long periods of time. In reality, this is a flaw in the application's logic somewhere that needs to be updated, so that's probably the way to approach it. (This PR doesn't fix the problem either and is essentially a workaround)

On a side note, I didn't know this "or" syntax was a thing (`loginButton.or(accountIndicator)`) - pretty cool!

Issue #5917, opened 2026-01-21
Flaky Playwright test caused by Web Feature ID and Blink Component input field


DanielRyanSmith:
[A Playwright test run](https://github.com/GoogleChrome/chromium-dashboard/actions/runs/21228963891/job/61083024743) experienced a test flake due to the Web Feature ID and Blink Component fields in the screenshot.

---
#### Expected

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/28c69a0a-840d-457d-88c9-d36158ddd778" />

#### Actual

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/a0bf07f8-67d3-4e45-aaf7-346281126364" />

#### Diff

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/a9c733be-eac1-46aa-9951-2639e7732ded" />

---

#### Expected
<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/237740b6-0639-4224-8631-94cf975ff124" />

#### Actual

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/f107b680-6614-48a2-aa24-2188b7d23e51" />

#### Diff

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/f2068c7d-6869-4500-a550-2492db121189" />

PR #5923, opened 2026-01-23
Clearer formatting guide for Test Suggestions section

DanielRyanSmith:
I noticed recently that the "Test Suggestions" section of the coverage reports would occasionally keep the quotes that are in the example, (stating "No gaps." rather than No gaps.)

Today, there was also a time where Gemini misinterpreted it enough to write "X test suggestions." rather than the actual number of test suggestions:
```
- Existence: X Test suggestion(s).
    - Create a test verifying the inheritance of text-justify. Set text-justify: inter-character on a parent and ensure a child element without the property specified renders with inter-character spacing.
- Commpn Use Cases: X Test suggestion(s).
    - ...
```

Additionally, this is the first time the "Commpn" typo was surfaced in a report (It seems that Gemini usually just knew to fix it 😅).

This change makes it clearer what the intended format is, and fixes the typo. After generating a few reports, I did not see the same issues after making this change.

### DIFF
--- a/templates/prompts/gap-analysis.html
+++ b/templates/prompts/gap-analysis.html
@@ -84,7 +84,7 @@ Do not output *any* other text or formatting if this rule is triggered.
 
 * **Existence:** [State the number of test suggestions or clarify there are no gaps. e.g. "No gaps." or "X Test suggestion(s)."]
     * [1 or more bullets containing individual test suggestions]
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
The Playwright test failure was an issue already captured in #5917 that hasn't been addressed yet

PR #5924, opened 2026-01-23
Enforce the 50-test file limit during report generation

DanielRyanSmith:
This change updates the logic in the WPT coverage evaluation pipeline to abort the process if there are more than 50 test files to be evaluated in a single test suite after all directories listed are evaluated.

If a report request is submitted with a test suite that is too large, the pipeline will fail and a short message explaining the test suite size issue will populate the report page. As we gather the test suite file requirements (including directories) in `utils.py` and find the test suite is too large, we throw a PipelineError that is bubbled up the stack (down the stack??)  and then caught to populate the report with the error message.

Some variables that lived in `gemini_helpers.py` have been moved to `utils.py` so they can be shared by both files.

image

### DIFF
--- a/framework/gemini_helpers_test.py
+++ b/framework/gemini_helpers_test.py
@@ -38,6 +38,9 @@ class GeminiHelpersTest(testing_config.CustomTestCase):
     self.mock_render_template = mock.patch(
         'framework.gemini_helpers.render_template').start()
     self.mock_utils = mock.patch('framework.gemini_helpers.utils').start()
+    # Ensure PipelineError is the real exception class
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
No hurry on this - can be looked at next week

framework/utils.py
ahfuckit: you guys are using TryIt as well? really?

ahfuckit: actual implies a fake exists... an explanation?

framework/gemini_helpers.py
ahfuckit: comments can be essential for many things. include them

PR #5927, opened 2026-01-26
Add functionality to copy WPT coverage reports to clipboard

### LINKED ISSUE #5921: Allow users to copy the WPT coverage reports to clipboard
Add a button at the bottom/top of the report page that copies the contents of the report to the user's clipboard.

DanielRyanSmith:
Fixes #5921


DanielRyanSmith:
> This reminds me that Shoelace has a sl-copy-button element, but I think it is optimized for simpler use cases where the value to be copied is taken directly from some other form field.
https://shoelace.style/components/copy-button

I actually really like how this component looks, with the checkmark replacing the copy icon after a click. If there is additional spare time, implementing that would be pretty cool 🙂

PR #5941, opened 2026-01-27
Do not allow confidential features to generate WPT coverage reports (API)

### LINKED ISSUE #5939: Don't let confidential features submit their information to Gemini
This shouldn't be possible at the moment, as only Enterprise features can be marked as confidential, and these features do not supply the correct information to use WPT coverage reports. But just to be safe, we should ensure that no feature marked as confidential can submit their feature information to Gemini to generate WPT coverage reports.

DanielRyanSmith:
Fixes #5939


ahfuckit:
Are you open-sourced as you claim or what is the agenda here?

PR #5943, opened 2026-01-29
Notify release notes reviewers after a flag reset.

jrobbins:
This should resolve item 3 of b/444431984.

Basically, there was already logic to reset the release notes review flags when a non-reviewer edited the milestones, name, or summary. This PR adds an email notification so that reviewers will know to go and re-review.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -31,7 +31,7 @@ from internals.core_enums import *
 from internals.core_models import FeatureEntry, MilestoneSet, Stage
 from internals.data_types import CHANGED_FIELDS_LIST_TYPE
 from internals import feature_links
-from internals import notifier_helpers
+from internals import notifier, notifier_helpers
 from internals.review_models import Gate, Activity
 from internals.data_types import VerboseFeatureDict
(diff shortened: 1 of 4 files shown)


internals/notifier.py
```diff
@@ -1334,3 +1334,30 @@ def build_email(self, feature_id, feature_name, owner_emails):
       'reply_to': None,
       'html': body,
     }
+
+
+RELEASENOTES_NOTIFY_ADDRS = [
+    'cbe-tpgms@google.com',
+    'siobhankeating@google.com',
```
DanielRyanSmith, line 1341: This is an individual's email, but I'm assuming this is an xWF email so it cannot be part of a Google group

DanielRyanSmith (approved):

internals/notifier.py
jrobbins: I am not sure about that rule, but that is the list of addresses that were requested.

api/features_api.py
ahfuckit: come on guys.

PR #5947, opened 2026-01-29
Some wording and formatting changes for WPT coverage reports

DanielRyanSmith:
This is an accumulation of some small changes to wording and formatting for the page and the reports. Some of these changes are subjective, so I'm willing to hear any objections 🙂 

- Section headers are a little smaller (Makes the report feel a little more compact)
- Line partitions for each section
- No more bold status result (I originally changed this to bold, but it felt like it actually made the status harder to find because all the subsections are in bold already)
- Note that the summary should not be in harsh language (more guardrails to not make the results sound aggressive).
- Update the summary example to say "may require" rather  than "requires" to soften language and add a note of possible uncertainty.

Other changes:
- Some wording of "Evaluation" changed to "Analysis" that was missed.
- A sentence in the instructions is changed from describe "evaluate if it meets standard minimum test coverage criteria" to the less combative "analyze coverage and suggest additional tests to increase test coverage"

### Example before:
Screenshot 2026-01-29 12 14 26 PM

### Example after:
Screenshot 2026-01-29 12 14 45 PM

### DIFF
--- a/templates/prompts/gap-analysis.html
+++ b/templates/prompts/gap-analysis.html
@@ -43,9 +43,20 @@ Do not output *any* other text or formatting if this rule is triggered.
 
 #### Output Format
 
-**Conclusion:** [Start with a brief conclusion (e.g., "**Minimum satisfactory coverage has been achieved!**", or "**Some additional tests may be required to meet minimum satisfactory coverage. See the test suggestions below.**")]
+**Conclusion:** [Start with one of two brief conclusions.
+  - "No test suggestions found. This feature has great test coverage!" If there are no gaps in test coverage.
+  - "Some test suggestions are available. See below." If there are available test suggestions to improve test coverage.
+]
 
-**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage in language that is not harsh. (e.g., "Feature existence is proven, but the suite may require tests for the specific error handling defined in the spec.")]
+**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage in language that is not harsh.
+  - Non-exhaustive guidance:
+    - Consider: Highlighting where happy paths are covered but edge cases are missing.
+    - Consider: Pointing out specific error handling defined in the spec that lacks tests.
+    - Avoid: Using strong negative wording, like "completely" or "lacking".
+  - Examples of output:
+    - "Feature existence is proven, but the suite may require tests for the specific error handling defined in the spec."
+    - "The provided test suite successfully verifies basic visual rendering the `@font-language-override` descriptor effectively. However, tests the `@font-face` descriptor mentioned in the spec are still needed."
+]
 
 -----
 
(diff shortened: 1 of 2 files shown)


templates/prompts/gap-analysis.html
```diff
@@ -43,43 +43,55 @@
 
 #### Output Format
 
-**Conclusion:** [Start with a brief, direct conclusion (e.g., "**Minimum satisfactory coverage has been achieved!**", or "**Some additional tests may be required to meet minimum satisfactory coverage. See the test suggestions below.**")]
+**Conclusion:** [Start with a brief conclusion (e.g., "**Minimum satisfactory coverage has been achieved!**", or "**Some additional tests may be required to meet minimum satisfactory coverage. See the test suggestions below.**")]
 
-**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage (e.g., "Feature existence is proven, but the suite currently requires tests for the specific error handling defined in the spec.")]
+**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage in language that is not harsh. (e.g., "Feature existence is proven, but the suite may require tests for the specific error handling defined in the spec.")]
```
jcscottiii: I think we need to make sure Gemini knows these examples aren't the only options. How about we format it like this?

```
**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage in language that is not harsh.
  - Non-exhaustive guidance:
    - Consider: Highlighting where happy paths are covered but edge cases are missing.
    - Consider: Pointing out specific error handling defined in the spec that lacks tests.
  - Examples of output:
    - The provided test suite .. (use the example from your screenshot in the PR body)
```

Also, we can easily add more cases to consider as needed.

templates/prompts/unified-gap-analysis.html
```diff
@@ -45,43 +45,55 @@
 
 #### Output Format
 
-**Conclusion:** [Start with a brief, direct conclusion (e.g., "**Minimum satisfactory coverage has been achieved!**", or "**Some additional tests may be required to meet minimum satisfactory coverage. See the test suggestions below.**")]
+**Conclusion:** [Start with a brief conclusion (e.g., "**Minimum satisfactory coverage has been achieved!**", or "**Some additional tests may be required to meet minimum satisfactory coverage. See the test suggestions below.**")]
```
jcscottiii: Similar to my above comment.

templates/prompts/unified-gap-analysis.html
```diff
@@ -45,43 +45,55 @@
...
 
-**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage (e.g., "Feature existence is proven, but the suite currently requires tests for the specific error handling defined in the spec.")]
+**Summary of Analysis:** [In 1-2 sentences, explain the state of the coverage in language that is not harsh. (e.g., "Feature existence is proven, but the suite may require tests for the specific error handling defined in the spec.")]
```
jcscottiii: Similar to my above comment.

templates/prompts/gap-analysis.html
DanielRyanSmith: This looks good. I added this format and it seems to work well. 🙂 

However, I noticed the "X test suggestions" wording error again, like it was not understanding my example, so I did the same thing to the last section of the output. 😊

PR #5948, opened 2026-01-29
WPT-AI: Skip failed fetch contents in coverage evaluation prompts

### LINKED ISSUE #5833: Don't pass test file contents that fail to fetch to be used in the coverage evaluation prompts.
Currently, the prompts contain some guardrail text to ignore bad data, but now that we're fetching both the test file contents and their dependency file contents, we can simply remove the text about failures from the prompts and not generate the prompts with the error text.

Optionally, it may be useful to add a caveat to the final report that "Failed to fetch the contents of X files", along with listing their file path.

pranjalisr:
Fixes #5833.

This change ensures that only valid, successfully fetched content is used when
constructing WPT coverage evaluation prompts.

Previously, failed fetches could result in missing, empty, or error placeholder
content being passed into coverage evaluation prompts, which could lead to
misleading or low-quality analysis.

This update addresses that by:
- Skipping WPT test and dependency files whose contents fail to fetch or are empty.
- Ensuring spec fetch failures return None instead of error placeholder strings.
- Failing fast when required spec content cannot be fetched, preventing prompt
  generation with invalid inputs.

Together, these changes ensure coverage evaluation prompts are constructed only
from valid test and specification content.

### DIFF
--- a/client-src/elements/chromedash-wpt-eval-page.ts
+++ b/client-src/elements/chromedash-wpt-eval-page.ts
@@ -2,6 +2,7 @@ import {LitElement, TemplateResult, css, html, nothing} from 'lit';
 import {customElement, property, state} from 'lit/decorators.js';
 import {unsafeHTML} from 'lit/directives/unsafe-html.js';
 import {marked} from 'marked';
+import DOMPurify from 'dompurify';
 import {SHARED_STYLES} from '../css/shared-css.js';
 import {Feature} from '../js-src/cs-client.js';
 import {parseRawTimestamp, showToastMessage} from './utils.js';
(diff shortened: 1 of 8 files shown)


DanielRyanSmith (commented):
This is adding logic to the function that fetches the spec content, but that is not what the issue described. We fetch multiple test file contents and add them to prompts. We need to avoid adding empty test file contents to prompts if they fail to be fetched.

pranjalisr:
@DanielRyanSmith Thanks for the clarification — I’ve updated the logic in get_mixed_wpt_contents_async to skip test files whose fetched contents are None or whitespace-only, so they aren’t added to prompts. I also added a unit test covering failed fetches, whitespace-only content, and valid content.

Also, apologies — I accidentally closed the PR while resolving conflicts earlier, but it’s reopened now.

DanielRyanSmith:
This is not quite what we were looking for, and I'm going to go ahead and close this. I encourage you to look at other repositories to find better opportunities for community contributions (maybe ones that have GSoC opportunities? We're not participating this year).

PR #5950, opened 2026-01-29
Add a token count check for unified prompt

### LINKED ISSUE #5949: Verify the number of tokens for the unified gap analysis prompt to ensure context is not overloaded
This was tangentially brought up by @jcscottiii for another project.

It would be wise to ensure we are not overloading the 1M token count context for Gemini before we try and submit the prompt for response. We can do this with the [countTokens method](https://ai.google.dev/api/tokens#method:-models.counttokens). If the prompt is too large, just revert to using the typical 3-prompt flow.

DanielRyanSmith:
Fixes #5949

### DIFF
--- a/framework/gemini_helpers.py
+++ b/framework/gemini_helpers.py
@@ -36,11 +36,6 @@ UNIFIED_GAP_ANALYSIS_TEMPLATE_PATH = 'prompts/unified-gap-analysis.html'
 # Otherwise, it's a directory.
 WPT_FILE_REGEX = re.compile(r"\/[^/]*\.[^/]*$")
 
-# Define the maximum number of listed tests for the feature that can be used in
-# the single prompt format. Any more than this will cause the trigger the use
-# of the three-prompt flow that utilizes separate test summaries.
-MAX_SINGLE_PROMPT_TEST_COUNT = 10
-
 
 def _create_feature_definition(feature: FeatureEntry) -> str:
   return f'Name: {feature.name}\nDescription: {feature.summary}'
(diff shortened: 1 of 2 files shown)


framework/gemini_helpers.py
```diff
@@ -40,6 +40,9 @@
 # the single prompt format. Any more than this will cause the trigger the use
 # of the three-prompt flow that utilizes separate test summaries.
 MAX_SINGLE_PROMPT_TEST_COUNT = 10
+# Maximum token count allowed for the unified prompt before falling back to
+# the multi-stage analysis.
+MAX_UNIFIED_PROMPT_TOKEN_COUNT = 1_000_000
```
DanielRyanSmith: TIL about this helpful notation!

jcscottiii: Each model in the Gemini has an input_token_count [1]

We should dynamically compare our token count against that

[1] https://ai.google.dev/gemini-api/docs/tokens?lang=python#context-windows

DanielRyanSmith: That's definitely how it should be done. I updated the implementation!

framework/gemini_helpers.py
```diff
@@ -254,7 +253,18 @@ async def run_wpt_test_eval_pipeline(feature: FeatureEntry) -> core_enums.AITest
     return core_enums.AITestEvaluationStatus.FAILED
 
   if len(test_files) <= MAX_SINGLE_PROMPT_TEST_COUNT:
```
jcscottiii: Question: Now that we are counting the tokens, do we need this outer limit check for the number of test_files?

Could we instead always try to:
1. build the _generate_unified_prompt_text
2. Check if the token limit exceeded
3. Fallback to `prompt_analysis` if limit is exceeded.

DanielRyanSmith: Great call! Yeah, there should be no reason to use the max test count anymore. I've just removed that check and it should check context token limits exclusively now 🙂

PR #5985, opened 2026-02-06
Update feature_links_tests.py to no longer call live APIs

DanielRyanSmith:
Several tests used real URLs (e.g., specific GitHub issues) and relied on the application to fetch live data from the internet via Link.parse(). The tests would occasionally flake if the data failed to be fetched. This change updates the implementation to mock Link.parse() and avoid interacting with external APIs for tests.


jrobbins (approved):
Thank you for doing this because those flakes were annoying.

PR #5998, opened 2026-02-10
New prompt rendering tests

DanielRyanSmith:
This adds a new test file that generates prompts from the templates and asserts their structure.


internals/prompts_test.py
```diff
@@ -0,0 +1,124 @@
...
+    data = {
+      'test_files': [
+        {'path':   'css/a.html', 'contents': 'contents of file a.html'},
+        {'path':   'css/b.html', 'contents': 'contents of file b.html'},
+      ],
+      'dependency_files': [
+        (  'css/c.html', 'contents of file c.html'),
+        (  'css/d.html', 'contents of file d.html'),
+        (  'css/e.html', 'contents of file e.html'),
+      ]
```
DanielRyanSmith, line 104: This makes it clear that there are some data type discrepancies in similarly-named arguments. I'll update this in a separate PR.

PR #6000, opened 2026-02-10
Make sure test and dependency data types are the same across prompts

DanielRyanSmith:
Follow up to https://github.com/GoogleChrome/chromium-dashboard/pull/5998#discussion_r2786002569

Make sure the provided data type for `test_files` and `dependency_files` is the same.

### DIFF
--- a/internals/prompts_test.py
+++ b/internals/prompts_test.py
@@ -21,7 +21,7 @@ from flask import Flask, render_template
 import settings
 
 # WARNING: Only do this when you are sure the output is correct!
-REGENERATE_GOLDENS = True
+REGENERATE_GOLDENS = False
 
 test_app = Flask(__name__,
   template_folder=settings.get_flask_template_path())


framework/gemini_helpers.py
```diff
@@ -116,8 +116,10 @@ def _generate_unified_prompt_text(
     'spec_content': _fetch_spec_content(feature.spec_link),
     'feature_name': feature.name,
     'feature_summary': feature.summary,
-    'test_files': wpt_contents.test_contents,
-    'dependency_files': wpt_contents.dependency_contents,
+    'test_files': [{'path': fpath, 'contents': fc}
```
jcscottiii, line 119: You may want to add a helper function to help with this:

```
def _format_files_for_template(file_dict):
    """Converts a dict of path->content into a list of dicts for templates."""
    return [{'path': path, 'contents': content} for path, content in file_dict.items()]
```

then this call would be like: `'test_files': _format_files_for_template(wpt_contents.test_contents),`

And similar for the other calls for test_files and dependency_files

DanielRyanSmith: I do agree - we do this formatting enough it should be a helper function. This PR is blocking today's deployment, so I'll address it in a follow-up PR

PR #6002, opened 2026-02-10
feat: Allow users to delete wpt coverage reports

szy196:
Key Changes:
Added a do_delete method to handle DELETE requests.

When a report is deleted, the ai_test_eval_report field is set to None, and the status is updated to a new DELETED state.

Modified the on_cooldown logic to allow re-running or managing reports if the previous status was DELETED.

Added DELETED = 4 to the AITestEvaluationStatus enumeration.

Added unit tests to verify successful deletion by authorized users.


DanielRyanSmith (approved):
LGTM!

PR #6003, opened 2026-02-11
feat: frontend: Allow users to delete wpt coverage reports

szy196:
Added a handleDeleteReport method in chromedash-wpt-eval-page.ts to manage the deletion flow, including a confirmation dialog to prevent accidental deletions.

UI Enhancements:
Introduced a "Delete Report" button with a trash icon in the analysis results section.
Added a success state to display "Report deleted" once the action is completed.
Defined new styles for .report-actions to layout the action buttons.

Enums & Assets:
Added DELETED to the AITestEvaluationStatus enum.
Added a new trash.svg icon asset.

Tested:
Delete button:https://screenshot.googleplex.com/5ZgLo3NVu5aM22J
Deletet successful: https://screenshot.googleplex.com/85xLsSBPr3JKc4W
Cooldown handling: https://screenshot.googleplex.com/9toyQ8TN45m8sSH

### DIFF
--- a/api/wpt_coverage_api_test.py
+++ b/api/wpt_coverage_api_test.py
@@ -224,7 +224,7 @@ class WPTCoverageAPITest(testing_config.CustomTestCase):
     params = {'feature_id': 123456}
     with test_app.test_request_context('/api/v0/wpt-coverage-reports',
                                        method='DELETE', json=params):
-      response = self.handler.do_delete()
+      response = self.handler.do_delete(feature_id=123456)
 
     self.assertEqual(response,
                          {'message': 'WPT coverage analysis report deleted.'})
@@ -258,7 +258,7 @@ class WPTCoverageAPITest(testing_config.CustomTestCase):
     with test_app.test_request_context('/api/v0/wpt-coverage-reports',
                                        method='DELETE', json=params):
       with self.assertRaises(werkzeug.exceptions.Forbidden):
-        self.handler.do_delete()
+        self.handler.do_delete(feature_id=123456)
 
     mock_can_edit.assert_called_once()
 


client-src/elements/chromedash-wpt-eval-page.ts
```diff
@@ -667,7 +696,19 @@ export class ChromedashWPTEvalPage extends LitElement {
         </section>
       `;
     }
-
+    if (
+      this.completedInThisSession &&
+      status == AITestEvaluationStatus.DELETED
```
DanielRyanSmith:
```suggestion
      status === AITestEvaluationStatus.DELETED
```

client-src/js-src/cs-client.js
```diff
@@ -781,6 +781,13 @@ export class ChromeStatusClient {
     });
   }
 
+  // Delete WPT Coverage Report
+  async deleteWPTCoverageEvaluation(featureId) {
+    return this.doDelete('/features/generate-wpt-coverage-analysis', {
+      feature_id: featureId,
+    });
+  }
+
```
DanielRyanSmith, line 788: Now that I'm thinking about it, the endpoint name `/features/generate-wpt-coverage-analysis` is not really conducive to the delete behavior.

(honestly, the original endpoint should probably have been named `/features/wpt-coverage-analysis/generate` or something similar).

For this, I would switch the endpoint to something like `/features/wpt-coverage-analysis/{featureId}/delete`. There should be examples of this kind of syntax in this file.

After that, you'll need to create the route for it in `main.py`.

client-src/js-src/cs-client.js
```diff
@@ -442,9 +442,9 @@ export class ChromeStatusClient {
     });
   }
 
-  async doDelete(resource) {
+  async doDelete(resource, body) {
     return this.ensureTokenIsValid().then(() => {
-      return this.doFetch(resource, 'DELETE', null);
+      return this.doFetch(resource, 'DELETE', body);
     });
   }
```
DanielRyanSmith: Once you update the `deleteWPTCoverageEvaluation` method, you won't need to pass the body to this method, so this change can be removed.

client-src/js-src/cs-client.js
DanielRyanSmith: Here is an example: https://github.com/GoogleChrome/chromium-dashboard/blob/14cdd3cdc543d0b178d99e325dc06bf294d37093/client-src/js-src/cs-client.js#L693-L695

DanielRyanSmith: This also will probably change the way you need to reference the feature ID in the backend code (`wpt_coverage_api.py`).

You'll probably need to reference it like this: https://github.com/GoogleChrome/chromium-dashboard/blob/14cdd3cdc543d0b178d99e325dc06bf294d37093/api/stages_api.py#L107

szy196: > Now that I'm thinking about it, the endpoint name `/features/generate-wpt-coverage-analysis` is not really conducive to the delete behavior.
> 
> (honestly, the original endpoint should probably have been named `/features/wpt-coverage-analysis/generate` or something similar).
> 
> For this, I would switch the endpoint to something like `/features/wpt-coverage-analysis/{featureId}/delete`. There should be examples of this kind of syntax in this file.
> 
> After that, you'll need to create the route for it in `main.py`.

created a issue to track renaming wpt-coverage-analysis generate endpoint https://github.com/GoogleChrome/chromium-dashboard/issues/6009

api/wpt_coverage_api.py
```diff
@@ -101,7 +101,7 @@ def do_post(self, **kwargs):
 
   def do_delete(self, **kwargs):
     """Delete the generated WPT coverage analysis report."""
-    feature_id = self.get_int_param('feature_id')
```
jcscottiii, line 104: Question: I see there are multiple places that use `self.get_int_param('feature_id')`

https://github.com/GoogleChrome/chromium-dashboard/blob/14cdd3cdc543d0b178d99e325dc06bf294d37093/api/wpt_coverage_api.py#L36

https://github.com/GoogleChrome/chromium-dashboard/blob/14cdd3cdc543d0b178d99e325dc06bf294d37093/api/wpt_coverage_api.py#L104

Was there a reason we switched this to kwargs?

I'm fine with changing it kwargs, but we should be consistent in other places too.

jcscottiii: The change to kwargs in other places can be done in a follow up PR too.

szy196: > The change to kwargs in other places can be done in a follow up PR too.

Sounds good! I'm planning to address this in https://github.com/GoogleChrome/chromium-dashboard/issues/6009

Issue #6009, opened 2026-02-12
Rename wpt-coverage-analysis generate endpoint


szy196:
According to comment https://github.com/GoogleChrome/chromium-dashboard/pull/6003#discussion_r2795995256 , we shall align endpoint naming for generate and delete

PR #6013, opened 2026-02-12
Rename wpt-coverage-analysis endpoint

szy196:
TESTED = local server run pass

API Endpoint:

Updated the routes in main.py from a generic POST endpoint (/api/v0/generate-wpt-coverage-analysis) to a more RESTful, resource-oriented structure: /api/v0/features/wpt-coverage-analysis/<int:feature_id>/generate and /api/v0/features/wpt-coverage-analysis/<int:feature_id>/delete.

Backend Logic Changes

In api/wpt_coverage_api.py, the do_post method was updated to retrieve feature_id directly from keyword arguments (kwargs) passed by the router, rather than as a request parameter.

Client-Side Update

Modified client-src/js-src/cs-client.js to point the generateWPTCoverageEvaluation function to the new URL format.

Cleanup

Removed deprecated "-evaluation" routes and added a TODO to remove the old "-analysis" route once the transition to the new path-based ID route is fully landed.

part of https://github.com/GoogleChrome/chromium-dashboard/issues/6009

### DIFF
--- a/main.py
+++ b/main.py
@@ -156,12 +156,14 @@ api_routes: list[Route] = [
           intents_api.IntentsAPI),
     Route(f'{API_BASE}/features/<int:feature_id>/<int:stage_id>/<int:gate_id>/intent',
           intents_api.IntentsAPI),
-    Route(f'{API_BASE}/features/wpt-coverage-analysis/<int:feature_id>/generate',
+    Route(f'{API_BASE}/features/<int:feature_id>/wpt-coverage-analysis',
           wpt_coverage_api.WPTCoverageAPI),
+    # TODO(suzyliu): Remove the "delete" route after the
+    # unified route had landed.
     Route(f'{API_BASE}/features/wpt-coverage-analysis/<int:feature_id>/delete',
           wpt_coverage_api.WPTCoverageAPI),
     # TODO(suzyliu): Remove the "generate-wpt-coverage-analysis" route after the
-    # "{featureId}/generate" route had landed.
+    # unified route had landed.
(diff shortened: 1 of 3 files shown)


main.py
```diff
@@ -156,14 +156,14 @@ class Route:
           intents_api.IntentsAPI),
     Route(f'{API_BASE}/features/<int:feature_id>/<int:stage_id>/<int:gate_id>/intent',
           intents_api.IntentsAPI),
+    Route(f'{API_BASE}/features/wpt-coverage-analysis/<int:feature_id>/generate',
+          wpt_coverage_api.WPTCoverageAPI),
     Route(f'{API_BASE}/features/wpt-coverage-analysis/<int:feature_id>/delete',
           wpt_coverage_api.WPTCoverageAPI),
```
jcscottiii, line 164: If we want to be RESTful, we should instead make these two routes like this:

```
POST {API_BASE}/features/<int:feature_id>/wpt-coverage-analysis
DELETE {API_BASE}/features/<int:feature_id>/wpt-coverage-analysis
```
So here it would be:

```
    Route(f'{API_BASE}/features/<int:feature_id>/wpt-coverage-analysis',
          wpt_coverage_api.WPTCoverageAPI),
```

Also both the POST (generate) and the DELETE (delete analysis) operations will then go to the same handler in a single route configuration instead of two.

DanielRyanSmith: Yes, the way James described it is the right way to do this - I gave you slightly incorrect information yesterday. I would follow this advice!

PR #6031, opened 2026-02-19
Exchange a member of the test gate review team

jrobbins:
This change was requested by a test gate reviewer.


DanielRyanSmith (approved):

jrobbins:
Note: This had a lot of playwright failures that looked like a 1px difference in the padding of a shoelace menu item.  Maybe we are somehow pulling different versions of the shoelace library on different test runs?

PR #6034, opened 2026-02-19
add helper function to fetch explainer link(s)

szy196:
_fetch_explainer_link

GitHub Optimization: If a GitHub "blob" URL is provided, it automatically converts it to a raw.githubusercontent.com link. This allows the system to fetch the raw Markdown source rather than the rendered HTML, which preserves code block integrity for better processing by Gemini.

Trafilatura Integration: For non-GitHub links, it uses the trafilatura library to extract main content. 
_fetch_explainer_content

This function acts as a wrapper to handle multiple links.

It uses regex to extract URLs from input strings, calls the fetch helper for each, and concatenates the resulting content into a single string.

It includes basic error handling to skip links that fail to fetch or return error messages.


framework/gemini_helpers.py
```diff
@@ -40,6 +40,71 @@
...
+def _fetch_explainer_link(url: str) -> str:
+    """
+    Fetches explainer content with specialized handling for technical data.
+    Prioritizes raw formats for GitHub to preserve code blocks.
+    """
+    parsed = urlparse(url)
+
+    # 1. Specialized GitHub Handling (Best for code blocks)
+    # Convert blob links to raw links to get original Markdown source.
+    if "github.com" in parsed.netloc and "/blob/" in parsed.path:
```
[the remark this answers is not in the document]
DanielRyanSmith: ### What Gemini thinks about this:

This is an **actual problem**, not a false positive. GitHub’s CodeQL scanner caught a very common—and potentially dangerous—security flaw in how you are validating the domain.

Here is a breakdown of why this happens and how you can fix it.

### Why It Is a Problem

The `in` operator checks if the substring exists *anywhere* within `parsed.netloc`. Because it doesn't anchor the string, an attacker can bypass your security check by using a maliciously crafted domain that simply contains `"github.com"`.

Here are a few examples of attacker domains that would pass your current `if "github.com" in parsed.netloc` check:

* `notgithub.com`
* `github.com.evil-empire.net`
* `fake-github.com`

Depending on what your code does with this URL afterward, this flaw could lead to Server-Side Request Forgery (SSRF), Open Redirects, or malicious file downloads.

---

### How to Fix It

To properly sanitize and validate the URL, you need to be much stricter about how you evaluate `parsed.netloc`.

**Option 1: Exact Match (Recommended)**
If you only ever expect the base `github.com` domain, use the `==` operator.

```python
if parsed.netloc == "github.com" and "/blob/" in parsed.path:
    # Safe to proceed

```

**Option 2: Allowing Subdomains**
If you need to support subdomains (like `gist.github.com` or `raw.githubusercontent.com`), checking the end of the string is the standard approach. You must include the dot (`.`) in the `.endswith()` check so you don't accidentally allow `notgithub.com`.

```python
def is_valid_github_domain(netloc):
    # Allow exact match OR valid subdomains
    return netloc == "github.com" or netloc.endswith(".github.com")

if is_valid_github_domain(parsed.netloc) and "/blob/" in parsed.path:
    # Safe to proceed

```

**Note on the path check:** While the warning specifically flagged the `github.com` substring, be mindful of `"/blob/" in parsed.path` as well. If you strictly expect the path to start with `/blob/`, using `parsed.path.startswith("/blob/")` is much safer than `in`.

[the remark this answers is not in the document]
DanielRyanSmith: We're taking this information from trusted sources (our users), but it's probably best that we update this check, and also the other location where this is checked (for spec URLs).

framework/gemini_helpers.py
```diff
@@ -40,6 +40,71 @@
...
+            return f"Error: No main content found at {url}"
+
+        return result
+    except Exception as e:
+        return f"Error: Unable to process explainer: {e}"
+
+def _fetch_explainer_content(explainer_links: list[str]) -> str:
+  """4
+  Fetches and concatenates content from multiple explainer links.
+  """
```
DanielRyanSmith:
```suggestion
  """
  Fetches and concatenates content from multiple explainer links.
  """
```

DanielRyanSmith:
Whoops, I reviewed the draft PR rather than this one. I was confused as to where my review had gone 🙃

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+        )
+
+        if not result:
+            return f"Error: No main content found at {url}"
+
+        return result
+    except Exception as e:
+        return f"Error: Unable to process explainer: {e}"
+
+def _fetch_explainer_content(explainer_links: list[str]) -> str:
```
DanielRyanSmith: This name function name and the `_fetch_explainer_link` name could be renamed to improve clarity. I might name them something like `_get_explainer_contents` (since it's assembling multiple contents) and `_fetch_explainer_content` (since that function is doing the actual fetching).

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+
+    url_match = re.search(r'https?://[^\s]+', item)
+    if url_match:
+      url = url_match.group(0)
+      content = _fetch_explainer_link(url)
+
+      if content and not content.startswith("Error:"):
+        contents.append(f"## Explainer Link: {url}\n{content}")
+
+  return "\n\n".join(contents)
```
DanielRyanSmith, line 116: I think it would be best to return the list of individual explainer contents, rather than joining them as a single string. That way we can clearly partition them in the prompt.

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
 
+def _validate_github_domain(netloc: str) -> bool:
+  """Allow exact match OR valid subdomains for github.com."""
+  return netloc == 'github.com' or netloc.endswith('.github.com')
+
+
 def _create_feature_definition(feature: FeatureEntry) -> str:
   return f'Name: {feature.name}\nDescription: {feature.summary}'
 
+def _fetch_explainer_link(url: str) -> str:
```
DanielRyanSmith: In reference to my other comment, this should probably be named `_fetch_explainer_content`

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
 WPT_FILE_REGEX = re.compile(r"\/[^/]*\.[^/]*$")
 
 
+def _validate_github_domain(netloc: str) -> bool:
+  """Allow exact match OR valid subdomains for github.com."""
+  return netloc == 'github.com' or netloc.endswith('.github.com')
```
DanielRyanSmith, line 42: Thank you for handling this and putting it in a reusable function!

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+
+    # 2. Standard Webpage Extraction
+    try:
+        downloaded = trafilatura.fetch_url(url)
+        if downloaded is None:
+            return f"Error: Could not fetch URL {url}"
+
+        # CRITICAL: Use include_formatting=True and include_tables=True
+        # This helps Trafilatura preserve <pre> and <code> blocks which
+        # are common in explainers like those on open-ui.org.
```
DanielRyanSmith:
```suggestion
        # Use include_formatting=True and include_tables=True
        # This helps Trafilatura preserve <pre> and <code> blocks.
```

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+            include_formatting=True,  # Added to preserve code structure
+            output_format="markdown"
+        )
+
+        if not result:
+            return f"Error: No main content found at {url}"
+
+        return result
+    except Exception as e:
+        return f"Error: Unable to process explainer: {e}"
```
DanielRyanSmith: We should probably just return an empty string and then omit the explainer from the prompt if it isn't found.

szy196: I would prefer to keep the error message and log_warning the errors. this can help us debug if we keep seeing empty explainer.

DanielRyanSmith: Actually, now that I'm thinking about it, and because the user is actually opting into using the explainer URLs,  if we fail to find any content for explainers, it should probably halt the entire pipeline and we should mark the pipeline run as failed. Maybe we can do that instead? We can surface an error to the user in the UI about the specific reason for the pipeline failure, even by saving that to `FeatureEntry.ai_test_eval_report`.

Right now, returning the error as a string will pass the error message to the prompt as if it were the explainer content itself, which is not the behavior we want.

szy196: based on offline conversation, if user decide to opt in, we should error out if explainer link is invalid. However, I found another issue with even processing spec link: return f"Error fetching GitHub Raw: {e}" will not error out anything. It will just pass the error message to the prompt. We should raise ValueError(f"No main content found at {url}") if we want to stop the process.

DanielRyanSmith: > We should raise ValueError(f"No main content found at {url}") if we want to stop the process.

I'm in total agreement. 🙂  Thanks for the investigation!

szy196: will wait for change https://github.com/GoogleChrome/chromium-dashboard/pull/6043/changes to submit

framework/gemini_helpers.py
```diff
@@ -37,9 +37,82 @@
...
 
+def _validate_github_domain(netloc: str) -> bool:
+  """Allow exact match OR valid subdomains for github.com."""
+  return netloc == 'github.com' or netloc.endswith('.github.com')
+
+
 def _create_feature_definition(feature: FeatureEntry) -> str:
   return f'Name: {feature.name}\nDescription: {feature.summary}'
 
+def _fetch_explainer_content(url: str) -> str:
```
DanielRyanSmith, line 48: This function is indented using 4 spaces instead of 2. 2-space indentation is needed to match the rest of the file

PR #6039, opened 2026-02-19
enable optional explainers in prompts

szy196:


templates/prompts/unified-gap-analysis.html
```diff
@@ -159,6 +160,12 @@
   Description: {{ feature_summary }}
 </feature_definition>
 
+{% if explainer_content %}
+<explainer_document>
+{{ explainer_content }}
+</explainer_document>
+{% endif %}
```
DanielRyanSmith: Ideally, we have a block here named something like `<explainer_documents>` that contains 1 or more `<explainer_document>` blocks, that way we're clearly partitioning different documents for the LLM. You can look at the `<test_suite>` block for an example on how to handle that.

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+        )
+
+        if not result:
+            return f"Error: No main content found at {url}"
+
+        return result
+    except Exception as e:
+        return f"Error: Unable to process explainer: {e}"
+
+def _fetch_explainer_content(explainer_links: list[str]) -> str:
```
DanielRyanSmith: This name function name and the `_fetch_explainer_link` name could be renamed to improve clarity. I might name them something like `_get_explainer_contents` (since it's assembling multiple contents) and `_fetch_explainer_content` (since that function is doing the actual fetching).

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+
+    url_match = re.search(r'https?://[^\s]+', item)
+    if url_match:
+      url = url_match.group(0)
+      content = _fetch_explainer_link(url)
+
+      if content and not content.startswith("Error:"):
+        contents.append(f"## Explainer Link: {url}\n{content}")
+
+  return "\n\n".join(contents)
```
DanielRyanSmith, line 116: I think it would be best to return the list of individual explainer contents, rather than joining them as a single string. That way we can clearly partition them in the prompt.

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
 
+def _validate_github_domain(netloc: str) -> bool:
+  """Allow exact match OR valid subdomains for github.com."""
+  return netloc == 'github.com' or netloc.endswith('.github.com')
+
+
 def _create_feature_definition(feature: FeatureEntry) -> str:
   return f'Name: {feature.name}\nDescription: {feature.summary}'
 
+def _fetch_explainer_link(url: str) -> str:
```
DanielRyanSmith: In reference to my other comment, this should probably be named `_fetch_explainer_content`

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
 WPT_FILE_REGEX = re.compile(r"\/[^/]*\.[^/]*$")
 
 
+def _validate_github_domain(netloc: str) -> bool:
+  """Allow exact match OR valid subdomains for github.com."""
+  return netloc == 'github.com' or netloc.endswith('.github.com')
```
DanielRyanSmith, line 42: Thank you for handling this and putting it in a reusable function!

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+
+    # 2. Standard Webpage Extraction
+    try:
+        downloaded = trafilatura.fetch_url(url)
+        if downloaded is None:
+            return f"Error: Could not fetch URL {url}"
+
+        # CRITICAL: Use include_formatting=True and include_tables=True
+        # This helps Trafilatura preserve <pre> and <code> blocks which
+        # are common in explainers like those on open-ui.org.
```
DanielRyanSmith:
```suggestion
        # Use include_formatting=True and include_tables=True
        # This helps Trafilatura preserve <pre> and <code> blocks.
```

framework/gemini_helpers.py
```diff
@@ -37,9 +37,79 @@
...
+        result = trafilatura.extract(
+            downloaded,
+            include_comments=False,
+            include_tables=True,
+            include_formatting=True,  # Added to preserve code structure
+            output_format="markdown"
+        )
+
+        if not result:
+            return f"Error: No main content found at {url}"
```
DanielRyanSmith: We should probably just return an empty string and then omit the explainer from the prompt if it isn't found.

templates/prompts/unified-gap-analysis.html
```diff
@@ -154,6 +157,12 @@
 {{ spec_content }}
 </spec_document>
 
+{% if explainer_content %}
+<explainer_documents>
+  {{ explainer_content }}
+</explainer_documents>
+{% endif %}
```
DanielRyanSmith:
```suggestion
{% if explainer_content %}
<explainer_documents>
  {{ explainer_content }}
</explainer_documents>
{% endif -%}
```

This should remove the extra spaces/newlines that are being generated in these templates when the explainer is not present.

templates/prompts/unified-gap-analysis.html
```diff
@@ -2,13 +2,16 @@
...
-**Core Directive:** I will provide you with three inputs:
+**Core Directive:** I will provide you with the following inputs:
 
-1.  `<spec_document>`: The technical specification text (this may contain the full spec, a large section of the spec, or a pull request diff highlighting the changes to the spec to include the new feature).
-2.  `<feature_definition>`: The Name and Description of the specific feature you are auditing. **You must use this to define the scope of your analysis.**
-3.  `<test_suite>`: One or more raw test files (including dependencies) written to test this feature.
+*   `<spec_document>`: The technical specification text (this may contain the full spec, a large section of the spec, or a pull request diff highlighting the changes to the spec to include the new feature).
+{% if explainer_content %}
+*   `<explainer_documents>`: Additional technical documentation and context for the feature.
+{% endif %}
```
DanielRyanSmith:
```suggestion
{% if explainer_content %}
*   `<explainer_documents>`: Additional technical documentation and context for the feature.
{% endif -%}
```
This should remove the extra spaces that are being generated in these templates when the explainer is not present.

templates/prompts/spec-synthesis.html
```diff
@@ -35,6 +35,11 @@
 <spec_document url="{{ spec_url }}">
   {{ spec_content }}
 </spec_document>
+{% if explainer_content %}
+<explainer_documents>
+  {{ explainer_content }}
+</explainer_documents>
+{% endif %}
```
DanielRyanSmith:
```suggestion
{% if explainer_content %}
<explainer_documents>
  {{ explainer_content }}
</explainer_documents>
{% endif -%}
```
This should remove the extra spaces that are being generated in these templates when the explainer is not present.

templates/prompts/spec-synthesis.html
```diff
@@ -1,9 +1,9 @@
 Role: You are an expert technical writer and web platform architect. You excel at deconstructing large, dense specification documents and synthesizing new, focused documentation for a specific feature.
 
-Core Directive: I will provide you with input data in XML format containing a `<spec_document>` and a `<feature_definition>`. Your task is to:
+Core Directive: I will provide you with input data in XML format containing a `<spec_document>`, {% if explainer_content %}`<explainer_documents>`, {% endif %}and a `<feature_definition>`. Your task is to:
```
DanielRyanSmith:
```suggestion
Core Directive: I will provide you with input data in XML format containing a `<spec_document>`{% if explainer_content %}, `<explainer_documents>`,{% endif %} and a `<feature_definition>`. Your task is to:
```

This should fix some spacing issues and only add a comma if there are 3 items.

PR #6059, opened 2026-02-26
frontend: optional explainer links for AI report gen

szy196:
TESTED:
Explainer provided:
https://screenshot.googleplex.com/8x7zbaPkRnnmiZn

Missing: https://screenshot.googleplex.com/8ZPx5sF9HbJEv8d
Can still generate

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -17,7 +17,7 @@ jobs:
         - name: Run your tests
           timeout-minutes: 30
           run: npm run pwtests-shutdown; npm run pwtests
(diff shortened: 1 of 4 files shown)


szy196:
will fix the test once we have more detailed UX design. we are allowing users to optionally have explainers (which could be provided already, or they can upload by edit) to generate the report. if they don't provide, I don't want to show red cross on them, since it's optional. and we want to make it very clear that it's optional.

szy196:
duplicates: https://github.com/GoogleChrome/chromium-dashboard/pull/6077

PR #6060, opened 2026-02-27
Update to use Gemini 3.1

DanielRyanSmith:
Gemini 3.0 will be discontinued from the Gemini API on March 9th.


DanielRyanSmith:
I confirmed that I was able to generate a report locally using Gemini 3.1.

PR #6065, opened 2026-03-02
Bypass Chromium file validation when file content is empty

DanielRyanSmith:
Adjust OriginTrialsAPI validation logic to skip checks for Chromium source files that are empty strings. This is useful for environments where these files may not be available. A new helper method `_is_in_file` is extracted to handle this bypass consistently across different file checks.

- Add `_is_in_file` helper to `OriginTrialsAPI`.
- Update `_validate_creation_args` to skip JSON parsing and substring checks for empty files.
- Add regression test for empty file bypass.

### DIFF
--- a/api/origin_trials_api.py
+++ b/api/origin_trials_api.py
@@ -19,6 +19,7 @@ import json5
 import re
 import requests
 import validators
+from typing import Any
 
 from chromestatus_openapi.models import (CreateOriginTrialRequest, GetOriginTrialsResponse, SuccessMessage)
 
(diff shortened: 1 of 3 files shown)


api/origin_trials_api.py
```diff
@@ -170,7 +177,8 @@ def _validate_creation_args(
         validation_errors['ot_webfeature_use_counter'] = (
             'No UseCounter specified for non-deprecation trial.')
       elif (uc_type == BlinkHistogramID.web_feature and
-            f'{webfeature_use_counter} =' not in chromium_files['webfeature_file']):
+            not self._is_in_file(chromium_files.get('webfeature_file'),
```
jrobbins, line 181: This will reject the user input if the file doesn't load because the `not` is on the whole expression.  You could try passing an `invert` keyword to _is_in_file.

DanielRyanSmith: I think this works the way it's supposed to when a file does not load

`is_in_file` returns  true (no file content) -> negated to false, so no validation error is added

jrobbins: Good point, I had it backwards.

PR #6076, opened 2026-03-02
Increase playwright snapshot timeout

jrobbins:
Try a longer timeout when checking screenshots because sometimes icons and CSS takes longer than 5s to load.

The longer timeout does not seem to increase the total run time by much.  Maybe from 6m to 7m.  It seems to reduce the number of flakey failures, but not eliminate them.   Also, having a screenshot helper function is a starting point to eventually having tests that support dark mode.

### DIFF
--- a/packages/playwright/tests/test_utils.js
+++ b/packages/playwright/tests/test_utils.js
@@ -380,6 +380,6 @@ export async function expectScreenshot(page, name, options, hoverElement) {
     await expect(hoverElement).toBeVisible();
     await hoverElement.hover();
   }
-  await page.waitForTimeout(100);
+  await page.waitForTimeout(250 + 100); // Some shoelace animations take 250ms.
   await expect(page).toHaveScreenshot(`${name}.png`, options);
 }


jrobbins:
> Actually , we might want to take screenshots that test states where the mouse hovers over an element. Maybe we can add the mouse movement adjustment as needed, separate from the screenshot function?

I added the ability to have a hover element, but it is currently not used.  Instead I have it always moving the mouse to (0, 0).  With that, and the longer timeout, and a brief delay, it seems to pass consistently.

packages/playwright/tests/__screenshots__/chromedash-guide-feature-page_pwtest.js/add-an-origin-trial-stage/create-origin-trial-stage-dialog-chromium-linux.png
DanielRyanSmith, line 1: I think these images are no longer used, since the screenshot check for this has been removed. The 4 images that depict this can be deleted. 🙂

jrobbins: Removed.

jrobbins:
There was one playwright failure because it took more than 30s to create a feature entry and display the feature detail page.   That happened today during the daytime, whereas I had many successes in a row yesterday evening.  So, it may have to do with GitHub CI load.  The way to resolve it might be to make our feature creation more efficient.

PR #6080, opened 2026-03-03
Dont try to delete the feature_freeze key if it is not there

jrobbins:
We started getting errors about this key not being found in the response from chromiumdash.  We don't need that field, so let's make the code that deleted it more robust.


jrobbins:
This is already live.

PR #6106, opened 2026-03-09
feat: return to eval page after editing

szy196:
**chromedash-wpt-eval-page.ts:**
Updated the "Edit" link to include a intent=wpt_eval query parameter.

**chromedash-guide-editall-page.ts**
Added getNextPage() logic to check for the intent parameter.

If the intent is wpt_eval, the user is redirected back to the /ai-coverage-analysis page.

Otherwise, it defaults to the standard /feature/ page, ensuring no regression for other editing workflows.


DanielRyanSmith:
After trying this on staging, this seems to work for all of the "Edit" links, except for the explainer "Edit" link.

PR #6111, opened 2026-03-12
Make our JS build process closer to webstatus

jrobbins:
We can't upgrade `eslint` because babel has a dependency on the old version.  So, this PR eliminates `babel`, `uglyJS`, `babel-minify`, and `gulp` and replaces them with a more direct call to `rollup` which now uses `terser`.  This makes chromestatus's JS build process closer to webstatus.dev and it should unblock our `eslint` upgrade.

---
*PR created automatically by Jules for task [9860918483722916965](https://jules.google.com/task/9860918483722916965) started by @jrobbins*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,6 @@ jobs:
 
       - name: installation
         run: |
-          npm install -g gulp
           npm config set script-shell bash --global
           make setup
 
--- a/.github/workflows/web_tests.yml
+++ b/.github/workflows/web_tests.yml
@@ -43,9 +43,7 @@ jobs:
 
       - name: installation
         run: |
-          npm install -g gulp
-
-      - run: npm config set script-shell bash --global
+          npm config set script-shell bash --global
       - run: make setup
       - name: Install playwright browsers
         run: npx playwright install --with-deps
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ For a one-click setup that leverages devcontainers, check out the devcontainer
     1. `gcloud components install beta`
 1. Install other developer tools commands
     1. node and npm.
-    1. Gulp: `npm install --global gulp-cli`
     1. Python virtual environment: `sudo apt install python3.11-venv`
 1. We recommend using node version 24
     1. Use `node -v` to check the default node version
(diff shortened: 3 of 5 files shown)


jrobbins:
@jules Good, now replace usage of the `babel-minify` module with the `terser` module for javascript minification.

jrobbins:
@jules.  Good.  Now, eliminate `gulp` and the file `gulpfile.babel.js`.  Instead have `Makefile` call roll-up directly and create a new roll-up config file that accomplishes the same goal as the old `gulpfile.babel.js` file.

jcscottiii:
@jules You'll want to adjust this comment on this line which still refers to babel
https://github.com/GoogleChrome/chromium-dashboard/blob/1af74ca3ff7a60e68f142ff195db9ed5d18e3dec/client-src/components.js#L5

jcscottiii:
It may not work because I prematurely sent it 😢. But here's another one.

@jules, Adjust this comment because it also refers to babel still: https://github.com/GoogleChrome/chromium-dashboard/blob/1af74ca3ff7a60e68f142ff195db9ed5d18e3dec/README.md?plain=1#L45

Makefile
```diff
@@ -37,13 +37,16 @@ stop-emulator:
 
 build: tsc-clean
 	npx tsc
-	npx gulp
+	mkdir -p static/css
+	cp node_modules/@shoelace-style/shoelace/dist/themes/light.css static/css/base.css
```
jcscottiii, line 41: One thing I do that is different is that in webstatus, we import our external css files at the same time when we import the components:

https://github.com/GoogleChrome/webstatus.dev/blob/d3a6450914563406764e20b336d752a069096892/frontend/src/static/js/index.ts#L39

Not sure if that matters. But something you may want to consider. You would need to use [rollup-plugin-css-only](https://github.com/GoogleChrome/webstatus.dev/blob/d3a6450914563406764e20b336d752a069096892/frontend/rollup.config.mjs#L20) though.

jrobbins:
@jules You'll want to adjust this comment on this line which still refers to babel

https://github.com/GoogleChrome/chromium-dashboard/blob/1af74ca3ff7a60e68f142ff195db9ed5d18e3dec/client-src/components.js#L5

Makefile
jrobbins: I like that.  I'll come back to that after I accomplish the main goal of getting eslint upgraded.

PR #6129, opened 2026-03-18
add explainer_links in form-definition.ts for all feature types

### LINKED ISSUE #5826: "Explainer Link" field should be available for all feature types
per conversations, this field is always useful to DevRel if it is available.

szy196:
resolves #5826


DanielRyanSmith (approved):
Thanks for fixing this! This will make it so that everyone can provide explainer links for the WPT coverage reports. 🙂

PR #6132, opened 2026-03-19
Use rollup for combined CSS.

jrobbins:
This is a follow-up to #6111 to address a review comment https://github.com/GoogleChrome/chromium-dashboard/pull/6111#discussion_r2926637815

In this PR:
* Use a `rollup` plugin to copy/combine CSS files rather than just unix `cp` in `Makefile`.  We only have one file to process this way, but this establishes a pattern that could handle more.
* Rename the generated CSS file to from `static/css/main.css` to `static/dist/styles.css` so that it is more like webstatus.dev and to avoid confusion over a previous hand-maintained main.css file.
*  Update golden HTML for unit tests and delete an old golden that was no longer being used.


DanielRyanSmith:
I'm going to merge this and put a new version up on staging. 🙂

jrobbins:
> I'm going to merge this and put a new version up on staging. 🙂

OK thanks!

PR #6140, opened 2026-03-20
Fix: Default user-facing boolean fields in FeatureEntry to None

DanielRyanSmith:
### Description

In the `FeatureEntry` model, several user-facing boolean fields automatically defaulted to `False`. This made it look like the user had explicitly selected "False" when they had actually not provided a value at all.

This PR removes `default=False` from these user-facing fields (such as `breaking_change`, `confidential`, `api_spec`, `automation_spec`, `requires_embedder_support`, etc.) so they correctly default to `None` when unset.

### Changes Made
* Removed `default=False` for user-facing fields in `FeatureEntry`.
* Retained `default=False` for internal backend metadata fields (e.g., `deleted`, `unlisted`) to preserve existing query behavior and prevent breakage.
* Updated `api/converters_test.py` expectations to check for `None` instead of `False`.
* Updated `templates/blink/intent_to_implement.html` to use `show_value()` for `requires_embedder_support` so it correctly renders "No information provided" instead of failing or displaying nothing.
* Updated golden HTML files for email templates and intent previews.


DanielRyanSmith:
Just a small change that came up occasionally, especially in the `requires_embedder_support` context.

jrobbins:
So, if a user sees that a field says "No information provided" for some boolean field, and they want that field to be false, what do they click to explicitly set it to false?

DanielRyanSmith:
> So, if a user sees that a field says "No information provided" for some boolean field, and they want that field to be false, what do they click to explicitly set it to false?

Ah yeah, glad you mentioned this. This is difficult to handle where the user explicitly wants to display "false", because, with this change, they would need to essentially toggle and untoggle the checkbox and then submit the change. That's not feasible at all.

If we really wanted to fix this then, my first thought is switching this to a drop-down with "No value", "True", or "False". But, all in all, this is a more complex change, and I was hoping to fix the problem without any  UI changes, but what you brought up makes it more difficult. I'm just going to create an issue for this and close this PR instead. 🙂

PR #6157, opened 2026-03-24
Hide `NEEDS_WORK` gates from /myfeatures/review

### LINKED ISSUE #4675: Default to hiding NEEDS_WORK gates from /myfeatures/review
The API Owners requested that the features with their gate in the NEEDS_WORK state not be shown.  They are basically just skipping over those feature because it is the feature owner's responsibility to resolve the needed work.

I am thinking that this would be done as some kind of filter widget that adds a query term.  Or, if we expose the search box, it could be an additional term in the search box.

That query term would need to somehow specify which team it relates to.  E.g., `pending-review-by:TEAM` could mean to exclude NEEDS_WORK because those gates are blocked on the feature owners at that time.  To include NEEDS_WORK gates, the term could be changed to something like `awaiting-review-by:TEAM`.  I think that I prefer a single alternate term over using a combination of query terms that could make it tricky for the user to add another term.

DanielRyanSmith:
Fixes #4675.

Changes the default query for `/myfeatures/review` to `pending-review-by:me` which excludes `NEEDS_WORK` states. 
Also adds `awaiting-review-by:me` to query gates including `NEEDS_WORK` states.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -235,7 +235,7 @@ SIMPLE_QUERY_TERMS = [
     'deleted_unlisted=false', 'deleted_unlisted_enterprise=false',
     'pending-approval-by:me', 'starred-by:me',
     'is:recently-reviewed', 'owner:me', 'editor:me', 'can_edit:me', 'cc:me',
-    'pending-review-by:me', 'awaiting-review-by:me']
+    'pending-review-by:me']
 
 
 def process_query_term(
--- a/internals/search_test.py
+++ b/internals/search_test.py
@@ -500,9 +500,6 @@ class SearchFunctionsTest(testing_config.CustomTestCase):
     actual_pending_rev, tc = search.process_query('pending-review-by:me')
     self.assertEqual(actual_pending_rev[0]['name'], 'feature 2')
 
-    actual_awaiting_rev, tc = search.process_query('awaiting-review-by:me')
-    self.assertEqual(actual_awaiting_rev[0]['name'], 'feature 2')
-
     actual_star_me, tc = search.process_query('starred-by:me')
     self.assertEqual(actual_star_me[0]['name'], 'feature 1')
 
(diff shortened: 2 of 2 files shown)


DanielRyanSmith:
This was my attempt at using the Gemini CLI to read in an issue and resolve it in ChromeStatus. It looks like this functions as expected, and works locally.

Prompts I used:
---

I want you to obtain an issue title and description from the `GoogleChrome/chromium-dashboard` repository. After that, I want you to read the issue carefully, and then implement a fix. Please ask me questions about the requirements whenever necessary. I will provide you with the issue number in the next prompt.

---
Please read the GitHub issue #4675, then implement the code changes needed to solve it. Ask me any questions you have if you don't have enough context to complete this task. After that, finalize your changes and then create a pull request. Use your skills to accomplish this. Please confirm with me before creating the pull request.

internals/search.py
```diff
@@ -250,6 +268,10 @@ def process_predefined_query_term(
 
   if query_term == 'pending-approval-by:me':
     return process_pending_approval_me_query()
+  if query_term == 'pending-review-by:me':
+    return process_pending_review_me_query()
+  if query_term == 'awaiting-review-by:me':
```
jrobbins, line 273: Why introduce this new query term when there is the existing code for `pending-approval-by:me`? (Note: we can't remove that right away even if we wanted to because people could have existing browser tabs open that still query `pending-appoval-by:me`).

DanielRyanSmith: I was assuming it was to match naming conventions, but I was incorrect

I asked Gemini:
---
Me: _Why introduce the new query "awaiting-review-by:me"? It doesn't seem to be used, and it results in the same result as "pending-approval-by:me"._

Gemini:
I added awaiting-review-by:me because it was explicitly suggested by the author in the original issue description. Jason Robbins wrote:

> "E.g., pending-review-by:TEAM could mean to exclude NEEDS_WORK... To include NEEDS_WORK gates, the term could be changed to something like awaiting-review-by:TEAM. I think that I prefer a single alternate term over using a combination of query terms..."

The intent was to give users a specific search term they could manually type into the search box if they did want to see features in the NEEDS_WORK state, pairing with the new pending-review-by nomenclature. 

You are completely right that it produces the exact same result as the pre-existing pending-approval-by:me query (both use Gate.PENDING_STATES which includes NEEDS_WORK). I left pending-approval-by:me in place so we wouldn't break backward compatibility for anyone who has it bookmarked.

If you prefer to keep the codebase leaner and just have users rely on the old pending-approval-by:me term when they want to see everything, I can easily remove awaiting-review-by:me from the PR.

---

Interesting! It tried to adhere to the issue description

DanielRyanSmith: This term has been removed now 🙂

PR #6158, opened 2026-03-24
Redirect unauthenticated and unauthorized users from restricted SPA pages

### LINKED ISSUE #6055: Pages that the user can't  view if they're not logged in should redirect to a login screen
At the moment, the user will see a generic 403 Forbidden message.

DanielRyanSmith:
Resolves #6055

## Overview
When users attempt to access restricted SPA pages while unauthenticated or unauthorized, they are now gracefully redirected to the login screen or a safe fallback page with an alert, rather than encountering a generic 403 Forbidden error page.

## Root Cause / Motivation
Previously, SPA routes that specified permission requirements like `require_edit_feature` or `require_create_feature` in `main.py` did not explicitly enforce a `require_signin` check if it wasn't defined alongside them. Unauthenticated users were redirected initially, but on the subsequent request, the permission check failed because they were still unauthenticated, causing the server to abort with a generic `403` error. Even when logged in, if a user lacked the necessary permissions for a page, the server would also throw a `403`, bypassing the SPA and rendering a blank white error page without context.

## Detailed Changelog
* **`framework/basehandlers.py`**:
  * Updated `get_spa_template_data` to ensure any route with a permission requirement (e.g., `require_create_feature`, `require_edit_feature`) acts as if `require_signin` is `True`. This correctly intercepts unauthenticated users and returns early so the SPA can render the login prompt.
  * Added a `try...except werkzeug.exceptions.Forbidden` block to catch permission failures for logged-in users. Instead of a hard 403 abort, it redirects them to the read-only feature details page (`/feature/{feature_id}?error=403`) or the features list (`/features?error=403`) depending on the context.
* **`templates/spa.html` & `templates/_base.html`**:
  * Added logic to the initial script block to check for the `error=403` URL parameter. If present, it displays an alert advising the user that they do not have permission and may need to sign in with an authorized account.
* **`framework/basehandlers_test.py` & `pages/testdata/users_test/test_html_rendering.html`**:
  * Updated tests to anticipate the new redirect tuples and fallback routes (`/features?error=403`) instead of the previous `werkzeug.exceptions.Forbidden` assertions. Updated HTML rendering snapshots to include the new JS logic.

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -766,44 +766,65 @@ class SPAHandler(FlaskHandler):
 
 def get_spa_template_data(handler_obj, defaults):
   """Check permissions then let spa.html do its thing."""
+  # Determine if this route requires a signed-in user.
+  requires_signin = (
+      defaults.get('require_signin') or
+      defaults.get('require_create_feature') or
+      defaults.get('require_edit_feature') or
+      defaults.get('require_admin_site') or
+      defaults.get('is_enterprise_page')
+  )
+
   # Check if the page requires user to sign in
-  if defaults.get('require_signin') and not handler_obj.get_current_user():
+  if requires_signin and not handler_obj.get_current_user():
     common_data = handler_obj.get_common_data()
     if 'loginStatus=False' in common_data['current_path']:
       return {}
     return flask.redirect(settings.LOGIN_PAGE_URL), handler_obj.get_headers()
 
-  # Check if the page requires create feature permission
-  if defaults.get('require_create_feature'):
-    redirect_resp = permissions.validate_feature_create_permission(handler_obj)
-    if redirect_resp:
-      return redirect_resp
+  try:
+    # Check if the page requires create feature permission
+    if defaults.get('require_create_feature'):
+      redirect_resp = permissions.validate_feature_create_permission(handler_obj)
+      if redirect_resp:
+        return redirect_resp
 
-  # Validate the user has edit permissions and redirect if needed.
-  if defaults.get('require_edit_feature'):
-    feature_id = defaults.get('feature_id')
-    if not feature_id:
-      handler_obj.abort(500, msg='Cannot get feature ID from the URL')
-    redirect_resp = permissions.validate_feature_edit_permission(
-        handler_obj, feature_id)
-    if redirect_resp:
-      return redirect_resp
+    # Validate the user has edit permissions and redirect if needed.
+    if defaults.get('require_edit_feature'):
+      feature_id = defaults.get('feature_id')
+      if not feature_id:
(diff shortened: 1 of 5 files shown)


framework/basehandlers.py
```diff
@@ -766,44 +766,65 @@ def get_template_data(self, **defaults):
 
 def get_spa_template_data(handler_obj, defaults):
```
jrobbins, line 767: Now that we are looking for `feature_id`, it is confusing to keep this argument named `defaults`.  It would be better to call it `kwargs` because it is actually a combination of the route defaults and values parsed from the URL by flask.

framework/basehandlers_test.py
```diff
@@ -1324,7 +1326,7 @@ def test_get_spa_template_data__edit_perm_anon(self):
       defaults = {
```
jrobbins, line 1326: Likewise in the tests: defaults -> kwargs.

jrobbins (approved):
LGTM after considering a parameter rename.

jrobbins (changes requested):
Also, I tested this PR interactively locally and it didn't seem to work.  When signed out, visiting /guide/new caused an alert() followed by a completely blank page.  Also, after creating a confidential enterprise feature entry, signing out while on that feature detail page resulted in an error.

DanielRyanSmith:
> Also, I tested this PR interactively locally and it didn't seem to work. When signed out, visiting /guide/new caused an alert() followed by a completely blank page. Also, after creating a confidential enterprise feature entry, signing out while on that feature detail page resulted in an error.

I'll try to replicate - I specifically tested using `feature/{feature_id}/ai-coverage-analysis`

DanielRyanSmith:
> Also, I tested this PR interactively locally and it didn't seem to work. When signed out, visiting /guide/new caused an alert() followed by a completely blank page. Also, after creating a confidential enterprise feature entry, signing out while on that feature detail page resulted in an error.

Thanks for finding that - there's definitely a regression on the `guide/new` page. I pushed a fix for the problem, so it should be taken care of 🙂 This was a side effect of my checking the logic for `?error=403` and not `?loginStatus=False`.

The confidential feature behavior seems to be pre-existing, because it works the same way on staging.  I'll see if something can be found to fix the underlying issue, though

DanielRyanSmith:
> Also, after creating a confidential enterprise feature entry, signing out while on that feature detail page resulted in an error.

Now that I'm investigating this, I'm realizing this is intended behavior, correct? The user shouldn't receive acknowledgement that the feature exists, but they can't view it. The feature is instead treated as if it doesn't exist, which is how it is behaving (this is the same behavior that happens if you try to view a non-existent feature ID)

PR #6160, opened 2026-03-24
Fix: Exclude archived stages from roadmap

### LINKED ISSUE #6135: Archived stages are still considered for the roadmap page
The feature https://chromestatus.com/feature/5962406356320256 is currently appearing on the roadmap page twice for M147 and M150.  It does have two rollout stages with those milestones, but they have `archived: true`, so they don't appear on the feature detail page.  It is incorrect for that feature to be on the roadmap page.

The root cause of the bug is that archived stages are not excluded in the `get_in_milestone()` method.

DanielRyanSmith:
Resolves #6135

## Overview
Exclude archived stages from being displayed on the roadmap page.

## Root Cause / Motivation
Currently, when generating features to be displayed on the roadmap, archived stages are not excluded by the `get_in_milestone` function. Because they're considered when fetching stages, features would be displayed on the roadmap twice if they had archived stages.

## Detailed Changelog
* **`internals/feature_helpers.py`**: Added `Stage.archived == False` filters for all stage queries inside `get_in_milestone`.
* **`internals/feature_helpers_test.py`**: Added unit test to verify that an archived stage is successfully excluded from `get_in_milestone` queries.

### DIFF
--- a/internals/feature_helpers_test.py
+++ b/internals/feature_helpers_test.py
@@ -371,6 +371,31 @@ class FeatureHelpersTest(testing_config.CustomTestCase):
     cached_result = rediscache.get(cache_key)
     self.assertEqual(cached_result, actual)
 
+  def test_get_in_milestone__archived(self):
+    """Archived stages should not be included."""
+    self.feature_1.impl_status_chrome = 5
+    # Set shipping milestone to 1.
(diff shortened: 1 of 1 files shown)


jrobbins:
I see in index.yaml that we already have some indexes that include stage_type and various milestones.  We should watch for errors on the staging roadmap page to see if any index is missing.

PR #6161, opened 2026-03-24
Fix: Create `BaseHandlers.abort` method to accept headers

### LINKED ISSUE #5688: Create `BaseHandler.abort_with_headers` helper method
It's a little tricky to add additional headers to a response using Flask, and our current `BaseHandler.abort` helper method does not support adding headers to the response. It would be useful down the road to create a new helper method that allows us to easily abort while providing additional headers to add to the response.

See: https://flask.palletsprojects.com/en/stable/api/#flask.abort

CC @jcscottiii

DanielRyanSmith:
Resolves #5688

## Overview
Added the `BaseHandler.abort_with_headers` helper method to allow aborting a request while providing custom headers in the response. It was implemented and tested within the `wpt_coverage_api`.

## Root Cause / Motivation
Currently, `BaseHandler.abort()` does not support adding additional headers to the response when an abort happens. This is necessary for some API responses that require custom headers when an error occurs, such as standard HTTP headers like `Retry-After`. 

## Detailed Changelog
* **`framework/basehandlers.py`**: Added `abort_with_headers` method. It delegates to `flask.abort` and intercepts the `HTTPException` to inject the custom headers into the response object.
* **`framework/basehandlers_test.py`**: Added unit tests to verify `abort_with_headers` works properly with both `401` and `500` status codes.
* **`api/wpt_coverage_api.py`**: Refactored the WPT AI coverage endpoint to use the newly created `abort_with_headers` to cleanly inject a `Retry-After` header when necessary.
* **`api/wpt_coverage_api_test.py`**: Updated unit tests to verify the `Retry-After` header is correctly returned in the response object when `abort_with_headers` is triggered.

### DIFF
--- a/framework/basehandlers.py
+++ b/framework/basehandlers.py
@@ -73,44 +73,34 @@ class BaseHandler(flask.views.MethodView):
   def request(self):
     return flask.request
 
-  def abort(self, status: int, msg: Optional[str|dict]=None, **kwargs) -> NoReturn:
+  def abort(
+      self, status: int, msg: str | None=None,
+      headers: dict[str, str] | None=None, **kwargs) -> NoReturn:
     """Support webapp2-style, e.g., self.abort(400)."""
     if msg:
       if status == 500:
         logging.error(f'ISE: {msg}')
       else:
         logging.info(f'Abort {status}: {msg}')
-      flask.abort(status, description=msg, **kwargs)
-    else:
-      logging.info(f'Abort {status}')
-      flask.abort(status, **kwargs)
-
-  def abort_with_headers(self, status: int, headers: dict[str, str], msg: Optional[str|dict]=None, **kwargs) -> NoReturn:
-    """Support webapp2-style with custom headers."""
-    if msg:
-      if status == 500:
-        logging.error(f'ISE: {msg}')
-      else:
-        logging.info(f'Abort {status}: {msg}')
     else:
       logging.info(f'Abort {status}')
 
-    if isinstance(msg, dict):
-      resp = flask.make_response(flask.jsonify(msg), status)
-      for k, v in headers.items():
-        resp.headers[k] = v
-      flask.abort(resp)
-
-    try:
+    if headers is not None:
+      try:
+        if msg:
+          flask.abort(status, description=msg, **kwargs)
+        else:
+          flask.abort(status, **kwargs)
+      except werkzeug.exceptions.HTTPException as e:
+        response = e.get_response()
+        for k, v in headers.items():
+          response.headers[k] = v
+        flask.abort(response)
+    else:
       if msg:
         flask.abort(status, description=msg, **kwargs)
       else:
         flask.abort(status, **kwargs)
-    except werkzeug.exceptions.HTTPException as e:
-      response = e.get_response()
-      for k, v in headers.items():
-        response.headers[k] = v
-      flask.abort(response)
 
 
   def redirect(self, url):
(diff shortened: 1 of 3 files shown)


framework/basehandlers.py
```diff
@@ -73,18 +73,46 @@ class BaseHandler(flask.views.MethodView):
   def request(self):
     return flask.request
 
-  def abort(self, status, msg=None, **kwargs) -> NoReturn:
+  def abort(self, status: int, msg: Optional[str|dict]=None, **kwargs) -> NoReturn:
```
jrobbins: Why allow msg to be a dict?  I think all callers pass strings.

framework/basehandlers.py
```diff
@@ -73,18 +73,46 @@ class BaseHandler(flask.views.MethodView):
...
       else:
-        logging.info('Abort %r: %s' % (status, msg))
+        logging.info(f'Abort {status}: {msg}')
       flask.abort(status, description=msg, **kwargs)
     else:
-      logging.info('Abort %r' % status)
+      logging.info(f'Abort {status}')
       flask.abort(status, **kwargs)
 
+  def abort_with_headers(self, status: int, headers: dict[str, str], msg: Optional[str|dict]=None, **kwargs) -> NoReturn:
```
jrobbins: Why a new function instead of an optional `headers=` keyword to `abort()`?

framework/basehandlers.py
```diff
@@ -73,18 +73,46 @@ class BaseHandler(flask.views.MethodView):
...
+    """Support webapp2-style with custom headers."""
+    if msg:
+      if status == 500:
+        logging.error(f'ISE: {msg}')
+      else:
+        logging.info(f'Abort {status}: {msg}')
+    else:
+      logging.info(f'Abort {status}')
+
+    if isinstance(msg, dict):
```
jrobbins: The if-statement should be checking `headers` rather than `msg`.

framework/basehandlers.py
DanielRyanSmith: This is true. This was me latching to hard onto the suggestion of keeping this a separate method. It seems more user-friendly to just add an optional headers arg to the existing abort method

framework/basehandlers.py
DanielRyanSmith: This is because this is how I was handling the error message in `wpt_coverage_api.py` - but it's not something that's required or needed. I'll update the implementation to avoid this

framework/basehandlers.py
```diff
@@ -73,17 +73,35 @@ class BaseHandler(flask.views.MethodView):
...
+      try:
+        if msg:
+          flask.abort(status, description=msg, **kwargs)
+        else:
+          flask.abort(status, **kwargs)
+      except werkzeug.exceptions.HTTPException as e:
+        response = e.get_response()
+        for k, v in headers.items():
+          response.headers[k] = v
+        flask.abort(response)
```
DanielRyanSmith, line 98: I didn't quite understand what was happening here, so I asked about it:

---

### My question

In the `abort` method, when headers are provided, why do we try to abort without providing the headers? If successful, doesn't that fail to provide the headers?

---

### Answer

flask.abort does not return on success; it instead immediately throws a werkzeug.exceptions.HTTPException. Because of this, it never fails to provide the headers.

  The try/except block is an idiomatic pattern in Flask/Werkzeug to add custom headers during an abort:
   1. flask.abort(status) is called.
   2. An HTTPException is raised (e.g., NotFound for a 404).
   3. The exception is caught in the except block.
   4. The response object is retrieved from the exception, and the custom headers are added.
   5. Finally, flask.abort(response) is called to re-raise the exception with the updated headers in place.

jrobbins: Ah, I guess that is how it gets a response object that has the desired response code.  Kind of odd.

PR #6162, opened 2026-03-24
Optimize test execution time using pytest and pytest-xdist

DanielRyanSmith:
## Overview
Optimizes test execution time from ~3 minutes down to ~1 minute by leveraging `pytest` and `pytest-xdist` to run tests in parallel, while establishing strict test isolation rules.

## Root Cause / Motivation
Previously, running `make test` would execute all 1,178 tests sequentially on a single thread against a single, uncleared Datastore emulator instance. By the end of the suite, the emulator was heavily burdened with thousands of lingering entities, making database interactions increasingly slow. Global states like the `fakeredis` cache and the mocked signed-in user (`os.environ`) were also bleeding across tests, causing false positives and masking flaky behavior under the strict sequential order. 

## Detailed Changelog
* **`requirements.dev.txt`**: Added `pytest`, `pytest-xdist`, and `requests` to support parallel test execution and emulator setup.
* **`Makefile`**: Updated the `do-tests` target to intelligently detect `pytest` and use `pytest -n auto` to parallelize tests across available CPU cores, falling back to `unittest` if `pytest` is not available.
* **`conftest.py`**: Created a new configuration file to programmatically spin up an isolated, dedicated Datastore emulator for each parallel worker thread (`pytest-xdist`), and clear it via a fixture before each test.
* **`testing_config.py`**: Switched static Datastore emulator environment variables (`os.environ[...]`) to `os.environ.setdefault(...)` to allow dynamic port assignment for parallel workers. Also added cache flushing (`rediscache.redis_client.flushall()`) and explicit user sign-outs (`sign_out()`) between tests to ensure strict test isolation.
* **`api/reviews_api_test.py`**: Fixed the `test_post__bad_state` test which was implicitly relying on state bleed from previous tests. Added an explicit `sign_in` call to ensure the handler correctly processes the test scenario as an authenticated user instead of throwing a 403 Forbidden.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,19 @@
+[pytest]
+filterwarnings =
+    # Python 3.12+ deprecated datetime.utcnow() and utcfromtimestamp().
+    # Legacy App Engine and google-cloud-ndb libraries still use them internally,
+    # causing thousands of warnings during a test run.
+    ignore:.*utcfromtimestamp.*:DeprecationWarning
+    ignore:.*utcnow.*:DeprecationWarning
+
+    # The local datastore emulator uses ruamel.yaml with an 'unsafe' parsing method,
+    # which triggers a PendingDeprecationWarning in the datastore_index module.
+    ignore::PendingDeprecationWarning:google\.appengine\.datastore\.datastore_index
+
+    # Pytest incorrectly tries to collect the 'test_app' fixture/variable
+    # as a test function because its name starts with 'test_'.
+    ignore:.*cannot collect 'test_app'.*
+
+    # The ghapi library warns when running tests locally without a GitHub token.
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
This seemed useful, since Gemini takes quite a while running `make test` each time to verify its work didn't break anything. 😅

DanielRyanSmith:
It seems the warnings are primarily caused by deprecation warnings:

> Python recently deprecated datetime.utcnow() in favor of timezone-aware datetimes. However, the Google Cloud NDB library (google-cloud-ndb) and the legacy App Engine libraries still use this under the hood. Every time a test saves a database model that has an auto_now or auto_now_add property, this warning is triggered. This accounts for thousands of the warnings you see.

So we can suppress them if we can't  really take action on them

PR #6168, opened 2026-03-25
Optimize test execution speed by removing redundant teardowns and sleeps

DanielRyanSmith:
## Overview
Optimizes unit test execution speed by removing redundant Datastore entity deletions from `tearDown()` methods and replacing the hardcoded `sleep` in the Makefile with a smart polling mechanism.

## Root Cause / Motivation
1. **Redundant Teardowns**: With the recent transition to running tests via `pytest` and `pytest-xdist`, an in-memory Google Cloud Datastore emulator runs natively in Java as an isolated background process. A pytest fixture automatically resets the emulator's entire memory before *every single test function executes*. Because of this, looping over Datastore entities to manually call `.key.delete()` is completely redundant and performs unnecessary network RPC calls.
2. **Hardcoded Makefile Sleeps**: The `test:` target in the `Makefile` had a hardcoded `sleep 6` command to wait for the emulator to boot. On modern machines, the emulator boots in ~2-3 seconds, wasting valuable execution time during CI and local testing.

## Detailed Changelog
* **`api/*_test.py`**, **`internals/*_test.py`**, **`framework/*_test.py`**, **`pages/*_test.py`**:
  - Removed loops inside `def tearDown(self)` that fetch all entities of a given Kind to individually delete them.
  - Removed `tearDown()` entirely if it contained no other cleanup code or contained unused variable declarations.
* **`Makefile`**:
  - Removed `sleep 6` from the `test` target.
  - Updated the `curl` health check to use `--retry-all-errors` and `--retry-delay 1` to smartly poll the emulator until it is available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,8 +73,8 @@ stop: pwtests-shutdown
 
 test:
 	($(MAKE) start-emulator > /dev/null 2>&1 &)
-	sleep 6
-	curl --retry 4 http://localhost:15606/ --retry-connrefused
+	
+	curl --retry 10 --retry-all-errors --retry-delay 1 http://localhost:15606/
 	$(MAKE) do-tests; status=$$?; $(MAKE) stop-emulator; exit $$status
 
 webtest: build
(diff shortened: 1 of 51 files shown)


DanielRyanSmith:
The test execution process went from ~53 seconds to ~42 seconds after making this change.

PR #6169, opened 2026-03-25
Configure tsc to use NodeNext

jrobbins:
This change makes chromestatus more similar to webstatus in terms of how tsc is configured.  It also seems to be a prerequisite for upgrading to typescript 6.


jrobbins:
Hang on, I have a bad merge that reverted some recently-landed changes.  I will fix that now.

jrobbins:
OK ready for review

PR #6215, opened 2026-04-01
Remove F403 and F405 from Ruff ignore and replace wildcard imports

DanielRyanSmith:
## Overview
Removes the `F403` and `F405` Ruff lint rules from the ignored list in `pyproject.toml` and replaces all wildcard imports (`from internals.core_enums import *` and `from internals.enterprise_helpers import *`) with explicit imports throughout the codebase.

## Root Cause / Motivation
Previously, wildcard imports were widely used in the codebase (particularly in tests), causing Ruff to flag undefined locals (`F403` and `F405`). These rules were ignored in `pyproject.toml`. By replacing wildcard imports with explicit imports, we make the codebase more robust, easier to read, and fully compliant with standard Ruff linting rules.

## Detailed Changelog
* **`pyproject.toml`**: Removed `F403` and `F405` from the ignored lint rules list.
* **`api/` & `internals/`**: Replaced wildcard imports with explicit imports for constants and enums from `internals.core_enums` and `internals.enterprise_helpers`. 
* **Various source files**: Removed all obsolete `# noqa: F403` and `# noqa: F405` inline comments to finalize the cleanup.

### DIFF
--- a/api/converters.py
+++ b/api/converters.py
@@ -22,46 +22,7 @@ from typing import Any, Optional, TypedDict
 from google.cloud import ndb  # type: ignore
 
 import settings
-from internals import approval_defs, self_certify, slo
-from internals.core_enums import (
-    BEHIND_A_FLAG,
-    DEV_NO_SIGNALS,
-    ENTERPRISE_PRODUCT_CATEGORY_CHROME_BROWSER_UPDATE,
-    FEATURE_CATEGORIES,
-    FEATURE_TYPE_ENTERPRISE_ID,
-    FEATURE_TYPES,
-    GECKO_DEFER,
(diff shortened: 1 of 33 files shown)


api/converters.py
```diff
@@ -23,7 +23,45 @@
 
 import settings
 from internals import approval_defs, self_certify, slo
-from internals.core_enums import *  # noqa: F403
+from internals.core_enums import (
```
jrobbins: I think that I would prefer
`from internals import core_enums` and then reference them like `core_enums.BEHIND_A_FLAG`.  That is the way that I have seen it in most Google python code and it means that you can see where a constant comes from directly on the line where it occurs.

jrobbins (approved):

PR #6251, opened 2026-04-11
Notify adoption team of late changes

jrobbins:
This is part of the work requested for data quality by the docs team.

The release-related blog posts include the name, summary, and relevant milestones of features that are launching soon.  This PR alerts the docs team to edits made to those feature fields after the feature has been announced.  Where "announced" is defined as the feature having some milestone in a shipping stage that is `<=` the milestone that is currently on the beta channel.

In this PR:
* Add test data when fetching channel info so as to avoid network requests while running tests, and to have consistent values over time.
* Refactor the existing subscription rules logic into separate functions to keep each one small.
* Add the new rule for notification of late changes.


internals/notifier.py
```diff
@@ -203,6 +203,84 @@ def convert_reasons_to_task(
...
+    return rule_results
+
+
+def apply_subscription_rule_docs(
+    fe: FeatureEntry,
+    ship_stages: list[Stage],
+    changes: list,
+    changed_field_names: set[str],
+) -> dict[str, list[str]]:
+    """Every change to an IWA feature notifies the IWA team."""
```
DanielRyanSmith: This docstring needs to be updated I think

internals/notifier.py
```diff
@@ -203,6 +203,84 @@ def convert_reasons_to_task(
...
+    rule_results: dict[str, list[str]] = {}
+    # Case A: name or summary changing while any milestone is post-beta.
+    beta_version = fetchchannels.get_channel_version('beta')
```
DanielRyanSmith: From Gemini:

## Optimization: Blocking Uncached HTTP Request on a Hot Path

In `internals/notifier.py`, the new `apply_subscription_rule_docs` function calls `fetchchannels.get_channel_version('beta')` to determine the current beta channel milestone:

```python
# Case A: name or summary changing while any milestone is post-beta.
beta_version = fetchchannels.get_channel_version('beta')
```

However, in `internals/fetchchannels.py`, `get_channel_version` performs a synchronous HTTP request to `versionhistory.googleapis.com` without any caching. This means that **every time a feature is updated**, the notification system will perform a blocking network request. This can slow down the save operation and introduce intermittent failures or rate-limiting issues if the API becomes unreachable.

**Recommendation:**
Instead of calling `get_channel_version` directly, you should retrieve the channel versions via the already-cached `fetchchannels.get_omaha_data()` function, or introduce a caching mechanism (like `rediscache` or `@utils.memoize`) directly to `get_channel_version`. 

If using `get_omaha_data()`, you can extract the beta version from the cached JSON object:
```python
omaha_data = fetchchannels.get_omaha_data()
beta_version = next(
    (v['version'] for v in omaha_data[0]['versions'] if v['channel'] == 'beta'),
    '0.0'
)
```

DanielRyanSmith (approved):
LGTM, but Gemini found a potential optimization

internals/notifier.py
jrobbins: Fixed.

internals/notifier.py
jrobbins: I went with Gemini's suggestion.  That approach had occurred to me, but I didn't previously choose it because this code executes in a task rather than a user's request.

I didn't go with adding caching on get_channel_version() because there is a chance that we would cache out-of-sync Omaha data if we get and cache different channels at different times.  If we are going to cache, I think it is better if we always get all three together so that they are consistent with each other.

PR #6303, opened 2026-04-24
Fix annual shipping features nav links breaking Playwright tests

### LINKED ISSUE #4490: annual shipping features nav links will break playwright tests twice a year
We now have date-based logic that changes the navigation links in chromedash-drawer.ts twice a year.  That will break our screenshot matching playwright tests twice a year.  We could redact part of the image to exclude that part from image comparison, but that won't help with this change because it adds a row to the menu, thus affecting the layout of all the items lower in the menu.

DanielRyanSmith:
Resolves #4490

## Overview
Fakes the current date in Playwright tests to prevent date-based UI changes in the navigation drawer from breaking screenshot tests.

## Root Cause / Motivation
Date-based logic in `chromedash-drawer.ts` changes navigation links twice a year (e.g., adding a link for the next year after September 1st). This adds a row to the menu and shifts layout, breaking visual regression tests. By faking the date to a fixed point (Jan 1, 2026), we stabilize the UI for tests.

## Detailed Changelog
* **`packages/playwright/tests/test_utils.js`**: Added `setupFakeNow` function to override `Date` constructor and `Date.now()` in the browser context using `page.addInitScript`. Called this function in `login` to apply it by default.

### DIFF
--- a/packages/playwright/tests/test_utils.js
+++ b/packages/playwright/tests/test_utils.js
@@ -126,6 +126,7 @@ export function acceptAlertDialogs(page) {
  * @param {import("playwright-core").Page} page
  */
 export async function login(page) {
+  await setupFakeNow(page);
   page.exposeFunction('isPlaywright', () => {});
 
   // Always reset to the roadmap page.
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
Hmm, I'm a little concerned that the time will always be frozen and not progress as the tests run, so any operation would appear to take zero time.  However, none of our tests depend on time progressing.  So, this seems like a simple fix.

DanielRyanSmith:
> Hmm, I'm a little concerned that the time will always be frozen and not progress as the tests run, so any operation would appear to take zero time. However, none of our tests depend on time progressing. So, this seems like a simple fix.

Yes, we'll need to adjust for that if we update our tests to check for this

PR #6319, opened 2026-04-29
Optimize API query performance by avoiding full keys-only Datastore fetch on empty queries

### LINKED ISSUE #6317: Optimize API query performance by avoiding full keys-only Datastore fetch on empty queries
## Background & Motivation
When a user visits the main page without a search query, `process_and_operations` returns an empty list for the user query clauses. Because it's empty, `process_or_operations` falls back to `fetch_all_feature_ids_set()`, triggering a keys-only Datastore fetch of every single feature in the database. Immediately after, this massive set is intersected with the permissions set. Intersecting "all features" with "permitted features" is completely redundant and causes unnecessary overhead in Datastore operations and CPU cycles.

## Proposed Changes
Update the orchestration logic in `internals/search.py` to bypass fetching all keys when the user query is empty.

Specifically, in `process_query` (Phase 3), we will refactor the combination of user query clauses and permission clauses:

```python
    # 3a. Process user query: negation, AND, and OR.
    feature_id_future_ops = process_negation_operations(feature_id_future_ops)
    query_clauses = process_and_operations(feature_id_future_ops)

    # 3b. Process all permission ops.
    permission_clauses = process_and_operations(permissions_future_ops)

    # Apply permissions without unnecessarily fetching all keys
    if not query_clauses and not permission_clauses:
        result_id_set = fetch_all_feature_ids_set()
    elif not query_clauses:
        result_id_set = process_or_operations(permission_clauses)
    elif not permission_clauses:
        result_id_set = process_or_operations(query_clauses)
    else:
        result_id_set = process_or_operations(query_clauses)
        permission_ids = process_or_operations(permission_clauses)
        result_id_set.intersection_update(permission_ids)
```

By adding this check, we avoid the fallback in `process_or_operations` that invokes `fetch_all_feature_ids_set()` when `query_clauses` is empty.

## Acceptance Criteria
- [ ] `process_query` in `internals/search.py` is updated to avoid unnecessary `fetch_all_feature_ids_set` calls.
- [ ] Empty searches (e.g., default dashboard view) return the correct, permission-filtered results without fetching all keys from the Datastore.
- [ ] Existing tests in `internals/search_test.py` pass, and any necessary coverage for this specific condition is added.

<sub>This issue description was drafted by Gemini</sub>

DanielRyanSmith:
Resolves #6317

## Overview
Optimizes the search API query pipeline to bypass fetching all Datastore feature keys when the user query is empty.

## Root Cause / Motivation
When a user visits the main dashboard page without a search query, `process_and_operations` returned an empty list of user query clauses. Because the list was empty, `process_or_operations` fell back to `fetch_all_feature_ids_set()`, triggering a keys-only Datastore fetch of every single feature in the database. This set was then intersected with the permissions set, which is redundant since intersecting "all features" with "permitted features" just yields the permitted features. Avoiding this global fetch drastically reduces Datastore read operations and CPU cycle overhead on default views.

## Detailed Changelog
* **`internals/search.py`**: Refactored `process_query` to check if `query_clauses` and `permission_clauses` are empty before combining them, entirely skipping the fallback call to `fetch_all_feature_ids_set` for empty searches.
* **`internals/search_test.py`**: Added `test_process_query__empty_query_optimization` to explicitly verify that `fetch_all_feature_ids_set` is not called when an empty string is provided, ensuring regression coverage.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -582,13 +582,22 @@ def process_query(
     # 3a. Process user query: negation, AND, and OR.
     feature_id_future_ops = process_negation_operations(feature_id_future_ops)
     query_clauses = process_and_operations(feature_id_future_ops)
-    result_id_set = process_or_operations(query_clauses)
-    logging.info('got %r result IDs w/o permissions', len(result_id_set))
 
-    # 3b. Process all permission ops, then interesect to apply permisisons.
+    # 3b. Process all permission ops.
     permission_clauses = process_and_operations(permissions_future_ops)
-    permission_ids = process_or_operations(permission_clauses)
-    result_id_set.intersection_update(permission_ids)
+
+    # Apply permissions without unnecessarily fetching all keys
+    if not query_clauses and not permission_clauses:
+        result_id_set = fetch_all_feature_ids_set()
+    elif not query_clauses:
+        result_id_set = process_or_operations(permission_clauses)
+    elif not permission_clauses:
+        result_id_set = process_or_operations(query_clauses)
+    else:
+        result_id_set = process_or_operations(query_clauses)
+        permission_ids = process_or_operations(permission_clauses)
+        result_id_set.intersection_update(permission_ids)
+
     logging.info('got %r result IDs with permissions', len(result_id_set))
 
     # 3c. Filter out confidential features that the user cannot view.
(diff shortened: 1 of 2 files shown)


jrobbins (approved):
This should help with the initial load time on the "All features" page.

I suppose that this could be a bit cleaner if we introduced a subclass (or wrapper) of Set that also had a special "everything" value.  But, I like using with standard python Sets still.

PR #6320, opened 2026-04-29
Optimize negation parsing in search queries

DanielRyanSmith:
## Overview
This PR optimizes how search queries with negations (e.g., `-text` or `-owner:me`) are parsed and executed, improving search performance and reducing Datastore read costs.

## Root Cause / Motivation
Previously, search queries containing negations preemptively executed a global, keys-only Datastore fetch of every `FeatureEntry` in the database just to invert a single term. Because negations are usually part of an AND chain, this pre-calculated inversion was inefficient. This PR addresses the memory and Datastore read overhead by applying the mathematical logic of set subtraction directly against the ongoing intersection of previous query terms. Additionally, the query terms are now sorted so that positive terms are evaluated before negative ones, guaranteeing that the global Datastore fetch is always avoided regardless of the order the user typed the query.

## Detailed Changelog
* **`internals/search.py`**: 
  * Removed the `process_negation_operations` function which performed the preemptive global Datastore fetch. 
  * Updated `process_and_operations` to handle negation operations (`-`) in-place using Python's native `difference_update` against the current result set.
  * Added logic in `process_and_operations` to group terms by `OR` clauses and sort each group to evaluate positive terms before negative terms.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -688,37 +688,55 @@ def process_or_operations(or_clauses):
 def process_and_operations(feature_id_future_ops):
     """Process all AND operations in between OR clauses."""
     or_clauses = []
-    current_result_set = None
 
+    # Split into groups separated by 'OR'.
+    groups = []
+    current_group = []
     for logical_op, future in feature_id_future_ops:
-        if logical_op == 'OR' and current_result_set is not None:
-            # Add the preceding AND result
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
This should help in the cases identified.  Sorting the term order is a nice detail.

PR #6321, opened 2026-04-29
Optimize memory and time complexity of search total order sorting

DanielRyanSmith:
## Overview
Optimizes the `_sort_by_total_order` function in `internals/search.py` to reduce memory allocation overhead and improve time complexity during search queries.

## Root Cause / Motivation
The previous implementation allocated a large dictionary to map thousands of IDs to their indices on every search query, and then executed an `O(K log K)` sort on the results. Since the database already provides the total order as a correctly sorted sequence of IDs, this was mathematically suboptimal. By taking advantage of `O(1)` set membership lookups, we can simply stream the pre-sorted list and extract our matching items in `O(N)` time without allocating temporary dictionaries.

## Detailed Changelog
* **`internals/search.py`**: Rewrote `_sort_by_total_order` to convert the `result_id_list` into a set. It then iterates through the massive pre-sorted `total_order_ids` sequence exactly once, extracting matching items in their correct order. It appends any remaining unseen IDs to the very end to guarantee safe fallback behavior.

### DIFF
--- a/internals/search.py
+++ b/internals/search.py
@@ -360,13 +360,13 @@ def _resolve_promise_to_id_list(
 
 
 def _sort_by_total_order(
-    result_id_list: list[int], total_order_ids: list[int]
+    result_id_set: set[int], total_order_ids: list[int]
 ) -> list[int]:
     """Sort the result_ids according to their position in the total order.
 
     This extracts the matching IDs in their exact sorted sequence.
     """
-    result_id_set = set(result_id_list)
+    result_id_set = result_id_set.copy()
     sorted_id_list = []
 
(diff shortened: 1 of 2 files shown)


internals/search.py
jrobbins: We have a result ID set in the caller on line 623 that we convert to a list when calling this function.  That step is not needed now that you now want it as a set.  However, I suppose you still need to copy the set since you modify it.

jrobbins (approved):
This is a clever improvement for both execution time and RAM usage.

internals/search.py
DanielRyanSmith: Updated to accommodate this 🙂

PR #6322, opened 2026-04-30
Add announced field to API responses.

jrobbins:
This implements a request from the docs team to add an `announced` boolean to our API responses for features.  The boolean should be true if any shipping stage has a milestone that is less than the milestone that is currently on the beta channel.

In this PR:
* Add the `announced` field to the JSON dict in converters.py.
* Implement a new `is_announced()` helper function.
* Refactor channel logic into fetchchannels.py and some JSON conversion logic out of feature_helpers.py to sort out our import dependencies.

### DIFF
--- a/api/converters.py
+++ b/api/converters.py
@@ -22,7 +22,13 @@ from typing import Any, Optional, TypedDict
 from google.cloud import ndb  # type: ignore
 
 import settings
-from internals import approval_defs, core_enums, self_certify, slo
+from internals import (
+    approval_defs,
+    core_enums,
+    self_certify,
+    slo,
+    stage_helpers,
+)
 from internals.core_models import (
     FeatureEntry,
     MilestoneSet,
@@ -113,6 +119,7 @@ class StagePrepResponse(TypedDict):
     ship: Stage | None
     rollout: Stage | None
     all_stages: list[StageDict]
+    raw_stages: list[Stage]
 
 
 def _prep_stage_info(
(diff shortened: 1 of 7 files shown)


DanielRyanSmith (approved):
Gemini suggested an optimization:

# Review of PR #6322

Overall, the logic introduced in this PR is sound and implements the requested behavior correctly. However, there is a significant performance optimization opportunity.

## Optimization: N+1 Performance Issue

In `api/converters.py`, the new `announced` field is populated by calling `stage_helpers.is_announced(ship_stages)` inside both `feature_entry_to_json_basic()` and `feature_entry_to_json_verbose()`.

Inside `is_announced()`, `fetchchannels.get_current_beta_milestone()` is evaluated, which does the following:
1. Calls `get_omaha_data()`.
2. Reads the `omaha_data` string from the cache (`rediscache.get`).
3. Parses the string using `json.loads()`.

Because `feature_entry_to_json_basic()` is frequently called within a loop across hundreds or thousands of features (such as when loading the dashboard's main feature list via `/api/v0/features`), `get_current_beta_milestone()` will be executed redundantly for every single feature in the response. This will result in thousands of unnecessary network roundtrips to the cache and repeated JSON parsing overhead, which will severely degrade the API's response time.

**Recommendation:**
To avoid this N+1 performance issue, `get_current_beta_milestone()` should only be evaluated once. You could either:
1. Utilize `@functools.lru_cache(maxsize=1)` (or `flask.g` / request-scoped caching) on `get_current_beta_milestone()` or `get_omaha_data()` to avoid parsing overhead.
2. Alternatively, pre-fetch the current beta milestone in the API handler before iterating over features, and pass it down as an optional parameter to `feature_entry_to_json_basic(..., current_beta_milestone=None)` and `is_announced(..., current_beta_milestone=None)`.

Other than that optimization opportunity, the refactoring and logic for checking the shipping stages look great!

PR #6324, opened 2026-04-30
Fix mobile sign-in button missing for logged-out users

### LINKED ISSUE #4149: On mobile, no sign-in button is offered
On mobile, we move the account-related stuff into the main menu.
When you are signed in, it correctly shows your address, settings link, and sign out link.
But if you are signed out, there is no way to sign in because the sign in button is not shown.

DanielRyanSmith:
Resolves #4149

## Overview
This PR fixes an issue where the Google sign-in button was not displaying on mobile devices for logged-out users and increases the drawer width slightly to accommodate the button size.

## Root Cause / Motivation
The Google Identity Services Library requires elements to be in the light DOM, so the sign-in button is dynamically created and injected as a child of the `chromedash-drawer` component on mobile. However, when a user was logged out, the drawer's `renderAccountMenu` method returned `nothing` instead of a `<slot></slot>`, preventing the projected sign-in button from being displayed. Additionally, the standard Google Sign-In button can be wider than the drawer's original hardcoded width of 200px, causing the button to be cut off.

## Detailed Changelog
* **`client-src/elements/chromedash-drawer.ts`**: 
  * Increased `DRAWER_WIDTH_PX` from `200` to `250` so the dynamically injected sign-in button is no longer cut off when the drawer opens.
  * Updated `renderAccountMenu()` to return `html'<slot></slot>'` when `!this.user`, ensuring the dynamically appended sign-in button is properly projected into the mobile drawer's shadow DOM.

### DIFF
--- a/client-src/elements/chromedash-drawer.ts
+++ b/client-src/elements/chromedash-drawer.ts
@@ -22,6 +22,7 @@ import {User} from '../js-src/cs-client.js';
 import {IS_MOBILE, showToastMessage} from './utils.js';
 
 export const DRAWER_WIDTH_PX = 200;
+export const MOBILE_DRAWER_WIDTH_PX = 250;
 
 @customElement('chromedash-drawer')
 export class ChromedashDrawer extends LitElement {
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
[Screen recording 2026-04-29 10.33.26 PM.webm](https://github.com/user-attachments/assets/ff50a821-a8a1-4ef0-b8b6-9b0d38883b63)

jrobbins (approved):

PR #6327, opened 2026-04-30
Fix App Engine deployment file limit by updating .gcloudignore

DanielRyanSmith:
## Overview
This PR updates the `.gcloudignore` file to exclude various caches, build artifacts, and local tool configuration directories. This drastically reduces the number of files uploaded during App Engine deployments from over 10,000 files to under 600 files, preventing the `INVALID_ARGUMENT: This deployment has too many files` error.

## Root Cause / Motivation
Google App Engine enforces a strict limit of 10,000 files per deployment version. Recently, our deployments were failing because the `gcloud app deploy` command was attempting to upload over 10,060 files. The vast majority of these were local artifacts and development dependencies that were never explicitly ignored by `.gcloudignore`. By explicitly ignoring these folders, we stay well under the limit and also significantly speed up our deployment times.

## Detailed Changelog
* **`.gcloudignore`**: 
  - Added `.mypy_cache/`: Ignored because mypy type-checking cache files are strictly for local development and accounted for ~8,300 files.
  - Added `.ruff_cache/`: Ignored because the Python linter/formatter cache is only relevant locally and accounted for ~300 files.
  - Added `build/`: Ignored because intermediate build artifacts (e.g., from TypeScript) are not required by the runtime.
  - Added `gen/js/`: Ignored because the frontend JS OpenAPI generated code is already bundled by Rollup before deployment.
  - Added `gen/py/chromestatus_openapi/build/`, `.openapi-generator/`, `docs/`, `test/`: Ignored because these are unused/unnecessary OpenAPI generator artifacts that aren't needed by the Python backend runtime.
  - Added `gen/py/webstatus_openapi/build/`, `.openapi-generator/`, `docs/`, `test/`: Ignored for the same reasons above.
  - Added `.github/`, `.devcontainer/`, and `.agents/`: Ignored because they contain CI/CD configurations, development container settings, and local AI agent configurations respectively, which are never utilized in the App Engine environment.

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -50,5 +50,24 @@ tutorial-env
 # Testdata for Python tests
 testdata
 
+# Caches and local tools
+.mypy_cache/
+.ruff_cache/
+.github/
+.devcontainer/
+.agents/
+
+# Build artifacts
+build/
+gen/js/
+gen/py/chromestatus_openapi/build/
+gen/py/chromestatus_openapi/.openapi-generator/
+gen/py/chromestatus_openapi/docs/
+gen/py/chromestatus_openapi/test/
+gen/py/webstatus_openapi/build/
+gen/py/webstatus_openapi/.openapi-generator/
+gen/py/webstatus_openapi/docs/
+gen/py/webstatus_openapi/test/
+
 # API Key for local development
 ot_api_key.txt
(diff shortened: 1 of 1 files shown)


jrobbins (approved):

jrobbins:
Maybe we could also ignore all the `*_test.py` files.

PR #6329, opened 2026-04-30
Optimize roadmap cache with asynchronous, debounced rebuilds

DanielRyanSmith:
## Overview

This PR optimizes the roadmap cache by migrating the massive JSON generation logic into an asynchronous background Cloud Task. It implements a robust debounce mechanism using Redis locks and a pending state flag to condense rapid updates into a minimal number of background rebuilds without losing any cache invalidation signals.

## Root Cause / Motivation

Currently, any mutation to a `FeatureEntry` or `Stage` often triggers a blanket clearing of feature caches via `rediscache.delete_keys_with_prefix(FeatureEntry.DEFAULT_CACHE_KEY)`. Because the milestone caches (which power the roadmap) generated by `get_in_milestone()` previously shared this same `DEFAULT_CACHE_KEY` namespace, they were wiped out on almost any feature or stage edit. This triggered expensive synchronous rebuilds of the milestone caches during subsequent user requests, causing multi-second UI freezes and spiking App Engine memory.

## Detailed Changelog

* **`api/features_api.py`**: Added calls to `feature_helpers.trigger_roadmap_rebuild_if_needed('FeatureEntry')` during feature creation, update, and deletion. In `do_patch`, replaced the aggressive global `delete_keys_with_prefix` for the default cache with targeted cache deletions for the individual feature and participant caches.
* **`api/stages_api.py`**: Added calls to `feature_helpers.trigger_roadmap_rebuild_if_needed('Stage')` during stage creation, update, and deletion. Replaced blanket cache wipes with targeted individual cache deletions.
* **`api/stages_api_test.py`**: Updated tests to reflect the changed cache clearing behavior.
* **`internals/core_enums.py`**: Added `RoadmapCachePendingState` enum to manage the pending state of roadmap cache rebuilds.
* **`internals/core_models.py`**: Added a new `ROADMAP_CACHE_KEY` constant to the `FeatureEntry` model to create an isolated namespace for roadmap caches.
* **`internals/feature_helpers.py`**:
  * Updated `get_in_milestone()` to use the new `ROADMAP_CACHE_KEY` prefix instead of `DEFAULT_CACHE_KEY`, decoupling the roadmap milestone cache from the general feature cache to prevent accidental synchronous wiping. Updated it to accept an `update_cache` flag.
  * Added a new helper function `rebuild_roadmap_cache()` which iterates through recent and upcoming milestones (from `current_mstone - 1` to `current_mstone + 7`) and asynchronously pre-warms their caches by calling `get_in_milestone(update_cache=True)`.
  * Added `trigger_roadmap_rebuild_if_needed()`, which enqueues the background rebuild task.
  * Added `enqueue_roadmap_rebuild_task()`, which uses a `RoadmapRebuildPending` flag in Redis to register intent, and a `RoadmapRebuildLock` to prevent Cloud Task flooding during rapid user edits.
* **`internals/maintenance_scripts.py`**: Added a new Cloud Task handler `RebuildRoadmapCache` to run the background rebuild. The worker consumes the pending flag before rebuilding, releases the lock upon completion, and automatically re-queues another task if the pending flag was set again during execution.
* **`main.py`**: Registered the new Cloud Task route `/tasks/rebuild-roadmap-cache`.
* **`pages/guide.py`**: Added calls to `feature_helpers.trigger_roadmap_rebuild_if_needed('FeatureEntry')` when creating or updating a feature via the guide.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -237,6 +237,8 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         search_fulltext.index_feature(feature)
         self._write_stages_and_gates_for_feature(id, feature.feature_type)
 
+        feature_helpers.trigger_roadmap_rebuild_if_needed('FeatureEntry')
+
         # Remove all feature-related cache.
         rediscache.delete_keys_with_prefix(FeatureEntry.DEFAULT_CACHE_KEY)
         rediscache.delete_keys_with_prefix(FeatureEntry.SEARCH_CACHE_KEY)
@@ -572,8 +574,20 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         notifier_helpers.notify_subscribers_and_save_amendments(
             feature, changed_fields, notify=True
         )
-        # Remove all feature-related cache.
-        rediscache.delete_keys_with_prefix(FeatureEntry.DEFAULT_CACHE_KEY)
+
+        feature_helpers.trigger_roadmap_rebuild_if_needed('FeatureEntry')
+
+        # Remove individual feature cache.
+        lookup_key = FeatureEntry.feature_cache_key(
+            FeatureEntry.DEFAULT_CACHE_KEY, feature_id
+        )
+        rediscache.delete(lookup_key)
+        rediscache.delete_keys_with_prefix(
+            f'{FeatureEntry.DEFAULT_CACHE_KEY}|participant'
+        )
+        rediscache.delete_keys_with_prefix(
+            f'{FeatureEntry.DEFAULT_CACHE_KEY}|impl_order'
+        )
         rediscache.delete_keys_with_prefix(FeatureEntry.SEARCH_CACHE_KEY)
         # Update full-text index.
         if feature:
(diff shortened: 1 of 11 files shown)


jrobbins:
Performance on the milestone queries is actually pretty good now, so I don't think that we should add a lot of complexity to improve it.  However, we could do more selective cache invalidation.  E.g., instead of invalidating everything, we could invalidate just `DEFAULT_KEY|milestone|N` when the user edits any milestone fields and the old or new values include `N`.

Allowing a stale cache and using a task to refresh stale cache info could cause us headaches because we don't know exactly when that task will run.  Humans might not see their edits show up on the roadmap page and get confused. And a hypothetical playwright test that checked that editing a milestone moves it to a different roadmap card would not know how long to wait.

DanielRyanSmith:
> Performance on the milestone queries is actually pretty good now, so I don't think that we should add a lot of complexity to improve it. However, we could do more selective cache invalidation. E.g., instead of invalidating everything, we could invalidate just DEFAULT_KEY|milestone|N when the user edits any milestone fields and the old or new values include N.
>
> Allowing a stale cache and using a task to refresh stale cache info could cause us headaches because we don't know exactly when that task will run. Humans might not see their edits show up on the roadmap page and get confused. And a hypothetical playwright test that checked that editing a milestone moves it to a different roadmap card would not know how long to wait.

I'm in agreement here - this was an optimization suggested by Gemini that got overly complex in the end, and the benefit doesn't outweigh the added complexity. I'm going to close this and conclude this async cache warming experiment 🙂

PR #6330, opened 2026-04-30
Improve editable features done UX and server-side filtering

hjanuschka:
allows to mark features as "done" to hide from list (and optionally enable a "include done")
idea is to get away stuff thats not really needed for attention anymore

https://github.com/user-attachments/assets/f0e1eb64-edaf-4376-93c3-5df3a495154f


jrobbins:
Thanks! I'm always glad to see a new contributor.  This PR look pretty well implemented and complete.

However, I don't think that ChromeStatus should have an explicit field for `done` because feature owners actually need to come back to update features after launch.  Most feature owners don't have a very long list of features that they can edit, and they can easily bookmark features that they are currently working on.  Most of the cross-feature work deals with current milestones, which is supported by the roadmap view and shipping year links. So, I don't think that the need outweighs the temptation to mark done a feature that is not really all the way done.

PR #6357, opened 2026-05-08
Cleanup: Fix state mutation in render in `chromedash-feature-detail.ts`

DanielRyanSmith:
## Overview
This PR fixes a critical anti-pattern in `client-src/elements/chromedash-feature-detail.ts` where component state was mutated during the render pass.

## Root Cause / Motivation
The component used `@state()` properties (`previousStageTypeRendered` and `sameTypeRendered`) to track and number consecutive stages of the same type. Modifying these properties during render triggered unnecessary re-renders and made the behavior brittle and dependent on render frequency.

## Detailed Changelog
* **`client-src/elements/chromedash-feature-detail.ts`**:
    * Removed `@state() previousStageTypeRendered` and `@state() sameTypeRendered`.
    * Refactored `render()` to compute consecutive stage counts locally using variables within the loop.
    * Refactored `renderProcessStage()` to accept `sameTypeCount` as a parameter and removed the state mutation logic.


jrobbins (approved):
Does this also actually fix a user-visible defect?
I thought that at one point we had something where toggling something cause the diferentiation  numbers to count up.

DanielRyanSmith:
> Does this also actually fix a user-visible defect?
I thought that at one point we had something where toggling something cause the diferentiation numbers to count up.

This does not fix a user-visible defect, but [it does fix a Lit anti-pattern](https://lit.dev/docs/components/rendering/#writing-a-good-render()-method). State should not be modified within the `render()` method, because it will trigger another invocation of `render()`, which is usually unnecessarily at best and infinitely recursive at worst.

PR #6361, opened 2026-05-08
Fix: Remove `.only` from WPT eval page test

DanielRyanSmith:
## Overview
Removes `.only` from a test in `chromedash-wpt-eval-page_test.ts` to ensure all tests in the file are run.

## Root Cause / Motivation
A `.only` was left over, which caused only that specific test to run and ignored others in the file.

## Detailed Changelog
* **`client-src/elements/chromedash-wpt-eval-page_test.ts`**: Removed `.only` from the test case on line 596.

### DIFF
--- a/client-src/elements/chromedash-wpt-eval-page_test.ts
+++ b/client-src/elements/chromedash-wpt-eval-page_test.ts
@@ -593,7 +593,7 @@ describe('chromedash-wpt-eval-page', () => {
         .be.true;
     });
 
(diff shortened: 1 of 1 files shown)


jrobbins (approved):
It looks like you also fixed up some of the other tests, but that is not in the PR description.

DanielRyanSmith:
> It looks like you also fixed up some of the other tests, but that is not in the PR description.

Yes, I should have mentioned that the `.only` method was hiding some incorrectly written tests because the other tests in the file were not actually running, so that has now been fixed. 🙂

PR #6365, opened 2026-05-08
Backend performance and scalability improvements

DanielRyanSmith:
## Overview
This PR addresses several performance bottlenecks and scalability risks in the backend. It fixes N+1 queries, adds batch operations, respects NDB query limits, and optimizes chunked NDB queries by utilizing asynchronous fetches to reduce sequential network latency.

## Root Cause / Motivation
The codebase had several patterns that would not scale well with a large number of features or users, such as queries inside loops and fetching all entities at once.

## Detailed Changelog
* **`framework/utils.py`**: Added `chunk_list` utility function.
* **`api/features_api.py`**: Batched `stage.put()` calls in `_write_stages_and_gates_for_feature`.
* **`api/feature_latency_api.py`**: Used `chunk_list` to avoid NDB `IN` query limit for `feature_ids`, and optimized with `fetch_async()` for concurrent execution.
* **`api/reviews_api.py`**: Used `chunk_list` to avoid NDB `IN` query limit for `stage_ids`, and optimized with `fetch_async()` for concurrent execution.
* **`internals/ot_process_reminders.py`**: Fixed N+1 query in `get_trials` by bulk fetching stages, and optimized chunked queries with `fetch_async()`.
* **`internals/reminders.py`**: Fixed N+1 query in `filter_by_milestones`, optimized chunked queries with `fetch_async()`, and used `get_multi` in `get_overdue_gates_and_features`. Added `Stage` import.
* **`internals/user_models.py`**: Batched `UserPref` puts in `get_prefs_for_emails`.
* **`internals/search_fulltext.py`**: Used pagination and cursors in `ReindexAllFeatures` to avoid memory issues, and optimized with `fetch_async()` for chunked queries.
* **`internals/ot_process_reminders_test.py`**: Updated test to pass `stage_map` to `build_trial_data`.

### DIFF
--- a/api/features_api.py
+++ b/api/features_api.py
@@ -250,18 +250,23 @@ class FeaturesAPI(basehandlers.EntitiesAPIHandler):
         # Obtain a list of stages and gates for the given feature type.
         stages_gates = core_enums.STAGES_AND_GATES_BY_FEATURE_TYPE[feature_type]
 
-        for stage_type, gate_types in stages_gates:
-            # Don't create a trial extension stage pre-emptively.
-            if (
-                stage_type
-                == core_enums.STAGE_TYPES_EXTEND_ORIGIN_TRIAL[feature_type]
-            ):
-                continue
+        # Filter out trial extension stages.
+        filtered_stages_gates = [
+            (st, gt)
+            for st, gt in stages_gates
+            if st != core_enums.STAGE_TYPES_EXTEND_ORIGIN_TRIAL[feature_type]
+        ]
 
-            stage = Stage(feature_id=feature_id, stage_type=stage_type)
-            stage.put()
-            new_gates: list[Gate] = []
-            # Stages can have zero or more gates.
+        # Create and put all stages in bulk to avoid N+1 puts.
+        stages = [
+            Stage(feature_id=feature_id, stage_type=st)
+            for st, _ in filtered_stages_gates
+        ]
+        ndb.put_multi(stages)
+
+        new_gates: list[Gate] = []
+        # Create gates referencing the saved stages.
+        for stage, (_, gate_types) in zip(stages, filtered_stages_gates):
             for gate_type in gate_types:
                 gate = Gate(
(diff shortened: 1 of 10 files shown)


api/features_api.py
```diff
@@ -250,18 +250,23 @@ def _write_stages_and_gates_for_feature(
...
+        ]
+        ndb.put_multi(stages)
 
-            stage = Stage(feature_id=feature_id, stage_type=stage_type)
-            stage.put()
-            new_gates: list[Gate] = []
-            # Stages can have zero or more gates.
+        new_gates: list[Gate] = []
+        # Create gates referencing the saved stages.
+        for stage, (_, gate_types) in zip(stages, filtered_stages_gates):
```
jrobbins, line 269: This is kind of tricky, but I think that it is correct.

internals/search.py
```diff
@@ -756,6 +756,13 @@ def process_and_operations(feature_id_future_ops):
 
 def fetch_all_feature_ids_set():
     """Fetch all FeatureEntry ids."""
-    all_feature_keys = FeatureEntry.query().fetch(keys_only=True)
-    feature_ids_set = set(key.integer_id() for key in all_feature_keys)
+    feature_ids_set = set()
+    cursor = None
+    while True:
+        keys, cursor, more = FeatureEntry.query().fetch_page(
+            1000, keys_only=True, start_cursor=cursor
```
jrobbins: I wonder if the 1000 should be a higher number given that the site is working with no limit right now and we have about 4000 features.  Maybe set it to 3000 so that we have only two round-trips for the near future.

jrobbins (approved):
LGTM after one change

internals/search.py
DanielRyanSmith: Sorry, I should have caught this earlier, but I am reverting this specific update because I think it's unnecessary overall and we should just maintain the approach we already have. There shouldn't be any pagination needed when we're fetching all features each time and we're already using `key_only=True`.

PR #6367, opened 2026-05-11
Fix web test warnings related to Lit lifecycle and layout

DanielRyanSmith:
## Overview
This PR addresses various warnings in web tests by improving the timing of state updates and layout measurements in several Lit components.

This removes all of the warnings from the web tests stating: `Element X scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled.`

## Root Cause / Motivation
Warnings were likely caused by:
1. Updating state in lifecycle methods like `updated` that can trigger extra render cycles.
2. Measuring layout (`offsetWidth`) before the DOM was fully ready or stable.
3. Potential null pointer access during initialization.

## Detailed Changelog
* **`client-src/elements/chromedash-form-field.ts`**: Moved `initialValue` setup to `willUpdate` and delayed `doSemanticCheck` to avoid synchronous updates during the first render.
* **`client-src/elements/chromedash-roadmap-page.ts`**: Deferred `handleResize` to `requestAnimationFrame` and added a guard for `sectionRef.value`.
* **`client-src/elements/chromedash-roadmap.ts`**: Added defensive checks for `numColumns` and channel existence.
* **`client-src/elements/chromedash-wpt-eval-page.ts`**: Changed `updated` to `willUpdate` to safely update state based on property changes before rendering.
* **`client-src/elements/chromedash-wpt-eval-page_test.ts`**: Removed an unnecessary `setTimeout` wait in the test.

### DIFF
--- a/client-src/elements/chromedash-wpt-eval-page_test.ts
+++ b/client-src/elements/chromedash-wpt-eval-page_test.ts
@@ -609,7 +609,6 @@ describe('chromedash-wpt-eval-page', () => {
       checkbox.checked = true;
       checkbox.dispatchEvent(new Event('sl-change'));
       await el.updateComplete;
-      await new Promise(resolve => setTimeout(resolve, 0));
 
       const button = el.shadowRoot!.querySelector(
         '.generate-button'
--- a/client-src/elements/chromedash-form-field.ts
+++ b/client-src/elements/chromedash-form-field.ts
@@ -15,7 +15,14 @@
  */
 
 import {SlDetails, SlIconButton, SlInput} from '@shoelace-style/shoelace';
-import {LitElement, TemplateResult, css, html, nothing} from 'lit';
+import {
+  LitElement,
+  TemplateResult,
+  css,
+  html,
+  nothing,
+  PropertyValues,
+} from 'lit';
 import {customElement, property, state} from 'lit/decorators.js';
 import {ref} from 'lit/directives/ref.js';
 
@@ -192,8 +199,14 @@ export class ChromedashFormField extends LitElement {
       });
   }
 
+  willUpdate(changedProperties: PropertyValues) {
+    super.willUpdate(changedProperties);
+    if (!this.hasUpdated) {
+      this.initialValue = JSON.parse(JSON.stringify(this.value));
+    }
+  }
+
   firstUpdated() {
-    this.initialValue = JSON.parse(JSON.stringify(this.value));
     // We need to wait until the entire page is rendered, so later dependents
     // are available to do the semantic check.
     if (document.readyState === 'loading') {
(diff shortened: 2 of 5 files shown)


DanielRyanSmith:
Tested manually and everything seems to work as intended here.

client-src/elements/chromedash-wpt-eval-page_test.ts
```diff
@@ -609,7 +609,6 @@ describe('chromedash-wpt-eval-page', () => {
       checkbox.checked = true;
       checkbox.dispatchEvent(new Event('sl-change'));
       await el.updateComplete;
-      await new Promise(resolve => setTimeout(resolve, 0));
```
jrobbins, line 612: What was this line supposed to do when it was there?

jrobbins (approved):

client-src/elements/chromedash-wpt-eval-page_test.ts
DanielRyanSmith: I wasn't 100% certain myself, so I asked about it:

That line was a common pattern used in tests to **yield execution to the browser's event loop**. It forces the test to wait for one "tick" (macrotask) to allow any pending asynchronous operations, microtasks, or deferred callbacks to execute before proceeding.

Here is why it was likely there and why it could be safely removed in this PR:

1.  **Why it was there:**
    In the test, after checking the checkbox and dispatching the `sl-change` event, the test awaited `el.updateComplete`. In Lit, `updateComplete` only waits for the element to finish its *current* update and render cycle. If the element's reaction to the `sl-change` event triggered *another* update (e.g., by setting state in the `updated` lifecycle method) or kicked off an asynchronous operation that wasn't tracked by a promise the test could await, the test would proceed before the element was actually ready. The `setTimeout(..., 0)` was likely added as a "hack" to give the browser time to settle those untracked async tasks.

2.  **Why it could be removed in this PR:**
    As seen in the diff for `client-src/elements/chromedash-wpt-eval-page.ts`, this PR changed the component to use `willUpdate` instead of `updated` for handling property changes:
    ```diff
    -  updated(changedProperties: Map<string | symbol, unknown>) {
    +  willUpdate(changedProperties: Map<string | symbol, unknown>) {
    ```
    When you update state inside `updated`, it triggers a brand new update cycle, which often causes the warning: *"Element X scheduled an update after an update completed."* By moving that logic to `willUpdate`, Lit handles the state changes *before* the render takes place.

    Because the state updates are now correctly sequenced within a single update cycle, a single `await el.updateComplete` is now sufficient to ensure the component has fully updated and rendered. The extra "tick" provided by `setTimeout(..., 0)` became redundant and could be safely removed.

---

Basically, it was a workaround for the anti-pattern of running the tests while mutating state during an update.

PR #6372, opened 2026-05-11
Refactor: Migrate server-rendered handlers to `api/`

DanielRyanSmith:
## Overview
Moves `ot_requests.py` and `users.py` from the legacy `pages/` directory to `api/` (renamed to `*_handler.py`), and updates `main.py` routes to use the new locations.

## Root Cause / Motivation
This is part of the effort to migrate away from the legacy `pages/` directory to comply with Project Rule #3. These handlers serve HTML templates but are moved to `api/` as a stopgap without full refactoring to APIs + SPAs, focusing on easy wins.

## Detailed Changelog
* **`api/ot_requests_handler.py`**: Migrated from `pages/ot_requests.py`.
* **`api/users_handler.py`**: Migrated from `pages/users.py`.
* **`api/ot_requests_test.py`**: Migrated from `pages/ot_requests_test.py`.
* **`api/users_test.py`**: Migrated from `pages/users_test.py`.
* **`api/testdata/users_test/test_html_rendering.html`**: Moved from `pages/testdata/`.
* **`main.py`**: Updated imports and routes for these handlers.

### DIFF
--- a/api/ot_requests_handler.py
+++ b/api/ot_requests_handler.py
@@ -1,0 +1,88 @@
+# Copyright 2023 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Provides the admin page handler for viewing Origin Trial requests."""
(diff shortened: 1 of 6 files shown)


jrobbins (changes requested):
These are still pages rather than APIs.  I think we should convert them to lit page elements on the client-side and have them use true API calls rather than server-side HTML templates.  Once that is done, we could delete these pages.  Moving the files would make it seem like that task is already done but it has not been done yet.

DanielRyanSmith:
> These are still pages rather than APIs. I think we should convert them to lit page elements on the client-side and have them use true API calls rather than server-side HTML templates. Once that is done, we could delete these pages. Moving the files would make it seem like that task is already done but it has not been done yet.

Agreed here, and this will take some work. I think we can retire `guide.py`, but not yet the entire `pages/` directory without some additional work converting Lit elements. (Sorry, this was a change I hadn't aggressively reviewed yet)

PR #6386, opened 2026-05-12
Standardize Client-Side Routing

DanielRyanSmith:
## Overview
Standardize client-side routing by replacing hard-coded `window.location.href` changes with a new `navigate` helper function that wraps `page.js`.

## Root Cause / Motivation
The project has standardized on `page.js` for routing, but many components were still using `window.location.href` for internal navigation, causing full page reloads and bypassing the SPA router.

## Detailed Changelog
* **`client-src/elements/utils.ts`**: Added `navigate` helper function.
* **`client-src/elements/chromedash-guide-metadata-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-guide-new-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-guide-stage-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-guide-verify-accuracy-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-ot-creation-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-ot-extension-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-post-intent-dialog.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-header.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-stack-rank-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-guide-editall-page.ts`**: Replaced `window.location.href` with `navigate`.
* **`client-src/elements/chromedash-feature-pagination.ts`**: Replaced `window.location.href` with `navigate`.


DanielRyanSmith:
This change caused a flaky Playwright test because we were fully reloading the page, which meant when the user created a feature, the editable_features list was not reloaded, which meant the user could not see the "Edit" buttons.

This has now been fixed in this PR, and is a scenario where the Playwright tests caught something important, rather than just being a nuisance. 🙂

client-src/elements/chromedash-feature-page.ts
```diff
@@ -161,6 +161,7 @@ export class ChromedashFeaturePage extends LitElement {
       window.csClient.getStars(),
       window.csClient.getFeatureProgress(this.featureId),
       window.csClient.getChannels(),
+      window.csClient.getPermissions(),
```
jrobbins, line 164: Why is this change in this PR?

jrobbins (changes requested):
There seems to be an unrelated change to get permissions.

Also, I would expect no screenshots to be different.  Does the behavior of `page(URL)` not scroll to the top of the page after navigation?  If so, the `navigate()` function might need to scroll to the top.

Another thought: If nothing causes the page to ever reload, then users are more likely to have an old version of the app running for multiple weeks.  It might be a good idea to introduce an age-check in `navigate()` that would do a full page load if the app had loaded from the server more than a day ago.

client-src/elements/chromedash-feature-page.ts
DanielRyanSmith: Sorry, I should have explained this one - this is a change that came out of the comment above regarding the Playwright tests. I asked Gemini to explain it to me, but didn't post it here (but I've still got the convo history)

> Can you explain to me what the problem was?

The problem was caused by the switch from full page reloads to client-side SPA navigation in the PR, which led to stale user permissions.

Here is the breakdown of what happened:
                                                                                                                                                                      
  1. Old Behavior: When a new feature was created, the application would set  window.location.href = '/feature/<id>' , triggering a full page reload. This caused the application to re-fetch the user's permissions from the server, including the list of features the user is allowed to edit ( editable_features ).                                                                                                                                                                                              
  2. New Behavior in PR: The PR changed this to use client-side navigation ( navigate('/feature/<id>') ). This updates the URL and renders the new page without a full reload.                                                                             
  3. The Bug: Because there was no reload, the global user state in the browser was not refreshed. The newly created feature's ID was not in the user's  editable_features  list.                                                                          
  4. The Symptom: When the feature page rendered, it checked if the user could edit the feature. Since the cached list was stale, it concluded the user could not edit it, and therefore did not display the "Edit all fields" button.                     
  5. Test Failure: The Playwright tests, expecting to see and click that button immediately after creating a feature, failed with a timeout because the button was not visible.
 
The Fix: I updated the feature page to always fetch fresh permissions whenever it loads its data. This ensures that even with client-side navigation, the page has the most up-to-date access rights for the user.

DanielRyanSmith:
> Also, I would expect no screenshots to be different. Does the behavior of page(URL) not scroll to the top of the page after navigation? If so, the navigate() function might need to scroll to the top.

Something that definitely needs to be checked

> Another thought: If nothing causes the page to ever reload, then users are more likely to have an old version of the app running for multiple weeks. It might be a good idea to introduce an age-check in navigate() that would do a full page load if the app had loaded from the server more than a day ago.

Agreed

client-src/elements/chromedash-feature-page.ts
jrobbins: OK, I can see how this is a good change.

DanielRyanSmith:
I'm going to close this for now, and if we want to revisit it later, we can. This PR has been sitting untouched for some time and there's no particular hurry to pick it back up.

PR #6389, opened 2026-05-12
Add extraction parameters and descriptions for enterprise release not…

ChangLyu:
Added new params to existing API [api/v0/features](https://chromestatus.com/api/v0/features?releaseNotesMilestone=138) to support l10n translation for chrome release notes.

### DIFF
--- a/internals/core_enums.py
+++ b/internals/core_enums.py
@@ -905,6 +905,17 @@ WEB_DEV_VIEWS = {
 }
 
 
+RELEASE_NOTES_L10N_FIELDS_DESCRIPTIONS = {
+    'features': {
+        'name': 'The primary title or name of the feature displayed at the top of each release note entry',
+        'summary': 'The main descriptive paragraph explaining the feature. It may contain Markdown formatting',
+    },
+    'stages': {
+        'rolloutDetails': 'Additional narrative or specific notes regarding a rollout step, often used for complex schedules',
+    },
+}
+
+
 PROPERTY_NAMES_TO_ENUM_DICTS = {
     'feature_type': FEATURE_TYPES,
     'category': FEATURE_CATEGORIES,
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -198,12 +198,20 @@ def _filter_out_wp_features_lacking_enterprise_approval(
     return result
 
 
-def get_features_in_release_notes(milestone: int):
-    """Fetches features slated for release notes in the specified milestone."""
+def get_features_in_release_notes(
+    milestone: int, end_milestone: int | None = None
+):
+    """Fetches features slated for release notes in the specified milestone or milestone range."""
+    actual_end_milestone = (
+        end_milestone if end_milestone is not None else milestone
+    )
+
     cache_key = '%s|%s|%s' % (
         FeatureEntry.SEARCH_CACHE_KEY,
         'release_notes_milestone',
-        milestone,
+        f'{milestone}_{end_milestone}'
+        if end_milestone is not None
+        else milestone,
     )
 
     cached_features = rediscache.get(cache_key)
@@ -287,7 +295,8 @@ def get_features_in_release_notes(milestone: int):
         for f in filter_unlisted_formatted(formatted_features)
         if (
             f['first_enterprise_notification_milestone'] is None
-            or f['first_enterprise_notification_milestone'] <= milestone
+            or f['first_enterprise_notification_milestone']
+            <= actual_end_milestone
         )
     ]
     logging.info('finished filtering')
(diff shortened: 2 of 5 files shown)


ChangLyu:
@yanndago Please take a look.

api/features_api.py
```diff
@@ -112,6 +112,14 @@ def do_search(self):
                 'total_count': len(features_in_release_notes),
             }
 
+        # Query-string parameter for l10n extraction is provided
+        l10n_start = self.get_int_arg('releaseNotesl10nFieldsStartMilestone')
+        l10n_end = self.get_int_arg('releaseNotesl10nFieldsEndMilestone')
+        if l10n_start and l10n_end:
```
yanndago: maybe we should return an error if we have one param and not the other ?

api/features_api_test.py
```diff
@@ -550,6 +550,33 @@ def test_get__in_milestone_invalid_query(self):
                 cm.exception.description,
             )
 
+    def test_get__l10n_extraction(self):
```
yanndago: in this test, please add more features:
1 feature that is valid and before the start milestone
1 feature that is valid and after the end milestone
1 feature that is in between start and end, but is not approved
Maybe create more tests

yanndago: Also test with invalid query parame, i.e one is missing or has a bad value

internals/core_enums.py
```diff
@@ -905,6 +905,17 @@ class SkyhookDashStatus(str, Enum):
 }
 
 
+L10N_DESCRIPTIONS = {
```
yanndago: nit: RELEASE_NOTES_L10N_FIELDS_DESCRIPTIONS

internals/feature_helpers.py
```diff
@@ -299,6 +299,58 @@ def get_features_in_release_notes(milestone: int):
     return formatted_features
 
 
+def get_features_for_l10n_extraction(
+    start_milestone: int, end_milestone: int
+) -> dict:
+    """Fetches and structures enterprise features in a milestone range for l10n extraction."""
```
yanndago, line 314: nit: update the comment to say that all valid features with a milestone > start_milestone are included, so in the remaining of this function we only have to check the end_milestone

internals/feature_helpers.py
yanndago, line 293: Take this in consideration in the new function.
Features that are shown fort the first time in milestones > start milestone will not be in you initial fetch, if we keep the code like this, we will be missing some features.

We can either pass the end_milestone as a optional argument for this (better idea IMO) , or do multiple iterative call to this function with all milestones, but it is probably not the best ides.

internals/feature_helpers.py
```diff
@@ -299,6 +299,58 @@ def get_features_in_release_notes(milestone: int):
...
+    # Explicitly enforce both content_reviewed and publish_ready for l10n.
+    formatted_features = [
+        f
+        for f in formatted_features
+        if f.get('is_releasenotes_content_reviewed')
+        and f.get('is_releasenotes_publish_ready')
+    ]
+
+    l10n_features = []
+    for f in formatted_features:
```
yanndago, line 329: add a comment here that we are only filtering the actual stages

yanndago (commented):
Added some comments

internals/feature_helpers.py
ChangLyu: Discussed in person, we don't filter stage anymore, we do filter in feature level, and return all stages for that filter.

internals/feature_helpers.py
ChangLyu: Added a end_milestone, also updated the cache to handle new case.
If end_milestone is none, this method should run old logic, if not none, we handle it in cache reading and filtering first_enterprise_notification_milestone before end_milestone.

internals/feature_helpers.py
ChangLyu: Since we updated get_features_in_release_notes to handle end_milestone, so we don't need to handle it here anymore.

internals/feature_helpers.py
```diff
@@ -198,12 +198,12 @@ def _filter_out_wp_features_lacking_enterprise_approval(
     return result
 
 
-def get_features_in_release_notes(milestone: int):
-    """Fetches features slated for release notes in the specified milestone."""
+def get_features_in_release_notes(milestone: int, end_milestone: int | None = None):
+    """Fetches features slated for release notes in the specified milestone or milestone range."""
```
yanndago, line 204: maybe the first thing here could be 
if not end_milestone:
  end_milestone = milestone
or a new variable named actual_end_milestone

yanndago (approved):

jrobbins (changes requested):
This code all looks like it would work and it aligns with the design doc that I previously approved.  However, now that I see the implementation I think that it is adding too much complexity to the existing features_api.py file and overloading the shape of the JSON returned from the `/api/v0/features` endpoint too much.  That endpoint is already overly complex, and it already cannot be described in OpenAPI, so I want to avoid making that problem worse.

So, I think that we should move this logic to a new endpoint, `/api/v0/releasenotes/i10n`.  I'll draft a PR right now to establish that endpoint.  If you agree with that change, we'll get it merged and you can rebase, and then move your new logic there.

ChangLyu:
> This code all looks like it would work and it aligns with the design doc that I previously approved. However, now that I see the implementation I think that it is adding too much complexity to the existing features_api.py file and overloading the shape of the JSON returned from the `/api/v0/features` endpoint too much. That endpoint is already overly complex, and it already cannot be described in OpenAPI, so I want to avoid making that problem worse.
> 
> So, I think that we should move this logic to a new endpoint, `/api/v0/releasenotes/i10n`. I'll draft a PR right now to establish that endpoint. If you agree with that change, we'll get it merged and you can rebase, and then move your new logic there.

Please go ahead with the API, that sounds good to me! @jrobbins

jrobbins (approved):

ChangLyu:
@yanndago  Hello, Could you please help review and merge this when you have a moment? (It seems I don't have merge permission)

yanndago (approved):

PR #6391, opened 2026-05-13
Use a pink background on non-prod.

jrobbins:
This should help us visually distinguish prod from non-prod sites when testing.

Since every page looks different, all the playwright screenshots are being updated.


DanielRyanSmith:
This seems like a good implementation, but the prod vs. non-prod color difference is so subtle that I can't tell much difference between them:

Screenshot 2026-05-13 2 51 41 PM

jrobbins:
How's this?
image

DanielRyanSmith (approved):
> How's this?

Love it! It actually looks pretty cool 😎

PR #6440, opened 2026-05-29
Correct the filename of the stale features test

jrobbins:
Filename `chromeash` -> `chromedash`. The old name was not mentioned anywhere because the test is run through test discovery.


DanielRyanSmith (approved):
Good catch!

PR #6450, opened 2026-06-02
Catch unicode errors in inbound emails.

jrobbins:
We had some exceptions this week for bad inbound email messages.  This adds logging so that we can see what those messages are, and catches the exception.

### DIFF
--- a/framework/sendemail.py
+++ b/framework/sendemail.py
@@ -248,8 +248,13 @@ def handle_incoming_mail(addr=None):
         # We only process plain text emails.
         if part.get_content_type() == 'text/plain':
             body = part.get_payload(decode=True)
+            logging.info('Considering: %r', body[:2000])
             if not isinstance(body, str):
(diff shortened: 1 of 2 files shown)


framework/sendemail.py
```diff
@@ -248,8 +248,13 @@ def handle_incoming_mail(addr=None):
         # We only process plain text emails.
         if part.get_content_type() == 'text/plain':
             body = part.get_payload(decode=True)
+            logging.info('Considering: %r', body[:2000])
             if not isinstance(body, str):
-                body = body.decode('utf-8')
+                try:
+                    body = body.decode('utf-8')
+                except UnicodeDecodeError:
+                    logging.info('Bad unicode')
```
DanielRyanSmith: Could this just be `logging.error` to make it easier to find?

DanielRyanSmith (approved):
LGTM, one small thought

PR #6466, opened 2026-06-03
Fix issue with rebuilding the devcontainer.

jstenback:
The devcontainer was failing to build for a couple of folks. Turns out the problem was a partial install of Chrome in Puppeteer. This fixes that by simply disabling Chrome downloads for Puppeteer as they're not actually needed, and it also makes the devcontainer gain access to npm credentials which are sometimes needed within cerain environments. This also makes git ignore the devcontainer lock file.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -68,6 +68,11 @@
 		8888
 	],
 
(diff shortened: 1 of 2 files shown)


jrobbins (approved):
I'm surprised that it was not actually needed.  Good discovery.

PR #6470, opened 2026-06-04
Add late change notifications for the enterprise team.

jrobbins:
This should resolve b/511135154.

The enterprise team requested this change to help keep on top of changes that sneak in after release notes have gone out.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -207,6 +207,10 @@ POST_BETA_RULE_REASON = (
     'Published aspects of this feature entry changed after it started beta'
 )
 POST_BETA_RULE_ADDRS = ['rachelandrew@google.com']
+ENTERPRISE_LATE_RULE_REASON = (
+    'Published aspects of this feature entry changed after it was first '
+    'included in release notes'
+)
 MILESTONE_FIELDS = {
     'shipped_milestone',
     'shipped_android_milestone',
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (commented):
It looked good to me, but Gemini found something:

---

### Logic Error: Overwriting Subscription Rule Reasons
In `internals/notifier.py`, the `apply_subscription_rules` function accumulates results from different rules using `results.update()`.

Both `apply_subscription_rule_docs` and the new `apply_subscription_rule_enterprise` can return results with the key `POST_BETA_RULE_REASON` (which corresponds to the reason "Published aspects of this feature entry changed after it started beta").

If a feature is relevant to both web platform and enterprise (e.g., a WP feature with enterprise impact), and a milestone changes post-beta, both rules will trigger:
- `apply_subscription_rule_docs` returns `{POST_BETA_RULE_REASON: POST_BETA_RULE_ADDRS}`
- `apply_subscription_rule_enterprise` returns `{POST_BETA_RULE_REASON: RELEASENOTES_NOTIFY_ADDRS}`

Because `results.update()` is used, the second update (enterprise) overwrites the first one (docs). As a result, the docs team (`rachelandrew@google.com`) will NOT be notified.

### Verification
I verified this by adding a test case `test_apply_subscription_rules__docs_and_enterprise_match` to `internals/notifier_test.py` which failed as expected because the docs team address was missing from the notifications.

### Proposed Fix
I implemented a helper `merge_rule_results` in `internals/notifier.py` to merge the lists of addresses for the same reason instead of overwriting them:

```python
def merge_rule_results(
    matching_rules: dict[str, list[str]], new_rules: dict[str, list[str]]
) -> None:
    """Merge new rule results into matching_rules, combining addresses."""
    for reason, addrs in new_rules.items():
        matching_rules[reason] = list(set(matching_rules.get(reason, []) + addrs))
```

jrobbins:
That's a copy-paste error.  I failed to update the dictionary key in one spot.  I will fix it now.

DanielRyanSmith (approved):

PR #6472, opened 2026-06-05
Consider more fields when checking for late change notifications.

jrobbins:
This is a follow-up to #6470.

I implemented the code for the Enterprise late-change notification rule by copying the similar rule for Web Platform features.  However, the Enterprise release notes display more fields, so the team will want to know if any of those fields change.

### DIFF
--- a/internals/notifier.py
+++ b/internals/notifier.py
@@ -312,6 +312,16 @@ def apply_subscription_rule_docs(
     return rule_results
 
 
+ENTERPRISE_RELEASENOTES_FIELDS = {
+    'name',
+    'summary',
+    'enterprise_product_category',
+    'rollout_platforms',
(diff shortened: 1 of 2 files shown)


internals/notifier.py
```diff
@@ -324,7 +334,7 @@ def apply_subscription_rule_enterprise(
         return {}
     rule_results: dict[str, list[str]] = {}
     # Case A: name or summary changing while any milestone is post-beta.
```
jcscottiii: Nit: Update the comment
```suggestion
    # Case A: Any enterprise release notes field changing while any milestone is post-beta.
```

jcscottiii (approved):
LGTM w/one nit to update the comment

PR #6473, opened 2026-06-05
feat: Serve localized release notes in FeaturesAPI

ChangLyu:
- Added internals/releasenotes_l10n_helpers.py with hash calculation, caching, translations fetching, and merging.
- Added tests in api/releasenotes_serving_test.py.
- Integrated merging logic into FeaturesAPI.
- Installed pymmh3 dependency.

### DIFF
--- a/internals/releasenotes_l10n_helpers.py
+++ b/internals/releasenotes_l10n_helpers.py
@@ -1,0 +1,172 @@
+# Copyright 2026 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License")
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


internals/releasenotes_l10n_helpers.py
```diff
@@ -0,0 +1,172 @@
...
+    # Check if cache is valid and fresh
+    if cached_entry and (
+        current_time - cached_entry.get('fetched_at', 0) < L10N_CACHE_TTL_SEC
+    ):
+        return cached_entry.get('data', {})
+
+    # Cache is missing or expired, attempt to fetch from SCS.
+    url = L10N_URL_TEMPLATE.format(lang=scs_lang_code)
+    try:
+        response = requests.get(url, timeout=L10N_FETCH_TIMEOUT_SEC)
```
jrobbins, line 86: How big could the response be?  Does it keep getting larger and larger as more years go by?

jrobbins (approved):
LGTM with one question

internals/releasenotes_l10n_helpers.py
ChangLyu: Right now it's about 180k data. It won't get into  unlimited size. Because in server side we will only keep around two-years translated data. When time goes by, we will also remove old data that won't show up in UI- so which don't need translation anymore.

PR #6568, opened 2026-07-06
Added self-cert checkbox for MWG.

jrobbins:
Add a 5th checkbox to the Adoption gate to ask feature owners to coordinate any needed changes to Modern Web Guidance.

### DIFF
--- a/openapi/api.yaml
+++ b/openapi/api.yaml
@@ -1191,6 +1191,8 @@ components:
           type: boolean
         adoption_mdn_drafted:
           type: boolean
+        adoption_mwg_drafted:
+          type: boolean
         launch_or_contact:
           type: string
         explanation:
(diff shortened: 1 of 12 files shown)


openapi/api.yaml
```diff
@@ -1191,6 +1191,8 @@ components:
           type: boolean
         adoption_mdn_drafted:
           type: boolean
+        adoption_mwg_drafted:
```
jcscottiii, line 1194: Question: Do we want to use an enum instead? Something like `adoption_mwg_status`? I guess the question is do we want to track more than yes or no?

I ask because we are combining yes and `no_impact` into true.

With an enum, we could have UNANSWERED, ISSUE_TRACKED, NO_IMPACT

jcscottiii (approved):
LGTM with one question to consider

openapi/api.yaml
jrobbins: I see what you mean, but I was hoping to keep the UX limited to check boxes.  If a requirement arises to be able to query the exact state of the survey questions, then I suppose we could add that complexity at that time.

PR #6581, opened 2026-07-08
Initial crawler content

jrobbins:
It turns out that (unlike Google's web search crawler) all the LLM crawlers seem to not execute JavaScript, so their training data was not including any content from ChromeStatus.com because of our SPA and web component-based implementation.  This PR is a first step in adding non-JS content for crawlers.

Rather than try to replicate our web component-based UI using SSR, we add  a new, invisible `<article>` to the bottom of the page that contains a dump of the feature details.  We use a simple semantic markup that should be easy for crawlers to parse.

### DIFF
--- a/templates/crawler_content.html
+++ b/templates/crawler_content.html
@@ -29,11 +29,11 @@
   {% endfor %}
 
   {% if crawler.pagination %}
-    <nav>
+    <div>
       {% for link in crawler.pagination %}
         <a href="{{link.href}}">{{link.name}}</a>
       {% endfor %}
-    </nav>
+    </div>
   {% endif %}
 
 </article>
--- a/templates/crawler_content.html
+++ b/templates/crawler_content.html
@@ -17,20 +17,24 @@
     <summary>{{sect.summary}}</summary>
     {% if sect.fields %}
     <dl>
-      {% for key, value in sect.fields.items() %}
-        <dt>{{key}}:</dt>
-        <dd>{{value | tojson}}</dd>
-      {% endfor %}
+      {%- for key, value in sect.fields.items() -%}
+        {% if value != None and value != [] %}
+          <dt>{{key}}:</dt>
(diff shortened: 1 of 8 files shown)


templates/crawler_content.html
```diff
@@ -0,0 +1,40 @@
...
+<article class="crawlers-only">
+
+  <header>
+    <h2>{{crawler.heading.title}}</h2>
+  </header>
+
+  <nav>
+    <a href="/roadmap">Features by milestone</a>
+    <a href="/features">All features</a>
+  </nav>
```
jcscottiii:
```suggestion
  <div>
    <a href="/roadmap">Features by milestone</a>
    <a href="/features">All features</a>
  </div>
```

Let's use div

templates/crawler_content.html
```diff
@@ -0,0 +1,40 @@
...
+    {% endif %}
+  </details>
+  {% endfor %}
+
+  {% if crawler.pagination %}
+    <nav>
+      {% for link in crawler.pagination %}
+        <a href="{{link.href}}">{{link.name}}</a>
+      {% endfor %}
+    </nav>
```
jcscottiii: Same as above
```suggestion
    <div>
      {% for link in crawler.pagination %}
        <a href="{{link.href}}">{{link.name}}</a>
      {% endfor %}
    </div>
```

jcscottiii (approved):
LGTM with two nits

jrobbins:
Chris pointed out that this could be seen as cloaking.  I'll put this on hold and explore SSR with visible semantic HTML.

jrobbins:
Also, note: the escaping was wrong.  I was doing JS escaping in an HTML context.  The `| tojson` filter causes JS escaping.  So, an input like `<b>bold</b>` ended up rendering as ` \u003cb\u003ebold\u003c/b\u003e`.

PR #6586, opened 2026-07-09
Upgrade ghapi to 2.0.0 and resolve compatibility with fastcore 2.0+

DanielRyanSmith:
## Overview
This PR upgrades the `ghapi` dependency from `1.0.14` to `2.0.0` and resolves compatibility issues with `fastcore` 2.0+ across the backend.

## Root Cause / Motivation
`ghapi` 1.x had compatibility issues with `fastcore` 2.x because the `L` class in `fastcore` 2.0+ removed the `starmap` method. Updating `ghapi` to 2.x resolves this since it natively supports the newer version of `fastcore`.
However, `ghapi` 2.x is asynchronous by default and raises `fastspec.errors.APIError` instead of `urllib.error.HTTPError`. This requires initializing the client with `sync=True` and updating the exception catching to support both error types.

## Detailed Changelog
* **`requirements.txt`**: Upgraded `ghapi` to `2.0.0` and removed the temporary `fastcore<2.0.0` pin.
* **`internals/link_helpers.py`**:
  * Added `from fastspec.errors import APIError` import.
  * Added `_get_error_code` helper function to extract HTTP status code from both `HTTPError` and `APIError`.
  * Initialized `GhApi` client with `sync=True`.
  * Caught both `HTTPError` and `APIError` in `_fetch_github_file` and `_fetch_github_issue` functions.

### DIFF
--- a/internals/link_helpers.py
+++ b/internals/link_helpers.py
@@ -21,13 +21,17 @@ import json
 import logging
 import re
 from typing import Any, Optional
-from fastspec.errors import APIError
 from urllib.error import HTTPError
 from urllib.parse import urlparse
 
 import requests
 import validators
+from fastspec.errors import APIError
 from ghapi.core import GhApi
+
+import settings
+from framework import secrets
+
 
 def _get_error_code(e: Exception) -> Optional[int]:
     """Get the HTTP status code from an exception."""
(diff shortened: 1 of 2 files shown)


jrobbins:
I just hit the same issue on head.

jrobbins (approved):

PR #6612, opened 2026-07-15
Exclude confidential features from shipping API.

jrobbins:
This addresses a concern raised in as section of an internal issue about `aggregate_shipping_features()`.

We only offer the `confidential` checkbox for enterprise features.  Enterprise features use rollout step stages.  WP features cannot normally be confidential, they can only be unlisted.  The query used to retrieve shipping stages only considers the milestone fields used with WP features.  So, it does not seem possible that any confidential features would be in the set of candidate features for this report.

However, to make that explicit, and to protect against any scenario in which there were confidential WP features in the future, this PR excludes them from the shipping features report response.

### DIFF
--- a/internals/feature_helpers.py
+++ b/internals/feature_helpers.py
@@ -1123,6 +1123,12 @@ def aggregate_shipping_features(
             logging.warning(f'Feature {stage.feature_id} not found.')
             continue
 
+        # WP feature entries cannot be confidential, so this should not matter.
+        # But, if they ever did exist, we exclude them from the report because
+        # the user is not authenticated.
+        if feature.confidential:
+            continue
+
         feature_info = build_feature_info(feature, stage)
 
         # PSA features do not require strict validation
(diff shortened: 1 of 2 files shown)


internals/feature_helpers.py
```diff
@@ -1123,6 +1123,12 @@ def aggregate_shipping_features(
             logging.warning(f'Feature {stage.feature_id} not found.')
             continue
 
+        # WP feature entries cannot be confidential, so this should not matter.
+        # But, if they ever did exist, we exclude them from the report because
+        # the user is not authenticated.
+        if feature.confidential:
```
jcscottiii: Question: Would it make sense to expand this guard to check `if feature.confidential or feature.deleted or stage.archived: continue`? That guarantees soft-deleted features or superseded shipping stages won't leak into `/api/v0/features/shipping`

jcscottiii (approved):
LGTM with one question

DanielRyanSmith (approved):

DanielRyanSmith:
Thank you for taking care of this for me!

PR #6613, opened 2026-07-15
Exclude confidential features from other WP reports.

jrobbins:
These reports were flagged in a section of an internal issue.
WP features cannot be confidential, so this is moot at present, but with this PR we explicitly exclude any confidential features.

### DIFF
--- a/api/external_reviews_api.py
+++ b/api/external_reviews_api.py
@@ -170,18 +170,18 @@ class ExternalReviewsAPI(basehandlers.APIHandler):
         # WP features cannot be confidential, so this should not matter.
         # But if any is ever somehow confidential, remove it.
         unreviewed_features = [
-            feature
-            for feature in unreviewed_features
-            if not feature.confidential
+            fe
+            for fe in unreviewed_features
+            if not fe.confidential and not fe.deleted
         ]
 
         # Remove features for which the review link isn't a request for the review group to review the  # noqa: E501
         # feature.
         unreviewed_features = [
-            feature
-            for feature in unreviewed_features
+            fe
+            for fe in unreviewed_features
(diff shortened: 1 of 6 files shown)


api/external_reviews_api.py
```diff
@@ -167,6 +167,14 @@ def do_get(self, **kwargs):
         reviewer_info = ExternalReviewerInfo(review_group)
         unreviewed_features = reviewer_info.unreviewed_features_query.fetch()
 
+        # WP features cannot be confidential, so this should not matter.
+        # But if any is ever somehow confidential, remove it.
+        unreviewed_features = [
+            feature
+            for feature in unreviewed_features
+            if not feature.confidential
```
jcscottiii: Same question as here: https://github.com/GoogleChrome/chromium-dashboard/pull/6612#discussion_r3589223843

api/spec_mentors_api.py
```diff
@@ -57,6 +57,9 @@ def do_get(self, **kwargs):
             if feature.unlisted:
                 # TODO: Consider showing these when the caller is logged in and has the right to see them.  # noqa: E501
                 continue
+            if feature.confidential:
```
jcscottiii: Same question as here: https://github.com/GoogleChrome/chromium-dashboard/pull/6612#discussion_r3589223843

jcscottiii (approved):
LGTM with two questions

api/spec_mentors_api.py
jrobbins: Done

api/external_reviews_api.py
jrobbins: Done

PR #6617, opened 2026-07-16
Replace external_reviewers use of active_stage_id

jrobbins:
I've been hoping to phase out `active_stage_id` since last year, however the external reviewers page uses it.  This CL replaces that logic with new logic that should give very similar results.

It used to be that users were supposed to check a checkbox to make a given stage the active stage.  But, of course, human users have no reason to do that when they are focused on making the changes that they need for approvals.  So, our active_stage_id field was usually wrong.  Last year I added logic to automatically set the `active_stage_id` to the "highest" stage touched in any edit.

This PR does an similar calculation to find the highest stage that has any milestones specified.  This seems better for the external reviewers report because it deals specifically with milestones.

After this PR merges, there will be no more usage of `active_stage_id`, so I will follow up with a PR that removes it from the UI.


api/external_reviews_api.py
```diff
@@ -264,3 +263,44 @@ def do_get(self, **kwargs):
...
+        self, unreviewed_features: list[FeatureEntry]
+    ) -> dict[int, Stage]:
+        """Build a map from each feature ID to the stage with the greatest stage_type that has at least one non-None milestone."""  # noqa: E501
+        # Collect all feature IDs.
+        feature_ids = [fe.key.id() for fe in unreviewed_features]
+
+        # Fetch all stages for these features in chunks to avoid query limits.
+        stages = []
+        for chunk in chunk_list(feature_ids, 30):
+            stages.extend(Stage.query(Stage.feature_id.IN(chunk)).fetch())
```
jcscottiii, line 277: Would it make sense to filter out archived stages? Or I guess, would we ever want to highlight an archived stage?

api/external_reviews_api.py
```diff
@@ -94,17 +96,39 @@ def stage_type(feature: FeatureEntry, stage: Stage | None) -> StageType:
     return tentative_type
 
 
-def min_of_present(*args):
-    """Return the minimum of the given arguments, ignoring None values."""
-    present_args = [arg for arg in args if arg is not None]
+def _min_of_first(stage: Stage | None) -> int | None:
```
jcscottiii, line 99: Gemini picked this one up. But I'm on the fence due to clarity:

---
  Instead of allocating two intermediate lists ([arg for arg in [...]]) and checking len(present_args) == 0, we can pass a generator directly to min(..., default=None) and max(..., default=None):

```py
    def _min_of_first(stage: Stage | None) -> int | None:
        """Return the minimum of the first milestones, ignoring None values."""
        if stage is None or stage.milestones is None:
            return None
        return min(
            (
                m
                for m in (
                    stage.milestones.desktop_first,
                    stage.milestones.android_first,
                    stage.milestones.ios_first,
                    stage.milestones.webview_first,
                )
                if m is not None
            ),
            default=None,
        )
```
  If stage.milestones has [100, None, 102, None], min cleanly returns 100. And if all milestones are None ([None, None, None, None]), min cleanly returns default=None without needing temporary lists or len == 0 checks!

jcscottiii (approved):

api/external_reviews_api.py
jrobbins: Good point.  Done.

api/external_reviews_api.py
jrobbins: True, that is simpler.  Changed.

PR #6618, opened 2026-07-16
Add sitemap.txt

jrobbins:
This is one of the preliminary steps in making ChromeStatus more crawlable.

This PR replaces the old, broken sitemap entry with one that points to a new page at /sitemap.txt.
It generates a list of all viewable features, one per line.


pages/sitemap.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+        keys = query.fetch(keys_only=True)
+
+        feature_ids = [k.integer_id() for k in keys]
+        feature_ids.sort()
+
+        urls = [
+            f'https://chromestatus.com/feature/{fid}' for fid in feature_ids
+        ]
+        content = ('\n'.join(urls) + '\n') if urls else ''
```
jcscottiii: Could we do this instead? 
```suggestion
        content = ''.join(f'https://chromestatus.com/feature/{fid}\n' for fid in feature_ids)
```

pages/sitemap.py
```diff
@@ -0,0 +1,48 @@
...
+        headers = self.get_headers()
+        headers['Content-Type'] = 'text/plain'
+
+        # Query for non-deleted, non-unlisted, and non-confidential features.
+        # We use keys_only=True for optimal performance and memory footprint.
+        query = FeatureEntry.query(
+            FeatureEntry.deleted == False,  # noqa: E712
+            FeatureEntry.unlisted == False,  # noqa: E712
+            FeatureEntry.confidential == False,  # noqa: E712
+        )
```
jcscottiii, line 38: Not in scope for this PR but I wonder if we want to DRY this up. Especially with other recent PRs.

jcscottiii: I gave Gemini the recent PRs and asked if that was a good idea and it said the following:

# Proposal: Centralizing Public Feature Eligibility Helpers (`get_public_features` / `filter_public_features`)

Across our recent PRs (`sitemap.txt`, `stalefeatures`, `spec_mentors`, `external_reviews`, and `shipping_features`), notice that every single public report or index endpoint needs to retrieve or filter features that are safe to show in public lists:
- `FeatureEntry.deleted == False`
- `FeatureEntry.unlisted == False`
- `FeatureEntry.confidential == False`

Because each handler writes these conditions ad-hoc (either in NDB queries or Python loop guards), we occasionally miss one (`deleted` or `unlisted`), leaving subtle leakage gaps across public dashboards.

Furthermore, `permissions.can_view_feature(user, feature)` isn't quite right for public index reports: because unlisted features are accessible by direct link (`/feature/123`), `can_view_feature()` intentionally allows them through. But for **sitemaps, roadmap lists, and canonical reports**, we need an invariant for *public listability*.

---

### Proposed Helpers in `internals/feature_helpers.py`

We could introduce four self-documenting helpers right in `internals/feature_helpers.py` to cover both NDB queries and in-memory list filtering:

```python
def get_public_features_query() -> ndb.Query:
    """Returns an NDB query for non-deleted, listed, and non-confidential features."""
    return FeatureEntry.query(
        FeatureEntry.deleted == False,
        FeatureEntry.unlisted == False,
        FeatureEntry.confidential == False,
    )


def get_public_features(keys_only: bool = False) -> list[Any]:
    """Fetches all publicly listable features (or their keys if keys_only=True)."""
    return get_public_features_query().fetch(keys_only=keys_only)


def is_publicly_listable(feature: FeatureEntry | None) -> bool:
    """Returns True if a feature is active, listed, and not confidential."""
    if not feature:
        return False
    return not (feature.deleted or feature.unlisted or feature.confidential)


def filter_public_features(
    features: Iterable[FeatureEntry],
) -> list[FeatureEntry]:
    """Filters an in-memory list to only include publicly listable features."""
    return [f for f in features if is_publicly_listable(f)]
```

---

### How much cleaner and safer our endpoints become:

#### 1. In `SitemapHandler` (`pages/sitemap.py` or `api/sitemap_api.py`)
```python
# Before (3 lines of NDB filtering)
query = FeatureEntry.query(
    FeatureEntry.deleted == False,
    FeatureEntry.unlisted == False,
    FeatureEntry.confidential == False,
)
keys = query.fetch(keys_only=True)

# After (1 clean, intention-revealing line)
keys = feature_helpers.get_public_features(keys_only=True)
```

#### 2. In `ExternalReviewsAPI` and `SpecMentorsAPI`
```python
# Before (Ad-hoc list comprehensions or loop guards that might omit .deleted or .unlisted)
unreviewed_features = [f for f in unreviewed_features if not f.confidential]

# After (Guarantees soft-deleted, unlisted, and confidential features are excluded)
unreviewed_features = feature_helpers.filter_public_features(unreviewed_features)
```

#### 3. In `StaleFeaturesAPI.do_get` and `aggregate_shipping_features`
```python
# Before (Multiple loop guards)
if feature.unlisted or feature.confidential:
    continue

# After (Single unified guard)
if not feature_helpers.is_publicly_listable(feature):
    continue
```

jcscottiii (approved):
LGTM with one suggestion/question for this PR.

Another suggestion for later too.

PR #6623, opened 2026-07-16
Fix formatting of amendments that include brackets.

jrobbins:
I happened to see a feature name that was like "Name of feature [SOMETHING]" and the acvitiy log showed it without the final "]".   This PR is a simple fix.


api/comments_api_test.py
jcscottiii, line 1: I know it's a simple helper. Did we want to assert the happy case of the stripping?

jcscottiii: Something like:
- `'[item1, item2]' -> 'item1, item2'`
- `'word [note]' -> 'word [note]'` (this is what you have)
- `'[word' - > '[word'`
- `'word]' -> 'word]'`

api/comments_api.py
```diff
@@ -36,12 +36,19 @@
 from internals.review_models import Activity, Amendment, Gate
 
 
+def _strip_brackets(s: str) -> str:
+    """Make serialized arrays look nicer."""
+    if s.startswith('[') and s.endswith(']'):
```
jcscottiii, line 41: Did you ever find any nested ones? 

May want something like this 
```py
    while s.startswith('[') and s.endswith(']'):
        s = s[1:-1]
    return s
```

But if you did not find it, we can leave that as-is.

jcscottiii (approved):
LGTM with a few questions. Non blocking though.

api/comments_api.py
jrobbins: I don't think that we have any field values that have nested arrays.  Also, that won't work because
`[[a, b], [c, d]]`
would give
`a, b], [c, d`

api/comments_api_test.py
jrobbins: Done

PR #6626, opened 2026-07-17
Initial implementation of SSR feature detail

jrobbins:
This is an initial implementation of implementing an SSR version of our feature detail page using our current technology stack.

In this PR:
* Copy the form and field definition constants from the client side to now be available on the server.  The SPA will continue to use the typescript constants, meaning that we now have two copies of these definitions.  I did not implement an API to serve these constants to the SPA because there are many other details in the typescript constants that are not relevant to the server, and I don't want to further complicate the client.  The skill to add a field now mentions both locations.
* Implement the new request handler at `/feature-ssr/<feature_id>`. It will eventually replace the existing URI, but for now it is separate so that we can compare the SPA and MPA versions during further development.
* Implement new templates and macros to render a bare-bones, unstyled, version of the page.  This is starting point for further refinement.
* Add a simple playwright test with a screenshot so that there is an image of the result in this PR.


pages/form_definitions.py
```diff
@@ -0,0 +1,488 @@
...
+
+    name: NotRequired[str]
+    fields: list[str]
+
+
+class FormDef(TypedDict):
+    """Represents a form definition with a name and a list of sections."""
+
+    name: str
+    sections: list[Section]
```
jcscottiii, line 37: Quick question for follow up:

Seeing this schema definition made me think: While there is some duplication, we could reduce some of it. Keep `Section` and FormDef` defined in both Python and TypeScript. But move the form definitions themselves to a common JSON file (not using anything like openapi).

Then this file becomes something like this:

```python
import json
import pathlib
from typing import Any, TypedDict

class Section(TypedDict):
    name: str
    fields: list[str]

class FormDef(TypedDict):
    sections: list[Section]

_JSON_PATH = pathlib.Path(__file__).parent / 'form_definitions.json'
with open(_JSON_PATH, 'r', encoding='utf-8') as f:
    _data: dict[str, Any] = json.load(f)

FLAT_METADATA_FIELDS: FormDef = _data['FLAT_METADATA_FIELDS']
```

and the frontend becomes something like this:

```typescript
export interface Section {
  name: string;
  fields: string[];
}

export interface FormDef {
  sections: Section[];
}

import formDefsJson from '../../pages/form_definitions.json';

export const FLAT_METADATA_FIELDS: FormDef = formDefsJson.FLAT_METADATA_FIELDS;
```

(may have to play with resolveJsonModule in tsconfig for this)

pages/form_definitions.py
```diff
@@ -0,0 +1,488 @@
...
+    core_enums.STAGE_PSA_IMPLEMENT: PSA_IMPLEMENT_FIELDS,
+    core_enums.STAGE_PSA_DEV_TRIAL: FLAT_DEV_TRIAL_FIELDS,
+    core_enums.STAGE_PSA_SHIPPING: PSA_PREPARE_TO_SHIP_FIELDS,
+    core_enums.STAGE_DEP_PLAN: DEPRECATION_PLAN_FIELDS,
+    core_enums.STAGE_DEP_DEV_TRIAL: DEPRECATION_DEV_TRIAL_FIELDS,
+    core_enums.STAGE_DEP_DEPRECATION_TRIAL: DEPRECATION_ORIGIN_TRIAL_FIELDS,
+    core_enums.STAGE_DEP_EXTEND_DEPRECATION_TRIAL: FLAT_TRIAL_EXTENSION_FIELDS,
+    core_enums.STAGE_DEP_SHIPPING: DEPRECATION_PREPARE_TO_SHIP_FIELDS,
+    core_enums.STAGE_ENT_ROLLOUT: FLAT_ENTERPRISE_PREPARE_TO_SHIP_FIELDS,
+    core_enums.STAGE_ENT_SHIPPED: FLAT_PREPARE_TO_SHIP_FIELDS,
```
jcscottiii, line 490: I see that there's a `STAGE_DEP_REMOVE_CODE` [stage](https://github.com/GoogleChrome/chromium-dashboard/blob/3f9d5e697b7e70665b1d889161cc9596dd796084/internals/core_enums.py#L280). But I also see that it's [commented out](https://github.com/GoogleChrome/chromium-dashboard/blob/3f9d5e697b7e70665b1d889161cc9596dd796084/client-src/elements/form-field-enums.ts#L319 ) in the frontend.

Question: Could we 1) add a comment as to why `STAGE_DEP_REMOVE_CODE` is missing here or 2) or add a mapping for it? (maybe `core_enums.STAGE_DEP_REMOVE_CODE: DEPRECATION_PREPARE_TO_SHIP_FIELDS`?)

pages/featuredetail.py
```diff
@@ -0,0 +1,178 @@
...
+    'extension_desktop_last': 'desktop_last',
+}
+
+
+def _get_form_fields(form_def: Any) -> list[str]:
+    """Flattens all fields across all sections in a FormDef, preserving order."""
+    fields = []
+    for section in form_def.get('sections', []):
+        for field in section.get('fields', []):
+            if field not in fields:
```
jcscottiii, line 67: Gemini spotted this one optimization suggestion in the event there are a lot of fields: (use a seen set instead of field not in fields to reduce the linear search to constant)

```python
fields: list[str] = []
seen: set[str] = set()
for section in form_def.get('sections', []):
    for field in section.get('fields', []):
        if field not in seen:
            seen.add(field)
            fields.append(field)
return fields
```

jcscottiii (approved):
LGTM with some questions and one suggestion to consider for long term.

pages/form_definitions.py
jrobbins: The same thought occurred to me.  I'll leave a TODO comment to consider coming back to try this.

pages/form_definitions.py
jrobbins: OK, I added a comment to state the facts of the situation: these stages exist in the DB but they are not currently displayed.
 
There is a "Remove code" step in https://www.chromium.org/blink/launching-features/#remove-code, however there is nothing for the feature owner to do in ChromeStatus at that point, so we don't display a stage that would have zero fields and zero gates.  If we ever do decide to display something for that stage, the DB entities are already there.

pages/featuredetail.py
jrobbins: Probably not worth it since the list of fields in a section is at most a could dozen.

jcscottiii (approved):

PR #6655, opened 2026-07-24
Catch APIErrors that exceed retry limit.

jrobbins:
This should resolve some of the cloud tasks errors that have been happening lately.

The link_helpers file tries to make various HTTP requests, and if they fail, the exception is caught and the request is retried, up to a limit.  If the limit is reached, the same exception is raised again to a higher layer in the call stack.  PR #6586 addressed a recent GH API change that caused a different type of exception to be raised by catching that new exception type.  However, it only added the new exception type in the inner retry loop, not the outer exception handler.

Also, the code seems to function differently when tested locally vs in prod, maybe because a different version of GH API is used.

In this PR,
* Catch the APIError exception in the outer scope.


internals/link_helpers.py
```diff
@@ -430,11 +430,18 @@ def parse(self):
...
+            ValueError,
+            APIError,
+        ) as e:
             logging.error(f'Error parsing {self.type} {self.url}: {e}')
             self.error = e
             self.is_error = True
             if isinstance(e, HTTPError):
                 self.http_error_code = e.code
+            if isinstance(e, APIError):
+                self.http_error_code = _get_error_code(e)
```
jcscottiii: Gemini had a style suggestion for this. Since get_error_code input is typed to any Exception, you could just always use get_error_code instead of doing check for each instance type.

https://github.com/GoogleChrome/chromium-dashboard/blob/95784b84824a54ad4e89b57568ae7263b8bfde5a/internals/link_helpers.py#L36-L44


So this could would be simplified like this: 


```python
        except (
            requests.RequestException,
            HTTPError,
            ValueError,
            APIError,
        ) as e:
            logging.error(f'Error parsing {self.type} {self.url}: {e}')
            self.error = e
            self.is_error = True
            self.http_error_code = _get_error_code(e)
            self.information = None
```

jcscottiii (approved):
LGTM with one optional suggestion

PR #6671, opened 2026-07-27
Fix failing feature links tasks

jrobbins:
Some of our feature_links tasks are failing over and over because they exceed the 10 minute time limit.  

These tasks check batches of URLs mentioned in ChromeStatus entries.  The batch size was 100, but this PR reduces it to 75.

Also, instead of all-or-nothing, the code now skips over recently updated feature_links entities, meaning that if a task fails for any reason, the next run of that task will skip over any URLs that succeeded, and thus picking up where the previous run left off.


internals/feature_links_test.py
```diff
@@ -362,6 +363,15 @@ def side_effect_2(self_link):
         mockParse.side_effect = side_effect_2
 
         link = FeatureLinks.query(FeatureLinks.url == url).get()
+        auto_now = FeatureLinks._properties['updated']._auto_now
+        try:
+            FeatureLinks._properties['updated']._auto_now = False
+            link.updated = datetime.datetime.now() - datetime.timedelta(
```
jcscottiii, line 369: Nit: we should use utc timezone like in the main code to be consistent. Here and the other places in the tests in this diff

jcscottiii (approved):
LGTM with one suggestion

Also Gemini found these to consider to further help with the reliability:
```python
            link.parse()
```
**Comment**:
If `link.parse()` raises an unexpected exception (e.g., `AttributeError` if the GitHub API response structure changes and lacks expected attributes), it will crash the task. Because the task crashes before `feature_link.put()` is called, this link will not be marked as updated. On retry, the task will attempt to process this link again, leading to an infinite retry loop.

Consider wrapping `link.parse()` in a `try-except Exception` block here to handle unexpected errors gracefully, log them, mark the link as failed, and allow the task to continue processing the rest of the batch:
```python
            try:
                link.parse()
            except Exception as e:
                logging.error(f'Unexpected error parsing {feature_link.url}: {e}')
                link.is_error = True
                link.error = e
```

#### Line 557 (Maintainability / Non-blocking)
```python
                # if error exists
                elif fe.is_error or fe.http_error_code:
```
**Comment**:
This selection logic in the cron job selects all links in an error state for updates on every run, regardless of how recently they were checked. For permanently broken links, this means they will be retried indefinitely on every cron run.

Consider adding a backoff mechanism, a maximum retry limit, or checking their `updated` timestamp (e.g., only retry failing links if they haven't been checked in the last 24 hours) to avoid wasting task queue resources.

PR #6672, opened 2026-07-28
[Backend APIs] Implement Summary Suggestion GET/PATCH CRUD API with OCC Version Token Enforcement

jcscottiii:
## Summary

Implements `GET` and `PATCH /api/v0/summary-suggestions/<int:feature_id>` for retrieving and updating AI summary suggestions and progress steps timelines.

Enforces Optimistic Concurrency Control (OCC) via `version_token` during `PATCH` operations to prevent concurrent edit overwrites. Validates string status updates against `SummarySuggestionStatus` enums and checks feature edit permissions via `permissions.can_edit_feature(user, feature)`.

## Key Changes

- **Controller (`api/summary_suggestion_api.py`)**: Adds `SummarySuggestionAPI` handling `do_get` and `do_patch`. Leverages repository helpers `self.get_specified_feature(**kwargs)` for feature retrieval & view authorization, and `self.get_int_param` for OCC `version_token` extraction. Orders progress steps timeline by `start_timestamp`. Validates OCC `version_token` matching (HTTP 409 Conflict on mismatch) and relies on Cloud NDB `auto_now=True` for automatic UTC timestamp updates.
- **Route (`main.py`)**: Registers `GET` and `PATCH /api/v0/summary-suggestions/<int:feature_id>`.
- **Testing (`api/summary_suggestion_api_test.py`, `api/converters_test.py`)**: Added `SummarySuggestionAPITest` suite verifying GET access, status updates, OCC version token conflict handling (HTTP 409), invalid enum validation (HTTP 400), and BOLA permission checks (HTTP 403) isolated via `@mock.patch`.

TAG=agy
CONV=86f63625-bdb5-4d50-ac8d-2f8ca5128ca9

### DIFF
--- a/api/summary_suggestion_api.py
+++ b/api/summary_suggestion_api.py
@@ -1,0 +1,154 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""API handler for AI summary suggestions (GET and PATCH /api/v0/summary-suggestions/<int:feature_id>)."""
+
+from datetime import datetime, timezone
+from typing import Any
+
+from chromestatus_openapi.models import SummarySuggestionResponse
+from google.cloud import ndb
+
+from api import converters
+from framework import basehandlers, permissions
+from internals import core_enums
+from internals.core_models import (
+    FeatureSummaryProgressStep,
+    FeatureSummarySuggestion,
+)
+
+
+class SummarySuggestionAPI(basehandlers.APIHandler):
+    """API handler for AI summary suggestions (GET and PATCH /api/v0/summary-suggestions/<int:feature_id>)."""
+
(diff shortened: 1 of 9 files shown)


api/summary_suggestion_api.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # 4. Build and return OpenAPI-compliant dictionary response.
+        payload = {
+            'suggestion': converters.feature_summary_suggestion_to_dict(
+                suggestion
+            ),
+            'progress_steps': [
+                converters.summary_progress_step_to_dict(s) for s in steps
+            ],
+            'access_level': access_level,
```
jrobbins: Remind me why this is needed.  The frontend already uses the permissions_api.

api/summary_suggestion_api.py
```diff
@@ -0,0 +1,154 @@
...
+            except (ValueError, TypeError):
+                self.abort(400, msg=f'Invalid status value: {raw_status}')
+            suggestion.status = status_enum.value
+
+        if 'suggested_summary' in request_body:
+            suggestion.suggested_summary = request_body['suggested_summary']
+
+        # 6. Increment OCC version token, update timestamp, and save entity.
+        suggestion.version_token += 1
+        suggestion.updated = datetime.now(timezone.utc)
```
jrobbins: I don't think that you need to update the `updated` field because it is defined with `auto_now=True`.

api/summary_suggestion_api.py
jcscottiii: This is leftover from the extra can_review_release_notes role. I will remove this (and also edit the openapi docs)

api/summary_suggestion_api.py
jcscottiii: good catch!

jrobbins (approved):

PR #6673, opened 2026-07-28
Handle GitHub API exceptions in feature-links task

### LINKED ISSUE #6632: feature-links task has exceptions
The feature-links makes a lot of requests to GitHub to check the status of issues.  We get rate limited, which is expected, but apparently we don't catch the exceptions properly.  So, the tasks can repeatedly fail.

love12yadav:
Summary

Fixes the feature-links cron/task pipeline so that a single failed GitHub API call (most commonly caused by rate limiting) no longer interrupts an entire batch of link updates.

Addresses #6632

Problem

internals/link_helpers.py's Link.parse() only handles a fixed set of exception types (requests.RequestException, HTTPError, ValueError, APIError).

When GitHub rate-limits requests (403/429), the underlying client may raise an exception outside this handled set. The exception then propagates through _index_feature_links_by_ids() in internals/feature_links.py, which previously had no exception handling.

This caused the /tasks/update-feature-links handler to return a 500 response, triggering Cloud Tasks retries on the same batch of links. The task would repeatedly fail instead of degrading gracefully.

Changes
internals/feature_links.py
Wrapped the per-link Link(...).parse() call inside _index_feature_links_by_ids() with try/except Exception so unexpected failures no longer terminate the entire batch.
Added _http_status_code_from_exception() and _normalise_http_status_code() helpers to extract HTTP status codes from common exception formats (.status_code, .status, .code, and nested .response objects).
Detect 403/429 rate-limit responses from:
raised exceptions, and
Link.parse() results (is_error / http_error_code).
Stop processing gracefully when rate limiting is detected, logging a warning with the number of remaining unprocessed links. Remaining links will be retried during the next scheduled run.
Handle non-rate-limit failures on individual links by recording the error on the corresponding FeatureLinks entity (is_error / http_error_code) and continuing processing of remaining links.

@jrobbins

Fixes #6632

### DIFF
--- a/internals/feature_links.py
+++ b/internals/feature_links.py
@@ -37,6 +37,8 @@ from internals.link_helpers import (
 LINK_STALE_MINUTES = 30
 CRON_JOB_LINK_STALE_DAYS = 8
 
+_RATE_LIMIT_HTTP_CODES = frozenset((403, 429))
+
 
 class FeatureLinks(ndb.Model):
     """Links that occur in the fields of the feature.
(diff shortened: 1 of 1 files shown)


jrobbins:
I looked at this but I will need to think about it more.  In the case of rate limit exceeded, we want the task to be rerun later.

jrobbins:
This was addressed in #6655.

PR #6702, opened 2026-08-06
Highlight links that are initially 404.

jrobbins:
I made a change to the feature_links functionality last week and I thought that I somehow broke the 404 detection.  It turns out that we were never detecting links that were initially 404, only links that were initially 200 and that later become 404.  

IIRC, that requirement was decided that way to avoid UI clutter for long text that includes link-like text like `https://foo.com/bar` or `http://DOMAIN.TLD/PATH`.  However, we never added the 404 labels to links inside the long text fields, they are only added to fields that are expected to contain only URLs.

In this PR:
* Record when a newly edited link is initially 404
* Don't try to preview them on the external reviewers page because they are 404

### DIFF
--- a/internals/feature_links.py
+++ b/internals/feature_links.py
@@ -121,8 +121,6 @@ def _get_index_link(
     else:
         if should_parse_new_link:
             link.parse()
-            if link.is_error:
-                return None
         feature_link = FeatureLinks(
             feature_ids=[feature_id],
             type=link.type,
--- a/api/external_reviews_api.py
+++ b/api/external_reviews_api.py
@@ -219,7 +219,11 @@ class ExternalReviewsAPI(basehandlers.APIHandler):
         review_links = {
             reviewer_info.review_link(fe) for fe in unreviewed_features
         }
-        previewable_urls = {fl['url'] for fl in feature_links}
+        previewable_urls = {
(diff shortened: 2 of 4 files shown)


internals/feature_links.py
```diff
@@ -121,8 +121,6 @@ def _get_index_link(
     else:
         if should_parse_new_link:
             link.parse()
-            if link.is_error:
```
jcscottiii, line 124: This change is fine here. But I have a question. When an existing link transitions from 200 to 404/error, we currently don't reset information to None. Is that intentional? Any risk? Maybe we could look at the logs maybe to make a decision?

We could also just clear it on 4xx errors. I also checked and saw there was a TODO to potentially notify users:

https://github.com/GoogleChrome/chromium-dashboard/blob/df5ab49401bd0d2afd4d84b5c2ebc39ffdec6b82/internals/feature_links.py#L354

jcscottiii (approved):
LGTM with one question that we may want to tackle now/soon too

internals/feature_links.py
jrobbins: Hmm, let's see if users ever mention it.  I think it could be potentially useful to see an old page `<title>` on a page that became 404.

PR #6765, opened 2026-08-18
Upgrade all codeql versions at same time.

jrobbins:
dependabot's version was failing because it didn't update everything at the same time.


jrobbins:
Now that I see that this works, I will make this change as part of #6761.

PR #6871, opened 2026-09-10
Avoid repeated writing of feature for feature_links

jrobbins:
This should further reduce latency when saving edits to a feature entry.   When updating a feature entry, we check all fields for URLs and then update FeatureLinks objects.  For vendor signals, we also check the URL that was entered and save some info about what was found at that URL.  However, the logic for that called `put()` inside a loop, which is redundant and slow.

In this PR:
* Just keep track of whether signals-related fields were edited and call `put()` once at the end.

This saves about 1 second for each field that was edited.  E.g., if the user edited 5 fields and then hit submit, we were spending 5 seconds on this part of the logic, which is now reduced to usually 0 or sometimes 1 second.


neilv-g (approved):
To the log, do you want to add the feature name or anything?
LGTM otherwise!

jrobbins:
> To the log, do you want to add the feature name or anything?

Other log lines give the feature ID, so that should be enough for me to visit the feature detail page.

PR #6877, opened 2026-09-11
End OTs at M+4 (usually).

jrobbins:
This is additional follow-through on the change from Chrome 4-week releases to 2-week releases.  The end date of an origin trial should usually be M+4 now, but there are some transitional details in the logic.

In this PR:
* Refactor the end-milestone logic into origin_trial_client
* Use it there to compute the end date when creating the OT


DanielRyanSmith (approved):
Awesome, glad to see the logic be reused. 🙂

jrobbins:
I'll merge what we have here, but thinking about it more M+3 might be aligned with the original intent of trial admin doc to have a 4-week buffer. 

Old calculation: M + 1 + 1 = M+2
New calculation: M + 1 + 2 = M+3
Where the first 1 was to make the entered end milestone number inclusive rather than exclusive.
And, the second value was to add the 4-week buffer.

However, we've been sending out reminders with the current logic, so we should stick with it.  By using M+4, I think we increased the buffer from 4 weeks to 6 weeks from M153 onward.

For future reference, this might all be more clear if we added the buffer as time rather than as milestones.  So that would be get_stable_date(M+1) + desired_buffer_duration.

MayLove31 (commented):
Common knowledge

PR #6878, opened 2026-09-12
Migrate from Shoelace to Cordwainer Elements via cw-migrate

matthandus:
## What this is

A real-world test of [`@cordwainer/cw-migrate`](https://cordwainer.dev/tooling/cw-migrate) — the CLI for migrating [Shoelace](https://shoelace.style) projects to [Cordwainer Elements](https://cordwainer.dev) — run against this project's actual Shoelace usage (80 files: custom element tags, CSS custom properties, custom events, TypeScript type references, and package imports).

This is **not** a proposal to change chromium-dashboard's dependencies. It's a demo/validation artifact for cw-migrate's own release plan, published on my fork so the diff is easy to review.

## What was found and fixed in cw-migrate along the way

Before this test, `cw-migrate` had never been run against a real Shoelace codebase. Running it here surfaced four gaps, all fixed before producing this diff:

1. **`elements` never scanned `.ts`/`.js` files.** 77 of this project's 80 Shoelace-referencing files are `.ts` files using Lit's `` html`...` `` tagged templates — a file type the tag-renaming command never looked at. `detect`'s element-match count went from **2 → 1102** after the fix (confirmed via a before/after comparison with the unpatched tool).
2. **No CSS type-selector renaming.** Selectors like `sl-button { ... }` (common in these components' `` css`...` `` blocks — see `chromedash-timeline.ts`, `chromedash-wpt-eval-page.ts`) were never renamed, which would have silently broken styling post-migration.
3. **Plain `style="..."` attributes inside Lit templates** (e.g. `chromedash-link.ts`, `chromedash-wpt-eval-page.ts`) weren't covered by the existing custom-property rewrite, leaving 6 `--sl-*` references behind after the first pass.
4. **`querySelector`/`querySelectorAll`/`closest`/`matches` string arguments** in test files (e.g. `chromedash-release-feature-card_test.ts`) weren't renamed — these tests would have silently started matching nothing once the DOM elements they query for were migrated to `cw-*`.

## What's in this diff

- Every `sl-*` custom element tag renamed to `cw-*`, including inside Lit `html`` templates.
- Every `--sl-*` CSS custom property renamed to `--cw-*`.
- Every `sl-*` CSS type selector renamed to `cw-*` (stylesheets and `css`` blocks).
- Every `sl-*` custom event renamed to `cw-*`.
- Every `Sl*` TypeScript type reference renamed to `Cw*`.
- `@shoelace-style/shoelace` imports (including CDN/component subpaths) rewritten to `@cordwainer/cw-elements`.
- `package.json`/`package-lock.json` updated to `@cordwainer/cw-elements@1.4.0` (latest published stable).

## Validation

- `npm install`, `npx tsc`, and `npx rollup -c rollup.config.js` all succeed against the migrated result (the JS/TS half of this repo's own `make build`).
- Zero Shoelace tag/token/event/import/type references remain anywhere in `client-src`. The only remaining "shoelace" string in the codebase is `/static/shoelace`, a local static-asset path convention — unrelated to the package itself, out of scope for a source-code migration tool.
- One `tsc` error surfaced along the way: a stale `@ts-expect-error` suppression in `chromedash-textarea.ts`, no longer needed because `CwTextarea` doesn't have a type quirk `SlTextarea` had. Removed here as a one-line manual follow-up — not a cw-migrate bug.
- Spot-checked several files by hand (split closing-tag formatting, compound CSS selectors, template-literal style attributes) — all correct.


matthandus:
Sorry didn't mean to add this to your contrib.

PR #6882, opened 2026-09-14
Auto-extend origin trials when an Intent to Ship is approved

### LINKED ISSUE #6875: Auto-extend OT on approved I2S
Sometimes we have cases where a feature is approved to ship for the milestone immediately after an OT expires and it doesn't leave enough time for users to upgrade. Eg. Chrome Android needs at least 2 weeks post stable release to get even a simple majority of traffic (and has a long tail after that). To avoid gaps for users which haven't updated, partners may need to update their token.

Ideally Chromestatus would automatically extend the lifetime of an OT for shipped features - maybe even for an extra 3 months? 

API owners discussed and agree that once an I2S is approved and feature shipped, there is no reason that approval is needed to extend the OT for old Chrome. 

**Additional context**
API owners can handle this manually for now (by approving an OT extension after approving an I2S), so this isn't high priority. I just wanted to record that we agreed that there is no process reason to require API owners to take action here and ideally we'd prefer it be automatic.

hjanuschka:
Fixes #6875

When a ship gate (API Owners I2S review) becomes newly approved, ChromeStatus now checks whether the feature has an active origin trial that ends before the approved shipping milestone. If so, it automatically creates a trial extension stage up to the shipping milestone, with its gate already in an approved (N/A) state and a vote recorded from the I2S approver, flags it for OT support processing (`ot_action_requested`), and enqueues the existing extension-approved notification.

This covers both approval paths: votes via the review UI and LGTM detection on intent threads.

Not covered:
- Trials that already run through the shipping milestone, trials never created, or ship stages without a milestone.
- Deprecation features (shipping there means removal; deprecation trial extensions follow a different process).
- The OT API call itself; the existing OT support flow processes the pre-approved request. With trial end dates already buffered to ~M+4 (#6877), this gives users on older versions several months to update.


jrobbins:
I wrote a comment on the issue to clarify the desired UX flow.  Let's see if Rick agrees with my read of it.

I don't think that it is a good idea to automatically create an extension after every approved I2S because that will clutter the feature detail page for features that don't need an OT extension.

jrobbins:
As per Rick's comment on the issue, let's not do auto-extension and instead do auto-approval of requested extensions.

PR #6588, opened 2026-07-10
Add some permission checks

jrobbins:
This addresses some problems reported to us in an internal issue.   Basically, ChromeStatus was missing some permission checks when getting objects related to features, e.g., review comments. WP feature entries are the ones that have reviews, and enterprise features are the ones that can be temporarily confidential, so there should not be overlap in practice but it is good to have the checks anyway.

In this PR:
* Call get_specified_feature() even if we don't need the FeatureEntry object because it does permission checks.
* When getting a stage of a feature, make sure that the stage is actually part of that feature.

PR #6909, opened 2026-09-21
Avoid permission check when serving images from no-cookie domain.

jrobbins:
This reverts part of #6588.  In that PR, I added permission checks consistently throughout the app, but I forgot that this one request handler is not supposed to check permissions.

Background: To prevent serving user-control content from our application domain, we serve images from a random-looking domain name that has no cookies. However, no cookies means no authentication.  Instead of checking the user's session cookie, we simply rely on the attachment ID being hard to guess, and check that the request has a matching attachment ID and feature ID pair. These IDs are not impossible to guess, but a matching pair is a more difficult, and an attacker would need to be motivated to do that for something that is going to be published fairly soon anyway.

In this PR:
* Don't use `get_specified_feature()` because it does a permission check on the feature entry.  Instead, just use the attachment ID in the URL (And, still check that it belongs to the provided feature ID).


jrobbins:
The Chrome Enterprise team said that they are OK with this scheme, and explained that certain feature entries are not put into ChromeStatus until after they are announced elsewhere.

jstenback (approved):

